PR #8, opened 2024-07-30
feat: add basic parsing from tools file

kurtisvg:


internal/sources/cloud_sql_pg.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+package sources
+
+const CloudSQLPgKind string = "cloud-sql-postgres"
+
+// validate interface
+var _ Config = CloudSQLPgConfig{}
+
+type CloudSQLPgConfig struct {
```
Yuan325, line 22: Does this db connection need user and pass? Or is it set in a different location for toolbox?

kurtisvg: I was planning on handling this like we do in the langchain library -- using IAM Auth by default, and optional user/password auth fields (to add latter)

PR #9, opened 2024-09-18
feat: stub basic control plane functionality 

kurtisvg:
Stub's out some basic control plane functionality. This also required setting up some Source and Tools initialization.

### DIFF
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -28,7 +28,7 @@ func apiRouter(s *Server) chi.Router {
 
 	r.Get("/toolset/{toolsetName}", toolsetHandler(s))
 
-	// TOOD: make this POST
+	// TODO: make this POST
 	r.Get("/tool/{toolName}", toolHandler(s))
 
 	return r


internal/server/api.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/go-chi/render"
+)
+
+// apiRouter creates a router that represents the routes under /api
+func apiRouter(s *Server) chi.Router {
+	r := chi.NewRouter()
+
+	r.Get("/toolset/{toolsetName}", toolsetHandler(s))
+
+	// TOOD: make this POST
```
duwenxin99: This should be `TODO`. And why do we want to make this a POST request? We are supposed to get tool names from this handler right?

internal/server/server.go
```diff
@@ -24,28 +24,60 @@ import (
...
+	fmt.Printf("Initalized %d sources.\n", len(sources))
+
+	// initalize and validate the tools
+	tools := make(map[string]tools.Tool)
+	for name, tc := range cfg.ToolConfigs {
+		t, err := tc.Initialize(sources)
+		if err != nil {
+			return nil, fmt.Errorf("Unable to initialize tool %s: %w", name, err)
+		}
+		tools[name] = t
```
duwenxin99, line 68: Do we want to handle the case where user specify duplicate tool names in the config?

internal/server/api.go
kurtisvg: Right now it's calling "tool.invoke" which is what we want the POST path to do. I left it as GET because it's easier to test in the browser -- we will need to add POST and make sure they each do the appropriate action.

internal/server/server.go
kurtisvg: Good catch -- The YAML spec actually specifically calls out that for mapping, [the restriction that each of the keys is unique](https://yaml.org/spec/1.2.2/#nodes).  So it's not value yaml to have duplicate tool names, and the YAML parser should error.

PR #10, opened 2024-09-30
feat: Add toolset manifest

duwenxin99:
1. Created a `toolsets` package.
2. Created `ToolManifest` and `ToolsetManifest` structs.
3. Added `DescribeTool()` method to tool `Config` interface.
4. In `toolsetHandler`, implement logic to return requested `ToolsetManifest`.

### DIFF
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -45,13 +45,8 @@ func toolsetHandler(s *Server) http.HandlerFunc {
 			var toolsetManifest toolsets.ToolsetManifest
 			toolsetManifest.Tools = make(map[string]tools.ToolManifest)
 			toolsetManifest.ServerVersion = s.version
-			for name, tool := range s.tools {
-				toolManifest, err := tool.Describe()
-				if err != nil {
-					fmt.Errorf("Error describing tool %s", err)
-					return
-				}
-				toolsetManifest.Tools[name] = toolManifest
+			for name, toolConfig := range s.conf.ToolConfigs {
+				toolsetManifest.Tools[name] = toolConfig.DescribeTool()
 			}
 			json.NewEncoder(w).Encode(toolsetManifest)
 		} else {
@@ -59,11 +54,14 @@ func toolsetHandler(s *Server) http.HandlerFunc {
 			var toolsetManifest toolsets.ToolsetManifest
 			toolsetManifest.Tools = make(map[string]tools.ToolManifest)
 			toolsetManifest.ServerVersion = s.version
-			if toolset, ok := s.toolsets[toolsetName]; ok {
-				toolset.Describe()
+			if toolset, ok := s.conf.ToolsetConfigs[toolsetName]; ok {
+				for _, toolName := range toolset {
+					toolsetManifest.Tools[toolName] = s.conf.ToolConfigs[toolName].DescribeTool()
+				}
 				json.NewEncoder(w).Encode(toolsetManifest)
 			}
 			fmt.Errorf("toolset not found: %s", toolsetName)
+
 		}
 	}
 }
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -29,7 +29,7 @@ import (
 	"github.com/googleapis/genai-toolbox/internal/toolsets"
 )
 
-const serverVersion = "1.0.0"
+const ServerVersion = "1.0.0"
 
 // Server contains info for running an instance of Toolbox. Should be instantiated with NewServer().
 type Server struct {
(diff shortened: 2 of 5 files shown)


internal/server/server.go
```diff
@@ -26,15 +26,20 @@ import (
...
-	root chi.Router
+	conf    Config
+	root    chi.Router
+	version string
 
-	sources map[string]sources.Source
-	tools   map[string]tools.Tool
+	sources  map[string]sources.Source
+	tools    map[string]tools.Tool
+	toolsets map[string]toolsets.Toolset
```
kurtisvg, line 42: I think it's worth it to separate the implementation of toolsets from the implementation of the manifest design into separate PRs -- smaller PRs are better and can make sure we get those things right independently first before bringing them together. 

My 2c is the core manifest endpoint should probably come first (even if it just serves "all"), because it'll help us unblock prototyping SDKs sooner.

internal/server/api.go
```diff
@@ -37,7 +40,29 @@ func apiRouter(s *Server) chi.Router {
 func toolsetHandler(s *Server) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		toolsetName := chi.URLParam(r, "toolsetName")
-		_, _ = w.Write([]byte(fmt.Sprintf("Stub for toolset %s manifest!", toolsetName)))
+		if toolsetName == "" {
+			// return all the tools if no toolset name is provided
+			var toolsetManifest toolsets.ToolsetManifest
+			toolsetManifest.Tools = make(map[string]tools.ToolManifest)
```
kurtisvg, line 46: This makes an allocation every-time this endpoint is called, which will slow things down significantly -- is there a way we can do this calculation once head of time?

Note that our manifest is currently only changed on server start up, so we don't need to be concerned with race conditions currently.

internal/server/api.go
```diff
@@ -37,7 +40,29 @@ func apiRouter(s *Server) chi.Router {
...
+			toolsetManifest.ServerVersion = s.version
+			if toolset, ok := s.conf.ToolsetConfigs[toolsetName]; ok {
+				for _, toolName := range toolset {
+					toolsetManifest.Tools[toolName] = s.conf.ToolConfigs[toolName].DescribeTool()
+				}
+				json.NewEncoder(w).Encode(toolsetManifest)
+			}
+			fmt.Errorf("toolset not found: %s", toolsetName)
```
kurtisvg, line 63: nit: This code has several duplicated parts -- leaving separate leaves room for drift to happen in the future (especially when it's more complicated). Can we try to combine the logic where it's the same?

internal/server/server.go
```diff
@@ -48,32 +53,45 @@ func NewServer(cfg Config) (*Server, error) {
 	})
 
 	// initalize and validate the sources
-	sources := make(map[string]sources.Source)
+	initialized_sources := make(map[string]sources.Source)
```
kurtisvg, line 56: This is not idiomatic naming, see https://google.github.io/styleguide/go/decisions#naming

Specifically, Go strongly prefers shorter names and uses MixedCase over snake_case. In this case "initialized" doesn't seem to add any additional value.

internal/server/server.go
```diff
@@ -48,32 +53,45 @@ func NewServer(cfg Config) (*Server, error) {
...
 		}
-		tools[name] = t
+		initialized_tools[name] = t
+	}
+	fmt.Printf("Initalized %d tools.\n", len(initialized_tools))
+
+	// initalize and validate the toolsets
+	initilized_toolsets := make(map[string]toolsets.Toolset)
+	for name, tc := range cfg.ToolsetConfigs {
+		t, err := tc.Initialize()
```
kurtisvg, line 80: This looks like it's throwing an error -- is this added to the interface?

internal/server/api.go
```diff
@@ -37,7 +40,29 @@ func apiRouter(s *Server) chi.Router {
...
+			}
+			json.NewEncoder(w).Encode(toolsetManifest)
+		} else {
+			// Describe the tools of the request toolset
+			var toolsetManifest toolsets.ToolsetManifest
+			toolsetManifest.Tools = make(map[string]tools.ToolManifest)
+			toolsetManifest.ServerVersion = s.version
+			if toolset, ok := s.conf.ToolsetConfigs[toolsetName]; ok {
+				for _, toolName := range toolset {
+					toolsetManifest.Tools[toolName] = s.conf.ToolConfigs[toolName].DescribeTool()
```
kurtisvg, line 59: You can implement custom JSON behavior by implementing the [json.MarshalJSON interface ](https://pkg.go.dev/encoding/json#example-package-CustomMarshalJSON)-- maybe worth using over a new interface? It would also let you remove this loop -- the JSON encoding pacakge will handling it for you if you have an array of items implementing the interface .

PR #11, opened 2024-09-30
feat: Add Toolset manifest endpoint

duwenxin99:
1. Calculate tool manifests when server starts.
2. Add toolset manifest endpoints.

### DIFF
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -17,67 +17,95 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
-	"github.com/googleapis/genai-toolbox/internal/tools"
 )
-
-func newToolsetManifest(s *Server, c tools.ToolsetConfig) tools.ToolsetManifest {
-	toolsManifest := make(map[string]tools.ToolManifest)
-	for _, name := range c.ToolNames {
-		manifest := s.conf.ToolConfigs[name].Manifest()
-		toolsManifest[name] = manifest
-	}
-	return tools.ToolsetManifest{ServerVersion: s.conf.Version, ToolsManifest: toolsManifest}
-}
 
 // apiRouter creates a router that represents the routes under /api
 func apiRouter(s *Server) (chi.Router, error) {
 	r := chi.NewRouter()
 
-	r.Get("/toolset/{toolsetName}", toolsetHandler(s))
+	r.Get("/toolset/", func(w http.ResponseWriter, r *http.Request) { toolsetHandler(s, w, r) })
+	r.Get("/toolset/{toolsetName}", func(w http.ResponseWriter, r *http.Request) { toolsetHandler(s, w, r) })
 
 	r.Route("/tool/{toolName}", func(r chi.Router) {
 		r.Use(middleware.AllowContentType("application/json"))
-		r.Post("/invoke", newToolHandler(s))
+		r.Post("/invoke", func(w http.ResponseWriter, r *http.Request) { toolInvokeHandler(s, w, r) })
 	})
 
-	// Convert tool configs to JSON for manifest
-	for name, c := range s.conf.ToolsetConfigs {
-		manifest, err := json.Marshal(newToolsetManifest(s, c))
-		if err != nil {
-			return nil, fmt.Errorf("unable to marshal toolset: %w", err)
-		}
-		s.manifests[name] = manifest
-	}
 	return r, nil
 }
 
-func toolsetHandler(s *Server) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		toolsetName := chi.URLParam(r, "toolsetName")
-		manifest, ok := s.manifests[toolsetName]
-		if !ok {
-			http.Error(w, fmt.Sprintf("Toolset %q does not exist", toolsetName), http.StatusNotFound)
-			return
-		}
-		_, _ = w.Write([]byte(manifest))
-
+// toolInvokeHandler handles the request for information about a Toolset.
+func toolsetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
+	toolsetName := chi.URLParam(r, "toolsetName")
+	toolset, ok := s.toolsets[toolsetName]
+	if !ok {
+		http.Error(w, fmt.Sprintf("Toolset %q does not exist", toolsetName), http.StatusNotFound)
+		return
 	}
+	b, err := json.Marshal(toolset.Manifest)
+	if err != nil {
+		log.Printf("unable to JSON the toolset manifest: %s", err)
+		http.Error(w, "", http.StatusInternalServerError)
+		return
+	}
+	_, _ = w.Write(b)
 }
+
+// toolInvokeHandler handles the API request to invoke a specific Tool.
+func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
+	toolName := chi.URLParam(r, "toolName")
+	tool, ok := s.tools[toolName]
+	if !ok {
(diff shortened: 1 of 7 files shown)


internal/server/api.go
```diff
@@ -36,8 +37,19 @@ func apiRouter(s *Server) chi.Router {
 
 func toolsetHandler(s *Server) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		toolsetName := chi.URLParam(r, "toolsetName")
-		_, _ = w.Write([]byte(fmt.Sprintf("Stub for toolset %s manifest!", toolsetName)))
+		type ToolsetManifest struct {
+			ServerVersion string `json:"serverVersion"`
+			Tools         []byte `json:"tools"`
+		}
```
kurtisvg: This defines this struct every-time we need to use it -- any reason not to define it once place?

internal/server/api.go
```diff
@@ -36,8 +37,19 @@ func apiRouter(s *Server) chi.Router {
 
 func toolsetHandler(s *Server) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		toolsetName := chi.URLParam(r, "toolsetName")
-		_, _ = w.Write([]byte(fmt.Sprintf("Stub for toolset %s manifest!", toolsetName)))
+		type ToolsetManifest struct {
+			ServerVersion string `json:"serverVersion"`
+			Tools         []byte `json:"tools"`
```
kurtisvg: I think you would be better off making this an array (or map) or tools and letting them JSON themselves

internal/server/server.go
```diff
@@ -28,13 +29,17 @@ import (
...
-	root chi.Router
+	conf    Config
+	root    chi.Router
+	version string
 
-	sources map[string]sources.Source
-	tools   map[string]tools.Tool
+	sources       map[string]sources.Source
+	tools         map[string]tools.Tool
+	toolsManifest []byte
```
kurtisvg: nit: maybe make this "rawManifest" if it's all tools and already ready for a string

internal/server/api.go
```diff
@@ -36,8 +37,19 @@ func apiRouter(s *Server) chi.Router {
...
+		}
+		toolsetManifest := ToolsetManifest{
+			ServerVersion: s.version,
+			Tools:         json.RawMessage(s.toolsManifest),
+		}
+		w.Header().Set("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(toolsetManifest); err != nil {
```
kurtisvg: if we store bytes for the already encoded format ahead of time, we can just write the bytes here

internal/server/api_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/http/httptest"
+	"testing"
+)
+
+func TestToolsetManifest(t *testing.T) {
+	serverVersion := "v1.2.3"
+	server := &Server{
+		version:       serverVersion,
+		toolsManifest: []byte(`{"tool1": "description"}`),
+	}
```
kurtisvg: It's worthwhile to test against the server initialization logic itself in this situation

cmd/root.go
```diff
@@ -76,6 +76,7 @@ func NewCommand() *Command {
 	}
 
 	flags := cmd.Flags()
+	flags.StringVarP(&cmd.cfg.Version, "version", "v", "1.0.0", "Server version.")
```
kurtisvg: should remove this -- will pass in as a compile time

internal/server/api.go
```diff
@@ -15,16 +15,29 @@
...
 	"fmt"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
+	"github.com/googleapis/genai-toolbox/internal/tools"
 )
 
+func createToolsetMarshalJSON(s *Server) func(*tools.ToolsetConfig) ([]byte, error) {
```
kurtisvg: should seperate into two functions: 

1. Initialize -- take ToolsetConfig convert to ToolsetManifest
2. Then can JSON Toolsetmanfiest (maybe doesn't need it's own function).

internal/server/api.go
```diff
@@ -34,13 +47,20 @@ func apiRouter(s *Server) chi.Router {
...
+	}
+	s.manifests[""] = allToolsManifest
+	return r, nil
 }
 
 func toolsetHandler(s *Server) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		toolsetName := chi.URLParam(r, "toolsetName")
-		_, _ = w.Write([]byte(fmt.Sprintf("Stub for toolset %s manifest!", toolsetName)))
+		// toolsetName := chi.URLParam(r, "toolsetName")
```
kurtisvg: need to update this

internal/server/api.go
```diff
@@ -34,13 +47,20 @@ func apiRouter(s *Server) chi.Router {
 		r.Post("/", toolHandler(s))
 	})
 
-	return r
+	// Convert tool configs to JSON for manifest
+	defaultToolsetConfig := s.conf.ToolsetConfigs[""]
+	allToolsManifest, err := createToolsetMarshalJSON(s)(&defaultToolsetConfig)
```
kurtisvg: Only needs to execute -- closure not needed

internal/tools/toolsets.go
```diff
@@ -67,3 +67,8 @@ func (t ToolsetConfig) Initialize(toolsMap map[string]Tool) (Toolset, error) {
 	}
 	return toolset, nil
 }
+
+type ToolsetManifest struct {
+	ServerVersion string          `json:"serverVersion"`
+	ToolsManifest []*ToolManifest `json:"tools"`
```
kurtisvg: Don't think you need a pointer here

internal/tools/cloud_sql_pg.go
```diff
@@ -38,6 +38,10 @@ func (r CloudSQLPgGenericConfig) toolKind() string {
 	return CloudSQLPgSQLGenericKind
 }
 
+func (t CloudSQLPgGenericConfig) Describe() ToolManifest {
```
kurtisvg: maybe ".Manifest()"

internal/server/api.go
duwenxin99: Changed to reateToolsetManifest() which returns ToolsetManifest

internal/server/api.go
```diff
@@ -15,16 +15,27 @@
...
 	"fmt"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
+	"github.com/googleapis/genai-toolbox/internal/tools"
 )
 
+func createToolsetManifest(s *Server, c tools.ToolsetConfig) tools.ToolsetManifest {
```
kurtisvg: nit: often idiomatic to use "newToolsetManifest" for initialize functions

internal/server/api_test.go
```diff
@@ -0,0 +1,164 @@
...
+					Parameters:  []tools.Parameter{{Name: "param2", Type: "type", Description: "description2"}},
+				},
+			},
+			toolsetConfigs: tools.ToolsetConfigs{
+				"toolset1": tools.ToolsetConfig{Name: "toolset1", ToolNames: []string{"tool1", "tool2"}},
+			},
+			wantStatusCode:    http.StatusOK,
+			wantServerVersion: "1.0.0",
+			wantManifests:     map[string]tools.ToolManifest{"tool1": tool1Manifest, "tool2": tool2Manifest},
+			wantErr:           false,
```
kurtisvg: nit: it's OK to group "wants" like these into their own struct when you have a bunch of them to try and making it more readable

duwenxin99: grouped

internal/server/api.go
duwenxin99: renamed

PR #12, opened 2024-10-07
feat: add toolset configuration

duwenxin99:
Add `Toolset` implementation to the `tools` package:
- struct and configs.
- Custom `UnmarshalYAML` function.
- Initialization function that validates if tools specified for the toolset exist.

### DIFF
--- a/internal/tools/toolsets.go
+++ b/internal/tools/toolsets.go
@@ -37,18 +37,13 @@ var _ yaml.Unmarshaler = &ToolsetConfigs{}
 func (c *ToolsetConfigs) UnmarshalYAML(node *yaml.Node) error {
 	*c = make(ToolsetConfigs)
 
-	for i := 0; i < len(node.Content); i += 2 {
-		nameNode := node.Content[i]
-		toolsNode := node.Content[i+1]
+	var raw map[string][]string
+	if err := node.Decode(&raw); err != nil {
+		return err
+	}
 
-		name := nameNode.Value
-
-		// Create ToolsetConfig
-		var tools []string
-		for _, tNode := range toolsNode.Content {
-			tools = append(tools, tNode.Value)
-		}
-		(*c)[name] = ToolsetConfig{Name: name, ToolNames: tools}
+	for name, toolList := range raw {
+		(*c)[name] = ToolsetConfig{Name: name, ToolNames: toolList}
 	}
 	return nil
 }


internal/tools/toolsets.go
```diff
@@ -0,0 +1,65 @@
...
+// validate interface
+var _ yaml.Unmarshaler = &ToolsetConfigs{}
+
+func (c *ToolsetConfigs) UnmarshalYAML(node *yaml.Node) error {
+	*c = make(ToolsetConfigs)
+
+	var raw map[string][]string
+	if err := node.Decode(&raw); err != nil {
+		return err
+	}
```
kurtisvg, line 43: In sources/tools, we do a "raw" parse to get the "kind" field to determine what type we need to make it. Since we only have a single type of toolset, I think we could skip this step and set the tool?

internal/tools/toolsets.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+	"slices"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Toolset struct {
+	Name  string   `yaml:"name"`
+	Tools []string `yaml:",inline"`
```
kurtisvg: Do we only want the name of the tool, or should we retain a pointer to it?

internal/tools/toolsets.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (c *ToolsetConfigs) UnmarshalYAML(node *yaml.Node) error {
+	*c = make(ToolsetConfigs)
+
+	var raw map[string][]string
+	if err := node.Decode(&raw); err != nil {
+		return err
+	}
+
+	for name, tools := range raw {
```
kurtisvg: One thing we need to think about here is -- what makes a valid tool name? It needs to fit in a url, so should we validate it here to make sure it's okay?

internal/tools/toolsets.go
```diff
@@ -0,0 +1,65 @@
...
+	toolNames := make([]string, 0, len(toolsMap))
+	for n := range toolsMap {
+		toolNames = append(toolNames, n)
+	}
+	// Validate each declared tool name exists
+	for _, name := range t.Tools {
+		exists := slices.Contains(toolNames, name) // exists will be true
+		if !exists {
+			return t, fmt.Errorf("invalide tool name: %s", t)
+		}
```
kurtisvg: You can check the map directly:
```go
tool, ok := toolsMap["tool_name"]
// If the key exists
if !ok {
    // throw error
}
```

internal/tools/toolsets.go
```diff
@@ -0,0 +1,65 @@
...
+	// fetch existing tool names
+	toolNames := make([]string, 0, len(toolsMap))
+	for n := range toolsMap {
+		toolNames = append(toolNames, n)
+	}
+	// Validate each declared tool name exists
+	for _, name := range t.Tools {
+		exists := slices.Contains(toolNames, name) // exists will be true
+		if !exists {
+			return t, fmt.Errorf("invalide tool name: %s", t)
```
kurtisvg: is the tool name invalid or does it not exist?

internal/tools/toolsets.go
duwenxin99: I couldn't save them as pointers because Tool is an interface and not a concrete type.

kurtisvg: Why not? https://go.dev/play/p/41iIfTc2dV5

Also -- interfaces are pretty much pointers themselves: https://go.dev/tour/methods/9

duwenxin99: The problem I was encountering before was that when I save the tools as pointers, I could not retrieve tool names later without implementing a new method `GetToolName()` for the `Tool` interface. But I think that is kind of redundant and I'm not sure how reusable it is. Is that something we want to do? Or is there a workaround?

duwenxin99: I want to retrieve the tool names because we check if the tools declared for the toolset exist in the `(t Toolset) Initialize` function, matching the tools initialization implementation. But if we want to save the tools as pointers, we should check if they exist in the `UnmarshalYAML` function. Do you think it is a better approach? In that case we can get rid of the toolset initialization function as it doesn't have anything to do.

internal/tools/toolsets.go
duwenxin99: I think the tool names should be validated in the tools' `UnmarshalYAML` function. Here we should validate the toolset names to make sure it fit in a URL. I have updated the code to check for valid URL characters.

internal/tools/toolsets.go
duwenxin99: updated error msg

internal/tools/toolsets.go
duwenxin99: updated

duwenxin99:
> This is just toolset configuration, correct? I don't think I see changes to the API for getting the manifest?

Yes. I'd like to keep them in a separate PR

internal/tools/toolsets.go
```diff
@@ -0,0 +1,82 @@
...
+	for i := 0; i < len(node.Content); i += 2 {
+		nameNode := node.Content[i]
+		toolsNode := node.Content[i+1]
+
+		name := nameNode.Value
+
+		// Validat Toolset name
+		var validName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
+		if !validName.MatchString(name) {
```
kurtisvg: offline discussion: Probably should happen Initialize

internal/tools/toolsets.go
```diff
@@ -0,0 +1,82 @@
...
+	*c = make(ToolsetConfigs)
+
+	for i := 0; i < len(node.Content); i += 2 {
+		nameNode := node.Content[i]
+		toolsNode := node.Content[i+1]
+
+		name := nameNode.Value
+
+		// Validat Toolset name
+		var validName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
```
kurtisvg: This should be a global -- doesn't need to be recompiled

internal/tools/toolsets.go
duwenxin99: moved

internal/tools/toolsets.go
duwenxin99: moved to a separate util file `common.go`

internal/tools/common.go
```diff
@@ -0,0 +1,25 @@
...
+
+package tools
+
+import (
+	"regexp"
+)
+
+var validName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
+func IsValidName(s string) bool {
```
kurtisvg, line 23: nit: Does this need to be exported?

internal/tools/toolsets.go
```diff
@@ -0,0 +1,74 @@
...
+type Toolset struct {
+	Name  string  `yaml:"name"`
+	Tools []*Tool `yaml:",inline"`
+}
+
+type ToolsetConfig struct {
+	Name      string   `yaml:"name"`
+	ToolNames []string `yaml:",inline"`
+}
+type ToolsetConfigs map[string]ToolsetConfig
```
kurtisvg, line 32: nit: Do we need ToolsetConfig to also store the name of the toolset? or could this be updated to just:
```go
type ToolsetConfigs map[string][]string
```

I think if we can get the name from the map itself, we can just skip the custom UnmarhsalYAML logic

internal/tools/toolsets.go
```diff
@@ -0,0 +1,74 @@
...
+
+		name := nameNode.Value
+
+		// Create ToolsetConfig
+		var tools []string
+		for _, tNode := range toolsNode.Content {
+			tools = append(tools, tNode.Value)
+		}
+		(*c)[name] = ToolsetConfig{Name: name, ToolNames: tools}
+	}
```
kurtisvg: Even if we want to preserve the name in the ToolsetConfig, I think this can be simplified to better rely on the YAML library to do the heavy lifting:

```suggestion
	var raw map[string][]string
	if err := node.Decode(&raw); err != nil {
		return err
	}
	
	for name, toolList := range raw {
	     c[name] = ToolsetConfig{Name: name, ToolNames: toolList}
	}
```

duwenxin99: I think I misunderstood you previously about the comment so I replaced the raw parsing with iteration: "In sources/tools, we do a "raw" parse to get the "kind" field to determine what type we need to make it. Since we only have a single type of toolset, I think we could skip this step and set the tool?" What did you mean that we should set the tool directly?

kurtisvg: I think I was thinking of it like this:

```
c[name] := Tool
```

e.i. without the name in the struct itself. But if it's easier to put the name on the struct to generate the manfiest, I'm okay with using a struct as in the above sample

internal/tools/toolsets.go
duwenxin99: We need the `ToolsetConfig` type because we need to call `initialize` on the type. If we replace it with `[]string`, we will need to replace `func (t ToolsetConfig) Initialize()` with `func InitilializeToolSet(toolsetConfig []string)` because we cannot create a receiver method on `[]string`. That will deviate from the tools and sources implementations and I think is not the best option.

internal/tools/common.go
duwenxin99: updated

internal/tools/toolsets.go
kurtisvg: Ok, that makes sense to me.

internal/tools/toolsets.go
duwenxin99: I see. I think it would be easier and also cleaner to put the name in the struct so we have a more coherent implementation to match other interfaces (reasoning in another comment). I have updated to use raw parsing.

PR #13, opened 2024-10-07
feat: add preliminary parsing of parameters

kurtisvg:
This PR adds preliminary parsing of parameters. Currently it only supports 4 types: string, int, float32, and bool. Almost certainly we will need to introduce more complicated parsing configuration (to handle objects and arrays), but my initial attempts got quickly complicated, so I simplified in the short term. 

This also makes 2 breaking changes to config.yaml:
- changes "parameters" to be a list over object -- this is because parameter ordering is important, and needs to be preserved
- removed the "required" field from parameter objects -- we need to determine how to handle optional parameters in SQL queries

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -63,13 +63,50 @@ func (c *Configs) UnmarshalYAML(node *yaml.Node) error {
 	return nil
 }
 
+type Tool interface {
+	Invoke([]any) (string, error)
+	ParseParams(data map[string]any) ([]any, error)
+}
+
 type Parameter struct {
 	Name        string `yaml:"name"`
 	Type        string `yaml:"type"`
 	Description string `yaml:"description"`
-	Required    bool   `yaml:"required"`
 }
 
-type Tool interface {
-	Invoke() (string, error)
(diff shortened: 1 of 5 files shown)


internal/tools/tools.go
```diff
@@ -63,13 +63,50 @@ func (c *Configs) UnmarshalYAML(node *yaml.Node) error {
...
+type Tool interface {
+	Invoke([]any) (string, error)
+	ParseParams(data map[string]any) ([]any, error)
+}
+
 type Parameter struct {
 	Name        string `yaml:"name"`
 	Type        string `yaml:"type"`
 	Description string `yaml:"description"`
-	Required    bool   `yaml:"required"`
```
Yuan325, line 70: Side note but I'm thinking having this might help improve latency too(?). Ideally, the LLM will prompt user to provide more information before moving on to the next steps since the field is required.

kurtisvg: I'm removing this because I want to default for all parameters to be required -- I'm unsure of how to handle optional parameters in the SQL query

PR #14, opened 2024-10-13
feat: add initial cloud sql pg invocation

kurtisvg:
Makes the following changes:
- Fills in previously stubbed "Invoke()" function for the Cloud SQL Generic Tool 
- Updates API to /{tool_name}/invoke to for invocation of said tool 
- Updates response to use JSON 
- Correctly returns error messages for invalid http codes


internal/sources/cloud_sql_pg.go
```diff
@@ -45,4 +67,33 @@ var _ Source = CloudSQLPgSource{}
...
+func initConnectionPool(project, region, instance, user, pass, dbname string) (*pgxpool.Pool, error) {
+	// Configure the driver to connect to the database
+	dsn := fmt.Sprintf("user=%s password=%s dbname=%s sslmode=disable", user, pass, dbname)
+	config, err := pgxpool.ParseConfig(dsn)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse connection uri: %w", err)
+	}
+
+	// Create a new dialer with any options
+	d, err := cloudsqlconn.NewDialer(context.Background())
```
Yuan325, line 82: Should we use the automatic IAM database authentication feature? (https://pkg.go.dev/cloud.google.com/go/cloudsqlconn#readme-automatic-iam-database-authentication, https://cloud.google.com/sql/docs/postgres/iam-authentication#auto-iam-auth) That way, users need to pass only the IAM database username in a connection request from the client. Seems like it will be more secure and reliable (also it will handle auth access tokens refresh auth access).

kurtisvg: Yes, but planning to add in a separate PR. The IAM Auth is more complicated for Go than Python because of the lack of the email in the credentials source. This PR was already sizable so I will follow up on that in another PR.

PR #15, opened 2024-10-15
ci: add realease-please

Yuan325:
Release-As: 0.1.0


.github/release-please.yml
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+handleGHRelease: true
+packageName: genai-toolbox
+releaseType: simple
+versionFile: "cmd/version.txt"
+extraFiles:
+  - README.md
+  - cmd/root.go
```
kurtisvg: Why is this an "extra" file?

Yuan325: This is to indicate extra files to track. Removed since we don't use this feature in our files currently. (Will need to add this if we are utilizing the release-please annotations, e.g. to automate releases URL in README).

PR #17, opened 2024-10-16
ci: Add Dockerfile for container images

Yuan325:


Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+WORKDIR /go/src/genai-toolbox
+COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG VERSION="0.1.0"
+
+RUN go get ./...
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+    go build -ldflags "-X main.versionString=$VERSION -X main.metadataString=container"
```
kurtisvg: For both variables, we need to set the variable in cmd/root.go. Not sure main.versionString does this. Can you validate this is setting the right thing? 

I don't think to need to set versionString -- it's already set from the file in cmd. 

For metadataString, can we also set the commit hash? something like `go build -ldflags="-X main.metadataString=$(git rev-parse HEAD).container"`?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Use the latest stable golang 1.x to compile to a binary
+FROM --platform=$BUILDPLATFORM golang:1 as build
+
+WORKDIR /go/src/genai-toolbox
+COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG VERSION="0.1.0"
```
kurtisvg: shouldn't set version -- let it use the file in cmd

Dockerfile
Yuan325: Updated.

Also tested locally and verified that the go build command worked (for setting the variable of both).

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+WORKDIR /go/src/genai-toolbox
+COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
+
+RUN go get ./...
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+    go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=$(git rev-parse HEAD).container"
```
kurtisvg:
```suggestion
    go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=container.$(git rev-parse HEAD)"
```

PR #21, opened 2024-10-18
ci: update release cloudbuild

Yuan325:
Renamed file.
Update versioned release cloudbuild config file to include version tags.
Added other OS and architecture binary.

### DIFF
--- a/versioned.release.cloudbuild.yaml
+++ b/versioned.release.cloudbuild.yaml
@@ -14,6 +14,7 @@
 steps:
   - id: "Build Docker"
     name: 'gcr.io/cloud-builders/docker'
+    waitFor: ['-']
     script: |
         #!/usr/bin/env bash
         export VERSION=$(cat ./cmd/version.txt)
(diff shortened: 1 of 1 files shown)


release.cloudbuild.yaml
```diff
@@ -21,14 +21,14 @@ steps:
...
-          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=binary.linux.amd64.$(git rev-parse HEAD)" -o genai-toolbox.linux.amd64
+          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=binary.linux.amd64.$(git rev-parse HEAD)" -o genai-toolbox.linux.amd64.$SHORT_SHA
 
 # Save binary to Google Cloud Storage
 artifacts:
   objects:
     location: gs://$_BUCKET_NAME/
     paths:
-      - genai-toolbox.linux.amd64
+      - genai-toolbox.linux.amd64.$SHORT_SHA
```
kurtisvg: it would be nice if the final url was more like "$BUCKET_NAME/$VERSION/linux/amd64/toolbox"

Yuan325: Updated

release.cloudbuild.yaml
```diff
@@ -17,26 +17,28 @@ steps:
     script: |
         docker buildx build -t ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/default-images:$SHORT_SHA .
 
+  - id: "Push Docker to Google Artifact Registry"
+    name: 'gcr.io/cloud-builders/docker'
+    script: |
+        docker push ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/default-images:$SHORT_SHA
```
kurtisvg: should 'default-images' be "toolbox"? What is the final URL to pull this down?

release.cloudbuild.yaml
```diff
@@ -17,26 +17,28 @@ steps:
...
-artifacts:
-  objects:
-    location: gs://$_BUCKET_NAME/
-    paths:
-      - genai-toolbox.linux.amd64
-
-# Store images in Google Artifact Registry
-images:
-  - ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/default-images:$SHORT_SHA
+  - id: "Save binary to Google Cloud Storage"
```
kurtisvg: Why did this change from the "artifacts" section?

release.cloudbuild.yaml
```diff
@@ -17,26 +17,28 @@ steps:
...
+    name: 'gcr.io/cloud-builders/docker'
+    script: |
+        docker push ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/default-images:$SHORT_SHA
+
   - id: "Build Go binary"
     name: golang:1
     script: |
         CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
-          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=binary.linux.amd64.$(git rev-parse HEAD)" -o genai-toolbox.linux.amd64
+          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=binary.linux.amd64.$(git rev-parse HEAD)" -o toolbox.$SHORT_SHA
```
kurtisvg: Why does the binary have the sha as part of the filename?

release.cloudbuild.yaml
Yuan325: It's so that we can get the version variable from version.txt

release.cloudbuild.yaml
Yuan325: url for docker pull: us-central1-docker.pkg.dev/toolbox-testing-438616/toolbox-repo/default-images:21b28d1

release.cloudbuild.yaml
Yuan325: to differentiate which commit is this release from. Do we want to remove this? else it will just replace the previous release.

versioned.release.cloudbuild.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  - id: "Store binary to Google Cloud Storage"
+    name: "gcr.io/cloud-builders/gcloud:latest"
+    script: |
+        #!/usr/bin/env bash
+        export VERSION=v$(cat ./cmd/version.txt)
+        gcloud storage cp toolbox.linux.amd64 gs://$_BUCKET_NAME/$VERSION/linux/amd64/toolbox
+        gcloud storage cp toolbox.darwin.arm64 gs://$_BUCKET_NAME/$VERSION/darwin/arm64/toolbox
+        gcloud storage cp toolbox.darwin.amd64 gs://$_BUCKET_NAME/$VERSION/darwin/amd64/toolbox
+        gcloud storage cp toolbox.windows.amd64 gs://$_BUCKET_NAME/$VERSION/windows/amd64/toolbox
```
kurtisvg: You can speed this build up considerably if you use concurrent steps: https://cloud.google.com/build/docs/configuring-builds/configure-build-step-order

Specifically you can break out into different steps for each build, including container and set "waitFor".

You can also add a "go get" step for the build steps so they only pull down dependencies one time (as long as it's into a shared data spot) and make the build steps dependent on that

PR #22, opened 2024-10-21
feat: Add Toolbox SDK for LangChain

anubhav756:
# Overview

This PR introduces the initial Python SDK for LangChain (`toolbox-langchain`).

This package provides the necessary interfaces to integrate MCP Toolbox tools directly into LangChain agentic workflows, matching the `BaseTool` interface required by the framework.

# Architecture Diagram

image

# Broad High-level Diagram

image

# Future Context & Strategy

This PR adds the initial implementation of the LangChain SDK. This package will serve as the foundation for our broader [Python SDK strategy](https://docs.google.com/document/d/1DHbZi6bkh4JkWLMHEYgpsKU-3Dzkp3xZzUCk2ZsQCu8/edit?usp=sharing).

This package will be refactored in [subsequent work](https://docs.google.com/document/d/1NXMeFz3jqKl6HZQCnRhiwjy1fwEhOCD8lCzaKGIppQk/edit?usp=sharing). That refactor will extract the orchestration-agnostic logic (such as manifest fetching, tool invocation, auth, and parameter binding) into a new foundational `toolbox-core` package.

Following that future refactor, this `toolbox-langchain` package will be converted into a thin integrated SDK that depends on the new core library.

### DIFF
--- a/sdks/langchain/src/toolbox_langchain_sdk/utils.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/utils.py
@@ -1,4 +1,4 @@
-from typing import Any, Type
+from typing import Any, Type, Optional
 
 import yaml
 from aiohttp import ClientSession
@@ -52,7 +52,8 @@ def _schema_to_model(model_name: str, schema: list[ParameterSchema]) -> Type[Bas
     field_definitions = {}
     for field in schema:
         field_definitions[field.name] = (
-            _parse_type(field.type),
+            # TODO: Remove the hardcoded optional types once optional fields are supported by Toolbox.
+            Optional[_parse_type(field.type)],
             Field(description=field.description),
         )
 
@@ -100,6 +101,17 @@ async def _invoke_tool(
         A dictionary containing the parsed JSON response from the tool invocation.
     """
     url = f"{url}/api/tool/{tool_name}/invoke"
-    async with session.post(url, json=data) as response:
+    async with session.post(url, json=_convert_none_to_empty_string(data)) as response:
         response.raise_for_status()
         return await response.json()
+
+
+# TODO: Remove this temporary fix once optional fields are supported by Toolbox.
+def _convert_none_to_empty_string(input_dict):
+    new_dict = {}
+    for key, value in input_dict.items():
+        if value is None:
+            new_dict[key] = ""
+        else:
+            new_dict[key] = value
+    return new_dict


sdk/pyproject.toml
```diff
@@ -0,0 +1,58 @@
+[project]
```
kurtisvg: folder should probably be "sdks/langchain/"

sdk/src/toolbox_sdk/toolbox.py
```diff
@@ -0,0 +1,91 @@
...
+from langchain_core.tools import StructuredTool
+from typing import List, Optional, Type
+
+from .utils import (
+    call_tool_api,
+    schema_to_model,
+)
+
+
+def load_manifest(url: str, toolset_name: Optional[str] = None) -> dict:
```
kurtisvg: There should be a client object that we can instantiate:
```
client = ToolboxClient("https://example.com")
```

Then I would like to see "load_tool" and "load_toolset" functions. Manifest is a "toolbox maintainer" only concept -- users of the SDK shouldn't need to worry about it. They should only get tools.

sdk/src/toolbox_sdk/toolbox.py
```diff
@@ -0,0 +1,91 @@
...
+        A StructuredTool object.
+    """
+    if not isinstance(tool_schema, dict):
+        raise ValueError("tool_schema must be a dictionary")
+
+    required_keys = ["parameters", "summary", "description"]
+    missing_keys: List[str] = []
+    for key in required_keys:
+        if key not in tool_schema:
+            missing_keys.append(key)
```
kurtisvg: Should be able to use pydantic to validate this

sdk/src/toolbox_sdk/toolbox.py
```diff
@@ -0,0 +1,91 @@
...
+        raise ValueError(
+            f"Missing required fields in tool schema: {', '.join(missing_keys)}"
+        )
+
+    tool_model: Type[BaseModel] = schema_to_model(
+        model_name=tool_name, schema=tool_schema["parameters"]
+    )
+
+    tool = StructuredTool.from_function(
+        func=lambda **kwargs: call_tool_api(url, tool_name, kwargs),
```
kurtisvg: what about async?

sdk/src/toolbox_sdk/utils.py
```diff
@@ -0,0 +1,76 @@
+from typing import Any, Dict, List, Type
+
+import requests
+from pydantic import create_model, Field, BaseModel
+
+
+def schema_to_model(model_name: str, schema: Dict[str, Any]) -> Type[BaseModel]:
```
kurtisvg: if we aren't exposing this function to users, let's preface with "_"

sdk/src/toolbox_sdk/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    if type_ == "string":
+        return str
+    elif type_ == "integer":
+        return int
+    elif type_ == "number":
+        return float
+    elif type_ == "boolean":
+        return bool
+    elif type_ == "array":
+        return List
```
kurtisvg: shouldn't list have a type?

sdk/src/toolbox_sdk/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    elif type_ == "object":
+        return Dict
```
kurtisvg: same with dict

sdk/src/toolbox_sdk/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    elif type_ == "null":
+        return None
```
kurtisvg: do we need a null type?

sdks/langchain/pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[project]
+name = "toolbox_langchain_sdk"
+version="0.1.0"
```
kurtisvg: nit: 

```suggestion
version="0.0.1"
```

sdks/langchain/src/toolbox_langchain_sdk/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    Asynchronously fetches and parses the YAML data from the given URL.
+
+    Args:
+        url: The base URL to fetch the YAML from.
+
+    Returns:
+        A dictionary representing the parsed YAML data.
+    """
+    async with aiohttp.ClientSession() as session:
```
kurtisvg: I believe sessions are like a connection pool, and should be reused?

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Args:
+            tool_name: The name of the tool to load.
+        """
+        url = f"{self._url}/api/tool/{tool_name}"
+
+        yaml = await _load_yaml(url)
+
+        if "tools" in self._manifest and "tools" in yaml and tool_name in yaml["tools"]:
+            self._manifest["tools"][tool_name] = yaml["tools"][tool_name]
```
kurtisvg: Would it make sense to use PyDantic to validate the response?

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,116 @@
...
+        Fetches and parses the YAML manifest from the Toolbox service.
+
+        Args:
+            toolset_name: The name of the toolset to load.
+                Default: None. If not provided, then all the tools are loaded.
+        """
+        if toolset_name:
+            url = f"{self._url}/api/toolset/{toolset_name}"
+        else:
+            url = f"{self._url}/api/toolset"
```
kurtisvg: does this work?
```suggestion
            url = f"{self._url}/api/toolset/{toolset_name or ""}"
```

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Args:
+            tool_name: The name of the tool.
+        """
+        tool_schema: dict = self._manifest["tools"][tool_name]
+
+        class ToolSchema(BaseModel):
+            summary: str
+            description: str
+            parameters: dict
```
kurtisvg: Should this classes be outside of the function?

sdks/langchain/src/toolbox_langchain_sdk/client.py
kurtisvg: It looks like you are already doing this for ToolSchema -- I think it makes sense to do for the manfiest?

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,116 @@
...
+
+        tool_model: Type[BaseModel] = _schema_to_model(
+            model_name=tool_name, schema=tool_schema["parameters"]
+        )
+
+        async def _tool_func(**kwargs) -> dict:
+            return await _call_tool_api(self._url, tool_name, kwargs)
+
+        tool: StructuredTool = StructuredTool.from_function(
+            func=_tool_func,
```
kurtisvg: Is this func or coro?

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,116 @@
...
+            toolset_name: The name of the toolset to load.
+                Default: None. If not provided, then all the tools are loaded.
+        """
+        if toolset_name:
+            url = f"{self._url}/api/toolset/{toolset_name}"
+        else:
+            url = f"{self._url}/api/toolset"
+        self._manifest = await _load_yaml(url)
+
+    def _generate_tool(self, tool_name: str) -> None:
```
kurtisvg: This should definitely return a tool

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,116 @@
...
+            tool_name: The name of the tool to load.
+        """
+        url = f"{self._url}/api/tool/{tool_name}"
+
+        yaml = await _load_yaml(url)
+
+        if "tools" in self._manifest and "tools" in yaml and tool_name in yaml["tools"]:
+            self._manifest["tools"][tool_name] = yaml["tools"][tool_name]
+        else:
+            self._manifest = yaml
```
kurtisvg: We're creating a race condition here between coros -- maybe just return the manifest directly and we can optimize later?

sdks/langchain/src/toolbox_langchain_sdk/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    elif type_ == "integer":
+        return int
+    elif type_ == "number":
+        return float
+    elif type_ == "boolean":
+        return bool
+    elif type_ == "array":
+        return list
+    elif type_ == "object":
+        return dict
```
kurtisvg: currently we aren't supporting object types

sdks/langchain/src/toolbox_langchain_sdk/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    else:
+        raise ValueError(f"Unsupported schema type: {type_}")
+
+
+async def _call_tool_api(url: str, tool_name: str, data: dict) -> dict:
```
kurtisvg: nit: 
```suggestion
async def _invoke_tool(url: str, tool_name: str, data: dict) -> dict:
```

sdks/langchain/src/toolbox_langchain_sdk/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Returns:
+        A dictionary containing the response from the Toolbox service.
+    """
+    url = f"{url}/api/tool/{tool_name}"
+    async with aiohttp.ClientSession() as session:
+        async with session.post(url, json=_filter_none_values(data)) as response:
+            return await response.json()
+
+
+def _filter_none_values(params: dict) -> dict:
```
kurtisvg: What's the point of this function?

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: This would have an extra `/` character at the end if the `toolset_name` is not provided. I've changed this line to be a bit more concise.

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,119 @@
...
+        tool: StructuredTool = StructuredTool.from_function(
+            func=_tool_func,
+            name=tool_schema["summary"],
+            description=tool_schema["description"],
+            args_schema=tool_model,
+        )
+
+        self._tools.append(tool)
+
+    async def load_tool(self, tool_name: str) -> StructuredTool:
```
twishabansal, line 79: Need to update docstring

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,119 @@
...
+        Initializes the ToolboxClient for the Toolbox service at the given URL.
+
+        Args:
+            url: The base URL of the Toolbox service.
+        """
+        self._url: str = url
+        self._manifest: dict = {}
+        self._tools: list[StructuredTool] = []
+
+    async def _load_tool_manifest(self, tool_name: str) -> None:
```
twishabansal: What happens if we have multiple versions of the same tool?

Would we want to structure manifest like:

```
manifest_object = {
	'0.0.1' : {
		'tool1': {},
		'tool2': {},
        },
        '0.1.1' : {
	        'tool1': {},
        },
}
```

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -0,0 +1,119 @@
...
+        Loads tools from the Toolbox service, optionally filtered by toolset name.
+
+        Args:
+            toolset_name: The name of the toolset to load.
+                Default: None. If not provided, then all the tools are loaded.
+
+        Returns:
+            A list of all tools loaded from the Toolbox.
+        """
+        self._manifest = {}
```
twishabansal: Do we want to reset this on each load_toolset call? Wouldn't it be better to allow user to load multiple toolsets without losing history?

sdks/langchain/src/toolbox_langchain_sdk/utils.py
anubhav756: I had added this function initially to filter out any potentially optional `None` values from the payload, but that doesn't seem to be necessary particularly because Toolbox doesn't support optional fields yet. Removed, thanks for pointing that out.

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: N/A. We are not storing the manifest in the ToolboxClient's internal state anymore. Thanks for pointing this out though.

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: N/A. We are not storing the returned manifest in the ToolboxClient's internal state anymore. Thanks for pointing this out though.

PR #23, opened 2024-10-21
feat: add AlloyDB source and tool

Yuan325:


internal/sources/alloydb_pg.go
```diff
@@ -0,0 +1,100 @@
...
+
+// validate interface
+var _ Config = AlloyDBPgConfig{}
+
+type AlloyDBPgConfig struct {
+	Name     string `yaml:"name"`
+	Kind     string `yaml:"kind"`
+	Project  string `yaml:"project"`
+	Region   string `yaml:"region"`
+    Cluster  string `yaml:"cluster"`
```
duwenxin99: This line seems to have weird indentation

kurtisvg: I think it's tabs vs spaces -- can you make sure to run lint?

PR #24, opened 2024-10-22
chore: refactor parameters logic to allow for more complex functionality

kurtisvg:
Refactors parameter logic to allow for more complex functionality (e.g. arrays) in the future.

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -103,7 +103,6 @@ type ParameterManifest struct {
 
 // CommonParameter are default fields that are emebdding in most Parameter implementations. Embedding this stuct will give the object Name() and Type() functions.
 type CommonParameter struct {
-	// These fields are prefixed with "P" for parameter to differntiate between the methods
 	Name string `yaml:"name"`
 	Type string `yaml:"type"`
 	Desc string `yaml:"description"`
@@ -139,7 +138,7 @@ func (e ParseTypeError) Error() string {
 	return fmt.Sprintf("Error parsing parameter %q: %q not type %q", e.Name, e.Value, e.Type)
 }
 
-// NewStringParameter is a convience function for initalizing a StringParameter.
+// NewStringParameter is a convenience function for initializing a StringParameter.
 func NewStringParameter(name, desc string) *StringParameter {
 	return &StringParameter{
 		CommonParameter: CommonParameter{
@@ -166,7 +165,7 @@ func (p *StringParameter) Parse(v any) (any, error) {
 	return v, nil
 }
 
-// NewIntParameter is a convience function for initalizing a IntParameter.
+// NewIntParameter is a convenience function for initializing a IntParameter.
 func NewIntParameter(name, desc string) *IntParameter {
 	return &IntParameter{
 		CommonParameter: CommonParameter{
@@ -192,7 +191,7 @@ func (p *IntParameter) Parse(v any) (any, error) {
 	return v, nil
 }
 
-// NewFloatParameter is a convience function for initalizing a FloatParameter.
+// NewFloatParameter is a convenience function for initializing a FloatParameter.
 func NewFloatParameter(name, desc string) *FloatParameter {
 	return &FloatParameter{
 		CommonParameter: CommonParameter{
@@ -218,7 +217,7 @@ func (p *FloatParameter) Parse(v any) (any, error) {
 	return v, nil
 }
 
-// NewBooleanParameter is a convience function for initalizing a BooleanParameter.
+// NewBooleanParameter is a convenience function for initializing a BooleanParameter.
 func NewBooleanParameter(name, desc string) *BooleanParameter {
 	return &BooleanParameter{
 		CommonParameter: CommonParameter{


internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+			*c = append(*c, a)
+		default:
+			return fmt.Errorf("%q is not valid type for a parameter!", p.GetName())
+		}
+	}
+
+	return nil
+}
+
+func generateManfiests(ps []Parameter) []ParameterManifest {
```
duwenxin99:
```suggestion
func generateManifests(ps []Parameter) []ParameterManifest {
```

internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+	rtn := make([]ParameterManifest, 0, len(ps))
+	for _, p := range ps {
+		rtn = append(rtn, p.Manifest())
+	}
+	return rtn
+}
+
+// ParameterManfiest represents parameters when served as part of a ToolManifest.
```
duwenxin99:
```suggestion
// ParameterManifest represents parameters when served as part of a ToolManifest.
```

internal/tools/cloud_sql_pg.go
```diff
@@ -60,7 +60,7 @@ func (cfg CloudSQLPgGenericConfig) Initialize(srcs map[string]sources.Source) (T
 		Source:     s,
 		Statement:  cfg.Statement,
 		Parameters: cfg.Parameters,
-		manifest:   Manifest{cfg.Description, cfg.Parameters},
+		manifest:   ToolManifest{cfg.Description, generateManfiests(cfg.Parameters)},
```
duwenxin99:
```suggestion
		manifest:   ToolManifest{cfg.Description, generateManifests(cfg.Parameters)},
```

internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+type ParameterManifest struct {
+	Name        string `json:"name"`
+	Type        string `json:"type"`
+	Description string `json:"description"`
+	// Parameter   *ParameterManifest `json:"parameter,omitempty"`
+}
+
+// CommonParameter are default fields that are emebdding in most Parameter implementations. Embedding this stuct will give the object Name() and Type() functions.
+type CommonParameter struct {
+	// These fields are prefixed with "P" for parameter to differntiate between the methods
```
duwenxin99:
```suggestion
	// These fields are prefixed with "P" for parameter to differentiate between the methods
```

internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+	Name  string
+	Type  string
+	Value any
+}
+
+func (e ParseTypeError) Error() string {
+	return fmt.Sprintf("Error parsing parameter %q: %q not type %q", e.Name, e.Value, e.Type)
+}
+
+// NewStringParameter is a convience function for initalizing a StringParameter.
```
duwenxin99:
```suggestion
// NewStringParameter is a convenience function for initializing a StringParameter.
```

internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+
+func (p *IntParameter) Parse(v any) (any, error) {
+	v, ok := v.(int64)
+	if !ok {
+		return nil, &ParseTypeError{p.Name, p.Type, v}
+	}
+	return v, nil
+}
+
+// NewFloatParameter is a convience function for initalizing a FloatParameter.
```
duwenxin99:
```suggestion
// NewFloatParameter is a convenience function for initializing a FloatParameter.
```

internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+
+func (p *FloatParameter) Parse(v any) (any, error) {
+	v, ok := v.(float64)
+	if !ok {
+		return nil, &ParseTypeError{p.Name, p.Type, v}
+	}
+	return v, nil
+}
+
+// NewBooleanParameter is a convience function for initalizing a BooleanParameter.
```
duwenxin99:
```suggestion
// NewBooleanParameter is a convenience function for initializing a BooleanParameter.
```

internal/tools/parameters.go
duwenxin99:
```suggestion
```

internal/tools/parameters.go
```diff
@@ -0,0 +1,262 @@
...
+// ParameterManfiest represents parameters when served as part of a ToolManifest.
+type ParameterManifest struct {
+	Name        string `json:"name"`
+	Type        string `json:"type"`
+	Description string `json:"description"`
+	// Parameter   *ParameterManifest `json:"parameter,omitempty"`
+}
+
+// CommonParameter are default fields that are emebdding in most Parameter implementations. Embedding this stuct will give the object Name() and Type() functions.
+type CommonParameter struct {
```
duwenxin99, line 105: Why is it better than having a Parameter struct and make the `Parse` and `Manifest` methods checking for types?

internal/tools/parameters.go
duwenxin99: Actually we may want to delete this line since there are not prefixes.

internal/tools/parameters.go
kurtisvg: If we have a single Parameter stuct, every parameter will have the same fields. With a Parameter interface, each struct can have different fields as long as they provide the same signature as the interface. 

This is important because when I introduce "arrays" we'll want a new field that specifies what type the values of the arrays are.

duwenxin99: makes sense, LGTM.

PR #25, opened 2024-10-22
feat: add Postgres source and tool

Yuan325:


internal/sources/postgres.go
```diff
@@ -0,0 +1,79 @@
...
+type PostgresSource struct {
+	Name string `yaml:"name"`
+	Kind string `yaml:"kind"`
+	Pool *pgxpool.Pool
+}
+
+func initPostgresConnectionPool(host, port, user, pass, dbname string) (*pgxpool.Pool, error) {
+    // urlExample := "postgres:dd//username:password@localhost:5432/database_name"
+	i := fmt.Sprintf("postgres://%s:%s@%s:%s/%s", user, pass, host, port, dbname)
+    pool, err := pgxpool.New(context.Background(), i)
```
kurtisvg: formatting issue here

internal/sources/postgres.go
```diff
@@ -0,0 +1,79 @@
...
+
+	err = pool.Ping(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("Unable to connect successfully: %w", err)
+	}
+
+	s := PostgresSource{
+		Name: r.Name,
+		Kind: PostgresKind,
+		Pool: pool,
```
kurtisvg, line 57: There is a lot of overlap between the PG implementations and the AlloyDB / Cloud SQL ones. Any idea on how we can dedupe?

PR #26, opened 2024-10-23
feat: add support for array type parameters

kurtisvg:
Adds support for "array" type parameters. Uses a subet of JSONSchema for specification, in that arrays can be specified in the following way:

```yaml
parameters:
    name: "my_array"
    type: "array"
    description: "some description"
    items:
       type: "integer"
```

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -214,11 +214,11 @@ type IntParameter struct {
 }
 
 func (p *IntParameter) Parse(v any) (any, error) {
-	v, ok := v.(int)
+	newV, ok := v.(int)
 	if !ok {
 		return nil, &ParseTypeError{p.Name, p.Type, v}
 	}
-	return v, nil
+	return newV, nil
 }
 
 // NewFloatParameter is a convenience function for initializing a FloatParameter.
@@ -240,11 +240,11 @@ type FloatParameter struct {
 }
 
 func (p *FloatParameter) Parse(v any) (any, error) {
-	v, ok := v.(float64)
+	newV, ok := v.(float64)
 	if !ok {
 		return nil, &ParseTypeError{p.Name, p.Type, v}
 	}
-	return v, nil
+	return newV, nil
 }
 
 // NewBooleanParameter is a convenience function for initializing a BooleanParameter.
@@ -266,11 +266,11 @@ type BooleanParameter struct {
 }
 
 func (p *BooleanParameter) Parse(v any) (any, error) {
-	v, ok := v.(bool)
+	newV, ok := v.(bool)
 	if !ok {
 		return nil, &ParseTypeError{p.Name, p.Type, v}
 	}
-	return v, nil
+	return newV, nil
 }
 
 // NewArrayParameter is a convenience function for initializing an ArrayParameter.


internal/tools/parameters.go
```diff
@@ -158,11 +188,11 @@ type StringParameter struct {
 
 // Parse casts the value "v" as a "string".
 func (p *StringParameter) Parse(v any) (any, error) {
-	v, ok := v.(string)
+	newV, ok := v.(string)
```
Yuan325, line 191: Wondering why we change this to `newV` whereas the other Parse is able to just use `v`?

internal/tools/parameters.go
```diff
@@ -243,19 +273,67 @@ func (p *BooleanParameter) Parse(v any) (any, error) {
...
 }
 
-// ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
-func ParseParams(ps Parameters, data map[string]any) ([]any, error) {
-	params := []any{}
-	for _, p := range ps {
-		v, ok := data[p.GetName()]
-		if !ok {
-			return nil, fmt.Errorf("Parameter %q is required!", p.GetName())
+// NewArrayParameter is a convenience function for initializing a ArrayParameter.
```
Yuan325:
```suggestion
// NewArrayParameter is a convenience function for initializing an ArrayParameter.
```

internal/tools/parameters.go
```diff
@@ -243,19 +273,67 @@ func (p *BooleanParameter) Parse(v any) (any, error) {
...
+var _ Parameter = &ArrayParameter{}
+
+// ArrayParameter is a parameter representing the "array" type.
+type ArrayParameter struct {
+	CommonParameter `yaml:",inline"`
+	Items           Parameter `yaml:"items"`
+}
+
+func (p *ArrayParameter) UnmarshalYAML(node *yaml.Node) error {
+	//	(*p) = ArrayParameter{}
```
Yuan325: Is this comment on purpose?

internal/tools/parameters.go
```diff
@@ -243,19 +273,67 @@ func (p *BooleanParameter) Parse(v any) (any, error) {
...
+	if err := node.Decode(&p.CommonParameter); err != nil {
+		return err
+	}
+	// Find the node that reprends the "items" field name
```
Yuan325:
```suggestion
	// Find the node that represents the "items" field name
```

duwenxin99: "represents"?

internal/tools/parameters.go
```diff
@@ -22,11 +22,29 @@ import (
...
+	typeArray  = "array"
 )
 
+// ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
+func ParseParams(ps Parameters, data map[string]any) ([]any, error) {
+	params := []any{}
+	for _, p := range ps {
+		v, ok := data[p.GetName()]
+		if !ok {
+			return nil, fmt.Errorf("Parameter %q is required!", p.GetName())
```
Yuan325:
```suggestion
			return nil, fmt.Errorf("parameter %q is required!", p.GetName())
```

nit

internal/tools/parameters.go
```diff
@@ -243,19 +273,67 @@ func (p *BooleanParameter) Parse(v any) (any, error) {
...
+	arrVal, ok := v.([]any)
+	if !ok {
+		return nil, &ParseTypeError{p.Name, p.Type, arrVal}
+	}
+	rtn := make([]any, 0, len(arrVal))
+	for idx, val := range arrVal {
+		val, err := p.Items.Parse(val)
 		if err != nil {
-			return nil, fmt.Errorf("unable to parse value for %q: %w", p.GetName(), err)
+			return nil, fmt.Errorf("Unable to parse element #%d: %w", idx, err)
```
Yuan325:
```suggestion
			return nil, fmt.Errorf("unable to parse element #%d: %w", idx, err)
```

nit

internal/tools/parameters.go
kurtisvg: @Yuan325 Which other parse?

Yuan325: e.g. under the other 'Parse' func (e.g. `func (p *IntParameter) Parse(v any)`) we did `v, ok := v.(int)`

kurtisvg: Fixed, they should all be consistent now.

PR #27, opened 2024-10-23
ci: add e2e testing with cloud build

duwenxin99:


integration.cloudbuild.yaml
```diff
@@ -0,0 +1,48 @@
...
+    entrypoint: go
+    args: ['build', '-v', './...']
+  
+  - id: Run integration tests
+    name: 'gcr.io/cloud-builders/go'
+    entrypoint: go
+    args: ['test', '-race', '-v', './tests/...']
+    env:
+      - "PROJECT_ID=$PROJECT_ID"
+      - "INSTANCE_ID=$_INSTANCE_ID"
```
Yuan325, line 24: This is not specified in substitutions?

integration.cloudbuild.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - "DATABASE_ID=$_DATABASE_ID"
+      - "REGION=$_REGION"
+    secretEnv: ["cloud_sql_pg_user", "cloud_sql_pg_pass"]
+
+availableSecrets:
+  secretManager:
+    - versionName: projects/$PROJECT_ID/secrets/DB_USER/versions/1
```
Yuan325:
```suggestion
    - versionName: projects/$PROJECT_ID/secrets/DB_USER/versions/latest
```

integration.cloudbuild.yaml
```diff
@@ -0,0 +1,48 @@
...
+      env: "cloud_sql_pg_user"
+    - versionName: projects/$PROJECT_ID/secrets/DB_PASSWORD/versions/1
```
Yuan325:
```suggestion
    - versionName: projects/$PROJECT_ID/secrets/DB_PASSWORD/versions/latest
```

wondering if we want to name it as `cloud_sql_pass` instead of `DB_PASSWORD` (same with `DB_USER`). Assuming that we will be setting up more integration tests for other db in the future. (unless we want the same pass for all db)

tests/cloud_sql_integration_test.go
```diff
@@ -0,0 +1,88 @@
...
+	resp, err := http.Post("http://127.0.0.1:5000/api/tool/tool2", "application/json", nil)
+	if err != nil {
+		t.Fatalf("Error sending POST request /api/tool/tool2: %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("Error handling POST request /api/tool/tool2: %s", err)
+	}
+
+	// Test toolset manifest request
+	resp, err = http.Get("http://127.0.0.1:5000/api/toolset/")
```
Yuan325:
```suggestion
	resp, err = http.Get("http://127.0.0.1:5000/api/toolset/toolset1")
```

we need to include the toolset name? (GET /api/toolset/<name>)

Yuan325: same for the 2 error message below.

tests/cloud_sql_integration_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"context"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/googleapis/genai-toolbox/internal/server"
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	"github.com/googleapis/genai-toolbox/internal/tools"
+	_ "github.com/jackc/pgx/v5/stdlib"
```
Yuan325: remove? seems like stdlib is not used?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/googleapis/genai-toolbox
 
-go 1.22.2
+go 1.22
```
Yuan325: is this automatically generated? i wonder if its because 1.22.2 doesn't have some of the test functionality hmm..

integration.cloudbuild.yaml
duwenxin99: updated

integration.cloudbuild.yaml
duwenxin99: Instance and database IDs are specified in Cloud Build trigger substitutions

integration.cloudbuild.yaml
duwenxin99: renamed

tests/cloud_sql_integration_test.go
duwenxin99: deleted

PR #32, opened 2024-10-28
ci: add script to generate release table

Yuan325:


.ci/generate_release_table.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+FILES=("linux.amd64" "darwin.arm64" "darwin.amd64" "windows.amd64")
+
```
kurtisvg: super-nit: 
```suggestion
#! /bin/bash

FILES=("linux.amd64" "darwin.arm64" "darwin.amd64" "windows.amd64")
```

PR #33, opened 2024-10-29
fix: handle content-type correctly

kurtisvg:
Fixes a bug where the content-type wasn't being set correctly in the response type. 

cc: @anubhav756

### DIFF
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -15,9 +15,7 @@
 package server
 
 import (
-	"encoding/json"
 	"fmt"
-	"log"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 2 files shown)


internal/server/api.go
```diff
@@ -45,16 +45,10 @@ func toolsetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
-	}
-	b, err := json.Marshal(toolset.Manifest)
-	if err != nil {
-		log.Printf("unable to JSON the toolset manifest: %s", err)
-		http.Error(w, "", http.StatusInternalServerError)
+		_ = render.Render(w, r, newErrResponse(fmt.Errorf("Toolset %q does not exist", toolsetName), http.StatusNotFound))
 		return
 	}
-	_, _ = w.Write(b)
+	render.JSON(w, r, toolset.Manifest)
```
Yuan325, line 51: Does this replace the json.Marshal(toolset.Manifest)?

kurtisvg: Yes -- it does the same bits as the code removed, but needs to be sent through render in order for the `r.Use(render.SetContentType(render.ContentTypeJSON))` line to take effect.

PR #34, opened 2024-10-29
chore: deduplicate postgres implementations

Yuan325:


internal/sources/cloud_sql_pg.go
```diff
@@ -70,6 +70,10 @@ type CloudSQLPgSource struct {
 	Pool *pgxpool.Pool
 }
 
+func (r CloudSQLPgSource) GetPgConnectionPool() *pgxpool.Pool {
+    return r.Pool
+}
+
```
kurtisvg: Do you need this as a function? Can you just refer to r.Pool?

PR #37, opened 2024-10-30
docs: add detailed documentation for sources

kurtisvg:
Adds more detailed documentation for sources.

### DIFF
--- a/docs/sources/postgres.md
+++ b/docs/sources/postgres.md
@@ -1,4 +1,4 @@
-# Cloud SQL for PostgreSQL Source 
+# PostgreSQL Source 
 
 [PostgreSQL][pg-docs] is a powerful, open source object-relational database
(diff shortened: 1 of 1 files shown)


docs/sources/postgres.md
```diff
@@ -0,0 +1,42 @@
+# Cloud SQL for PostgreSQL Source
```
Yuan325:
```suggestion
# Non-Cloud PostgreSQL Source 
```

kurtisvg:
> Should we include something like "Set up Database" with the db's getting-started guide within each Requirement section? or do we expect that they already have that setup?

I'm assuming they already have a database set up, or that they can refer to the documentation linked at the beginning for getting started. If I try to document everything with setting up the database it's going to be a long section for all of those :).

PR #43, opened 2024-11-01
feat!: consolidate "x-postgres-generic" tools to "postgres-sql" tool

kurtisvg:
This PR introduces the following breaking change: The `alloydb-pg-generic`, `cloud-sql-pg-generic`, and `postgres-generic-tool` have been replaced by the `postgres-sql` tool, which works with all 3 Postgres sources.

If you were using of the the previous tools, you will need to update it as follows:
```diff
example_tool:
-    kind: cloud-sql-pg-generic
+    kind: postgres-sql
     source: my-cloud-sql-pg-instance
     description: some description
        statement: |
            SELECT * FROM SQL_STATEMENT;
        parameters:
        - name: country
          type: string
          description: some description
```

I'm proposing this change for the following reasons:
1. It provides greater flexibility between postgres-compatible sources -- you can change between "postgres" and "alloydb-postgres" without issue
2. The name "postgres-sql" is more clear that "postgres-generic" -- it indicates it's a tool that runs SQL on the source 
3. It's easier for us to maintain feature compatibility across a single "postgres-sql" tool

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -19,7 +19,7 @@ import (
 )
 
 type ToolConfig interface {
-	ToolKind() string
+	ToolConfigKind() string
 	Initialize(map[string]sources.Source) (Tool, error)
 }
 
--- a/internal/tools/postgressql/postgressql.go
+++ b/internal/tools/postgressql/postgressql.go
@@ -52,7 +52,7 @@ type Config struct {
 // validate interface
 var _ tools.ToolConfig = Config{}
 
-func (cfg Config) ToolKind() string {
+func (cfg Config) ToolConfigKind() string {
 	return ToolKind
 }
 


docs/tools/postgres-sql.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Reference
+
+| **field**   | **type** | **required** | **description**                                                                                     |
+|-------------|:--------:|:------------:|-----------------------------------------------------------------------------------------------------|
+| kind        |  string  |     true     | Must be "postgres-generic".                                                                         |
+| source      |  string  |     true     | Name of the source the SQL should execute on.                                                       |
+| description |  string  |     true     | Port to connect to (e.g. "5432")                                                                    |
+| statement   |  string  |     true     | SQL statement to execute on.                                                                        |
+| parameters  |  string  |     true     | List of [parameters](README.md#specifying-parameters) that will be inserted into the SQL statement. |
```
Yuan325: parameters is a list/array type?

docs/tools/README.md
```diff
@@ -0,0 +1,116 @@
+# Tools
+
+Tools represent an action your agent can take, such as running a SQL statement.
+You can define Tools as a map in the `sources` section of your `tools.yaml`
+file. Typically, a tool will require a source to act on:
+
+```yaml
+tools:
+ search_flights_by_number:
+    kind: cloud-sql-postgres-generic
```
Yuan325:
```suggestion
    kind: postgres-sql
```

Also made this suggestion in another PR.

docs/tools/README.md
```diff
@@ -0,0 +1,116 @@
...
+      - name: flight_number
+        type: string
+        description: 1 to 4 digit number
+```
+
+## Kinds of Tools
+
+We currently support the following types of kinds of tools:
+
+* [alloydb-postgres](./alloydb-pg.md) - Run a SQL statement against an AlloyDB
```
Yuan325: This will probably be updated to only`postgres-sql`?

internal/tools/tools.go
```diff
@@ -15,73 +15,22 @@
 package tools
 
 import (
-	"fmt"
-
 	"github.com/googleapis/genai-toolbox/internal/sources"
-	"gopkg.in/yaml.v3"
 )
 
 type Config interface {
```
Yuan325: nit: maybe we can make this consistent by updating this to `ToolConfig`, similar to `SourceConfig`? (or change `SourceConfig` to `Config`.

internal/tools/tools.go
```diff
@@ -15,73 +15,22 @@
...
-	toolKind() string
+	ToolKind() string
```
Yuan325: nit: same with this, in source we call it `SourceConfigKind`

internal/tools/tools.go
kurtisvg: Agreed

docs/tools/README.md
```diff
@@ -46,11 +46,7 @@ tools:
 
 We currently support the following types of kinds of tools:
 
-* [alloydb-postgres](./alloydb-pg.md) - Run a SQL statement against an AlloyDB
-  for PostgreSQL instance.
-* [cloud-sql-postgres](./cloud-sql-pg.md) - Run a SQL statement against a Cloud
-  SQL for PostgreSQL instance.
-* [postgres-generic](./postgres.md) - Run a PostgreSQL statement against a
+* [postgres-sql](./sql.md) - Run a PostgreSQL statement against a
```
Yuan325, line 49:
```suggestion
* [postgres-sql](./postgres-sql.md) - Run a PostgreSQL statement against a
```

kurtisvg: Fixed in #44

PR #46, opened 2024-11-04
feat: add support for structured logs

Yuan325:
Logging support 4 different types of logging (debug, info, warn, error). The default logging level is Info.

User will be able to set flag for log level (allowed values: "debug", "info", "warn", "error"), example:
`go run . --log-level debug`

User will be able to set flag for logging format (allowed values: "standard", "JSON"), example:
`go run . --logging-format json`

Example of standard logger:
`2024-11-12T15:08:11.451147-08:00 ERROR "unable to parse tool file at \"tools.yaml\": \"cloud-sql-postgres1\" is not a valid kind of data source"`

`2024-11-12T15:08:11.451377-08:00 INFO "Initalized 0 sources.\n"`

Example of structured logger:

```
{
  "timestamp":"2024-11-04T16:45:11.987299-08:00",
  "severity":"ERROR",
  "logging.googleapis.com/sourceLocation":{
    "function":"github.com/googleapis/genai-toolbox/internal/log.(*StructuredLogger).Errorf",
    "file":"/Users/yuanteoh/github/genai-toolbox/internal/log/log.go","line":157
  },
  "message":"unable to parse tool file at \"tools.yaml\": \"cloud-sql-postgres1\" is not a valid kind of data source"
}
```

```
{
  "timestamp":"2024-11-04T16:45:11.987562-08:00",
  "severity":"INFO",
  "logging.googleapis.com/sourceLocation":{
    "function":"github.com/googleapis/genai-toolbox/internal/log.(*StructuredLogger).Infof",
    "file":"/Users/yuanteoh/github/genai-toolbox/internal/log/log.go","line":147
  },
  "message":"Initalized 0 sources.\n"
}
```

### DIFF
--- a/docs/sources/alloydb-pg.md
+++ b/docs/sources/alloydb-pg.md
@@ -5,7 +5,11 @@ database for demanding transactional workloads. It provides enterprise-grade
 performance and availability while maintaining 100% compatibility with
 open-source PostgreSQL.
 
+If you are new to AlloyDB for PostgreSQL, you can try [creating and connecting
+to a database by following these instructions][alloydb-quickstart].
+
 [alloydb-docs]: https://cloud.google.com/alloydb/docs
+[alloydb-quickstart]: https://cloud.google.com/alloydb/docs/quickstart/create-and-connect
 
 ## Requirements 
 
--- a/docs/sources/cloud-sql-pg.md
+++ b/docs/sources/cloud-sql-pg.md
@@ -4,7 +4,11 @@
 that helps you set up, maintain, manage, and administer your PostgreSQL
 relational databases on Google Cloud Platform.
 
+If you are new to Cloud SQL for PostgreSQL, you can try [creating and connecting
+to a database by following these instructions][csql-pg-quickstart].
+
 [csql-pg-docs]: https://cloud.google.com/sql/docs/postgres
+[csql-pg-quickstart]: https://cloud.google.com/sql/docs/postgres/connect-instance-local-computer
 
 ## Requirements 
 


kurtisvg:
Some high level feedback:

- I think the `--structured-logs` might make more sense as `--logging-format=` with options for `standard` and `JSON`.
- With the standard logging, I would expect the output to be more like `${TIME} [${LEVEL}] ${MESSAGE}` instead of something like`time=${TIME} level=${LEVEL} message=${MESSAGE}` 
- For log_level, I'd like to see something like `--log-level=WARN` so we can set the log level up as well as down

internal/server/server.go
```diff
@@ -57,7 +59,7 @@ func NewServer(cfg ServerConfig) (*Server, error) {
 		}
 		sourcesMap[name] = s
 	}
-	fmt.Printf("Initalized %d sources.\n", len(sourcesMap))
+	l.Infof(fmt.Sprintf("Initalized %d sources.\n", len(sourcesMap)))
```
kurtisvg: nit: "l" has a longer scope, I think you could use "log" instead. see https://google.github.io/styleguide/go/decisions#variable-names on when longer names are okay

toolbox/toolbox.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package toolbox
+
+// Logger is the interface used throughout the project for logging.
+type Logger interface {
+	// Debug is for reporting additional information about internal operations.
+	Debugf(format string, args ...interface{})
```
kurtisvg: It looks like the slog struct uses just "Debug" and not "Debugf"?

toolbox/toolbox.go
```diff
@@ -0,0 +1,27 @@
...
+	// Info is for reporting informational messages.
+	Infof(format string, args ...interface{})
```
kurtisvg: Should we add context's here?

Yuan325: Context is needed for tracing. Will be adding this when we support trace.

cmd/root.go
```diff
@@ -61,23 +63,29 @@ type Command struct {
...
+		logger: logger,
 	}
 
 	flags := cmd.Flags()
 	flags.StringVarP(&cmd.cfg.Address, "address", "a", "127.0.0.1", "Address of the interface the server will listen on.")
 	flags.IntVarP(&cmd.cfg.Port, "port", "p", 5000, "Port the server will listen on.")
 
-	flags.StringVar(&cmd.tools_file, "tools_file", "tools.yaml", "File path specifying the tool configuration")
+	flags.StringVar(&cmd.tools_file, "tools_file", "tools.yaml", "File path specifying the tool configuration.")
+	flags.VarP(&cmd.cfg.LogLevel, "log-level", "l", "Specify the log level. Allowed: 'DEBUG', 'INFO', 'WARN', 'ERROR'.")
```
kurtisvg: nit:
```suggestion
	flags.VarP(&cmd.cfg.LogLevel, "log-level", "ll", "Specify the log level. Allowed: 'DEBUG', 'INFO', 'WARN', 'ERROR'.")
```

cmd/root.go
```diff
@@ -104,24 +112,35 @@ func run(cmd *Command) error {
...
+		cmd.logger.Error(errMsg.Error())
 	}
 
 	// run server
-	s, err := server.NewServer(cmd.cfg)
+	s, err := server.NewServer(cmd.cfg, cmd.logger)
 	if err != nil {
-		return fmt.Errorf("toolbox failed to start with the following error: %w", err)
+        errMsg := fmt.Errorf("toolbox failed to start with the following error: %w", err)
+		cmd.logger.Error(errMsg.Error())
```
kurtisvg: This is a behavior change -- it needs to return the error after logging this message

cmd/root.go
```diff
@@ -104,24 +112,35 @@ func run(cmd *Command) error {
...
 	}
 	err = s.ListenAndServe(ctx)
 	if err != nil {
-		return fmt.Errorf("toolbox crashed with the following error: %w", err)
+        errMsg := fmt.Errorf("toolbox crashed with the following error: %w", err)
+		cmd.logger.Error(errMsg.Error())
```
kurtisvg: same here -- need to return the error after logging this message

cmd/root.go
```diff
@@ -61,23 +63,29 @@ type Command struct {
...
 
 	cfg        server.ServerConfig
+	logger     toolbox.Logger
 	tools_file string
 }
 
 // NewCommand returns a Command object representing an invocation of the CLI.
 func NewCommand() *Command {
+	// default log level to info
+	logger := log.NewStdLogger(os.Stdout, os.Stderr, "")
```
kurtisvg: I'm not sure we should initialize this until we've processed the flags

internal/log/handler.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (h *ValueTextHandler) WithGroup(name string) slog.Handler {
+	return &ValueTextHandler{h: h.h.WithGroup(name), out: h.out, mu: h.mu}
+}
+
+// Handle formats its argument [Record] as a single line of space-separated values.
+func (h *ValueTextHandler) Handle(ctx context.Context, r slog.Record) error {
+	buf := make([]byte, 0, 1024)
+
+	// time
```
kurtisvg: nit: suggest adding a comment somewhere saying what the final out put is

internal/log/log.go
```diff
@@ -0,0 +1,169 @@
...
+func severityToLevel(s string) slog.Level {
+	switch strings.ToLower(s) {
+	case "debug":
+		return slog.LevelDebug
+	case "info":
+		return slog.LevelInfo
+	case "warn":
+		return slog.LevelWarn
+	default:
+		return slog.LevelError
```
kurtisvg: We should explicitly check for "error" and throw an error if it doesn't match any of out preferred settings.

internal/log/log.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+// Returns severity string based on level.
+func levelToSeverity(s string) string {
+	switch s {
+	case slog.LevelDebug.String():
+		return "DEBUG"
```
kurtisvg: nit: suggest making these values constant and sharing them between here and severityToLevel

internal/log/log.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return &StructuredLogger{outLogger: slog.New(outHandler), errLogger: slog.New(errHandler)}
+}
+
+// Debug logs debug messages
+func (sl *StructuredLogger) Debug(msg string, keysAndValues ...interface{}) {
+	sl.outLogger.Debug(msg, keysAndValues...)
+}
+
+// Info logs debug messages
```
kurtisvg:
```suggestion
// Info logs info messages
```

internal/server/api_test.go
```diff
@@ -78,7 +80,8 @@ func TestToolsetEndpoint(t *testing.T) {
 		toolsets[name] = m
 	}
 
-	server := Server{conf: ServerConfig{}, tools: toolsMap, toolsets: toolsets}
+    testLogger := log.NewStdLogger(os.Stdout, os.Stderr, "")
+	server := Server{conf: ServerConfig{}, logger: testLogger, tools: toolsMap, toolsets: toolsets}
```
kurtisvg: indentation seems off?

internal/log/log.go
Yuan325: This is already checked in config.go during the `func (s *StringLevel) Set(v string)` function. Do we still have to do this again here?

cmd/root.go
Yuan325: Removed this shorthand. Go only allow single letter shorthand.

PR #55, opened 2024-11-08
feat(sdk): make ClientSession optional when initializing ToolboxClient

anubhav756:


sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -8,16 +9,41 @@
...
         """
         self._url: str = url
-        self._session = session
+
+        if session:
+            self._is_session_managed: bool = False
+            self._session: ClientSession = session
+        else:
+            self._is_session_managed: bool = True
+            self._session: ClientSession = ClientSession()
```
kurtisvg: nit: also, instead of "is_session_managed" it's probably better to say something like "_should_close_session"
```suggestion
        self._is_session_managed: bool = session is not None
        self._session: ClientSession = session or ClientSession()
```

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -8,16 +9,41 @@
...
+        # lifecycle.
+        if self._session and self._is_session_managed:
+            await self._session.close()
+
+    def __del__(self):
+        loop = asyncio.get_event_loop()
+        if loop.is_running():
+            loop.create_task(self.close())
+        else:
+            loop.run_until_complete(self.close())
```
kurtisvg: You want to wrap this in the try/catch "pass" because if the event loop is no longer running it'll throw an error (but at that point the Session should have been closed already anyway).

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: Thanks for the suggestion!

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: Thanks for the suggestion :)

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -8,20 +9,50 @@
...
+
+    def __del__(self):
+        try:
+            loop = asyncio.get_event_loop()
+            if loop.is_running():
+                loop.create_task(self.close())
+            else:
+                loop.run_until_complete(self.close())
+        except Exception:
+            # We "pass" assuming that the exception is thrown because  the event loop is no longer running, but at that point the Session should have been closed already anyway.
```
kurtisvg: nit: This comment seems like the lint should have wrapped it?

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -8,20 +9,50 @@
...
         """
         self._url: str = url
-        self._session = session
+
+        if session:
+            self._should_close_session: bool = False
+            self._session: ClientSession = session
+        else:
+            self._should_close_session: bool = True
+            self._session: ClientSession = ClientSession()
```
kurtisvg: nit: I think you can still simplify this 
```suggestion
      self._should_close_session: bool = session != None
      self._session: ClientSession = session or ClientSession()

```

anubhav756: Thanks for the suggestion! :)

PR #66, opened 2024-11-12
ci: improve releasing

kurtisvg:
Does a few things to improve releasing process:
- Moves file to .ci folder
- Updates release script to check for version
- Uses printf instead of "echo" to work better between linux/mac
- removes comments about non-mac (worked on linux for me)

I've also updated the tigger to run on new tags.

### DIFF
--- a/.ci/versioned.release.cloudbuild.yaml
+++ b/.ci/versioned.release.cloudbuild.yaml
@@ -18,7 +18,7 @@ steps:
     script: |
         #!/usr/bin/env bash
         export VERSION=$(cat ./cmd/version.txt)
-        docker buildx build -t ${_DOCKER_URI}:$REF_NAME -t ${_DOCKER_URI}:latest .
+        docker buildx build --build-arg METADATA_TAGS=$(git rev-parse HEAD) -t ${_DOCKER_URI}:$REF_NAME -t ${_DOCKER_URI}:latest .
 
   - id: "push-docker"
     name: "gcr.io/cloud-builders/docker"
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,10 +20,11 @@ COPY . .
 
 ARG TARGETOS
 ARG TARGETARCH
+ARG METADATA_TAGS=dev
 
 RUN go get ./...
 RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
-    go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=container.$(git rev-parse HEAD)"
+    go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=container.${METADATA_TAGS}"
 
 # Final Stage
 FROM gcr.io/distroless/static:nonroot


.ci/generate_release_table.sh
```diff
@@ -20,15 +28,13 @@ do
...
 
     # Calculate the SHA256 checksum of the file
-    # Running from mac
     SHA256=$(shasum -a 256 toolbox | awk '{print $1}')
-    # Running from non-mac
-    # SHA256=$(sha256sum toolbox | awk '{print $1}')
 
     # Write the table row
     output_string+="| [$OS/$ARCH]($URL)   | $SHA256 |\n"
+    output_string+="| %-93s | %-65s |\n"
```
Yuan325: This will create a new row. Is it intentional?

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -124,22 +131,24 @@ steps:
...
-  dynamic_substitutions: true
+  dynamicSubstitutions: true
   logging: CLOUD_LOGGING_ONLY # Necessary for custom service account
 
 substitutions:
   _REGION: us-central1
   _AR_HOSTNAME: ${_REGION}-docker.pkg.dev
   _AR_REPO_NAME: toolbox
   _BUCKET_NAME: genai-toolbox
+  _DOCKER_URI: ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/toolbox
```
Yuan325:
```suggestion
  _DOCKER_URI: ${_AR_HOSTNAME}/${PROJECT_ID}/${_AR_REPO_NAME}/toolbox
```

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -12,40 +12,44 @@
...
-    name: 'gcr.io/cloud-builders/docker'
+  - id: "build-docker"
+    name: "gcr.io/cloud-builders/docker"
     waitFor: ['-']
     script: |
         #!/usr/bin/env bash
         export VERSION=$(cat ./cmd/version.txt)
-        docker buildx build -t ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/toolbox:$VERSION \
-            -t ${_AR_HOSTNAME}/$PROJECT_ID/${_AR_REPO_NAME}/toolbox:latest .
+        docker buildx build -t ${_DOCKER_URI}:$REF_NAME -t ${_DOCKER_URI}:latest .
```
Yuan325: Tested this locally manually and $REF_NAME doesn't work. error with "invalid reference format"

.ci/generate_release_table.sh
kurtisvg: Somehow I committed and earlier version. Should be fixed now.

.ci/versioned.release.cloudbuild.yaml
kurtisvg: It's listed here: https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions

It says "When starting a build using gcloud builds submit, you can specify variables that would normally come from triggered builds with the [--substitutions argument](https://cloud.google.com/sdk/gcloud/reference/builds/submit).". Did you try that?

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -55,20 +58,21 @@ steps:
         CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
           go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=binary.linux.amd64.$VERSION.$(git rev-parse HEAD)" -o toolbox.linux.amd64
```
Yuan325, line 59: This line have the `$(git rev-parse HEAD)` substitution too. Should we replace all these with the $REF_NAME?

.ci/versioned.release.cloudbuild.yaml
Yuan325: Yeap, tried that. If we want to be able to run from local, we'll have to update the Dockerfile as well. We have this substitution `$(git rev-parse HEAD)`.

PR #73, opened 2024-11-14
ci(langchain): Added auto lint and integration test config

twishabansal:
This PR adds CI workflows for linting and integration tests to the Langchain SDK.

Linting is failing due to type checking errors in the SDK (to be fixed separately).
Integration tests are failing and need updates to match the latest SDK code (also to be fixed separately).

### DIFF
--- a/.github/workflows/lint-langchain.yaml
+++ b/.github/workflows/lint-langchain.yaml
@@ -51,4 +51,4 @@ jobs:
       - name: Run type-check
         env:
           MYPYPATH: 'sdks/langchain/src'
-        run: mypy --install-types --non-interactive sdks/langchain
+        run: mypy --install-types --non-interactive --explicit-package-bases sdks/langchain


twishabansal:
Created a new PR: https://github.com/googleapis/genai-toolbox/pull/94 to pin requirement versions correctly.

sdks/langchain/integration.cloudbuild.yaml
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+  - id: Install module (and test requirements)
+    name: 'python:${_VERSION}'
+    args:
+      - install
+      - 'sdks/langchain[test]'
+      - '--user'
```
kurtisvg: I think we need to install the requirements.txt here first, then the test dependencies.

PR #83, opened 2024-11-19
feat: add authenticated parameters to tools

duwenxin99:
Add auth params parsing implementation and tool calling API changes to support auth header.

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -200,9 +200,10 @@ func (ps Parameters) Manifest() []ParameterManifest {
 
 // ParameterManifest represents parameters when served as part of a ToolManifest.
 type ParameterManifest struct {
-	Name        string `json:"name"`
-	Type        string `json:"type"`
-	Description string `json:"description"`
+	Name        string   `json:"name"`
+	Type        string   `json:"type"`
+	Description string   `json:"description"`
+	AuthSources []string `json:"authSources"`
 	// Parameter   *ParameterManifest `json:"parameter,omitempty"`
 }
 
@@ -226,10 +227,16 @@ func (p *CommonParameter) GetType() string {
 
 // GetType returns the type specified for the Parameter.
 func (p *CommonParameter) Manifest() ParameterManifest {
+	// only list ParamAuthSource names (without fields) in manifest
+	authNames := make([]string, len(p.AuthSources))
+	for i, a := range p.AuthSources {
+		authNames[i] = a.Name
+	}
 	return ParameterManifest{
 		Name:        p.Name,
 		Type:        p.Type,
 		Description: p.Desc,
+		AuthSources: authNames,
 	}
 }
 


cmd/root_test.go
```diff
@@ -132,6 +132,40 @@ func TestAddrPort(t *testing.T) {
 		})
 	}
 }
+func TestToolFileFlag(t *testing.T) {
```
kurtisvg: nit: rebase?

internal/server/api.go
```diff
@@ -74,6 +76,22 @@ func toolGetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	render.JSON(w, r, m)
 }
 
+func parseAuthHeader(s *Server, h http.Header) (map[auth.AuthSource]string, error) {
+	authInfo := make(map[auth.AuthSource]string)
+	for name, token := range h {
```
kurtisvg: Optimize: loop through authsources rather than headers

internal/server/api_test.go
```diff
@@ -37,12 +38,12 @@ func (t MockTool) Invoke([]any) (string, error) {
 	return "", nil
 }
 
-func (t MockTool) ParseParams(data map[string]any, claims map[string]any) ([]any, error) {
+func (t MockTool) ParseParams(data map[string]any, claims map[string]map[string]interface{}) ([]any, error) {
```
kurtisvg: nit: 
```suggestion
func (t MockTool) ParseParams(data map[string]any, claims map[string]map[string]any) ([]any, error) {
```

kurtisvg: Add a comment to explain what claims is (authSourceName -> map of claims)

internal/tools/parameters.go
```diff
@@ -29,12 +29,33 @@ const (
...
 // ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
-func ParseParams(ps Parameters, data map[string]any, claims map[string]any) ([]any, error) {
+func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[string]any) ([]any, error) {
+	// claimsMap maps an AuthSource name to its decoded claims (user info)
 	params := []any{}
 	for _, p := range ps {
-		v, ok := data[p.GetName()]
-		if !ok {
-			return nil, fmt.Errorf("parameter %q is required!", p.GetName())
+		var v interface{}
```
kurtisvg: What is v in this case?

also nit: should prefer `any` over `interface{}`
```suggestion
		var v any
```

internal/tools/parameters.go
```diff
@@ -29,12 +29,33 @@ const (
...
+		var ok bool
```
kurtisvg: nit:
```suggestion
		var v interface{}, ok bool
```

internal/tools/parameters.go
```diff
@@ -29,12 +29,33 @@ const (
...
+						if !ok {
+							return nil, fmt.Errorf("claims returned from authentication do not contain field %q", p.GetName())
+						}
+						break
+					}
+				}
+			}
+			if v == nil {
+				return nil, fmt.Errorf("missing authentication header for parameter %q", p.GetName())
+			}
```
kurtisvg: I'm not sure I follow this logic -- can we add some comments?

internal/tools/parameters.go
```diff
@@ -29,12 +29,33 @@ const (
...
+			for _, a := range authSources {
+				for authName, claims := range claimsMap {
+					if a.Name == authName {
+						v, ok = claims[a.Field]
+						if !ok {
+							return nil, fmt.Errorf("claims returned from authentication do not contain field %q", p.GetName())
+						}
+						break
+					}
+				}
```
kurtisvg: Why is this looping? Can it just get the claims from claimsMap by authSource name?

```suggestion


func parseFromAuthSource(authSources, claimsMap) {
    for _, a := range authSources {
       claims, ok := claimsMap[authName]
       if !ok {
           // not validated for this authsource, skip to next one
	  continue
       }     
       v, ok := claims[a.Field] 
       if !ok {
           // not validated for this authsource, skip to next one
           return nil, fmt.Errorf("no field named X in claim")  
       }
     return nil, fmt.Errorf("no auth sources authenticated")
   }
}
```

internal/server/api.go
```diff
@@ -74,6 +75,17 @@ func toolGetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	render.JSON(w, r, m)
 }
 
+func parseAuthHeader(s *Server, h http.Header) (map[auth.AuthSource]string, error) {
+	// parse auth header into an {auth sources --> auth token} map
+	authInfo := make(map[auth.AuthSource]string)
+	for name := range s.authSources {
+		if token, ok := h[name+"_token"]; ok {
+			authInfo[s.authSources[name]] = token[0]
+		}
```
kurtisvg: use http.Header.Get -- https://pkg.go.dev/net/http#Header

internal/server/api.go
```diff
@@ -84,15 +96,28 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+		err := fmt.Errorf("failure parsing auth header: %w", err)
+		_ = render.Render(w, r, newErrResponse(err, http.StatusUnauthorized))
+		return
+	}
+	claimsMap, err := tool.Authenticate(authInfo)
+	if err != nil {
+		err := fmt.Errorf("tool authentication failure: %w", err)
+		_ = render.Render(w, r, newErrResponse(err, http.StatusUnauthorized))
+		return
+	}
```
kurtisvg:
```go
claimsFromAuth := make(map[auth.AuthSource]string)
for _, aS := range s.authSources {
    claims, ok := aS.GetClaimsFrom(r.Header)
    if ok {
        claimsFromAuth[aS.Name] = claims
    }
}


// authZ (added later)
// this can just check if auth_required source names are in claimsFromAuth, and return false if not
isAuthorized := tool.Authorized(claimsFromAuth)
If !isAuthorized {
    // return 401
}
```

internal/server/api_test.go
```diff
@@ -37,12 +38,13 @@ func (t MockTool) Invoke([]any) (string, error) {
...
-	return tools.ParseParams(t.Params, data, claims)
+// claims is a map of user info decoded from an auth token
+func (t MockTool) ParseParams(data map[string]any, claimsMap map[string]map[string]any) ([]any, error) {
+	return tools.ParseParams(t.Params, data, claimsMap)
 }
 
-func (t MockTool) Authenticate(authSource string, authToken string) (map[string]any, error) {
-	return make(map[string]any), nil
+func (t MockTool) Authenticate(map[auth.AuthSource]string) (map[string]map[string]any, error) {
+	return nil, nil
```
kurtisvg: We should make sure to add TODO's when we stub functions out like this

internal/tools/postgressql/postgressql.go
```diff
@@ -124,14 +125,20 @@ func (t Tool) Invoke(params []any) (string, error) {
...
+func (t Tool) Authenticate(authInfo map[auth.AuthSource]string) (map[string]map[string]any, error) {
+	allClaims := make(map[string]map[string]any)
+	for authSource, token := range authInfo {
+		claims, err := authSource.Verify(token)
+		if err != nil {
+			return allClaims, fmt.Errorf("auth source %s verification failure: %w", authSource.GetName(), err)
+		}
+		allClaims[authSource.GetName()] = claims
+	}
+	return allClaims, nil
```
kurtisvg: This is probably the same for every tool -- should it be a UtilityModule like "tools.ParseParams"?

internal/tools/parameters.go
duwenxin99: v is the parameter value in this case

internal/tools/parameters.go
duwenxin99: I tried it and this syntax doesn't work. I think go requires variables to be of the same type to declare in one line: https://go.dev/tour/basics/8

internal/server/api.go
duwenxin99: updated

internal/tools/postgressql/postgressql.go
duwenxin99: updated

internal/tools/parameters.go
duwenxin99: updated

duwenxin99: updated

internal/server/api.go
```diff
@@ -84,15 +84,34 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var data map[string]interface{}
+	// Tool authentication
+	// Maps AthSources --> claims
```
kurtisvg:
```suggestion
	// claimsFromAuth maps the name of the authsource to the claims retrieved from it. 
```

internal/tools/common.go
```diff
@@ -23,3 +23,24 @@ var validName = regexp.MustCompile(`^[a-zA-Z0-9_-]*$`)
 func IsValidName(s string) bool {
 	return validName.MatchString(s)
 }
+
+func Authorized(parameters []Parameter, claims map[string]map[string]any) bool {
```
kurtisvg: Need a docstring here. What's this function do?

internal/server/api.go
```diff
@@ -84,15 +84,34 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+	for _, aS := range s.authSources {
+		claims, err := aS.GetClaimsFromHeader(r.Header)
+		if err != nil {
+			err := fmt.Errorf("Failure getting claims from header: %w", err)
+			_ = render.Render(w, r, newErrResponse(err, http.StatusBadRequest))
+			return
+		}
+		claimsFromAuth[aS.GetName()] = claims
+	}
+	isAuthorized := tool.Authorized(claimsFromAuth)
```
kurtisvg: Is this for tool based authorization? should that be a later PR?

duwenxin99: removed

internal/tools/common.go
duwenxin99: removed

internal/tools/parameters.go
```diff
@@ -68,18 +68,49 @@ func (p ParamValues) AsMapByOrderedKeys() map[string]interface{} {
...
-			return nil, fmt.Errorf("parameter %q is required!", p.GetName())
+		if paramAuthSources == nil {
+			// parse non auth-required parameter
+			var ok bool
+			v, ok = data[name]
+			if !ok {
+				return nil, fmt.Errorf("parameter %q is required", name)
+			}
+		} else {
+			// parse auth-required parameter
```
```suggestion
			// parse authenticated parameter
```

PR #90, opened 2024-11-23
feat: add Spanner source and tool

Yuan325:
Add Spanner source and tool.

Spanner source is initialize with the following config:
```
sources:
    my-spanner-source:
        kind: spanner
        project: my-project-name
        instance: my-instance-name
        database: my_db
        # dialect: postgresql # The default dialect is google_standard_sql.
```

Spanner tool (with gsql dialect) is initialize with the following config.
```
tools:
    get_flight_by_id:
        kind: spanner
        source: my-cloud-sql-source
        description: >
            Use this tool to list all airports matching search criteria. Takes 
            at least one of country, city, name, or all and returns all matching
            airports. The agent can decide to return the results directly to 
            the user.
        statement: "SELECT * FROM flights WHERE id = @id"
        parameters:
        - name: id
          type: int
          description: 'id' represents the unique ID for each flight. 
```

Spanner tool (with postgresql dialect) is initialize with the following config.
```
tools:
    get_flight_by_id:
        kind: spanner
        source: my-cloud-sql-source
        description: >
            Use this tool to list all airports matching search criteria. Takes 
            at least one of country, city, name, or all and returns all matching
            airports. The agent can decide to return the results directly to 
            the user.
        statement: "SELECT * FROM flights WHERE id = $1"
        parameters:
        - name: id
          type: int
          description: 'id' represents the unique ID for each flight. 
```

Note: the only difference in config for both dialects is the sql statement.

### DIFF
--- a/internal/sources/dialect.go
+++ b/internal/sources/dialect.go
@@ -21,6 +21,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// Dialect represents the dialect type of a database.
 type Dialect string
 
 func (i *Dialect) String() string {


internal/sources/dialect.go
```diff
@@ -0,0 +1,45 @@
...
+package sources
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Dialect string
```
kurtisvg, line 25: nit: add comment

internal/tools/parameters.go
```diff
@@ -56,6 +56,18 @@ func (p ParamValues) AsMap() map[string]interface{} {
 	return params
 }
 
+// AsMapByOrderedKeys returns a map of ParamValue's order keys to values.
+// Example of keys: p1, p2, p3,...
```
```suggestion
// AsMapByOrderedKeys returns a map of a key's position to it's value, as neccesary for Spanner PSQL. 
// Example { $1 -> "value1", $2 -> "value2" }
```

PR #94, opened 2024-11-26
chore(langchain-sdk): pin versions in requirements.txt

twishabansal:


sdks/langchain/pyproject.toml
```diff
@@ -8,10 +8,8 @@ authors = [
     {name = "Google LLC", email = "googleapis-packages@google.com"}
 ]
 dependencies = [
-    "aiohttp==3.10.10",
-    "PyYAML==6.0.2",
-    "langchain-core==0.3.18",
-    "pydantic==2.9.2",
+    "aiohttp>=3.0.8,<=3.11.7",
+    "langchain-core>=0.2.23,<=0.3.21",
 ]
```
kurtisvg: I think most of these dependencies need to be specified, just with the right min version (and optionally <next major version max)

sdks/langchain/pyproject.toml
```diff
@@ -8,10 +8,10 @@ authors = [
...
 dependencies = [
-    "aiohttp==3.10.10",
-    "PyYAML==6.0.2",
-    "langchain-core==0.3.18",
-    "pydantic==2.9.2",
+    "aiohttp>=3.7.0,<4.0.0",
+    "langchain-core>=0.2.23,<1.0.0",
+    "PyYAML>=5.3,<7.0.0",
+    "pydantic>=1.10.0,<3.0.0",
```
kurtisvg: pydantic should def be >= something in the 2.x range?

sdks/langchain/pyproject.toml
```diff
@@ -8,10 +8,10 @@ authors = [
     {name = "Google LLC", email = "googleapis-packages@google.com"}
 ]
 dependencies = [
-    "aiohttp==3.10.10",
-    "PyYAML==6.0.2",
-    "langchain-core==0.3.18",
-    "pydantic==2.9.2",
+    "aiohttp>=3.7.0,<4.0.0",
+    "langchain-core>=0.2.23,<1.0.0",
+    "PyYAML>=5.3,<7.0.0",
```
kurtisvg: pyyaml looks like it should be >= something in the 6.x range?

twishabansal: I tried running a test when in my system version 5.3.0 is installed. This version works with the code. Should we be changing the version to 6.x and above?

sdks/langchain/pyproject.toml
twishabansal: I tried running a test when in my system version 1.10.0 is installed. This version works with the code. Should we be changing the version to 2.x and above?

sdks/langchain/pyproject.toml
kurtisvg: I'd rather err on the side of caution and work within a specific major version -- never know when something might behave differently than we think it should

twishabansal: Alright. I'll update both of these

PR #95, opened 2024-11-26
feat: add std logger

Yuan325:
Logging support 4 different types of logging (debug, info, warn, error).

Example of standard logger:
`2024-11-12T15:08:11.451147-08:00 ERROR "unable to parse tool file at \"tools.yaml\": \"cloud-sql-postgres1\" is not a valid kind of data source"`

`2024-11-12T15:08:11.451377-08:00 INFO "Initalized 0 sources.\n"`

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -86,6 +86,6 @@ func severityToLevel(s string) (slog.Level, error) {
 	case Error:
 		return slog.LevelError, nil
 	default:
-		return slog.Level(-1), fmt.Errorf("invalid log level")
+		return slog.Level(-5), fmt.Errorf("invalid log level")
 	}
 }


internal/log/log.go
```diff
@@ -0,0 +1,91 @@
...
+	case Debug:
+		return slog.LevelDebug, nil
+	case Info:
+		return slog.LevelInfo, nil
+	case Warn:
+		return slog.LevelWarn, nil
+	case Error:
+		return slog.LevelError, nil
+	default:
+		return slog.Level(-1), fmt.Errorf("invalid log level")
```
kurtisvg: it looks like Debug is -4 -- maybe we should use -4 or something lower for the non-zero value? https://pkg.go.dev/log/slog#Level

PR #98, opened 2024-11-28
chore: update params type

Yuan325:
Different databases require different types for `Params` field when adding parameters to their statement. e.g. alloydb, cloudsql, and postgres uses `pgxpool` to query and build sql statement, whereas spanner uses `Spanner` library.

Added a new `ParamValue` struct. `ParseParams` helper function parses arbitraryJSON object into `[]ParamValue`, and the tool's invoke will convert `[]ParamValue` into it's required type.

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -31,13 +31,13 @@ const (
 // ParamValues is an ordered list of ParamValue
 type ParamValues []ParamValue
 
-// ParamValue represents the parameter's name and value. 
+// ParamValue represents the parameter's name and value.
 type ParamValue struct {
 	Name  string
 	Value any
 }
 
-// SliceValues returns a slice of the Param's values (in order).
+// AsSlice returns a slice of the Param's values (in order).
 func (p ParamValues) AsSlice() []any {
 	params := []any{}
 
@@ -47,8 +47,8 @@ func (p ParamValues) AsSlice() []any {
 	return params
 }
 
-// MapValues returns a map of ParamValue's names to values.
-func MapValues(p ParamValues) map[string]interface{} {
+// AsMap returns a map of ParamValue's names to values.
+func (p ParamValues) AsMap() map[string]interface{} {
 	params := make(map[string]interface{})
 	for _, p := range p {
 		params[p.Name] = p.Value
@@ -56,7 +56,7 @@ func MapValues(p ParamValues) map[string]interface{} {
 	return params
 }
 
-// ParseParams parses specified Parameters from data and returns them as ParamValues. 
+// ParseParams parses specified Parameters from data and returns them as ParamValues.
 func ParseParams(ps Parameters, data map[string]any) (ParamValues, error) {
 	params := make([]ParamValue, 0, len(ps))
 	for _, p := range ps {
--- a/internal/tools/parameters_test.go
+++ b/internal/tools/parameters_test.go
@@ -270,8 +270,8 @@ func TestParamValues(t *testing.T) {
 	}
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			gotSlice := tools.SliceValues(tc.in)
-			gotMap := tools.MapValues(tc.in)
+			gotSlice := tc.in.AsSlice()
+			gotMap := tc.in.AsMap()
 
 			for i, got := range gotSlice {
 				want := tc.wantSlice[i]
--- a/internal/tools/postgressql/postgressql.go
+++ b/internal/tools/postgressql/postgressql.go
@@ -106,7 +106,7 @@ type Tool struct {
 }
 
 func (t Tool) Invoke(params tools.ParamValues) (string, error) {
-	sliceParams := tools.SliceValues(params)
+	sliceParams := params.AsSlice()
 	fmt.Printf("Invoked tool %s\n", t.Name)
 	results, err := t.Pool.Query(context.Background(), t.Statement, sliceParams...)
 	if err != nil {


internal/tools/parameters.go
```diff
@@ -28,19 +28,55 @@ const (
 	typeArray  = "array"
 )
 
-// ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
-func ParseParams(ps Parameters, data map[string]any) ([]any, error) {
+type ParamValues []ParamValue
+
+type ParamValue struct {
```
kurtisvg, line 35: nit: Can we add a comment here? Probably for the other type as well.

internal/tools/parameters.go
```diff
@@ -28,19 +28,55 @@ const (
...
+	Name  string
+	Value any
+}
+
+func (p ParamValues) SliceValuesPG() []any {
```
kurtisvg: nit: should also have a comment here

kurtisvg: Does this need to have PG in the title? I would think "SliceValues" should be pretty agnostic?

internal/tools/parameters.go
```diff
@@ -28,19 +28,55 @@ const (
...
+		params = append(params, p.Value)
+	}
+	return params
+}
+
+func (p ParamValues) MapValuesPG() map[string]interface{} {
+	params := make(map[string]interface{})
+
+	for i, p := range p {
+		key := fmt.Sprintf("p%d", i+1)
```
kurtisvg: I'm not sure I understand what this is doing. The formatting seems specific to Spanner maybe -- can we push this formatting into the Spanner tool?

internal/tools/parameters.go
```diff
@@ -28,19 +28,55 @@ const (
...
+		params[key] = p.Value
+	}
+	return params
+}
+
+func (p ParamValues) MapValuesGSQL() map[string]interface{} {
```
kurtisvg: Similarly here -- can this just be "MapValues", and if SQL uses special formatting can we push it into that tool?

internal/tools/parameters.go
```diff
@@ -28,19 +28,55 @@ const (
...
+
+	for _, p := range p {
+		params[p.Name] = p.Value
+	}
+	return params
+}
+
+// ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
+func ParseParams(ps Parameters, data map[string]any) (ParamValues, error) {
+	params := []ParamValue{}
```
kurtisvg: nit: could optimize slightly by reserving some capacity 
```suggestion
	params := make([]ParamValue, 0, len(ps))
```

internal/tools/parameters.go
Yuan325: Removed the PG ones for now, and made the GSQL ones as MapValues that is general (param name as keys).

internal/tools/parameters.go
```diff
@@ -28,19 +28,49 @@ const (
 	typeArray  = "array"
 )
 
-// ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
-func ParseParams(ps Parameters, data map[string]any) ([]any, error) {
+// ParamValues is an ordered list of ParamValue
+type ParamValues []ParamValue
+
+// ParamValue represents the parameter's name and value. This struct is parsed from arbitraryJSON object by the ParseParam helper function.
```
```suggestion
// ParamValue represents the parameter's name and value. 
```

internal/tools/parameters.go
```diff
@@ -28,19 +28,49 @@ const (
...
+type ParamValue struct {
+	Name  string
+	Value any
+}
+
+// SliceValuesPG is a helper function to convert ParamValues into an ordered slice.
```
```suggestion
// SliceValues returns a slice of the Param's values (in order).
```

internal/tools/parameters.go
```diff
@@ -28,19 +28,49 @@ const (
...
+func SliceValues(p ParamValues) []any {
 	params := []any{}
+
+	for _, p := range p {
+		params = append(params, p.Value)
+	}
+	return params
+}
+
+// MapValues is a helper function to convert ParamValues into a map. It uses param name as keys.
```
```suggestion
// MapValues returns a map of ParamValue's names to values.
```

internal/tools/parameters.go
kurtisvg: nit: https://tip.golang.org/doc/comment#func

internal/tools/parameters.go
```diff
@@ -28,19 +28,49 @@ const (
...
+func MapValues(p ParamValues) map[string]interface{} {
+	params := make(map[string]interface{})
+
+	for _, p := range p {
```
kurtisvg: super-nit:
```suggestion
	for _, p := range p {
```

internal/tools/parameters.go
```diff
@@ -28,19 +28,49 @@ const (
...
+		params[p.Name] = p.Value
+	}
+	return params
+}
+
+// ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
+func ParseParams(ps Parameters, data map[string]any) (ParamValues, error) {
```
kurtisvg: nit: not really json at this point
```suggestion
// ParseParams parses specified Parameters from data and returns them as ParamValues. 
func ParseParams(ps Parameters, data map[string]any) (ParamValues, error) {
```

internal/tools/parameters.go
```diff
@@ -28,19 +28,49 @@ const (
...
+type ParamValues []ParamValue
+
+// ParamValue represents the parameter's name and value. This struct is parsed from arbitraryJSON object by the ParseParam helper function.
+type ParamValue struct {
+	Name  string
+	Value any
+}
+
+// SliceValuesPG is a helper function to convert ParamValues into an ordered slice.
+func SliceValues(p ParamValues) []any {
```
kurtisvg: nit: would these be better as a ParamValues receiver? same for the other one
```suggestion
func (p ParamValues) AsSlice() []any {
```

PR #100, opened 2024-12-02
chore: Remove redundant requirements.txt files.

anubhav756:
The dependencies are already specified in the respective pyproject.toml files.

### DIFF
--- a/sdks/langchain/requirements.txt
+++ b/sdks/langchain/requirements.txt
@@ -1,4 +1,0 @@
-langchain-core==0.3.21
-PyYAML==6.0.2
-pydantic==2.10.2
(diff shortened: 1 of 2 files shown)


twishabansal:
We're specifying dependency version ranges in the [.toml ](https://github.com/googleapis/genai-toolbox/blob/a12973ea360a335a0ae8b05f07000bfee96817e6/sdks/langchain/pyproject.toml#L10) files. This is to tell the user that our app is compatible with the versions that lie within this range.

However, requirements.txt is used for testing purposes. Eg. What exact version of dependencies do we need to install to run the tests.

So, we should not be deleting the requirements.txt file

PR #103, opened 2024-12-02
feat(source/alloydb-pg): add configuration for public and private IP

Yuan325:
Allow user to set if their database uses private or public ip. The reason we add this is because the dialer require different initialization with private and public ip.

By default, toolbox will use public ip.

### DIFF
--- a/internal/sources/alloydbpg/alloydb_pg.go
+++ b/internal/sources/alloydbpg/alloydb_pg.go
@@ -31,16 +31,16 @@ const SourceKind string = "alloydb-postgres"
 var _ sources.SourceConfig = Config{}
 
 type Config struct {
-	Name     string          `yaml:"name"`
-	Kind     string          `yaml:"kind"`
-	Project  string          `yaml:"project"`
-	Region   string          `yaml:"region"`
-	Cluster  string          `yaml:"cluster"`
-	Instance string          `yaml:"instance"`
-	IP_type  sources.IP_type `yaml:"ip_type"`
-	User     string          `yaml:"user"`
-	Password string          `yaml:"password"`
-	Database string          `yaml:"database"`
+	Name     string         `yaml:"name"`
+	Kind     string         `yaml:"kind"`
+	Project  string         `yaml:"project"`
+	Region   string         `yaml:"region"`
+	Cluster  string         `yaml:"cluster"`
+	Instance string         `yaml:"instance"`
+	IP_type  sources.IPType `yaml:"ip_type"`
+	User     string         `yaml:"user"`
+	Password string         `yaml:"password"`
+	Database string         `yaml:"database"`
 }
 
 func (r Config) SourceConfigKind() string {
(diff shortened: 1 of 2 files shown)


internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -36,6 +37,7 @@ type Config struct {
 	Region   string `yaml:"region"`
 	Cluster  string `yaml:"cluster"`
 	Instance string `yaml:"instance"`
+	Ip_type  string `yaml:"ip_type"`
```
kurtisvg: https://google.github.io/styleguide/go/decisions#initialisms

```suggestion
	IP_type  string `yaml:"ip_type"`
```

kurtisvg: We should have validation on this, and explicitly default to public. Potentially using a custom type with MarshallYAML and UnmarshalYaml?

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -80,7 +82,22 @@ func (s *Source) PostgresPool() *pgxpool.Pool {
...
+		// alloydbconn create a dialer with private IP by default
+		return alloydbconn.NewDialer(context.Background())
+	default:
+		return alloydbconn.NewDialer(
+			context.Background(),
+			alloydbconn.WithDefaultDialOptions(
+				alloydbconn.WithPublicIP(),
+			),
+		)
+	}
```
kurtisvg: Would prefer this to be explicitly type checked -- public, private, or error

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -80,7 +82,24 @@ func (s *Source) PostgresPool() *pgxpool.Pool {
...
+		return alloydbconn.NewDialer(
+			context.Background(),
+			alloydbconn.WithDefaultDialOptions(
+				alloydbconn.WithPublicIP(),
+			),
+		)
+	default:
+		return nil, fmt.Errorf("invalid ip_type %s", ip_type)
+	}
+}
```
kurtisvg: small-nit: slightly more explicit
```suggestion
        var dialOpts []alloydbconn.DialOption
        switch strings.ToLower(ip_type) {
	case "private":
		dialOpts := append(dialOpts, alloydbconn.WithPrivateIP())
	case "public":
		dialOpts := append(dialOpts, alloydbconn.WithPublicIP())
	default:
		return nil, fmt.Errorf("invalid ip_type %s", ip_type)
	}
	ctx := context.Background()
	return alloydbconn.NewDialer(ctx, alloydbconn.WithDefaultDialOptions(dialOpts...))
}
```

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -89,7 +108,7 @@ func initAlloyDBPgConnectionPool(project, region, cluster, instance, user, pass,
 	}
 
 	// Create a new dialer with any options
-	d, err := alloydbconn.NewDialer(context.Background())
+	d, err := getDialer(ip_type)
```
kurtisvg: super-nit: we might want to return opts here and just pass them to the connector?

internal/sources/ip_type.go
```diff
@@ -0,0 +1,45 @@
...
+package sources
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+type IP_type string
```
kurtisvg: https://go.dev/doc/effective_go#mixed-caps
```suggestion
type IPType string
```

PR #104, opened 2024-12-03
chore: type check fix

twishabansal:


sdks/langchain/tests/test_client.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 from langchain_core.tools import StructuredTool
 from pydantic import ValidationError
-from toolbox_langchain_sdk import ToolboxClient
+from toolbox_langchain_sdk import ToolboxClient # type: ignore
```
kurtisvg, line 8: Can we provide some context in the PR description -- why is this required?

PR #107, opened 2024-12-03
ci: Add Google OAuth provider test

duwenxin99:
1. OAuth provider test must run on Cloud Build to be able to print ID tokens of the GCE VM attached service account.
2. Borrowing gcloud CLI team's public `Client ID` in the Google ID token [documentation](https://cloud.google.com/docs/authentication/token-types#id-contents)

### DIFF
--- a/tests/auth_test.go
+++ b/tests/auth_test.go
@@ -18,6 +18,7 @@ package tests
 import (
 	"io"
 	"net/http"
+	"os/exec"
 
 	"testing"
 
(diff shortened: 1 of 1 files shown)


tests/auth_test.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	return string(body), nil
+}
+
+func TestGoogleAuthVerification(t *testing.T) {
+	clientId := "32555940559.apps.googleusercontent.com"
```
Yuan325, line 56: Is this secure to be included? Or should we put this as a secret?

tests/auth_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	for _, tc := range tcs {
+
+		token, err := getGoogleIdToken(clientId)
+		if err != nil {
+			t.Fatalf("ID token generation error: %s", err)
+		}
+		headers := http.Header{}
+		headers.Add("my-google-auth_token", token)
+		_, err = tc.authSource.GetClaimsFromHeader(headers)
```
Yuan325: Is there something that's returned if it's successful? should we test that too?

tests/auth_test.go
kurtisvg: +1 -- using an env var here should be good?

tests/auth_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+
+	"github.com/googleapis/genai-toolbox/internal/auth"
+	"github.com/googleapis/genai-toolbox/internal/auth/google"
+)
+
+// Returns an ID token of the service account
+func getGoogleIdToken(audience string) (string, error) {
+	url := "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=" + audience
+	req, err := http.NewRequest(http.MethodGet, url, nil)
```
kurtisvg: This works with the metadata service, but what about the other ADC? specifically the gcloud login

tests/auth_test.go
duwenxin99: This is an example client ID provided in the public [documentation](https://cloud.google.com/docs/authentication/token-types#id-contents) so it should be ok.

tests/auth_test.go
duwenxin99: Good point. Added.

tests/auth_test.go
duwenxin99: added support for local testing using gcloud login credentials

PR #112, opened 2024-12-05
chore: set out and err io.Writer during initialization

Yuan325:
Logger might be created pre- or post- flag. The logger defined pre- and post- flag should be using the same out and err writer.

This will ensure that the server uses the same out and err writer defined with the `WithStreams` options.


cmd/options.go
```diff
@@ -15,15 +15,19 @@
...
+
 	"github.com/googleapis/genai-toolbox/internal/log"
 )
 
 // Option is a function that configures a Command.
 type Option func(*Command)
 
 // WithLogger overrides the default logger.
-func WithLogger(l log.Logger) Option {
+func WithLogger(l log.Logger, out, err io.Writer) Option {
```
kurtisvg: I think I would remove logger as an input -- just make this "WithStreams" and take in out/err. 

There isn't much point in taking a logger if we're going to replace it right away anyway.

cmd/root.go
```diff
@@ -64,22 +65,34 @@ type Command struct {
 	cfg        server.ServerConfig
 	logger     log.Logger
 	tools_file string
+	out        io.Writer
+	err        io.Writer
```
kurtisvg: nit: 
```suggestion
	outStream        io.Writer
	errStream        io.Writer
```

PR #114, opened 2024-12-05
feat(source/cloudsql-pg): add configuration for public and private IP

Yuan325:
Allow user to set if their database uses private or public ip. The reason we add this is because the dialer require different initialization with private and public ip.

By default, toolbox will use public ip.

### DIFF
--- a/internal/server/api_test.go
+++ b/internal/server/api_test.go
@@ -35,11 +35,11 @@ type MockTool struct {
 	Params      []tools.Parameter
 }
 
-func (t MockTool) Invoke([]any) (string, error) {
+func (t MockTool) Invoke(tools.ParamValues) (string, error) {
 	return "", nil
 }
 
-func (t MockTool) ParseParams(data map[string]any) ([]any, error) {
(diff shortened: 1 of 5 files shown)


internal/sources/cloudsqlpg/cloud_sql_pg.go
```diff
@@ -30,22 +31,23 @@ const SourceKind string = "cloud-sql-postgres"
...
-	Instance string `yaml:"instance"`
-	User     string `yaml:"user"`
-	Password string `yaml:"password"`
-	Database string `yaml:"database"`
+	Name     string         `yaml:"name"`
+	Kind     string         `yaml:"kind"`
+	Project  string         `yaml:"project"`
+	Region   string         `yaml:"region"`
+	Instance string         `yaml:"instance"`
+	IP_type  sources.IPType `yaml:"ip_type"`
```
kurtisvg:
```suggestion
	IPType  sources.IPType `yaml:"ip_type"`
```

PR #117, opened 2024-12-06
feat(langchain-sdk): Support authentication in LangChain Toolbox SDK.

anubhav756:
This PR adds support for authentication in the `ToolboxClient`.

* We add API support for providing methods that fetch ID tokens for given auth sources. The SDK fetches tokens using these methods and formats and includes them in the headers of tool invocation requests.
* The SDK also validates whether the tool's authentication requirements are met with the registered auth sources both after loading tool/toolset (warning) and right before tool invocation (error).
* We do this by filtering out all the parameters of all the tools which have auth sources specified in their respective parameter schema. This filtering is done while loading the tool/toolset.
* We also warn the user in case ID tokens (containing user's information in the form of claims) are sent over HTTP and not HTTPS.

Note: We need to add the same functionality for the LlamaIndex SDK. We
also need to update the documentations for the same.

### DIFF
--- a/sdks/langchain/tests/test_client.py
+++ b/sdks/langchain/tests/test_client.py
@@ -1,3 +1,4 @@
+import asyncio
 import warnings
 from unittest.mock import AsyncMock, Mock, call, patch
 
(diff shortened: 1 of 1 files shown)


sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -106,21 +110,83 @@ async def _tool_func(**kwargs) -> dict:
             args_schema=tool_model,
         )
 
-    async def load_tool(self, tool_name: str) -> StructuredTool:
+    def _validate_auth_sources(self, manifest: ManifestSchema) -> None:
+        """
+        Validates that each parameter in the given manifest has at least one
```
duwenxin99: authSource are optional in parameters and most parameters are probably not authenticated. Why do we think throwing warning here is needed? Even if we do, I think it should be combined with `_remove_auth_params()` so that we don't look through the parameters twice.

anubhav756: It seems to be suggested in [this part](https://screenshot.googleplex.com/3oFJMfAQotKEZHq) of the design document. Am I missing any details?

duwenxin99: This part in the design doc is suggesting that we need to throw a warning when a tool contains auth-required parameters but a getter function is not provided during tool loading. Here `_validate_auth_sources` seems to throw a warning when a tool  contains a parameter that is not auth-required, which is a different case.

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -77,6 +80,33 @@ async def _load_toolset_manifest(
...
+        Returns:
+            True if at least one permitted auth source of each of the auth
+            params, of the given tool, is registered. Also returns True if the
+            given tool does not require any auth sources.
+        """
+
+        if tool_name not in self._tool_param_auth:
+            return True
+
+        return all(
```
duwenxin99: nit: Could you improve the readability of this return logic? I think we could sacrifice some conciseness for more readability (i.e putting for loop in the front)

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -106,21 +144,89 @@ async def _tool_func(**kwargs) -> dict:
...
+                    param.name
+                ] = param.authSources
+
+            tool_schema.parameters = non_auth_params
+
+            # If none of the permitted auth sources of a parameter are
+            # registered, raise a warning message to the user.
+            if not self._validate_auth(tool_name):
+                warnings.warn(
+                    f"The parameter {param.name} of tool {tool_name} requires authentication, but none of its permitted auth sources are currently registered. Please ensure the necessary auth source is registered before invoking this tool."
```
duwenxin99: Where is this `param` variable since we are outside of the for loop?

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: Done. Thanks for the suggestion.

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: Thanks for pointing that out. Fixed!

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: I have refactored this function to combine extraction and validation of auth params in a single method. Could you please take a look now. Please note that we throw a warning when a getter function of a given auth services is not provided while a tool is loaded having a parameter that requires that auth service's getter function registered.

Note that I've also added test cases for this logic.

anubhav756: Per offline discussion, this looks good, so closing this for now.

PR #119, opened 2024-12-09
ci(langchain-sdk): Update file format check to run inside the sdk directory

twishabansal:


.github/workflows/lint-langchain.yaml
```diff
@@ -45,10 +45,10 @@ jobs:
...
-          isort --check .
+          black --check sdks/langchain/
+          isort --check sdks/langchain/
 
       - name: Run type-check
         env:
           MYPYPATH: 'sdks/langchain/src'
-        run: mypy --install-types --non-interactive --explicit-package-bases sdks/langchain
\ No newline at end of file
+        run: mypy --install-types --non-interactive --explicit-package-bases sdks/langchain
```
kurtisvg, line 54: Would it be easier to set working-directory? 

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iddefaultsrunworking-directory

PR #121, opened 2024-12-09
chore: separate listen and server into two functions

Yuan325:
Separate listen and server into two distinct functions to prevent race condition during testing.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -167,13 +167,14 @@ func run(cmd *Command) error {
 	}
 	l, err := s.Listen(ctx)
 	if err != nil {
-		errMsg := fmt.Errorf("toolbox crashed with the following error: %w", err)
+		errMsg := fmt.Errorf("toolbox failed to mount listener: %w", err)
 		cmd.logger.Error(errMsg.Error())
 		return errMsg
 	}
+	cmd.logger.Info("Server ready to serve")
 	err = s.Serve(l)
 	if err != nil {
-		errMsg := fmt.Errorf("toolbox creash with the following error: %w", err)
+		errMsg := fmt.Errorf("toolbox crashed with the following error: %w", err)
 		cmd.logger.Error(errMsg.Error())
 		return errMsg
 	}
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -165,12 +165,18 @@ func run(cmd *Command) error {
...
-	err = s.ListenAndServe(ctx)
+	l, err := s.Listen(ctx)
 	if err != nil {
 		errMsg := fmt.Errorf("toolbox crashed with the following error: %w", err)
 		cmd.logger.Error(errMsg.Error())
 		return errMsg
 	}
+	err = s.Serve(l)
+	if err != nil {
+		errMsg := fmt.Errorf("toolbox creash with the following error: %w", err)
```
```suggestion
		errMsg := fmt.Errorf("toolbox crashed with the following error: %w", err)
```

cmd/root.go
```diff
@@ -165,12 +165,18 @@ func run(cmd *Command) error {
 		cmd.logger.Error(errMsg.Error())
 		return errMsg
 	}
-	err = s.ListenAndServe(ctx)
+	l, err := s.Listen(ctx)
 	if err != nil {
 		errMsg := fmt.Errorf("toolbox crashed with the following error: %w", err)
```
```suggestion
		errMsg := fmt.Errorf("toolbox failed to mount listener: %w", err)
```

internal/server/server.go
```diff
@@ -138,17 +138,22 @@ func NewServer(cfg ServerConfig, log logLib.Logger) (*Server, error) {
...
 	defer cancel()
 
 	addr := net.JoinHostPort(s.conf.Address, strconv.Itoa(s.conf.Port))
 	lc := net.ListenConfig{KeepAlive: 30 * time.Second}
 	l, err := lc.Listen(ctx, "tcp", addr)
 	if err != nil {
-		return fmt.Errorf("failed to open listener for %q: %w", addr, err)
+		return nil, fmt.Errorf("failed to open listener for %q: %w", addr, err)
 	}
+	s.logger.Info("Server ready to serve")
```
kurtisvg: nit: I would move this up a level (between .Listen() and .Serve()). 

Ideally we try to keep logging at the highest level -- usually you return errors until you cant anymore, and then you log them.

PR #146, opened 2024-12-17
doc(langchain-sdk): Add guidance for configuring and using authenticated tools in Toolbox.

anubhav756:


sdks/langchain/README.md
```diff
@@ -163,6 +163,183 @@ graph.invoke(
...
+> user information and man-in-the-middle attacks, where sensitive data can be
+> intercepted.
+
+Some tools in your Toolbox configuration might require user authentication to
+access sensitive data. This section guides you on how to configure tools for
+authentication and use them with the SDK.
+
+### Supported Authentication Mechanisms
+The Toolbox SDK currently supports authentication using [Google OAuth
+2.0](https://cloud.google.com/apigee/docs/api-platform/security/oauth/oauth-home).
```
duwenxin99, line 194: We should mention that we only support the OIDC tokens. Specifically ID token and not access token for Google OAuth

sdks/langchain/README.md
```diff
@@ -163,6 +163,183 @@ graph.invoke(
...
+
+### Configuring Tools for Authentication
```
duwenxin99, line 196: This is a little repetitive to the Toolbox service's doc. Instead of creating new examples, could you link to them instead? https://github.com/googleapis/genai-toolbox/blob/main/docs/tools/README.md#authenticated-parameters

sdks/langchain/README.md
```diff
@@ -163,6 +163,183 @@ graph.invoke(
...
+            - my_auth_service
+```
+
+> [!TIP]
+> You can configure multiple authentication sources with the same `kind` (e.g.,
+> Google OAuth) to accommodate different applications using the same service
+> with distinct client IDs.
+>
+> Here's an example:
+>
```
duwenxin99: This example is really helpful. Could you add it to `authSource` [doc](https://github.com/googleapis/genai-toolbox/tree/main/docs/authSources) and link to it here?

sdks/langchain/README.md
anubhav756: Thanks for the suggestion.

sdks/langchain/README.md
anubhav756: Good catch. Thanks!

sdks/langchain/README.md
anubhav756: Sure thing!

PR #147, opened 2024-12-17
feat: update log with given context

Yuan325:
Update logging with the given context.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -44,7 +44,7 @@ type Server struct {
 }
 
 // NewServer returns a Server object based on provided Config.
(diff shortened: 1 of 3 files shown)


internal/server/server.go
```diff
@@ -89,7 +89,7 @@ func NewServer(cfg ServerConfig, log logLib.Logger) (*Server, error) {
 		}
 		sourcesMap[name] = s
 	}
-	log.Info(fmt.Sprintf("Initialized %d sources.", len(sourcesMap)))
+	log.InfoContext(context.Background(), fmt.Sprintf("Initialized %d sources.", len(sourcesMap)))
```
kurtisvg: Is context background the right thing here?

PR #149, opened 2024-12-18
feat(tools): add trace to tools

Yuan325:
Add trace for toolbox initialization within tools:

toolbox/server/tool/init
toolbox/server/tool/get
toolbox/server/tool/invoke

### DIFF
--- a/internal/tools/spanner/spanner.go
+++ b/internal/tools/spanner/spanner.go
@@ -130,8 +130,7 @@ func (t Tool) Invoke(params tools.ParamValues) (string, error) {
 	fmt.Printf("Invoked tool %s\n", t.Name)
 	var out strings.Builder
 
-	ctx := context.Background()
-	_, err = t.Client.ReadWriteTransaction(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
+	_, err = t.Client.ReadWriteTransaction(context.Background(), func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
 		stmt := spanner.Statement{
 			SQL:    t.Statement,
 			Params: mapParams,
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -21,6 +21,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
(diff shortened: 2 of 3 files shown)


internal/telemetry/trace.go
```diff
@@ -25,6 +25,10 @@ const (
 
 	SourceInit    = "toolbox/server/source/init"
 	SourceConnect = "toolbox/server/source/connect"
+
+	ToolInit   = "toolbox/server/tool/init"
+	ToolGet    = "toolbox/server/tool/get"
+	ToolInvoke = "toolbox/server/tool/invoke"
```
kurtisvg: nit: I think these are only used in once place, so I'm not totally use the use of a constant is justified.

internal/tools/postgressql/postgressql.go
```diff
@@ -57,7 +57,10 @@ func (cfg Config) ToolConfigKind() string {
 	return ToolKind
 }
 
-func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
+func (cfg Config) Initialize(ctx context.Context, srcs map[string]sources.Source) (tools.Tool, error) {
+	_, span := tools.ToolInitTrace(ctx, ToolKind, cfg.Name)
```
kurtisvg:
```suggestion
	ctx, span := tools.ToolInitTrace(ctx, ToolKind, cfg.Name)
```

internal/tools/spanner/spanner.go
```diff
@@ -55,7 +55,10 @@ func (cfg Config) ToolConfigKind() string {
 	return ToolKind
 }
 
-func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
+func (cfg Config) Initialize(ctx context.Context, srcs map[string]sources.Source) (tools.Tool, error) {
+	_, span := tools.ToolInitTrace(ctx, ToolKind, cfg.Name)
```
kurtisvg:
```suggestion
	ctx, span := tools.ToolInitTrace(ctx, ToolKind, cfg.Name)
```

internal/tools/postgressql/postgressql.go
kurtisvg: nit: we had discussed trying to move this up -- aka around the initialize function instead of inside of it

PR #156, opened 2024-12-19
feat(llamaindex-sdk): Implement OAuth support for LlamaIndex.

anubhav756:
* Update docs for Authenticated Tools.
* Implement test cases for OAuth in LlamaIndex.
* Sync SDK dependency versions b/w LangChain and LlamaIndex.

### DIFF
--- a/sdks/llamaindex/README.md
+++ b/sdks/llamaindex/README.md
@@ -4,6 +4,8 @@ This SDK allows you to seamlessly integrate the functionalities of
 [Toolbox](https://github.com/googleapis/genai-toolbox) into your LLM
 applications, enabling advanced orchestration and interaction with GenAI models.
 
+<!-- TOC ignore:true -->
+## Table of Contents
 <!-- TOC -->
 
 - [Installation](#installation)
--- a/sdks/langchain/README.md
+++ b/sdks/langchain/README.md
@@ -4,6 +4,8 @@ This SDK allows you to seamlessly integrate the functionalities of
 [Toolbox](https://github.com/googleapis/genai-toolbox) into your LLM
 applications, enabling advanced orchestration and interaction with GenAI models.
 
+<!-- TOC ignore:true -->
+## Table of Contents
 <!-- TOC -->
 
 - [Installation](#installation)


sdks/llamaindex/README.md
```diff
@@ -2,41 +2,40 @@
 
 This SDK allows you to seamlessly integrate the functionalities of
 [Toolbox](https://github.com/googleapis/genai-toolbox) into your LLM
-applications, enabling advanced orchestration and interaction with GenAI
-models.
-
-<!-- TOC ignore:true -->
-
-## Table of Contents
+applications, enabling advanced orchestration and interaction with GenAI models.
```
twishabansal, line 5: I think this got copied down by mistake. Please correct

sdks/llamaindex/README.md
```diff
@@ -2,41 +2,40 @@
...
 
 <!-- TOC -->
 
-- [Installation](#installation)
-- [Usage](#usage)
-- [Load a toolset](#load-a-toolset)
-- [Load a single tool](#load-a-single-tool)
-- [Use with LlamaIndex](#use-with-llamaindex)
-- [Manual usage](#manual-usage)
+- [GenAI Toolbox SDK](#genai-toolbox-sdk)
```
twishabansal: GenAI Toolbox SDK should not be a part of the TOC

sdks/llamaindex/README.md
```diff
@@ -2,41 +2,40 @@
...
+        - [Configuring Tools for Authentication](#configuring-tools-for-authentication)
+        - [Configure SDK for Authentication](#configure-sdk-for-authentication)
+        - [Complete Example](#complete-example)
 
 <!-- /TOC -->
-
 ## Installation
 
+> [!IMPORTANT]
+> This SDK is not yet available on PyPI. For now, install it from source by following these [installation instructions](DEVELOPER.md).
```
twishabansal: Is this linted? Might have more than 80 chars in line 26.

sdks/llamaindex/README.md
```diff
@@ -99,6 +98,97 @@ response = agent.query("Get some response from the agent.")
 
 You can also execute a tool manually using the `acall` method.
 
-```python
-result = await tools[0].acall({"param1": "value1", "param2": "value2"})
+```py
+result = await tools[0].arun({ "name": "Alice", "age": 30 })
```
twishabansal: Why are we using the arun method instead of the acall method?

sdks/llamaindex/README.md
```diff
@@ -99,6 +98,97 @@ response = agent.query("Get some response from the agent.")
 
 You can also execute a tool manually using the `acall` method.
```
twishabansal, line 102: Need to update documentation here to mention using the `arun` method

sdks/llamaindex/pyproject.toml
```diff
@@ -1,17 +1,17 @@
...
-description = "Python SDK for interacting with the Toolbox service with Llamaindex"
+description = "Python SDK for interacting with the Toolbox service with LlamaIndex"
 license = {file = "LICENSE"}
 requires-python = ">=3.9"
 authors = [
     {name = "Google LLC", email = "googleapis-packages@google.com"}
 ]
 dependencies = [
     "llama-index>=0.11.2,<1.0.0",
-    "PyYAML>=6.0.1,<7.0.0",
```
twishabansal, line 12: When I looked into lower bounding these versions, versions like `pydantic==6.0.0` were breaking. Are they working now?

sdks/llamaindex/README.md
```diff
@@ -99,6 +98,97 @@ response = agent.query("Get some response from the agent.")
...
+    # request = requests.Request()
+    # id_token_string = id_token.fetch_id_token(request, "YOUR_AUDIENCE")# Replace with your audience
+    # return id_token_string
+    return "YOUR_ACTUAL_ID_TOKEN" # placeholder
+
+async def main():
+    toolbox = ToolboxClient("http://localhost:5000")
+    toolbox.add_auth_header("my_auth_service", get_auth_header)
+    tools = await toolbox.load_toolset()
+    result = await tools[0].arun({"input": "some input"})
```
twishabansal: Why are we using `arun` here? and not other methods like `acall`?

sdks/llamaindex/src/toolbox_llamaindex_sdk/client.py
```diff
@@ -75,7 +78,36 @@ async def _load_toolset_manifest(
...
+
+        for permitted_auth_sources in self._tool_param_auth[tool_name].values():
+            found_match = False
+            for registered_auth_source in self._id_token_getters:
+                if registered_auth_source in permitted_auth_sources:
+                    found_match = True
+                    break
+            if not found_match:
+                return False
+        return True
```
twishabansal, line 110: Why do we return True here? Isn't false a better default value here?

anubhav756:
> What is the current test coverage?

```
---------- coverage: platform darwin, python 3.11.9-final-0 ----------
Name                                                     Stmts   Miss  Cover
----------------------------------------------------------------------------
sdks/langchain/src/toolbox_langchain_sdk/__init__.py         2      0   100%
sdks/langchain/src/toolbox_langchain_sdk/client.py          79      0   100%
sdks/langchain/src/toolbox_langchain_sdk/utils.py           64      0   100%
sdks/langchain/tests/test_client.py                        256      0   100%
sdks/langchain/tests/test_utils.py                         123      0   100%
sdks/llamaindex/src/toolbox_llamaindex_sdk/__init__.py       2      0   100%
sdks/llamaindex/src/toolbox_llamaindex_sdk/client.py        79      0   100%
sdks/llamaindex/src/toolbox_llamaindex_sdk/utils.py         64      0   100%
sdks/llamaindex/tests/test_client.py                       248      0   100%
sdks/llamaindex/tests/test_utils.py                        123      0   100%
----------------------------------------------------------------------------
TOTAL                                                     1040      0   100%
```

sdks/llamaindex/src/toolbox_llamaindex_sdk/client.py
anubhav756: This statement is reached only in the cases where an auth source is required, and is fulfilled by the currently registered ID token getters, which is a valid scenario, hence `True` is returned. If we return `False` here, then a `True` value will be returned only for the non-auth params, and a `False` for all auth params regardless of whether their ID token getters are registered or not.

Thoughts?

sdks/llamaindex/README.md
anubhav756: I had been having a couple of issues with Rewrap. Fixed it. Thanks for pointing this out.

sdks/llamaindex/README.md
anubhav756: Agree, although Rewrap seems to add it back even if I remove it or putting the ignore comment. Removed it manually for now.

sdks/llamaindex/README.md
anubhav756: Good catch. Thanks!

sdks/llamaindex/README.md
anubhav756: Another miss while syncing the doc with LangChain's. Thanks for pointing that out!

sdks/llamaindex/README.md
anubhav756: `acall` seems to be the correct one. Thanks anyways.

sdks/llamaindex/pyproject.toml
anubhav756: Although I don't see a major [change](https://github.com/yaml/pyyaml/blob/c42fa3bff1eabdb64763bb1526d9ea1ccb708479/CHANGES#L9), I think it's safer to go with `6.0.1` considering the gap between both these versions is around 2 years.
image

sdks/llamaindex/README.md
anubhav756: Missed it. Thanks for the call-out.

sdks/llamaindex/pyproject.toml
twishabansal: The same goes for the other deps as well. I remember them erroring out on any lower versions. Although your changes might have allowed some versions to work, we should test these lower bounds out. Are you able to run the code on all deps with their lower bound versions?

anubhav756: Fixed in #159. Thanks for the suggestion!

PR #157, opened 2024-12-19
feat: add custom metrics to toolbox

Yuan325:
Add metrics for toolbox:

toolbox.server.toolset.get.count - Count of toolset manifest request served
toolbox.server.tool.get.count - Count of tool manifest request served
toolbox.server.tool.invoke.count - Count of invocation request served

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -151,7 +151,7 @@ func run(cmd *Command) error {
 	}
 
 	// Set up OpenTelemetry
-	otelShutdown, err := telemetry.SetupOTel(ctx, cmd.Command.Version, cmd.cfg)
+	otelShutdown, err := telemetry.SetupOTel(ctx, cmd.Command.Version, cmd.cfg.TelemetryOTLP, cmd.cfg.TelemetryGCP)
 	if err != nil {
 		errMsg := fmt.Errorf("error setting up OpenTelemetry: %w", err)
 		cmd.logger.Error(errMsg.Error())
@@ -164,12 +164,6 @@ func run(cmd *Command) error {
 			cmd.logger.Error(errMsg.Error())
 		}
 	}()
-	metrics, err := server.CreateCustomMetrics(cmd.Command.Version)
-	if err != nil {
-		errMsg := fmt.Errorf("unable to create custom metrics: %w", err)
-		cmd.logger.Error(errMsg.Error())
-		return errMsg
-	}
 
 	// Read tool file contents
 	buf, err := os.ReadFile(cmd.tools_file)
@@ -187,7 +181,7 @@ func run(cmd *Command) error {
 	}
 
 	// run server
-	s, err := server.NewServer(cmd.cfg, cmd.logger, metrics)
+	s, err := server.NewServer(cmd.cfg, cmd.logger)
 	if err != nil {
 		errMsg := fmt.Errorf("toolbox failed to start with the following error: %w", err)
 		cmd.logger.Error(errMsg.Error())
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -83,11 +83,9 @@ func toolGetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	tool, ok := s.tools[toolName]
 	var err error
 	defer func() {
-		var status string
+		status := "success"
 		if err != nil {
 			status = "error"
-		} else {
-			status = "success"
 		}
 		s.metrics.ToolGet.Add(
 			r.Context(),
@@ -124,11 +122,9 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	tool, ok := s.tools[toolName]
 	var err error
 	defer func() {
-		var status string
+		status := "success"
 		if err != nil {
 			status = "error"
-		} else {
-			status = "success"
 		}
 		s.metrics.ToolInvoke.Add(
 			r.Context(),
--- a/internal/server/api_test.go
+++ b/internal/server/api_test.go
@@ -63,6 +63,7 @@ func (t MockTool) Authorized(verifiedAuthSources []string) bool {
 func TestToolsetEndpoint(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+
 	// Set up resources to test against
 	tool1 := MockTool{
 		Name:   "no_params",
@@ -91,25 +92,26 @@ func TestToolsetEndpoint(t *testing.T) {
 		toolsets[name] = m
 	}
 
-	otelShutdown, err := telemetry.SetupOTel(ctx, fakeVersionString)
-	if err != nil {
-		t.Fatalf("unable to initialize telemetry: %s", err)
-	}
-	defer func() {
-		err := otelShutdown(ctx)
-		if err != nil {
-			t.Fatalf("unexpected error: %s", err)
-		}
-	}()
-	metrics, err := CreateCustomMetrics(fakeVersionString)
-	if err != nil {
-		t.Fatalf("unable to initialize custom metrics: %s", err)
-	}
-
 	testLogger, err := log.NewStdLogger(os.Stdout, os.Stderr, "info")
 	if err != nil {
 		t.Fatalf("unable to initialize logger: %s", err)
 	}
+
+	otelShutdown, err := telemetry.SetupOTel(ctx, fakeVersionString, "", false)
+	if err != nil {
+		t.Fatalf("unable to setup otel: %s", err)
+	}
+	defer func() {
+		err := otelShutdown(ctx)
+		if err != nil {
+			t.Fatalf("error shutting down OpenTelemetry: %s", err)
+		}
+	}()
+	metrics, err := CreateCustomMetrics(fakeVersionString)
+	if err != nil {
+		t.Fatalf("unable to create custom metrics: %s", err)
+	}
+
 	server := Server{conf: ServerConfig{Version: fakeVersionString}, logger: testLogger, metrics: metrics, tools: toolsMap, toolsets: toolsets}
 	r, err := apiRouter(&server)
 	if err != nil {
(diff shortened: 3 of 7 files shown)


internal/server/api.go
```diff
@@ -32,12 +35,27 @@ func apiRouter(s *Server) (chi.Router, error) {
 	r.Use(middleware.StripSlashes)
 	r.Use(render.SetContentType(render.ContentTypeJSON))
 
-	r.Get("/toolset", func(w http.ResponseWriter, r *http.Request) { toolsetHandler(s, w, r) })
-	r.Get("/toolset/{toolsetName}", func(w http.ResponseWriter, r *http.Request) { toolsetHandler(s, w, r) })
-
+	r.Get("/toolset", func(w http.ResponseWriter, r *http.Request) {
+		telemetry.OperationActiveUpDownCounter().Add(r.Context(), 1)
+		toolsetHandler(s, w, r)
+		telemetry.OperationActiveUpDownCounter().Add(r.Context(), -1)
```
kurtisvg: I think this is counting "current active requests" and not "requests served". "requests served" would only go up.

internal/server/api.go
```diff
@@ -48,10 +66,22 @@ func toolsetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
 		err := fmt.Errorf("Toolset %q does not exist", toolsetName)
 		_ = render.Render(w, r, newErrResponse(err, http.StatusNotFound))
 		return
 	}
+	telemetry.ToolsetGetCounter().Add(
+		r.Context(),
+		1,
+		metric.WithAttributes(attribute.String("toolbox.name", toolsetName)),
+		metric.WithAttributes(attribute.String("toolbox.operation.status", "success")),
+	)
```
kurtisvg: Can you combine these into a single "defer" function, and just set "status" before the return?

internal/server/api_test.go
```diff
@@ -56,6 +58,21 @@ func (t MockTool) Authorized(verifiedAuthSources []string) bool {
 	return true
 }
 
+func setupServer(ctx context.Context) (context.Context, Server, func(context.Context) error, error) {
+	otelShutdown, err := telemetry.SetupOTel(ctx, "0.0.0")
```
kurtisvg: 0.0.0?

internal/server/api_test.go
```diff
@@ -56,6 +58,21 @@ func (t MockTool) Authorized(verifiedAuthSources []string) bool {
...
+		return nil, Server{}, nil, err
+	}
+
+	testLogger, err := log.NewStdLogger(os.Stdout, os.Stderr, "info")
+	if err != nil {
+		return nil, Server{}, nil, err
+	}
+	server := Server{conf: ServerConfig{Version: "0.0.0"}, logger: testLogger}
+
+	return ctx, server, otelShutdown, nil
```
kurtisvg: Why do we return context here? Doesn't look changed?

internal/telemetry/metric.go
```diff
@@ -0,0 +1,111 @@
...
+	operationActiveName = "toolbox.server.operation.active"
+)
+
+// meter generates metrics for toolbox server
+var (
+	meter                        = otel.Meter("")
+	toolsetGetCounter            metric.Int64Counter
+	toolGetCounter               metric.Int64Counter
+	toolInvokeCounter            metric.Int64Counter
+	operationActiveUpDownCounter metric.Int64UpDownCounter
```
kurtisvg: This is creating global state, which might break if multiple commands are initialized. Can you put these in a struct and attach them to the server objects somewhere instead?

internal/server/api.go
Yuan325: yes, operation active is counting current active request

internal/server/api_test.go
Yuan325: Yes, that's the version string, changed that as a const instead.

internal/server/api.go
kurtisvg: Is their a reason we care about active requests? I think that total requests served would be the metric I would expect with "count".

internal/server/api.go
```diff
@@ -48,8 +65,23 @@ func apiRouter(s *Server) (chi.Router, error) {
...
 	toolsetName := chi.URLParam(r, "toolsetName")
 	toolset, ok := s.toolsets[toolsetName]
+	var err error
+	defer func() {
+		var status string
+		if err != nil {
+			status = "error"
+		} else {
+			status = "success"
+		}
```
kurtisvg: super-nit: 
```suggestion
		status := "success"
		if err != nil {
			status = "error"
		} 
```

internal/server/api_test.go
```diff
@@ -24,11 +25,14 @@ import (
 	"testing"
 
 	"github.com/googleapis/genai-toolbox/internal/log"
+	"github.com/googleapis/genai-toolbox/internal/telemetry"
 	"github.com/googleapis/genai-toolbox/internal/tools"
 )
 
 var _ tools.Tool = &MockTool{}
 
+const versionString = "0.0.0"
```
kurtisvg:
```suggestion
const fakeVersionString = "0.0.0"
```

internal/server/api_test.go
```diff
@@ -78,18 +97,27 @@ func TestToolsetEndpoint(t *testing.T) {
...
-		m, err := tc.Initialize("0.0.0", toolsMap)
+		m, err := tc.Initialize(versionString, toolsMap)
 		if err != nil {
 			t.Fatalf("unable to initialize toolset %q: %s", name, err)
 		}
 		toolsets[name] = m
 	}
 
-	testLogger, err := log.NewStdLogger(os.Stdout, os.Stderr, "info")
+	ctx := context.Background()
```
kurtisvg: nit: best practice to declare this at the top of the function, and defer a cancel with it (just in case)

internal/server/api_test.go
```diff
@@ -78,18 +97,27 @@ func TestToolsetEndpoint(t *testing.T) {
...
+	server, otelShutdown, err := setupServer(ctx)
```
kurtisvg: nit: as is, this helper function creates some indirection that I'm not sure is warranted. We're using a function here, but not for the logic about with tools/toolsets. Is it worth trying to move some of that logic and make it "setupDefaultServer"?

internal/server/server_test.go
```diff
@@ -50,12 +51,22 @@ func TestServe(t *testing.T) {
 		Port:    port,
 	}
 
+	customMetric, otelShutdown, err := telemetry.SetupOTel(ctx, "0.0.0")
```
kurtisvg: version string?

internal/telemetry/metric.go
```diff
@@ -0,0 +1,108 @@
...
+const (
+	InstrumentationName = "github.com/googleapis/genai-toolbox/internal/opentel"
+
+	toolsetGetCountName = "toolbox.server.toolset.get.count"
+	toolGetCountName    = "toolbox.server.tool.get.count"
+	toolInvokeCountName = "toolbox.server.tool.invoke.count"
+	operationActiveName = "toolbox.server.operation.active"
+)
+
+type Metric struct {
```
kurtisvg: should this be "metrics"? or maybe even "ServerMetrics" with a "metrics" fieldName? 

Also please add a comment to she struct

internal/telemetry/metric.go
```diff
@@ -0,0 +1,108 @@
...
+		toolsetGetCounter:            toolsetGetCounter,
+		toolGetCounter:               toolGetCounter,
+		toolInvokeCounter:            toolInvokeCounter,
+		operationActiveUpDownCounter: operationActiveUpDownCounter,
+	}
+	return customMetric, nil
+}
+
+// ToolsetGetCounter retrieves the toolsetGetCounter metric
+func (m *Metric) ToolsetGetCounter() metric.Int64Counter {
```
kurtisvg: any reason not to just access the struct directly?

internal/telemetry/metric.go
```diff
@@ -0,0 +1,108 @@
...
+	toolInvokeCountName = "toolbox.server.tool.invoke.count"
+	operationActiveName = "toolbox.server.operation.active"
+)
+
+type Metric struct {
+	meter                        metric.Meter
+	toolsetGetCounter            metric.Int64Counter
+	toolGetCounter               metric.Int64Counter
+	toolInvokeCounter            metric.Int64Counter
+	operationActiveUpDownCounter metric.Int64UpDownCounter
```
kurtisvg: nit: You _probably_ don't need to include Counter in the name here -- should be indicated by the type

cmd/root.go
```diff
@@ -162,6 +162,12 @@ func run(cmd *Command) error {
 			cmd.logger.Error(errMsg.Error())
 		}
 	}()
+	metrics, err := server.CreateCustomMetrics(cmd.Command.Version)
+	if err != nil {
+		errMsg := fmt.Errorf("unable to create custom metrics: %w", err)
+		cmd.logger.Error(errMsg.Error())
+		return errMsg
+	}
```
kurtisvg: nit: I would just move this into NewServer

internal/server/api.go
```diff
@@ -66,8 +81,23 @@ func toolGetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
 	toolName := chi.URLParam(r, "toolName")
 	tool, ok := s.tools[toolName]
+	var err error
+	defer func() {
+		var status string
+		if err != nil {
+			status = "error"
+		} else {
+			status = "success"
+		}
```
kurtisvg: nit: can you update these as well:
```suggestion
		status := "success"
		if err != nil {
			status = "error"
		}
```

internal/server/api.go
```diff
@@ -92,8 +122,23 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
```
kurtisvg: here too

PR #160, opened 2024-12-20
feat: add OTLP and GCP exporter to providers

Yuan325:
If user would like to use OTLP exporter with Collector, it could be set by adding the `--telemetry-otlp` flag. 

```
./toolbox --telemetry-otlp="https://127.0.0.1:4553"
```

If user would like to use the GCP exporter, it could be set by adding the `--telemetry-gcp` flag.

```
./toolbox --telemetry-gcp
```

If none of the flag above is added, no exporter will be added to the telemetry `Provider`s.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -101,8 +101,8 @@ func NewCommand(opts ...Option) *Command {
 	flags.StringVar(&cmd.tools_file, "tools_file", "tools.yaml", "File path specifying the tool configuration.")
 	flags.Var(&cmd.cfg.LogLevel, "log-level", "Specify the minimum level logged. Allowed: 'DEBUG', 'INFO', 'WARN', 'ERROR'.")
 	flags.Var(&cmd.cfg.LoggingFormat, "logging-format", "Specify logging format to use. Allowed: 'standard' or 'JSON'.")
-	flags.BoolVar(&cmd.cfg.TelemetryGCP, "telemetry-gcp", false, "Enable google cloud exporter.")
-	flags.StringVar(&cmd.cfg.TelemetryOTLP, "telemetry-otlp", "", "Define OTLP collector url for telemetry exports.")
+	flags.BoolVar(&cmd.cfg.TelemetryGCP, "telemetry-gcp", false, "Enable exporting directly to Google Cloud Monitoring.")
+	flags.StringVar(&cmd.cfg.TelemetryOTLP, "telemetry-otlp", "", "Enable exporting using OpenTelemetry Protocol (OTLP) to the specified endpoint (e.g. 'http://127.0.0.1:4318')")
 
 	// wrap RunE command so that we have access to original Command object
 	cmd.RunE = func(*cobra.Command, []string) error { return run(cmd) }


internal/opentel/opentel.go
```diff
@@ -109,10 +109,10 @@ func newResource(versionString string) (*resource.Resource, error) {
 	return r, nil
 }
 
-func newTraceProvider(r *resource.Resource) (*trace.TracerProvider, error) {
-	// TODO: stdout used for dev, will be updated to OTLP exporter
-	traceExporter, err := stdouttrace.New(
-		stdouttrace.WithPrettyPrint())
+func newTraceProvider(r *resource.Resource, ctx context.Context) (*trace.TracerProvider, error) {
+	// otlptracehttp provides an OTLP span exporter using HTTP with protobuf payloads.
```
kurtisvg: comments should go above the function.

internal/opentel/opentel.go
```diff
@@ -109,10 +109,10 @@ func newResource(versionString string) (*resource.Resource, error) {
...
+	// By default, the telemetry is sent to https://localhost:4318/v1/traces.
```
kurtisvg: Does this need to be configurable?

Yuan325: Updated.

internal/telemetry/telemetry.go
```diff
@@ -62,7 +67,7 @@ func SetupOTel(ctx context.Context, versionString string) (shutdown func(context
 		return
 	}
 
-	tracerProvider, err := newTracerProvider(res)
+	tracerProvider, err := newTracerProvider(res, context.Background(), cfg)
```
kurtisvg: is background the right context here?

internal/telemetry/telemetry.go
```diff
@@ -108,38 +113,51 @@ func newResource(versionString string) (*resource.Resource, error) {
 
 // newTracerProvider creates TracerProvider.
 // TracerProvider is a factory for Tracers and is responsible for creating spans.
-func newTracerProvider(r *resource.Resource) (*trace.TracerProvider, error) {
-	// TODO: stdout used for dev, will be updated to OTLP exporter
-	traceExporter, err := stdouttrace.New(
-		stdouttrace.WithPrettyPrint())
-	if err != nil {
-		return nil, err
+func newTracerProvider(r *resource.Resource, ctx context.Context, cfg server.ServerConfig) (*trace.TracerProvider, error) {
```
kurtisvg: nit: ctx first

internal/telemetry/telemetry.go
```diff
@@ -108,38 +113,51 @@ func newResource(versionString string) (*resource.Resource, error) {
...
+		// otlptracehttp provides an OTLP span exporter using HTTP with protobuf payloads.
+		// By default, the telemetry is sent to https://localhost:4318/v1/traces.
+		otlpExporter, err := otlptracehttp.New(ctx, otlptracehttp.WithEndpoint(cfg.TelemetryOTLP))
+		if err != nil {
+			return nil, err
+		}
+		traceOpts = append(traceOpts, trace.WithBatcher(otlpExporter))
 	}
+	if cfg.TelemetryGCP {
+		gcpExporter, err := texporter.New(texporter.WithProjectID(projectID))
```
kurtisvg: Do we actually need to specify ProjectID here? Will it default to something from the credentials if we dont?

cmd/root.go
```diff
@@ -101,6 +101,8 @@ func NewCommand(opts ...Option) *Command {
 	flags.StringVar(&cmd.tools_file, "tools_file", "tools.yaml", "File path specifying the tool configuration.")
 	flags.Var(&cmd.cfg.LogLevel, "log-level", "Specify the minimum level logged. Allowed: 'DEBUG', 'INFO', 'WARN', 'ERROR'.")
 	flags.Var(&cmd.cfg.LoggingFormat, "logging-format", "Specify logging format to use. Allowed: 'standard' or 'JSON'.")
+	flags.BoolVar(&cmd.cfg.TelemetryGCP, "telemetry-gcp", false, "Enable google cloud exporter.")
```
```suggestion
	flags.BoolVar(&cmd.cfg.TelemetryGCP, "telemetry-gcp", false, "Enable exporting directly to Google Cloud Monitoring.")
```

cmd/root.go
```diff
@@ -101,6 +101,8 @@ func NewCommand(opts ...Option) *Command {
...
+	flags.StringVar(&cmd.cfg.TelemetryOTLP, "telemetry-otlp", "", "Define OTLP collector url for telemetry exports.")
```
```suggestion
	flags.StringVar(&cmd.cfg.TelemetryOTLP, "telemetry-otlp", "", "Enable exporting using OpenTelemetry Protocol (OTLP) to the specified endpoint (e.g. 'http://127.0.0.1:4318')")
```

internal/server/api.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
-	"github.com/googleapis/genai-toolbox/internal/telemetry"
+	telemetrytrace "github.com/googleapis/genai-toolbox/internal/telemetry/trace"
```
kurtisvg, line 24: nit: unsure on why we are calling this telemetrytrace

Yuan325: cause there are another library name trace. Going to remove this in the next PR.

PR #163, opened 2024-12-23
chore: fix comments

twishabansal:


internal/server/api.go
```diff
@@ -43,7 +43,7 @@ func apiRouter(s *Server) (chi.Router, error) {
 	return r, nil
 }
 
-// toolInvokeHandler handles the request for information about a Toolset.
+// toolsetHandler handles the request for information about a Toolset.
```
anubhav756, line 46:
```suggestion
// Handles the request for information about a Toolset.
```

kurtisvg: The preferred Go style is actually "Comments should begin with the name of the thing being described and end in a period" -- https://go.dev/wiki/CodeReviewComments#comment-sentences

PR #165, opened 2024-12-23
tests(sdk): Added auth tests for toolbox SDK

twishabansal:
This PR is based on [#176](https://github.com/googleapis/genai-toolbox/pull/176)

The e2e tests cover the following cases:
[Added as a part of [#176](https://github.com/googleapis/genai-toolbox/pull/176)]
1. Loading a tool.
2. Loading a specific toolset.
3. Loading all toolsets.

[Newly added in this PR]
1. Running a tool with no required auth, with auth provided.
2. Running a tool with required auth:
      - No auth provided.
      - Wrong auth provided.
      - Correct auth provided.
3. Running a tool with a parameter that requires auth:
      - No auth provided.
      - Correct auth provided.
      - Auth provided does not contain the required field.


sdks/langchain/tests/conftest.py
```diff
@@ -74,6 +78,29 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
...
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        # If gcloud fails or is not found, use the metadata server
+        request = google.auth.transport.requests.Request()
+        credentials = compute_engine.IDTokenCredentials(
+            request=request,
+            target_audience=client_id,
+            use_metadata_identity_endpoint=True,
+        )
+        credentials.refresh(request)
+        return credentials.token
```
kurtisvg: We should prefer not relying on gcloud here. Can we use similar logic to what we did in the retrieval app: https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/blob/0f23a65d9ce5556f0448d1e3f046aca8e5b0878a/llm_demo/orchestrator/langchain_tools/tools.py#L35

PR #170, opened 2024-12-27
fix(toolbox-sdk)!: deprecate 'add_auth_headers' in favor of 'add_auth_tokens' 

anubhav756:
Client code should not be concerned with how authentication tokens are handled internally. This commit refactors method and parameter names to use the more abstract term "token" instead of "header,"  improving the developer experience and hiding implementation details.

### DIFF
--- a/sdks/langchain/src/toolbox_langchain_sdk/client.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/client.py
@@ -188,9 +188,7 @@ class ToolboxClient:
                     f"Some parameters of tool {tool_name} require authentication, but no valid auth sources are registered. Please register the required sources before use."
                 )
 
-    @deprecated(
-        "Please use `add_auth_token` instead."
-    )
+    @deprecated("Please use `add_auth_token` instead.")
     def add_auth_header(
         self, auth_source: str, get_id_token: Callable[[], str]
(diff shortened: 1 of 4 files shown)


sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -183,13 +184,19 @@ def _process_auth_params(self, manifest: ManifestSchema) -> None:
             # If none of the permitted auth sources of a parameter are
             # registered, raise a warning message to the user.
             if not self._validate_auth(tool_name):
-                warnings.warn(
+                warn(
                     f"Some parameters of tool {tool_name} require authentication, but no valid auth sources are registered. Please register the required sources before use."
                 )
 
+    @deprecated(
+        "Method `add_auth_header` is deprecated. Please use `add_auth_token` instead."
```
kurtisvg: It looks like the warning for this is something like (so we can remove the first part):
```
client.py:192: DeprecationWarning: Call to deprecated method some_old_method.
(Method `add_auth_header` is deprecated. Please use `add_auth_token` instead.) -- Deprecated since version 1.3.0.
  add_auth_header(...)
```
```suggestion
        Please use `add_auth_token` instead."
```

PR #171, opened 2024-12-27
feat(langchain-sdk)!: Migrate existing state and APIs to a tools level class.

anubhav756:
# `ToolboxTool` Class

The newly implemented `ToolboxTool` class manages tool state and supports features like bound params and OAuth.

We have also added the logic as well as state related to bound params to `ToolboxTool` in #192.

`ToolboxTool` follows a functional approach, meaning it ensures that the internal tool state remains unchanged and a new copy of the tool is updated with the new values of auth tokens (or bound params from #192).

## `ToolboxTool` Class Diagram

To better explain, here's a visual diagram of the ToolboxTool class.

The boxes have member function names, and the "red" box means a function is mutating the internal state, while "green" means it does not change the internal state (like a C++ `const` function or JS immutable function). The red box also shows the member variable they mutated.

Image of a diagram showing the `ToolboxTool` class with member functions and their colors indicating whether they mutate the internal state

## Constructor

From the diagram above, it is evident that all the state changes, or rather setting of the state, is done in the class constructor.

For further reference, here's a visual explanation of the functionality of the constructor showing which params we add as auth tokens, which params we add to the underlying schema, and for what the error/warnings are thrown.

Image of a diagram showing the functionality of the `ToolboxTool` constructor

> [!NOTE]
> Documentation updates are done by #193 and the corresponding changes to the LlamaIndex SDK are done by #203.

### DIFF
--- a/sdks/langchain/tests/test_e2e.py
+++ b/sdks/langchain/tests/test_e2e.py
@@ -146,7 +146,10 @@ class TestE2EClient:
     async def test_run_tool_param_auth_no_auth(self, toolbox):
         """Tests running a tool with a param requiring auth, without auth."""
         tool = await toolbox.load_tool("get-row-by-email-auth")
-        with pytest.raises(PermissionError, match="Parameter\(s\) `email` of tool get-row-by-email-auth require authentication\, but no valid authentication sources are registered\. Please register the required sources before use\."):
+        with pytest.raises(
+            PermissionError,
+            match="Parameter\(s\) `email` of tool get-row-by-email-auth require authentication\, but no valid authentication sources are registered\. Please register the required sources before use\.",
+        ):
             await tool.ainvoke({})
 
     @pytest.mark.asyncio


sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,312 @@
...
+                continue
+
+            # Check if the parameter is missing from the tool schema.
+            if not param_name in [param.name for param in self._schema.parameters]:
+                missing_params.append(param_name)
+                continue
+
+            self._bound_params[param_name] = param_value
+
+        self._remove_bound_params()
```
twishabansal:
```suggestion
        # Bound parameters are handled internally, so remove them from the schema 
        # to prevent validation errors and present a cleaner schema to the user.
        self._remove_bound_params()
```

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,312 @@
...
+
+        A tool is considered authenticated if all of its parameters meet at
+        least one of the following conditions:
+
+            * The parameter has at least one registered authentication source.
+            * The parameter requires no authentication.
+
+        Args:
+            strict: If True, raises a PermissionError if any required
+                authentication sources are not registered. If False, only issues
```
twishabansal, line 125: What's the use of the false case? Is it to allow default values for authenticated params?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,312 @@
...
+        for auth_source, get_id_token in auth_tokens.items():
+
+            # Check if the authentication source is already registered.
+            if auth_source in self._auth_tokens:
+                dupe_sources.append(auth_source)
+                continue
+
+            self._auth_tokens[auth_source] = get_id_token
+
+        self._process_auth_params()
```
twishabansal:
```suggestion
        # Remove auth params from the schema to prevent data validation
        # errors since their values are inferred by the Toolbox service, not
        # provided by the user.
        self._process_auth_params()
```

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: We require the user to provide all the necessary auth tokens while invoking the tool. However, during initialization, we just want to warn the user in case a tool is loaded and we detect some missing authentication (eg. when the user does not provide all necessary auth tokens in the `auth_tokens` field of the `ToolboxTool` constructor).

During both tool invocation as well as initialization, we use the same method to validate whether all auth is provided or not (via this `_validate_auth` method). So the false case is applicable during the tool initialization.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Done. Thanks for the suggestion.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Added. Thanks for the suggestion!

anubhav756:
> Are you planning to update the e2e tests to include bound params cases?

Yes. I will add those e2e test cases in a future PR.

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -85,158 +85,32 @@ async def _load_toolset_manifest(
...
-        """
-        self._id_token_getters[auth_source] = get_id_token
-
     async def load_tool(
         self,
         tool_name: str,
         auth_tokens: dict[str, Callable[[], str]] = {},
         auth_headers: Optional[dict[str, Callable[[], str]]] = None,
-    ) -> StructuredTool:
+    ) -> ToolboxTool:
```
kurtisvg, line 102: This makes the PR a breaking change. You'll need to use `feat!`.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+        Args:
+            name: The name of the tool.
+            schema: The tool schema.
+            url: The base URL of the Toolbox service.
+            session: The HTTP client session.
+            auth_tokens: A mapping of authentication source names to functions
+                that retrieve ID tokens.
+        """
+        if not isinstance(schema, ToolSchema):
+            schema = ToolSchema(**schema)
```
kurtisvg, line 63: What's this attempting to accomplish? Can we add a comment?

If the type is `ToolSchema`, isn't this always a ToolSchema?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+
+        self._name: str = name
+        self._schema: ToolSchema = schema
+        self._url: str = url
+        self._session: ClientSession = session
+        self._auth_tokens: dict[str, Callable[[], str]] = {}
+        self._auth_params: dict[str, list[str]] = {}
+
+        self.add_auth_tokens(auth_tokens)
+        self._validate_auth(strict=False)
```
kurtisvg: should this be self.__validate_auth?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        dupe_sources: list[str] = []
+        for auth_source, get_id_token in auth_tokens.items():
+
+            # Check if the authentication source is already registered.
+            if auth_source in self._auth_tokens:
+                dupe_sources.append(auth_source)
+                continue
+
+            self._auth_tokens[auth_source] = get_id_token
```
kurtisvg: IIUC, this is checking to see if the user added an auth source over an existing auth_source name. Are we sure we don't want the user to be able to override this?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Remove auth params from the schema to prevent data validation errors
+        # since their values are inferred by the Toolbox service, not provided
+        # by the user.
+        self._process_auth_params()
```
kurtisvg: Should this be __?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+        self._schema: ToolSchema = schema
+        self._url: str = url
+        self._session: ClientSession = session
+        self._auth_tokens: dict[str, Callable[[], str]] = {}
+        self._auth_params: dict[str, list[str]] = {}
+
+        self.add_auth_tokens(auth_tokens)
+        self._validate_auth(strict=False)
+
+    def _update_params(self, params: list[ParameterSchema]) -> None:
```
kurtisvg: for a number of these fields, I think we should consider strongly if they should be single unders `_function_name` or double unders `__function_name`.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+        `_auth_params` for efficient validation in `_validate_auth`.
+        """
+        non_auth_params: list[ParameterSchema] = []
+        for param in self._schema.parameters:
+            if param.authSources:
+                self._auth_params[param.name] = param.authSources
+            else:
+                non_auth_params.append(param)
+
+        self._update_params(non_auth_params)
```
kurtisvg: I _think_ this is only used in one place, and only uses `schema.parameters` as input. Can it be converted into something that's more functional?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+        args schema.
+
+        Args:
+            params: The new list of parameters.
+        """
+        self._schema.parameters = params
+
+        self.args_schema = _schema_to_model(
+            model_name=self._name, schema=self._schema.parameters
+        )
```
kurtisvg: Does this need a function? It's only 1 line

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+        for param_name, auth_sources in self._auth_params.items():
+            found_match = False
+            for registered_auth_source in self._auth_tokens:
+                if registered_auth_source in auth_sources:
+                    found_match = True
+                    break
+            if not found_match:
+                unauth_params.append(param_name)
+
+        if not unauth_params:
```
kurtisvg: Generally it's consider a readibility improvement to keep the happy path to the left (sometimes called a guard clause): https://en.wikipedia.org/wiki/Guard_(computer_science)

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+            return
```
kurtisvg: I feel like this logic seems overly complex. Can we do something like:

```suggestion
        params_missing_auth: list[str] = []

       # check each parameter for at least 1 required auth_source
        for param_name, required_auth in self._auth_params.items():
            has_auth = False
            for src in required_auth:
                # find first auth_source that is specified
                if src in self._auth_tokens:
                    found = true
                    continue
            if has_auth:
                params_missing_auth += param_name
                     
        if params_missing_auth:
            # throw error here
            
         # happy path here
```

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -0,0 +1,197 @@
...
+
+        message = f"Parameter(s) `{', '.join(unauth_params)}` of tool {self._name} require authentication, but no valid authentication sources are registered. Please register the required sources before use."
+
+        if strict:
+            raise PermissionError(message)
```
kurtisvg: Is the point of the strict behavior to just validate the auth before sending?

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: Makes sense. Thanks for pointing that out!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: To maintain consistency with our handling of Bound Params (as detailed [here](https://docs.google.com/document/d/1zNgcu7_x-uUtQwWgrVNk0E3GX4dUxqnEHlGm-V9KzMs/edit?resourcekey=0-szIFxw2LCVQjXPO8hDreWg&tab=t.0#heading=h.6mnm163sn17w)), I propose we disallow overriding auth token values.

Since auth params are essentially a specialized type of bound param, this approach seems like a more intuitive and unified user experience.

Thoughts?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: I'm not sure I fully understand your suggestion to make this code "more functional".

For example, do you suggest:
* Creating an independent function in `utils.py`?
* Inlining this function since (currently) it's only used in one place?
* Reducing the use of `for`, `if`, and `else` and using higher-order functions like `map`, `filter`, and `reduce` instead?
  * Something like:
```
def __process_auth_params(self) -> None:
    auth_params = {param.name: param.authSources for param in self._schema.parameters if param.authSources}
    non_auth_params: List[ParameterSchema] = list(filter(lambda param: not param.authSources, self._schema.parameters))

    self._auth_params.update(auth_params)
    self._update_params(non_auth_params)
```

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: In the following PR (#192), we reuse this function to update the params while binding params [here](https://github.com/googleapis/genai-toolbox/pull/192/files#diff-e6aabf3f0b08b0964204e7069b24540610f7832b583506774df1b4bad297518cR182).

It's actually 2 lines which are kind of tied together, first line updates the tool level state of `self._schema` and the second line updates the orchestration specific schema.

We keep the first one as an orchestration agnostic state, and the second one is what actually validates the input data, so using this function ensures they both are in-sync.

Thoughts?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Good point! I originally intended to make an early exit and reduce the nesting, but I see how that could make things harder to read. Thanks for pointing that out!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Thanks for the helpful suggestions! I especially appreciate the shorter variable names and the added comments, which definitely improve readability. Flipping the for loops to bring `required_auth` closer to its relevant logic is a smart move too.

I've incorporated your changes, along with a few minor tweaks of my own [[screenshot](https://screenshot.googleplex.com/59aA3GEHDBYjRXg)].

Let me know what you think!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: To improve the end-user experience and provide more informative error messages, I've added this check to ensure the `schema` parameter is properly validated, even when provided as a dictionary (added the comment [here](https://github.com/googleapis/genai-toolbox/pull/171/files#diff-e6aabf3f0b08b0964204e7069b24540610f7832b583506774df1b4bad297518cR36-R39)).

Previously, if a user passed a dictionary they would encounter a generic error like `'dict' object has no attribute 'description'`.

This change would not only enable end-users to pass in the `schema` as a dictionary, but also provide more specific and helpful error messages in case that dictionary has wrong keys for instance, like the following:
```
E           pydantic_core._pydantic_core.ValidationError: 1 validation error for ToolSchema
E           description
E             Field required [type=missing, input_value={'bla': 'Test Tool...scription': 'Param 2'}]}, input_type=dict]
E               For further information visit https://errors.pydantic.dev/2.10/v/missing
```

So this makes the class more flexible and user-friendly without impacting internal functionality.

Thoughts?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Agree. Thanks for the suggestion!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Agree. Thanks for the suggestion!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: That is correct. This ensures consistent validation logic during both initialization and invocation.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Agree. Thanks for the suggestion!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
kurtisvg: I think I have 2 concerns here:
1. I would prefer _process_auth_params not mutate the underlying object. I'd prefer that mutation operations return new copies of the object with the state changed. 
2. This function (`_process_auth_params`) is attached to the object and relies on fields to do some filtering before actually mutating the object. Can we update the function to just be a helper instead, and not mutate anything outside of in/out?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
kurtisvg: SGTM

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Thanks for the suggestion Kurtis!

I've updated this class to be more functional.

To better explain, here's a visual before and after view my recent change. It has boxes which are member function names.

Note that "red" box means a function mutating, and these boxes also show the member variable they mutated.

# Before
image

# After
image

For further reference, here's a visual explanation of the functionality of the constructor showing which params we add as auth tokens, which params we add to the underlying schema, and for what the error/warnings are thrown.

image

Hope this looks good.

PR #172, opened 2024-12-27
ci(auth): Add OAuth integration test to Cloud SQL

duwenxin99:
1. Add generic integration test helpers for authenticated parameters and auth-required tool invocations.
2. Call generic test helpers as Cloud SQL integration tests.

### DIFF
--- a/tests/auth_test.go
+++ b/tests/auth_test.go
@@ -101,7 +101,6 @@ func RunGoogleAuthenticatedParameterTest(t *testing.T, sourceConfig map[string]a
 			},
 		},
 	}
-
 	// Initialize a test command
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -56,7 +56,8 @@ steps:
       - "ALLOYDB_POSTGRES_INSTANCE=$_ALLOYDB_POSTGRES_INSTANCE"
       - "ALLOYDB_POSTGRES_DATABASE=$_DATABASE_NAME"
       - "ALLOYDB_POSTGRES_REGION=$_REGION"
-    secretEnv: ["ALLOYDB_POSTGRES_USER", "ALLOYDB_POSTGRES_PASS"]
+      - "SERVICE_ACCOUNT_EMAIL=$SERVICE_ACCOUNT_EMAIL"
+    secretEnv: ["ALLOYDB_POSTGRES_USER", "ALLOYDB_POSTGRES_PASS", "CLIENT_ID"]
     volumes:
       - name: "go"
         path: "/gopath"
@@ -85,17 +86,17 @@ steps:
 
   - id: "spanner"
     name: golang:1
-    waitFor: ['install-dependencies']
+    waitFor: ["install-dependencies"]
     entrypoint: /bin/bash
     env:
-      - 'GOPATH=/gopath'
+      - "GOPATH=/gopath"
       - "SPANNER_PROJECT=$PROJECT_ID"
       - "SPANNER_DATABASE=$_DATABASE_NAME"
       - "SPANNER_INSTANCE=$_SPANNER_INSTANCE"
     volumes:
-      - name: 'go'
-        path: '/gopath'
-    args: 
+      - name: "go"
+        path: "/gopath"
+    args:
       - -c
       - |
         go test -race -v -tags=integration,spanner ./tests
--- a/tests/alloydb_pg_integration_test.go
+++ b/tests/alloydb_pg_integration_test.go
@@ -121,7 +121,7 @@ func initAlloyDBPgConnectionPool(project, region, cluster, instance, ip_type, us
 }
 
 func TestAlloyDBSimpleToolEndpoints(t *testing.T) {
-	requireAlloyDBPgVars(t)
+	sourceConfig := requireAlloyDBPgVars(t)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
@@ -130,16 +130,7 @@ func TestAlloyDBSimpleToolEndpoints(t *testing.T) {
 	// Write config into a file and pass it to command
 	toolsFile := map[string]any{
 		"sources": map[string]any{
-			"my-pg-instance": map[string]any{
-				"kind":     "alloydb-postgres",
-				"project":  ALLOYDB_POSTGRES_PROJECT,
-				"instance": ALLOYDB_POSTGRES_INSTANCE,
-				"cluster":  ALLOYDB_POSTGRES_CLUSTER,
-				"region":   ALLOYDB_POSTGRES_REGION,
-				"database": ALLOYDB_POSTGRES_DATABASE,
-				"user":     ALLOYDB_POSTGRES_USER,
-				"password": ALLOYDB_POSTGRES_PASS,
-			},
+			"my-pg-instance": sourceConfig,
 		},
 		"tools": map[string]any{
 			"my-simple-tool": map[string]any{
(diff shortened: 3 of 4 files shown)


integration.cloudbuild.yaml
```diff
@@ -43,6 +43,22 @@ steps:
       - -c
       - |
         go test -race -v -tags=integration,alloydb ./tests
+  
+  - id: "auth test"
```
kurtisvg:
```suggestion
  - id: "auth-test"
```

internal/sources/cloudsqlpg/cloud_sql_pg.go
```diff
@@ -92,7 +92,7 @@ func getDialOpts(ip_type string) ([]cloudsqlconn.DialOption, error) {
 	}
 }
 
-func initCloudSQLPgConnectionPool(project, region, instance, ip_type, user, pass, dbname string) (*pgxpool.Pool, error) {
+func InitCloudSQLPgConnectionPool(project, region, instance, ip_type, user, pass, dbname string) (*pgxpool.Pool, error) {
```
kurtisvg: If this really needs to be public, it should have a comment (but does it really need to be public?)

tests/auth_test.go
```diff
@@ -17,23 +17,63 @@
...
+	CLOUD_SQL_POSTGRES_PROJECT  = os.Getenv("CLOUD_SQL_POSTGRES_PROJECT")
+	CLOUD_SQL_POSTGRES_REGION   = os.Getenv("CLOUD_SQL_POSTGRES_REGION")
+	CLOUD_SQL_POSTGRES_INSTANCE = os.Getenv("CLOUD_SQL_POSTGRES_INSTANCE")
+	CLOUD_SQL_POSTGRES_DATABASE = os.Getenv("CLOUD_SQL_POSTGRES_DATABASE")
+	CLOUD_SQL_POSTGRES_USER     = os.Getenv("CLOUD_SQL_POSTGRES_USER")
+	CLOUD_SQL_POSTGRES_PASS     = os.Getenv("CLOUD_SQL_POSTGRES_PASS")
+	SERVICE_ACCOUNT_EMAIL       = os.Getenv("SERVICE_ACCOUNT_EMAIL")
+)
+
+func requireCloudSQLPgVars(t *testing.T) {
```
kurtisvg: To me, this function indicates that these are postgres tests. Any reason to just not integrate them with the existing pg tests?

tests/auth_test.go
```diff
@@ -17,23 +17,63 @@
...
+		t.Fatal("'CLOUD_SQL_POSTGRES_PASS' not set")
+	}
+}
+
 // Get a Google ID token
 func getGoogleIdToken(audience string) (string, error) {
 	// For local testing
 	cmd := exec.Command("gcloud", "auth", "print-identity-token")
 	output, err := cmd.Output()
 	if err == nil {
```
kurtisvg, line 45: nit: You can improve readability by keeping "the happy path to the left" here (https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)

tests/auth_test.go
```diff
@@ -108,3 +148,316 @@ func TestGoogleAuthVerification(t *testing.T) {
...
+	statement := `
+		INSERT INTO auth_table (name, email) 
+		VALUES ($1, $2), ($3, $4)
+	`
+	params := []any{"Alice", SERVICE_ACCOUNT_EMAIL, "Jane", "janedoe@gmail.com"}
+	_, err = pool.Query(context.Background(), statement, params...)
+	if err != nil {
+		t.Fatalf("unable to insert test data: %s", err)
+	}
+	defer pool.Exec(context.Background(), `DROP TABLE auth_table;`)
```
kurtisvg: This might be better to move into a one-time setup -- e.g. https://stackoverflow.com/a/34102842/9414803

tests/auth_test.go
```diff
@@ -108,3 +148,316 @@ func TestGoogleAuthVerification(t *testing.T) {
...
+
+	var args []string
+
+	cmd, cleanup, err := StartCmd(ctx, toolsFile, args...)
+	if err != nil {
+		t.Fatalf("command initialization returned an error: %s", err)
+	}
+	defer cleanup()
+
+	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
kurtisvg: nit: 
```suggestion
	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```

internal/sources/cloudsqlpg/cloud_sql_pg.go
duwenxin99: To use it from the test package, it needs to be exportable. I think being able to reuse this function makes writing our integration tests a lot easier.

tests/auth_test.go
duwenxin99: integrated

tests/auth_test.go
duwenxin99: moved

tests/auth_test.go
duwenxin99: updated

tests/auth_test.go
```diff
@@ -17,94 +17,322 @@
...
+	"time"
 
 	"testing"
-
-	"github.com/googleapis/genai-toolbox/internal/auth"
-	"github.com/googleapis/genai-toolbox/internal/auth/google"
 )
 
+var SERVICE_ACCOUNT_EMAIL = os.Getenv("SERVICE_ACCOUNT_EMAIL")
+var clientId string = "32555940559.apps.googleusercontent.com"
```
Yuan325: Should we have this as a secret?

tests/auth_test.go
```diff
@@ -17,94 +17,322 @@
...
+	case strings.EqualFold(toolKind, "postgres-sql"):
+		statement = "SELECT * FROM auth_table WHERE email = $1;"
+	default:
+		t.Fatalf("invalid tool kind: %s", toolKind)
+	}
+
+	// Write config into a file and pass it to command
+	toolsFile := map[string]any{
+		"sources": map[string]any{
+			"my-pg-instance": sourceConfig,
```
Yuan325:
```suggestion
			"my-source-instance": sourceConfig,
```

not necessarily pg. Or just `my-instance` is okay too~

tests/auth_test.go
```diff
@@ -17,94 +17,322 @@
...
+		"authSources": map[string]any{
+			"my-google-auth": map[string]any{
+				"kind":     "google",
+				"clientId": clientId,
+			},
+		},
+		"tools": map[string]any{
+			"my-auth-tool": map[string]any{
+				"kind":        toolKind,
+				"source":      "my-pg-instance",
```
Yuan325: the source name here will need to be updated as well.

tests/auth_test.go
duwenxin99: This is the public client ID of google cloud SDK: https://github.com/twistedpair/google-cloud-sdk/blob/60ac4f7cdec0d0897386657ad59cabd31f719fff/google-cloud-sdk/lib/googlecloudsdk/core/config.py#L182 so it should be ok.

kurtisvg: There is a TODO there to move it into a secret, so maybe we should just go ahead and do that?

duwenxin99: Oops yea I'll create a secret for it.

tests/auth_test.go
```diff
@@ -17,94 +17,322 @@
...
-		}
-		defer resp.Body.Close()
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return "", err
-		}
-		return string(body), nil
+		return strings.TrimSpace(string(output)), nil
+	}
+	// Cloud Build testing
```
kurtisvg: I don't know what this comment means. Can we add more context?

tests/auth_test.go
```diff
@@ -17,94 +17,322 @@
...
+	url := "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=" + audience
```
kurtisvg: Is this the best way to get an IdToken? I also see a library here: https://pkg.go.dev/google.golang.org/api/idtoken#NewTokenSource

tests/auth_test.go
```diff
@@ -17,94 +17,322 @@
...
+	}
+	return string(body), nil
 }
 
-func TestGoogleAuthVerification(t *testing.T) {
-	clientId := "32555940559.apps.googleusercontent.com"
-	tcs := []struct {
-		authSource auth.AuthSource
-		isErr      bool
+func GoogleAuthenticatedParameterTestHelper(t *testing.T, sourceConfig map[string]any, toolKind string) {
```
kurtisvg:
```suggestion
func RunGoogleAuthenticatedParameterTest(t *testing.T, sourceConfig map[string]any, toolKind string) {
```

tests/auth_test.go
duwenxin99: updated

tests/auth_test.go
duwenxin99: added

tests/auth_test.go
duwenxin99: updated

tests/auth_test.go
duwenxin99: updated

PR #176, opened 2024-12-30
tests(sdk): e2e tests for langchain sdk

twishabansal:
This PR contains toolbox Set up and e2e tests for the following functionality:

- Load tools
- Load toolsets

Follow up PRs would include testing for
- Running tools
- Auth for tools
- Auth for tool parameters

### DIFF
--- a/sdks/langchain/tests/test_e2e.py
+++ b/sdks/langchain/tests/test_e2e.py
@@ -22,7 +22,8 @@ This file covers the following use cases:
 4. Running a tool with no required auth, with auth provided.
 5. Running a tool with required auth:
     a. No auth provided.
-    b. Wrong auth provided.
+    b. Wrong auth provided: The tool requires a different authentication
+                            than the one provided.
     c. Correct auth provided.
 6. Running a tool with a parameter that requires auth:
     a. No auth provided.


twishabansal:
Will add auth tests in a follow up PR

sdks/langchain/tests/test_e2e.py
```diff
@@ -0,0 +1,45 @@
+import conftest as conftest
```
anubhav756: This looks a bit redundant, like `import x as x` could simply be `import x`. In this specific case however, this import isn't needed at all since pytest will automatically discover them (see [here](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files)).

sdks/langchain/tests/test_e2e.py
```diff
@@ -0,0 +1,45 @@
...
+    @pytest_asyncio.fixture(scope="function")
+    async def toolbox(self):
+        toolbox = ToolboxClient("http://localhost:5000")
+        yield toolbox
+        await toolbox.close()
+
+    @pytest.mark.asyncio
+    async def test_load_tool(self, toolbox):
+        tool = await toolbox.load_tool("get-n-rows")
+        response = await tool.arun({"num_rows": "2"})
```
anubhav756:
```suggestion
        response = await tool.ainvoke({"num_rows": "2"})
```
`arun` is deprecated.

sdks/langchain/tests/test_e2e.py
twishabansal: Done.

sdks/langchain/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+    name = f"projects/{project_id}/secrets/{secret_id}/versions/{version_id}"
+    response = client.access_secret_version(request={"name": name})
+    return response.payload.data.decode("UTF-8")
+
+
+def create_tmpfile(content: str) -> str:
+    """Creates a temporary file with the given content."""
+    with tempfile.NamedTemporaryFile(delete=False, mode="w") as tmpfile:
+        tmpfile.write(content)
+        return tmpfile.name
```
kurtisvg, line 56: Does this work? I would somewhat expect this file to be closed when the with statement returns

sdks/langchain/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+        # Make toolbox executable
+        os.chmod("toolbox", 0o700)
+        # Run toolbox binary
+        toolbox_server = subprocess.Popen(
+            ["./toolbox", "--tools_file", tools_file_path]
+        )
+        # Wait for server to start
+        time.sleep(10)
+        print("Checking if toolbox is successfully started...")
+        assert not toolbox_server.poll(), "Toolbox server failed to start"
```
kurtisvg: nit: Can we optimize this -- rather than sleep for 10s, can we sleep smaller amounts in a loop until we see Toolbox as started?

sdks/langchain/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+    tools_manifest = access_secret_version(
+        project_id=project_id, secret_id="sdk_testing_tools"
+    )
+    tools_file_path = create_tmpfile(tools_manifest)
+    yield tools_file_path
+    os.remove(tools_file_path)
+
+
+@pytest_asyncio.fixture(scope="session")
+def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator:
```
kurtisvg: nit: should it be Generator[SomeType]?

sdks/langchain/tests/conftest.py
kurtisvg: "It will be destroyed as soon as it is closed (including an implicit close when the object is garbage collected)." -- https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile

sdks/langchain/tests/conftest.py
twishabansal: Fixed.

sdks/langchain/tests/conftest.py
twishabansal: Done.

sdks/langchain/tests/conftest.py
twishabansal: The delete=False option does not delete the temp file upon closing: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile.

We're cleaning up this file in the tools_file_path fixture (line 113)

sdks/langchain/tests/test_e2e.py
```diff
@@ -12,8 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""End-to-end tests for the toolbox SDK interacting with the toolbox server.
+
+This file covers the following use cases:
+
+1. Loading a tool.
+2. Loading a specific toolset.
+3. Loading all toolsets.
```
anubhav756: nit: This scenario loads the default toolset, which currently has all the tools.

So “loading all tools” seems more accurate than “loading all toolsets”, and “loading the default toolset” seems yet more accurate to me.

sdks/langchain/tests/test_e2e.py
```diff
@@ -12,8 +12,27 @@
...
+4. Running a tool with no required auth, with auth provided.
+5. Running a tool with required auth:
+    a. No auth provided.
+    b. Wrong auth provided.
```
anubhav756: Would it be possible to make this a bit more clear, like “Auth provided other than the ones configured” or something like that?

sdks/langchain/tests/test_e2e.py
```diff
@@ -12,8 +12,27 @@
...
+    c. Correct auth provided.
+6. Running a tool with a parameter that requires auth:
+    a. No auth provided.
+    b. Correct auth provided.
+    c. Auth provided does not contain the required field.
```
anubhav756:
```suggestion
    c. Auth provided does not contain the required claim.
```

sdks/langchain/tests/test_e2e.py
twishabansal: Done

PR #178, opened 2024-12-30
fix: handle shutdown gracefully

kurtisvg:
Adds logic to make the server shutdown gracefully, including better respecting cancelled contexts and providing up to 10 seconds to finish current connections.

### DIFF
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -102,7 +102,7 @@ options:
   substitutionOption: 'ALLOW_LOOSE'
   dynamicSubstitutions: true
   pool:
-    name: projects/$PROJECT_ID/locations/us-central1/workerPools/alloy-private-pool # Necessary for VPC network connection
+    name: projects/$PROJECT_ID/locations/us-central1/workerPools/integration-testing # Necessary for VPC network connection
 
 substitutions:
   _DATABASE_NAME: test_database


cmd/root.go
```diff
@@ -132,6 +135,26 @@ func run(cmd *Command) error {
...
 	defer cancel()
 
+	// watch for sigterm / sigint signals
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		var s os.Signal
+		select {
+		case <-ctx.Done():
+			// this should only happen when the context supplied in tests in canceled
```
duwenxin99:
```suggestion
			// this should only happen when the context supplied in tests is canceled
```

internal/server/server.go
```diff
@@ -26,16 +26,18 @@ import (
...
 )
 
 // Server contains info for running an instance of Toolbox. Should be instantiated with NewServer().
 type Server struct {
-	conf   ServerConfig
-	root   chi.Router
-	logger logLib.Logger
+	version string
+	srv     *http.Server
+	l       net.Listener
```
duwenxin99: Using `l` as listener here but as logger in NewServer(l log.Logger) is less readable. Maybe `listener` instead?

internal/server/server.go
```diff
@@ -131,42 +145,47 @@ func NewServer(ctx context.Context, cfg ServerConfig, log logLib.Logger) (*Serve
...
 	}
-	log.InfoContext(ctx, fmt.Sprintf("Initialized %d toolsets.", len(toolsetsMap)))
+	l.InfoContext(ctx, fmt.Sprintf("Initialized %d toolsets.", len(toolsetsMap)))
+
+	addr := net.JoinHostPort(cfg.Address, strconv.Itoa(cfg.Port))
+	srv := &http.Server{Addr: addr, Handler: r}
 
-	s := &Server{
-		conf:        cfg,
+	s = &Server{
```
duwenxin99: I think we need to set the `version`  here to `cfg.Version` otherwise it's lost

cmd/root.go
```diff
@@ -132,6 +135,26 @@ func run(cmd *Command) error {
 	ctx, cancel := context.WithCancel(cmd.Context())
 	defer cancel()
 
+	// watch for sigterm / sigint signals
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
```
duwenxin99, line 141: QQ: why do we need to watch for interrupt signals before the server even gets started?

kurtisvg: We want to start watching as soon as the command starts to be sure that if something takes a long time while the command runs, it doesn't hold up the sigterm signal.

internal/server/server.go
kurtisvg: Good catch!

kurtisvg: Added a test for this as well

PR #185, opened 2025-01-06
ci(auth): Add OAuth integration tests to AlloyDB

duwenxin99:
Call the generic auth test helpers in AlloyDB integration tests.

### DIFF
--- a/tests/alloydb_pg_integration_test.go
+++ b/tests/alloydb_pg_integration_test.go
@@ -334,7 +334,8 @@ func setupAlloyDBAuthTest(t *testing.T, ctx context.Context, tableName string) f
 	_, err = pool.Query(ctx, fmt.Sprintf(`
 		CREATE TABLE %s (
 			id SERIAL PRIMARY KEY,
-			name TEXT
+			name TEXT,
+			email TEXT
 		);
 	`, tableName))
 	if err != nil {


tests/alloydb_pg_integration_test.go
```diff
@@ -187,3 +190,64 @@ func TestAlloyDBPostgres(t *testing.T) {
 		})
 	}
 }
+
+// Set up auth test database table
+func setupAuthTest(t *testing.T) func(*testing.T) {
+	// set up testt
```
kurtisvg:
```suggestion
	// set up test
```

tests/alloydb_pg_integration_test.go
```diff
@@ -187,3 +190,64 @@ func TestAlloyDBPostgres(t *testing.T) {
...
+	pool, err := alloydbpg.InitAlloyDBPgConnectionPool(ALLOYDB_POSTGRES_PROJECT, ALLOYDB_POSTGRES_REGION, ALLOYDB_POSTGRES_CLUSTER, ALLOYDB_POSTGRES_INSTANCE, "public", ALLOYDB_POSTGRES_USER, ALLOYDB_POSTGRES_PASS, ALLOYDB_POSTGRES_DATABASE)
+	if err != nil {
+		t.Fatalf("unable to create ALLOYDB connection pool: %s", err)
+	}
+
+	err = pool.Ping(context.Background())
```
kurtisvg: should use a shared context -- maybe from the original test with a timeout?

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -93,7 +93,7 @@ func getDialOpts(ip_type string) ([]alloydbconn.DialOption, error) {
 	}
 }
 
-func initAlloyDBPgConnectionPool(project, region, cluster, instance, ip_type, user, pass, dbname string) (*pgxpool.Pool, error) {
+func InitAlloyDBPgConnectionPool(project, region, cluster, instance, ip_type, user, pass, dbname string) (*pgxpool.Pool, error) {
```
kurtisvg: nit: I'm not entirely sold on exporting this -- this might be a an example of [A little copying is better than a little dependency.](https://go-proverbs.github.io/)

tests/alloydb_pg_integration_test.go
duwenxin99: updated

internal/sources/alloydbpg/alloydb_pg.go
duwenxin99: My concern is that we will have to copy over two helper functions (InitAlloyDBPgConnectionPool() and getDialOpts()) for every source. Maintaining two copies of these functions adds to the burden of maintenance. It also makes our test package more verbose and harder to navigate.

.envrc
```diff
@@ -0,0 +1,15 @@
...
+export CLIENT_ID=32555940559.apps.googleusercontent.com
+export CLOUD_SQL_POSTGRES_USER=postgres
+export CLOUD_SQL_POSTGRES_PASS=postgres
+export ALLOYDB_POSTGRES_PROJECT=toolbox-testing-438616
+export ALLOYDB_POSTGRES_CLUSTER=alloydb-pg-testing
+export ALLOYDB_POSTGRES_INSTANCE=alloydb-pg-testing-instance
+export ALLOYDB_POSTGRES_DATABASE=test_database
+export ALLOYDB_POSTGRES_REGION=us-central1
+export ALLOYDB_POSTGRES_USER=postgres
+export ALLOYDB_POSTGRES_PASS=alloydb-testing
```
kurtisvg: @duwenxin99 this shouldn't have been committed.

internal/sources/alloydbpg/alloydb_pg.go
kurtisvg: I think the flip side is that if we need to update this function signature in the future it would be a breaking change. I think this is in _internal_ so that's probably okay for now, but if we moved these things into more public APIs as some point and need to break it it would be an issue. 

I'm okay with a little bit of copy, doubly so if the chance of needing to change it is low -- we can just add a comment that says the function was copied from the other location and try to manually keep them in sync.

tests/alloydb_pg_integration_test.go
```diff
@@ -187,3 +190,70 @@ func TestAlloyDBPostgres(t *testing.T) {
 		})
 	}
 }
+
+// Set up auth test database table
+func setupAlloyDBAuthTest(t *testing.T, ctx context.Context) func(*testing.T) {
+	// set up dn connection pool
```
kurtisvg:
```suggestion
	// set up a connection pool
```

.envrc
duwenxin99: removed

internal/sources/alloydbpg/alloydb_pg.go
duwenxin99: Makes sense. Copied over.

PR #187, opened 2025-01-06
ci: Increase AlloyDB integration test coverage

duwenxin99:
add tests for tool invocation with params, connection over public IP, and connection over private IP.

### DIFF
--- a/tests/alloydb_pg_integration_test.go
+++ b/tests/alloydb_pg_integration_test.go
@@ -300,7 +300,7 @@ func setupParamTest(t *testing.T, ctx context.Context, tableName string) func(*t
 
 	return func(t *testing.T) {
 		// tear down test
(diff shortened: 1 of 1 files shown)


tests/alloydb_pg_integration_test.go
```diff
@@ -187,3 +199,69 @@ func TestAlloyDBPostgres(t *testing.T) {
...
+		t.Fatalf("unable to create ALLOYDB connection pool: %s", err)
+	}
+
+	err = pool.Ping(ctx)
+	if err != nil {
+		t.Fatalf("unable to connect to test database: %s", err)
+	}
+
+	_, err = pool.Query(ctx, `
+		CREATE TABLE param_test_table (
```
kurtisvg: nit: consider a uuid pre or postfix on the table name so that this test can be run in parallel if needed.

duwenxin99: added uuid

PR #188, opened 2025-01-06
ci(spanner): add integration tests

Yuan325:


tests/spanner_integration_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Test tool invoke endpoint
+	invokeTcs := []struct {
+		name        string
+		api         string
+		requestBody io.Reader
+		want        string
+	}{
+		{
+			name:        "invoke my-simple-tool",
```
duwenxin99, line 141: Should we add some invocation tests with parameters?

Yuan325: Will add it in a separate PR to improve current tests since it requires also adding test dataset etc.)

tests/spanner_integration_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"net/http"
+	"os"
+	"reflect"
+	"regexp"
+	"testing"
+	"time"
+)
+
+var (
+    SPANNER_PROJECT = os.Getenv("SPANNER_PROJECT")
```
kurtisvg: make sure to run lint

PR #189, opened 2025-01-07
feat: Added Neo4j Source and Tool

jexp:
- configure neo4j source with url, username, password, database
- configure neo4j tools with cypher statement and paramters
- tests based on the postgres tests
- neo4j.yaml for integration tests


kurtisvg:
@duwenxin99 the lint is failing on this because of repo permissions - can you make sure you run manually before we import it in?

jexp:
Awesome, thank you all so much !!!

PR #191, opened 2025-01-08
tests(sdk): Run tests on toolbox dev version

twishabansal:
This PR adds support for testing on both development (dev) and released (numbered) versions of the toolbox.

When testing against a `dev` version, the toolbox code is compiled directly from the repository, ensuring the latest changes are included. SDK tests are then executed against this newly compiled binary.

This is based on https://github.com/googleapis/genai-toolbox/pull/165

### DIFF
--- a/sdks/langchain/tests/conftest.py
+++ b/sdks/langchain/tests/conftest.py
@@ -25,6 +25,7 @@ import time
 from typing import Generator
 
 import google
+import pytest
 import pytest_asyncio
 from google.auth import compute_engine
 from google.cloud import secretmanager, storage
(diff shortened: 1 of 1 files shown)


sdks/langchain/tests/conftest.py
```diff
@@ -140,19 +141,36 @@ def auth_token2(project_id: str) -> str:
...
-    print("Toolbox binary downloaded successfully.")
+    # Get a temp dir for toolbox data
+    tmp_path = tmp_path_factory.mktemp("toolbox_data")
+    toolbox_binary_path = str(tmp_path) + "/toolbox"
+
+    # Get toolbox binary
+    if toolbox_version == "dev":
+        print("Compiling the current dev toolbox version...")
+        subprocess.Popen(["go", "get", "./..."])
+        subprocess.Popen(["go", "build", "-o", toolbox_binary_path])
```
anubhav756, line 146: optional: Should we also support an option for users to define source file locations independent of the working directory?

twishabansal:
> Do we plan to document the configuration and execution of these integration tests?

Yes. Once these tests are pushed, all this will be documented in an SDK Testing Playbook

sdks/langchain/tests/conftest.py
twishabansal: Do you mean a location to save the toolbox binary? That should not be needed. This compilation is just used to run the tests.
If the user needs a persistent dev version of toolbox, there are more straightforward ways to directly compile the code and run tests locally. All of that will be documented in the playbook: [go/toolbox-sdk-testing-playbook](https://goto.google.com/toolbox-sdk-testing-playbook)

anubhav756: No, I mean the location the source code, eg. `go build -o <DST_PATH> <SRC_PATH>`, or by using the `-C` flag (https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies).

twishabansal: As discussed offline, the build directly takes the source code and so we won't allow users to configure the source binary here.

Yuan325:
migrate PR to https://github.com/googleapis/genai-toolbox-langchain-python/pull/19

PR #192, opened 2025-01-08
feat(langchain-sdk): Add features for binding parameters to ToolboxTool class.

anubhav756:
The newly implemented `ToolboxTool` class manages tool state and supports this new feature of binding parameters along wiith the existing OAuth.

> [!NOTE]
> These changes are done in the LlamaIndex SDK as well in #203, along with documentation updates in #193.

### DIFF
--- a/sdks/langchain/src/toolbox_langchain_sdk/client.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/client.py
@@ -99,6 +99,8 @@ class ToolboxClient:
         tool_name: str,
         auth_tokens: dict[str, Callable[[], str]] = {},
         auth_headers: Optional[dict[str, Callable[[], str]]] = None,
+        bound_params: dict[str, Union[Any, Callable[[], Any]]] = {},
+        strict: bool = True,
     ) -> ToolboxTool:
         """
         Loads the tool with the given tool name from the Toolbox service.
@@ -108,6 +110,11 @@ class ToolboxClient:
             auth_tokens: An optional mapping of authentication source names to
                 functions that retrieve ID tokens.
             auth_headers: Deprecated. Use `auth_tokens` instead.
+            bound_params: An optional mapping of parameter names to their
+                bound values.
+            strict: If True, raises a ValueError if any of the given bound
+                parameters are missing from the schema or require
+                authentication. If False, only issues a warning.
 
         Returns:
             A tool loaded from the Toolbox.
@@ -132,6 +139,8 @@ class ToolboxClient:
             self._url,
             self._session,
             auth_tokens,
+            bound_params,
+            strict,
         )
 
     async def load_toolset(
@@ -139,6 +148,8 @@ class ToolboxClient:
         toolset_name: Optional[str] = None,
         auth_tokens: dict[str, Callable[[], str]] = {},
         auth_headers: Optional[dict[str, Callable[[], str]]] = None,
+        bound_params: dict[str, Union[Any, Callable[[], Any]]] = {},
+        strict: bool = True,
     ) -> list[ToolboxTool]:
         """
         Loads tools from the Toolbox service, optionally filtered by toolset
@@ -150,6 +161,11 @@ class ToolboxClient:
             auth_tokens: An optional mapping of authentication source names to
                 functions that retrieve ID tokens.
             auth_headers: Deprecated. Use `auth_tokens` instead.
+            bound_params: An optional mapping of parameter names to their
+                bound values.
+            strict: If True, raises a ValueError if any of the given bound
+                parameters are missing from the schema or require
+                authentication. If False, only issues a warning.
 
         Returns:
             A list of all tools loaded from the Toolbox.
@@ -178,6 +194,8 @@ class ToolboxClient:
                     self._url,
                     self._session,
                     auth_tokens,
+                    bound_params,
+                    strict,
                 )
             )
         return tools
--- a/sdks/langchain/src/toolbox_langchain_sdk/tools.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/tools.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from copy import deepcopy
-from typing import Any, Callable
+from typing import Any, Callable, Union
 from warnings import warn
 
 from aiohttp import ClientSession
@@ -24,6 +24,7 @@ from .utils import (
     ParameterSchema,
     ToolSchema,
     _find_auth_params,
+    _find_bound_params,
     _invoke_tool,
     _schema_to_model,
 )
@@ -32,7 +33,7 @@ from .utils import (
 class ToolboxTool(StructuredTool):
     """
     A subclass of LangChain's StructuredTool that supports features specific to
-    Toolbox, like authenticated tools.
+    Toolbox, like bound parameters and authenticated tools.
     """
 
     def __init__(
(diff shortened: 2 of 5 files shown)


sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -99,6 +99,7 @@ async def load_tool(
         tool_name: str,
         auth_tokens: dict[str, Callable[[], str]] = {},
         auth_headers: Optional[dict[str, Callable[[], str]]] = None,
+        bound_params: dict[str, Union[Any, Callable[[], Any]]] = {},
```
kurtisvg, line 102: I think Any already includes Callable. If you want to just signal this more explicitly to the user do:
```suggestion
        bound_params: dict[str, Union[Callable[Any], Any]] = {},
```

Also I think recently (maybe python 3.10? You can just do `dict[str, Callable | Any]`

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -148,6 +164,23 @@ def _validate_auth(self, strict: bool = True) -> None:
...
+
+        Raises:
+            ValueError: If attempting to bind a value on an authenticated tool.
+        """
+        non_bound_params: list[ParameterSchema] = []
+        for param in self._schema.parameters:
+            if param.name not in self._bound_params:
+                non_bound_params.append(param)
+
+        self._update_params(non_bound_params)
```
kurtisvg: is this update or set? Update implies "in addition" too, but I think this is overriding set params?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -148,6 +164,23 @@ def _validate_auth(self, strict: bool = True) -> None:
             raise PermissionError(message)
         warn(message)
 
+    def _remove_bound_params(self) -> None:
```
kurtisvg: This function is pretty short, and only used in 1 place. Does it need to be a function?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -195,3 +228,91 @@ def add_auth_token(self, auth_source: str, get_id_token: Callable[[], str]) -> N
...
+        # the tool.
+        self._remove_bound_params()
+
+        if dupe_params:
+            raise ValueError(
+                f"Parameter(s) `{', '.join(dupe_params)}` already bound in tool `{self._name}`."
+            )
+        if invalid_params:
+            raise ValueError(
+                f"Value(s) for `{', '.join(invalid_params)}` automatically handled by authentication source(s) and cannot be modified."
```
kurtisvg: Maybe a better error message would be "Parameter X x is authenticated and cannot be bound"

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -195,3 +228,91 @@ def add_auth_token(self, auth_source: str, get_id_token: Callable[[], str]) -> N
...
+
+        # Bound parameters are handled internally, so remove them from the
+        # schema to prevent validation errors and present a cleaner schema in
+        # the tool.
+        self._remove_bound_params()
+
+        if dupe_params:
+            raise ValueError(
+                f"Parameter(s) `{', '.join(dupe_params)}` already bound in tool `{self._name}`."
+            )
```
kurtisvg, line 347: I'm still fuzzy on why this is the expected behavior -- we don't want to allow a user to rebind their parameter?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
```diff
@@ -195,3 +228,91 @@ def add_auth_token(self, auth_source: str, get_id_token: Callable[[], str]) -> N
                 registered.
         """
         return self.add_auth_tokens({auth_source: get_id_token})
+
+    def bind_params(
+        self,
+        bound_params: dict[str, Union[Any, Callable[[], Any]]],
+        strict: bool = True,
+    ) -> None:
```
kurtisvg: I would strongly prefer if we preferred non-mutating functions over mutating, and return a new version of the tool instead of modifying the existing one.

sdks/langchain/src/toolbox_langchain_sdk/client.py
anubhav756: Yes, the intention is to make it explicit to the user ([here](https://docs.google.com/document/d/1zNgcu7_x-uUtQwWgrVNk0E3GX4dUxqnEHlGm-V9KzMs/edit?resourcekey=0-szIFxw2LCVQjXPO8hDreWg&disco=AAABaLYB0lI) is the relevant suggestion for reference).

With `dict[str, Union[Callable[Any], Any]]`, I get a type error as follows:
```
Expected parameter type list or "..."Pylance[reportInvalidTypeForm](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInvalidTypeForm)
```

Also, since we are targeting to support Python 3.9 and above, I think we should not go for `dict[str, Callable | Any]` for now.

Thoughts?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: `_update_params` method is removed now.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: `_remove_bound_params` is removed now.

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: Updated. Thanks for the suggestion!

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: I assume you mean you want to allow overwriting a bound parameter. We had decided earlier [here](https://docs.google.com/document/d/1zNgcu7_x-uUtQwWgrVNk0E3GX4dUxqnEHlGm-V9KzMs/edit?resourcekey=0-szIFxw2LCVQjXPO8hDreWg&tab=t.0#heading=h.tjkaj8v4bhnp) that we should prevent accidental modification of an already bound parameter. And if a user wants to override values, they can just use dynamic values instead.

Thoughts?

sdks/langchain/src/toolbox_langchain_sdk/tools.py
anubhav756: I've incorporated this change. [Here's](https://github.com/googleapis/genai-toolbox/pull/171#discussion_r1920802837) a more detailed explanation.

Let me know if that looks off.

sdks/langchain/src/toolbox_langchain_sdk/client.py
kurtisvg: SGTM

sdks/langchain/src/toolbox_langchain_sdk/tools.py
kurtisvg: I guess I'm having second thoughts on that decision :)

LGTM for now, but let's farm for feedback on this decision

anubhav756: Ok. It shouldn't be super hard to enable modification of bound params, but it would be interesting to see the feedback on this. I'll keep an eye on this. Thanks!

PR #197, opened 2025-01-12
feat!: replace Source field `ip_type` with `ipType` for consistency

duwenxin99:
Replace `ip_type` with `ipType` to match the camel-case of the other fields.
Update docs since we support both private and public IP connections.

### DIFF
--- a/docs/sources/alloydb-pg.md
+++ b/docs/sources/alloydb-pg.md
@@ -32,15 +32,16 @@ IAM identity has been given the following IAM permissions:
 
 ### Network Path
 
-Currently, this source only supports [connecting over Private
-IP][private-ip]. Most notably, this means
-you need to connect from a VPC that AlloyDB has been connected to.
+Currently, AlloyDB supports connection over both [private IP][private-ip] and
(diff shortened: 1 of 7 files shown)


docs/sources/alloydb-pg.md
```diff
@@ -32,15 +32,16 @@ IAM identity has been given the following IAM permissions:
...
 
-Currently, this source only supports [connecting over Private
-IP][private-ip]. Most notably, this means
-you need to connect from a VPC that AlloyDB has been connected to.
+Currently, AlloyDB supports connection over both [private IP][private-ip] and
+[public IP][public-ip]. Set the `ipType` parameter in your source
+configuration to `public` or `private`.
 
 [private-ip]: https://cloud.google.com/alloydb/docs/private-ip
+[public-ip]: https://cloud.google.com/alloydb/docs/connect-public-ipa
```
kurtisvg:
```suggestion
[public-ip]: https://cloud.google.com/alloydb/docs/connect-public-ip
```

kurtisvg:
This is a breaking change, so we need to use `feat!`.

PR #207, opened 2025-01-14
tests(sdk): e2e tests for llamaindex sdk

twishabansal:


sdks/llamaindex/tests/test_e2e.py
```diff
@@ -0,0 +1,155 @@
...
+        """Provides a ToolboxClient instance for each test."""
+        toolbox = ToolboxClient("http://localhost:5000")
+        yield toolbox
+        await toolbox.close()
+
+    #### Basic e2e tests
+    @pytest.mark.asyncio
+    async def test_load_tool(self, toolbox):
+        tool = await toolbox.load_tool("get-n-rows")
+        response = await tool.acall(**{"num_rows": "2"})
```
anubhav756:
```suggestion
        response = await tool.acall(num_rows="2")
```
optional: Wondering if tool invocations (here and below) could simply be written like this.

PR #211, opened 2025-01-14
chore: migrate yaml library to goccy/go-yaml

Yuan325:
Migrate yaml library from [gopkg.in/yaml.v3](https://pkg.go.dev/gopkg.in/yaml.v3) to [goccy/go-yaml](https://github.com/goccy/go-yaml).

We are trying to add validation feature to the unmarshaling of yaml. However, `yaml.v3` currently does not support that and the library had not been actively maintained for awhile. This migration will allow us to use validation processing such as [go-playground/validator](https://github.com/go-playground/validator)


cmd/root.go
```diff
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	// "bytes"
```
kurtisvg: nit: should remove

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/go-chi/chi/v5 v5.1.0
 	github.com/go-chi/httplog/v2 v2.1.1
 	github.com/go-chi/render v1.0.3
+	github.com/goccy/go-yaml v1.15.13
```
kurtisvg, line 16: Should the other yaml package be removed as well?

kurtisvg: (would expect go mod tidy to remove this, else we probably have it as a dep somewhere)

internal/server/config.go
```diff
@@ -115,52 +115,66 @@ func (s *StringLevel) Type() string {
 type SourceConfigs map[string]sources.SourceConfig
 
 // validate interface
-var _ yaml.Unmarshaler = &SourceConfigs{}
+var _ yaml.InterfaceUnmarshaler = &SourceConfigs{}
 
-func (c *SourceConfigs) UnmarshalYAML(node *yaml.Node) error {
+// RawConfig is a type used to allow unmarshal another InterfaceUnmarshaler
```
kurtisvg:
```suggestion
// DelayedUnmarshaler is struct that saves the provided unmarshal function passed to UnmarshalYAML so it can be re-used later once the target interface is known. 
```

kurtisvg: Also consider putting it into a util module and de-dupe with RawConfig -- if not, it probably doesn't need to be exported.

internal/server/config.go
```diff
@@ -115,52 +115,66 @@ func (s *StringLevel) Type() string {
...
+type RawConfig struct {
```
kurtisvg: nit: Should probably also add `var _ yaml.InterfaceUnmarshaler = &RawConfig{}`

internal/tools/parameters.go
```diff
@@ -131,15 +129,29 @@ type Parameter interface {
 // Parameters is a type used to allow unmarshal a list of parameters
 type Parameters []Parameter
 
-func (c *Parameters) UnmarshalYAML(node *yaml.Node) error {
+// RawParam is a type used to allow unmarshal another InterfaceUnmarshaler
+type RawParam struct {
```
kurtisvg: similar comments as the other one

internal/tools/parameters.go
```diff
@@ -148,42 +160,42 @@ func (c *Parameters) UnmarshalYAML(node *yaml.Node) error {
 	return nil
 }
 
-func parseParamFromNode(node *yaml.Node) (Parameter, error) {
+func parseParamFromRawParam(rawParam *RawParam) (Parameter, error) {
 	// Helper function that is required to parse parameters
 	// because there are multiple different types
```
kurtisvg: nit: Comment should probably be outside of the function

internal/util/util.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2024 Google LLC
```
kurtisvg: nit: 
```suggestion
// Copyright 2025 Google LLC
```

PR #218, opened 2025-01-15
docs: add local quickstart

twishabansal:
This quickstart helps users to onboard to toolbox quickly.

### DIFF
--- a/docs/en/getting-started/quickstart.md
+++ b/docs/en/getting-started/quickstart.md
@@ -87,20 +87,24 @@ sources:
         user: test_user
         password: test-password
 tools:
-  search-hotels:
+  search-hotels-by-name:
     kind: postgres-sql
     source: my-pg-source
-    description: Search for hotels based on location and name. 
-      Returns a list of hotel dictionaries matching the search criteria.
+    description: Search for hotels based on name. 
+    parameters:
+      - name: name
+        type: string
+        description: The name of the hotel.
+    statement: SELECT * FROM hotels WHERE name ILIKE '%' || $1 || '%';
+  search-hotels-by-location:
+    kind: postgres-sql
+    source: my-pg-source
+    description: Search for hotels based on location. 
     parameters:
       - name: location
         type: string
         description: The location of the hotel.
-      - name: name
-        type: string
-        description: The name of the hotel.
-    statement: SELECT * FROM hotels WHERE location ILIKE '%' || $1 || '%' AND
-        name ILIKE '%' || $2 || '%';
+    statement: SELECT * FROM hotels WHERE location ILIKE '%' || $1 || '%';
   book-hotel:
     kind: postgres-sql
     source: my-pg-source
@@ -139,8 +143,8 @@ tools:
     statement: UPDATE hotels SET booked = B'0' WHERE id = $1;
 ```
 
-The config file defines four tools:
-`search-hotels`, `book-hotel`, `update-hotel` and `cancel-hotel`.
+The config file defines five tools:
+`search-hotels-by-name`, `search-hotels-by-location`, `book-hotel`, `update-hotel` and `cancel-hotel`.
 
 Each tool specifies its description,
 [kind](https://github.com/googleapis/genai-toolbox/tree/main/docs/sources#kinds-of-sources),
@@ -206,11 +210,11 @@ and the corresponding SQL statements to execute upon tool invocation.
     asyncio.run(main())
     ```
 
-1. Try out the `search-hotels` tool.
+1. Try out the `search-hotels-by-name` tool.
 
     ```python
-    search_tool = await client.load_tool('search-hotels')
-    response = await search_tool.ainvoke({"location": "Zurich", "name": ""})
+    search_tool = await client.load_tool('search-hotels-by-name')
+    response = await search_tool.ainvoke({"name": "Hilton"})
     print(response)
     ```
 


docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
+---
+title: "Toolbox Quickstart"
```
kurtisvg: nit: Probably just Quickstart, or maybe "Local Quickstart"

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+type: docs
+weight: 1
+description: This guide would help you set up a basic agentic application using toolbox.
+
+---
+
+## Step 1: Set up a database
```
kurtisvg, line 9: Should their be a "before you begin" step?  Are we assuming the user has any tools installed or running a specific environment?

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+
+[Install postgres and configure a
+database](https://neon.tech/postgresql/postgresql-getting-started) for your
+system.
+
+This process creates a database `postgres` with superuser `postgres`.
```
kurtisvg: I would like to see if we can capture this set up ourselves, without requiring the user to go to a new page. I think using docker might also be useful in streamlining?

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Step 2: Import data into the database
+
+1. Connect to the database using command line.
+
+    ```bash
+    psql -U postgres -d postgres
```
kurtisvg: let's create a new database (using the default pg one is not a best practice).

Similarly, can we create a user for that database?

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+    source: my-pg-source
+    description: Search for hotels based on location and name. 
+      Returns a list of hotel dictionaries matching the search criteria.
+    parameters:
+      - name: location
+        type: string
+        description: The location of the hotel.
+      - name: name
+        type: string
+        description: The name of the hotel.
```
kurtisvg: nit: curious if this should be 2 separate tools. one for search_by_location and another that's search_by_name?

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+1. Install the `toolbox_langchain_sdk` package.
+
+    ```bash
+    pip install toolbox_langchain_sdk
+    ```
+
+    > **_NOTE:_**
+    > Right now, the toolbox_langchain_sdk package is not available on pip. To
+    > use the sdk, download the source code from [git](https://github.com/googleapis/genai-toolbox/tree/main/sdks/langchain) and install it locally
+    > using the command:
```
kurtisvg, line 187: There is a module that does infoboxs in a bit nicer format: https://github.com/martignoni/hugo-notice

Maybe we can look at adding it in a follow up PR?

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+
+    ```python
+    tools = await client.load_toolset()
+    ```
+
+1. Install required modules.
+
+    ```bash
+    pip install langgraph langchain-google-vertexai
+    ```
```
kurtisvg: nIt: I would expect this to be more towards the top

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,254 @@
...
+        response = await agent.ainvoke(inputs, stream_mode="values", config=config)
+        print(response["messages"][-1].content)
+    ```
+
+    To verify the agent's actions, you can examine the hotels table. You should
+    observe that the `booked` column for the `Hyatt Regency Basel` has changed
+    from `0` to `1`, indicating that the hotel has been successfully booked.
+    Additionally, the `checkin_date` and `checkout_date` have been updated to
+    `2024-04-10` and `2024-04-19` from `2024-04-02` and `2024-04-20`
+    respectively.
```
kurtisvg: Would it be better to combine this into a single file users can copy and hack on locally?

docs/en/getting-started/quickstart.md
twishabansal: Raised a bug: b/390650129. Will create a follow up PR for the same.

docs/en/getting-started/quickstart.md
twishabansal: This structure could help users understand and customize the steps involved. 
However, we could upload the code and YAML files into the github repo and add a link for users to access them.

We could add a note at the beginning, like:

```
The code and configuration files used in this quickstart are available [here](link_to_files).
```

What do you think?

docs/en/getting-started/quickstart.md
twishabansal: Right now, we're not assuming anything about the user environment, except that they have python installed.
We're going over the whole process from scratch.

docs/en/getting-started/quickstart.md
twishabansal: In that case, can we assume basic familiarity with docker and have docker installation as a prerequisite? 
Do you think the friction of docker installation is worth ignoring?

docs/en/getting-started/quickstart.md
twishabansal: Done.

docs/en/getting-started/quickstart.md
twishabansal: We don't support optional parameters yet, so this sounds like a good idea. Changed the example.

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,276 @@
...
+    model = ChatVertexAI(model_name="gemini-pro", project="my-project") # Change the GCP project here
+    agent = create_react_agent(model, tools, checkpointer=MemorySaver())
+    ```
+
+1. Define the initial prompt and user queries.
+
+    ```python
+    prompt = """
+      You're a helpful hotel assistant. You handle hotel searching, booking and
+      cancellations. When the user searches for a hotel, mention it's name, id,
```
anubhav756, line 244:
```suggestion
      cancellations. When the user searches for a hotel, mention its name, id, 
```

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,276 @@
...
+      cancellations. When the user searches for a hotel, mention it's name, id, 
+      location and price tier. Always mention hotel ids while performing any 
+      searches. This is very important for any operations. For any bookings or 
+      cancellations, please provide the appropriate confirmation. Be sure to update 
+      checkin or checkout dates if mentioned by the user.
+      Don't ask for confirmations from the user.
+    """
+    
+    queries = [
+        "Find hotels in Basel with Basel in it's name.",
```
anubhav756, line 253:
```suggestion
        "Find hotels in Basel with Basel in its name.",
```

docs/en/getting-started/quickstart.md
```diff
@@ -0,0 +1,276 @@
...
+        "Can you book the Hilton Basel for me?",
+        "Oh wait, this is too expensive. Please cancel it and book the Hyatt Regency instead.",
+        "My check in dates would be from April 10, 2024 to April 19, 2024.",
```
anubhav756, line 256:
```suggestion
        "My check-in dates would be from April 10, 2024 to April 19, 2024.",
```

kurtisvg:
Superseded by #232

PR #219, opened 2025-01-16
test(langchain-sdk): Add more test cases to e2e tests

twishabansal:
Test for tool calls with

1. Missing params
2. Wrong param type

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -25,11 +25,11 @@ import (
 	"syscall"
 	"time"
 
+	yaml "github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/log"
 	"github.com/googleapis/genai-toolbox/internal/server"
 	"github.com/googleapis/genai-toolbox/internal/telemetry"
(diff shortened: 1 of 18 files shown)


sdks/langchain/tests/test_e2e.py
```diff
@@ -85,6 +89,18 @@ async def test_load_toolset_all(self, toolbox):
...
+    async def test_run_tool_missing_params(self, toolbox):
+        tool = await toolbox.load_tool("get-n-rows")
+        with pytest.raises(ValidationError, match="Field required"):
+            await tool.ainvoke({})
+
+    @pytest.mark.asyncio
+    async def test_run_tool_wrong_param_type(self, toolbox):
+        tool = await toolbox.load_tool("get-n-rows")
+        with pytest.raises(ValidationError, match="Input should be a valid string"):
+            await tool.ainvoke({"num_rows": 2})
```
anubhav756, line 102: optional: We can parameterize similar tests, eg. in these two we can have the tool input and error string as parameters. But since this is a very small test, it's optional here.

PR #221, opened 2025-01-17
feat: add cloudsql mysql source and tool

Yuan325:
Add CloudSQL for MySQL source and tool.

CloudSQLMySQL source is initialize with the following config:
```
sources:
    my-cloudsqlmysql-source:
        kind: cloud-sql-mysql
        project: my-project-name
        region: my-region
        instance: my-instance-name
        user: my_user
        password: my_pass
        database: my_db
        # ipType: public # The default dialect is public.
```

MySQL tool is initialize with the following config.
```
tools:
    test_tool:
        kind: mysql
        source: my-cloudsqlmysql-source
        description: >
            Testing tool.
        statement: "SELECT 1;"
```


internal/tools/mysql/mysql.go
```diff
@@ -0,0 +1,158 @@
...
+	for i := range v {
+		pointers[i] = &v[i]
+	}
+	var out strings.Builder
+	for results.Next() {
+		err := results.Scan(pointers...)
+		if err != nil {
+			return "", fmt.Errorf("unable to parse row: %w", err)
+		}
+		out.WriteString(fmt.Sprintf("%s", v))
```
kurtisvg, line 133: @duwenxin99 added some functionality for SQL Server here for the typing: https://github.com/googleapis/genai-toolbox/pull/223/files#diff-e1a12645f42304e79c7b473c58f5c25747ecfe4f1b1f5b5c0bd23a61f341e4abR148-R186

Should we do something similar here? Or should the SQL Server tool just use interface like we are doing here?

internal/tools/mysql/mysql.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	cols, err := results.Columns()
+	if err != nil {
+		return "", fmt.Errorf("unable to retrieve rows column name: %w", err)
+	}
+
+	cl := len(cols)
+	v := make([]interface{}, cl)
+	pointers := make([]interface{}, cl)
```
kurtisvg: nit: s/interace{}/any

PR #222, opened 2025-01-17
tests(langchain-sdk): Update conftest to be able to run tests in local

twishabansal:
[Getting an id token from metadata server](https://cloud.google.com/docs/authentication/get-id-token#metadata-server) does not work on local systems. This causes test failures for auth tests locally. This PR fixes the issue.

This unblocks https://b.corp.google.com/issues/385051620


sdks/langchain/tests/conftest.py
```diff
@@ -80,15 +80,25 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
...
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        request = google.auth.transport.requests.Request()
+        credentials = compute_engine.IDTokenCredentials(
+            request=request,
+            target_audience=client_id,
+            use_metadata_identity_endpoint=True,
+        )
+        if not credentials.valid:
+            credentials.refresh(request)
+        return credentials.token
```
kurtisvg, line 101: I'm not sure I agree with this flow. I think ideally it should be trying the metadata server first, and falling back to gcloud if it's unable to reach. An alternative would be use an env var to signal which to use. 

Ideally, we should not rely on "gcloud auth" as it's not semantically versioned (and I've personally been bit by it changing int the past)

Yuan325:
Migrated PR to https://github.com/googleapis/genai-toolbox-langchain-python/pull/18

PR #223, opened 2025-01-17
feat: Add Cloud SQL for SQL Server Source and Tool

duwenxin99:
1. `sql/database` provides a `Scan()`interface to scan query results into typed variables. Therefore we have to create a slice of typed variables (types retrieved from rows.ColumnTypes()) to pass them into `Scan()`. Using []byte works but makes the printing result different from other tools (e.g [1] instead of %!s(int32=1)]
2. MS SQL supports both named (e.g @name) and positional args (e.g @p2), so we have to check if the name is contained in the original statement before passing them into `db.Query()` as either named arg or as values.

### DIFF
--- a/internal/tools/mssql/mssql.go
+++ b/internal/tools/mssql/mssql.go
@@ -141,13 +141,6 @@ func (t Tool) Invoke(params tools.ParamValues) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("unable to parse row: %w", err)
 		}
-		for i := range v {
-			// Database/sql scans both `INT` and `BIGINT` into `int64``
-			// Cast the result into `int32` if the column type is `INT`
-			if types[i].DatabaseTypeName() == "INT" {
-				v[i] = int32(v[i].(int64))
-			}
-		}
 		out.WriteString(fmt.Sprintf("%s", v))
 	}
 	err = rows.Close()
--- a/tests/auth_test.go
+++ b/tests/auth_test.go
@@ -129,8 +129,14 @@ func RunGoogleAuthenticatedParameterTest(t *testing.T, sourceConfig map[string]a
 		t.Fatalf("error getting Google ID token: %s", err)
 	}
 
-	// Create wanted string
-	wantResult := fmt.Sprintf("Stub tool call for \"my-auth-tool\"! Parameters parsed: [{\"email\" \"%s\"}] \n Output: [%%!s(int32=1) Alice %s]", SERVICE_ACCOUNT_EMAIL, SERVICE_ACCOUNT_EMAIL)
+	// Tools using database/sql interface only outputs `int64` instead of `int32`
+	var wantString string
+	switch toolKind {
+	case "mssql":
+		wantString = fmt.Sprintf("Stub tool call for \"my-auth-tool\"! Parameters parsed: [{\"email\" \"%s\"}] \n Output: [%%!s(int64=1) Alice %s]", SERVICE_ACCOUNT_EMAIL, SERVICE_ACCOUNT_EMAIL)
+	default:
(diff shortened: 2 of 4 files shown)


internal/sources/cloudsqlmssql/cloud_sql_mssql.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const SourceKind string = "cloud-sql-mssql"
+
+// validate interface
+var _ sources.SourceConfig = Config{}
+
+type Config struct {
```
kurtisvg, line 35: nit: It would be nice to have a comment (at minimum on exported interfaces)

internal/tools/mssql/mssql.go
```diff
@@ -0,0 +1,209 @@
...
+
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	"github.com/googleapis/genai-toolbox/internal/sources/cloudsqlmssql"
+	"github.com/googleapis/genai-toolbox/internal/tools"
+)
+
+const ToolKind string = "mssql"
+
+type compatibleSource interface {
+	MssqlDb() *sql.DB
```
kurtisvg: nit: Should it be `MSSQLDB()`?

internal/tools/mssql/mssql.go
```diff
@@ -0,0 +1,209 @@
...
+
+	namedArgs := make([]any, 0, len(params))
+	paramsMap := params.AsReversedMap()
+	// To support both named args (e.g @id) and positional args (e.g @p1), check if arg name is contained in the statement.
+	for _, v := range params.AsSlice() {
+		paramName := paramsMap[v]
+		if strings.Contains(t.Statement, "@"+paramName) {
+			namedArgs = append(namedArgs, sql.Named(paramName, v))
+		} else {
+			namedArgs = append(namedArgs, v)
```
kurtisvg, line 118: Just to confirm, can we add a test for both positional args and named args together?

duwenxin99: Our `RunToolInvocationWithParamsTest()` test in common.test is testing for both types together in one statement: 
```case "mssql":
		statement = fmt.Sprintf("SELECT * FROM %s WHERE id = @id OR name = @p2;", tableName)```

Comment (kurtisvg) [reply internal/tools/mssql/mssql.go:118]: Missed that, thank you!

File: internal/tools/mssql/mssql.go (hunk 2 above)
Comment (duwenxin99) [reply internal/tools/mssql/mssql.go]: updated

File: internal/sources/cloudsqlmssql/cloud_sql_mssql.go (hunk 1 above)
Comment (duwenxin99) [reply internal/sources/cloudsqlmssql/cloud_sql_mssql.go:35]: added comments

File: internal/tools/mssql/mssql.go [hunk 4]
```diff
@@ -0,0 +1,209 @@
...
+				out.WriteString(fmt.Sprintf("%s", resValue.Int32)) //nolint:all
+				continue
+			}
+			if resValue, ok := res.(*sql.NullInt64); ok {
+				out.WriteString(fmt.Sprintf("%s", resValue.Int64)) //nolint:all
+				continue
+			}
+			if resValue, ok := res.(*sql.NullFloat64); ok {
+				out.WriteString(fmt.Sprintf("%s", resValue.Float64)) //nolint:all
+				continue
```
kurtisvg: Could this part use a type switch instead? https://go.dev/tour/methods/16

duwenxin99: removed type check

PR #226, opened 2025-01-20
feat(langchain-sdk)!: Add support for sync operations in Toolbox LangChain SDK

anubhav756:
## Overview

This PR introduces asynchronous support for the `ToolboxClient` by introducing a new `AsyncToolboxClient` class that provides a purely asynchronous interface. `ToolboxClient` class acts as a wrapper around `AsyncToolboxClient`, providing both synchronous and asynchronous methods.

The synchronous methods are implemented by utilizing a dedicated background event loop running in a separate thread. This background event loop in a background thread is managed by a new class called `_BackgroundLoop`. When a synchronous method is called, it delegates the operation to the underlying `AsyncToolboxClient` using `asyncio.run_coroutine_threadsafe` and blocks until the result is available. This approach allows existing synchronous code to seamlessly interact with the new asynchronous capabilities.

## Key changes:

* `AsyncToolboxClient`: A fully asynchronous interface for toolbox operations. This class is similar to the `ToolboxClient` class before this change. It adds the new async interfaces `aload_tool` and `aload_toolset`, and uses `create` to initialize.
* `ToolboxClient`: A wrapper class exposing both synchronous and asynchronous methods, delegating to `AsyncToolboxClient` under the hood.
* Background Event Loop: A dedicated event loop in a separate thread to handle asynchronous operations for synchronous method calls. This loop-thread combo is managed by a new class called `_BackgroundLoop`.

## Benefit
This approach allows us to modernize the `ToolboxClient` while ensuring backward compatibility and providing users with the flexibility to choose between synchronous and asynchronous methods.

## Diagram
image

> [!NOTE]
> README updates are done in #237. Updates to test cases and to LlamaIndex SDK will be done in a future PR.

### DIFF
--- a/sdks/langchain/src/toolbox_langchain_sdk/client.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/client.py
@@ -61,6 +61,8 @@ class ToolboxClient:
 
         asyncio.run_coroutine_threadsafe(coro, ToolboxClient.__loop).result()
 
+        if not ToolboxClient.__session:
+            raise ValueError("Session cannot be None.")
         self.__async_client = AsyncToolboxClient(url, ToolboxClient.__session)
 
     def __run_as_sync(self, coro: Awaitable[T]) -> T:
@@ -114,6 +116,8 @@ class ToolboxClient:
             )
         )
 
+        if not self.__loop or not self.__thread:
+            raise ValueError("Background loop or thread cannot be None.")
         return ToolboxTool(async_tool, self.__loop, self.__thread)
 
     async def aload_toolset(
@@ -150,6 +154,9 @@ class ToolboxClient:
         )
 
         tools: list[ToolboxTool] = []
+
+        if not self.__loop or not self.__thread:
+            raise ValueError("Background loop or thread cannot be None.")
         for async_tool in async_tools:
             tools.append(ToolboxTool(async_tool, self.__loop, self.__thread))
         return tools
@@ -185,6 +192,8 @@ class ToolboxClient:
             )
         )
 
+        if not self.__loop or not self.__thread:
+            raise ValueError("Background loop or thread cannot be None.")
         return ToolboxTool(async_tool, self.__loop, self.__thread)
 
     def load_toolset(
@@ -220,6 +229,8 @@ class ToolboxClient:
             )
         )
 
+        if not self.__loop or not self.__thread:
+            raise ValueError("Background loop or thread cannot be None.")
         tools: list[ToolboxTool] = []
         for async_tool in async_tools:
             tools.append(ToolboxTool(async_tool, self.__loop, self.__thread))
--- a/sdks/langchain/src/toolbox_langchain_sdk/async_tools.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/async_tools.py
@@ -195,7 +195,8 @@ class AsyncToolboxTool(BaseTool):
 
         # Check each parameter for at least 1 required auth source
         for param in self.__auth_params:
-            assert param.authSources is not None
+            if not param.authSources:
+                raise ValueError("Auth sources cannot be None.")
             has_auth = False
             for src in param.authSources:
 


sdks/langchain/src/toolbox_langchain_sdk/async_client.py
```diff
@@ -0,0 +1,204 @@
+# Copyright 2024 Google LLC
```
kurtisvg: nit: 2025 for new files

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
```diff
@@ -0,0 +1,204 @@
...
+        session: Optional[ClientSession] = None,
+        bg_loop: Optional[_BackgroundLoop] = None,
+    ) -> "AsyncToolboxClient":
+
+        # Use a default session if none is provided. This leverages connection
+        # pooling for better performance by reusing a single session throughout
+        # the application's lifetime.
+        if session is None:
+            if cls.__default_session is None:
+                cls.__default_session = ClientSession()
```
kurtisvg: ClientSession() doesn't seem to require await. Do we really need to use create/create_async for this?

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
```diff
@@ -0,0 +1,204 @@
...
+from warnings import warn
+
+from aiohttp import ClientSession
+
+from .background_loop import _BackgroundLoop
+from .tools import ToolboxTool
+from .utils import ManifestSchema, _load_manifest
+
+
+class AsyncToolboxClient:
```
kurtisvg, line 24: Should this be exported? I thought we were only presenting ToolboxClient to users.

sdks/langchain/src/toolbox_langchain_sdk/background_loop.py
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+
+import asyncio
+from threading import Thread
+from typing import Awaitable, Optional, TypeVar
+
+T = TypeVar("T")
+
+
+class _BackgroundLoop:
```
kurtisvg: not sure a class is needed for this. Any reason Toolbox Client shouldn't just hold onto a loop / thread?

sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -170,32 +211,9 @@ async def load_toolset(
...
-                ToolboxTool(
-                    tool_name,
-                    tool_schema,
-                    self._url,
-                    self._session,
-                    auth_tokens,
-                    bound_params,
-                    strict,
-                )
+        assert self.__bg_loop
```
kurtisvg: assert is not safe to use in non-testing environments

kurtisvg: https://stackoverflow.com/a/1838411/9414803

sdks/langchain/src/toolbox_langchain_sdk/background_loop.py
anubhav756: This class currently handles loop/thread management, simplifying the process of passing these to `ToolboxTool` instances created by `ToolboxClient`. More importantly, it also allows us to reuse the `run_as_sync`/`run_as_async` methods. However, I'm wondering if there's a better way to achieve this.

kurtisvg: I think the ran_as_sync and run_as_async are only 1-2 lines. I think we could just inline them.

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
twishabansal: Removed the create methods

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
twishabansal: Removed AsyncClient export

twishabansal:
Addressed comments

1. Split classes into: `ToolboxClient` and `AsyncToolboxClient`, `ToolboxTool` and `AsyncToolboxTool`. 
1. Changed license year to 2025 for new files.
1. Removed create methods from ToolboxClient and AsyncToolboxClient.
1. Removed AsyncClient export. (Not to be exported by the user)
1. Removed assert statements from files

Detailed commits in PR: https://github.com/googleapis/genai-toolbox/pull/245

sdks/langchain/src/toolbox_langchain_sdk/background_loop.py
twishabansal: We are using the `run_as_sync` and `run_as_async_methods` currently in the `ToolboxClient` and `ToolboxTool` class.

In the future, we might want to add more wrapper classes around the async classes, so this would be reusable there too.

Do you think it makes sense to inline this function in each class or use a separate background loop class?

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
```diff
@@ -0,0 +1,200 @@
...
+        Initializes the AsyncToolboxClient for the Toolbox service at the given URL.
+
+        Args:
+            url: The base URL of the Toolbox service.
+            bg_loop: Optional background async event loop used to create
+            ToolboxTool.
+            session: An HTTP client session.
+        """
+        self.__url = url
+        self.__bg_loop = bg_loop
```
kurtisvg: I don't think we need a Background loop -- think it would be easier to just initialize the loop and use run_as_threadsafe where needed

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
```diff
@@ -0,0 +1,200 @@
...
+                warn(
+                    "Argument `auth_headers` is deprecated. Use `auth_tokens` instead.",
+                    DeprecationWarning,
+                )
+                auth_tokens = auth_headers
+
+        url = f"{self.__url}/api/tool/{tool_name}"
+        manifest: ManifestSchema = await _load_manifest(url, self.__session)
+
+        if self.__bg_loop is None:
```
kurtisvg: async client should not know about loop reference

sdks/langchain/src/toolbox_langchain_sdk/background_loop.py
kurtisvg: Imo, I think we just hold onto the thread/loop in the class. If we really need a connivence function we don't need to attach it to a class for this. It's too much abstraction for 2-3 lines of code.

sdks/langchain/src/toolbox_langchain_sdk/async_client.py
```diff
@@ -28,7 +27,6 @@ class AsyncToolboxClient:
     def __init__(
         self,
         url: str,
-        bg_loop: _BackgroundLoop,
```
anubhav756: We want to clear this from the docstring as well.

anubhav756:
Continued in https://github.com/googleapis/genai-toolbox-langchain-python/pull/15

PR #203, opened 2025-01-13
feat(llamaindex-sdk)!: Add Bound Params support to Toolbox LlamaIndex SDK

anubhav756:

PR #229, opened 2025-01-21
fix(langchain-sdk): Improve session cleanup

anubhav756:
This PR improves session cleanup by using `asyncio.get_running_loop` instead of `asyncio.get_event_loop`. The latter is deprecated and has been known to cause complications. If no running loop is found, then a `RuntimeError` is thrown and caught, and a new event loop is spun up to close the session.

Key changes:
* Use `asyncio.get_running_loop` instead of `asyncio.get_event_loop`.
* Catch and handle a specific `RuntimeError` that can occur if no running loop is found.
* Spin up a new event loop to close the session if no running loop is found.

> [!NOTE]
> The corresponding change for LlamaIndex SDK is done in #203.

### DIFF
--- a/sdks/langchain/src/toolbox_langchain_sdk/client.py
+++ b/sdks/langchain/src/toolbox_langchain_sdk/client.py
@@ -53,16 +53,10 @@ class ToolboxClient:
         collected.
         """
         try:
-            loop = asyncio.get_event_loop()
-            if loop.is_running():
-                loop.create_task(self.close())
-            else:
-                loop.run_until_complete(self.close())
-        except Exception:
-            # We "pass" assuming that the exception is thrown because the event
-            # loop is no longer running, but at that point the Session should
-            # have been closed already anyway.
-            pass
+            loop = asyncio.get_running_loop()
+            loop.create_task(self.close())
+        except RuntimeError:
+            asyncio.run(self.close())
 
     async def _load_tool_manifest(self, tool_name: str) -> ManifestSchema:
         """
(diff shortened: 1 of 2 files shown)


sdks/langchain/src/toolbox_langchain_sdk/client.py
```diff
@@ -53,16 +53,10 @@ def __del__(self):
...
-                loop.run_until_complete(self.close())
-        except Exception:
-            # We "pass" assuming that the exception is thrown because the event
-            # loop is no longer running, but at that point the Session should
-            # have been closed already anyway.
-            pass
+            loop = asyncio.get_running_loop()
+            loop.create_task(self.close())
+        except RuntimeError:
+            asyncio.run(self.close())
```
kurtisvg, line 59: I think we want to leave the broader "except Exception" in place -- I don't think we want to raise an other exceptions.

Also, I'm not entirely sure that we should even try to do 'asyncio.run' -- I believe the session needs to be closed in then same loop it was opened in. Can we validate that is the case?

anubhav756: You are completely right, it seems like the `ClientSession` is bound to the event loop that created it. If the loop is not running, the session is implicitly closed. At best, it should throw `Session is closed` error anyway. At worst, we might get a `ResourceWarning: unclosed transport` warning: https://docs.aiohttp.org/en/stable/client_advanced.html#graceful-shutdown

I originally thought that `loop.run_until_complete(...)` would throw an error if the loop is already closed, but I double-checked the docs and it seems like it won't: image

I was having second-thoughts about this because there was no clear documentation for this case apart from requiring the user to use `with ToolboxClient(...) as client:` or enforcing them to pass in a `session` mandatorily. Both of these options seem like too much inconvenience for the end-user.

I think given the lack of clear guidance, we can rely on the current way, which is in fact how `aioredis` python library handle this: https://github.com/aio-libs-abandoned/aioredis-py/blob/ff5a8fe068ebda837d14c3b3777a6182e610854a/aioredis/client.py#L1012-L1020

Thanks for pointing this out. Closing this PR for now.

PR #233, opened 2025-01-23
feat: add dgraph tool and source

shivaji-kharse:
* add dgraph tool and source

### DIFF
--- a/docs/en/resources/sources/dgraph.md
+++ b/docs/en/resources/sources/dgraph.md
@@ -1,0 +1,55 @@
+---
+title: "Dgraph"
+type: docs
+weight: 1
+description: >
+ Dgraph is a horizontally scalable and distributed graph database.
+
+---
+
+## About
+
+[Dgraph][dgraph-docs] is a horizontally scalable and distributed graph database.
+It provides ACID transactions, consistent replication, and linearizable reads.
+
+This source can connect to either a self-managed Dgraph cluster or one hosted on
+Dgraph Cloud. If you're new to Dgraph, the fastest way to get started is to
+[sign up for Dgraph Cloud][dgraph-login].
+
+[dgraph-docs]: https://dgraph.io/docs
+[dgraph-login]: https://cloud.dgraph.io/login
+
+## Requirements 
+
+### Database User
+
+When **connecting to a hosted Dgraph database**, this source uses the API key
+for access. If you are using a dedicated environment, you will additionally need
+the namespace and user credentials for that namespace.
+
+For **connecting to a local or self-hosted Dgraph database**, use the namespace
+and user credentials for that namespace.
+
+## Example
+
+```yaml
+sources:
+    my-dgraph-source:
+        kind: "dgraph"
+        dgraphUrl: "https://xxxx.cloud.dgraph.io"
+        user: "groot"
+        password: "password"
+        apiKey: abc123
+        namepace : 0
+```
+
+## Reference
+
+| **Field**   | **Type** | **Required** | **Description**                                                                                  |
+|-------------|:--------:|:------------:|--------------------------------------------------------------------------------------------------|
+| kind        |  string  |     true     | Must be "dgraph".                                                                                |
+| dgraphUrl   |  string  |     true     | Connection URI (e.g. "https://xxx.cloud.dgraph.io", "https://localhost:8080").                   |
+| user        |  string  |     false    | Name of the Dgraph user to connect as (e.g., "groot").                                           |
+| password    |  string  |     false    | Password of the Dgraph user (e.g., "password").                                                  |
+| apiKey      |  string  |     false    | API key to connect to a Dgraph Cloud instance.                                                   |
+| namespace   |  uint64  |     false    | Dgraph namespace (not required for Dgraph Cloud Shared Clusters).                                |
--- a/docs/tools/dgraph-dql.md
+++ b/docs/tools/dgraph-dql.md
@@ -1,0 +1,105 @@
+# Dgraph DQL Tool 
+
+
+A "dgraph-dql" tool executes a pre-defined DQL statement against a Dgraph database. It's compatible with any of the following 
+sources:
+- [dgraph](../sources/dgraph.md)
+
+To run a statement as a query, you need to set the config isQuery=true. For upserts or mutations, set isQuery=false.
+You can also configure timeout for a query.
+
+## Example
+
+### Query:
+
+```yaml
+tools:
+  search_user:
+    kind: dgraph-dql
+    source: my-dgraph-source
+    statement: |
+      query all($role: string){
+        users(func: has(name)) @filter(eq(role, $role) AND ge(age, 30) AND le(age, 50)) {
+          uid
+          name
+          email
+          role
+          age
+        }
+      }
+    isQuery: true
+    timeout: 20s
+    description: |
+     Use this tool to retrieve the details of users who are admins and are between 30 and 50 years old.
+     The query returns the user's name, email, role, and age.
+     This can be helpful when you want to fetch admin users within a specific age range.
+     Example: Fetch admins aged between 30 and 50:
+      [
+        {
+          "name": "Alice",
+          "role": "admin",
+          "age": 35
+        },
+        {
+          "name": "Bob",
+          "role": "admin",
+          "age": 45
+        }
+      ]
+    parameters:
+      - name: $role
+        type: string
+        description: admin
+```
+
+### Mutation: 
+
+```yaml
+tools:
+  dgraph-manage-user-instance:
+    kind: dgraph-dql
+    source: my-dgraph-source
+    isQuery: false
+    statement: |
+           {
+            set {
+            _:user1 <name> $user1 .
+            _:user1 <email> $email1 .
+            _:user1 <role> "admin" .
+            _:user1 <age> "35" .
+
+            _:user2 <name> $user2 .
+            _:user2 <email> $email2 .
+            _:user2 <role> "admin" .
+            _:user2 <age> "45" .
+            }
+           }
+    description: |
+     Use this tool to insert or update user data into the Dgraph database.
+     The mutation adds or updates user details like name, email, role, and age.
+     Example: Add users Alice and Bob as admins with specific ages.
+    parameters:
+      - name: user1
+        type: string
+        description: Alice
+      - name: email1
+        type: string
+        description: alice@email.com
+      - name: user2
+        type: string
+        description: Bob
+      - name: email2
+        type: string
+        description: bob@email.com
+```
+
+## Reference
+| **field**   |  **type** | **required** | **description**                                                                                    |
+|-------------|----------:|:------------:|----------------------------------------------------------------------------------------------------|
+| kind        |    string |     true     | Must be "dgraph-dql".                                                                              |
+| source      |    string |     true     | Name of the source the dql query should execute on.                                                |
+| description |    string |     true     | Description of the tool                                                                            |
+| statement   |    string |     true     | dql statement to execute                                                                           |
+| isQuery     |   boolean |     false    | To run statment as query set true otherwise false                                                  |
+| timeout     |   string  |     false    | To set timout for query                                                                            |
+| parameters  | parameter |     true     | List of [parameters](README.md#specifying-parameters) that will be used with the dql statement.    |
--- a/internal/sources/dgraph/dgraph.go
+++ b/internal/sources/dgraph/dgraph.go
@@ -1,0 +1,379 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dgraph
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const SourceKind string = "dgraph"
+
+// validate interface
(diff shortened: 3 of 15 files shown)


ryanfoxtyler:
@kurtisvg is there an account we need to give access to for the fork so that the workflows can check out the code since it's private?

docs/en/resources/sources/_index.md
```diff
@@ -26,4 +26,15 @@ sources:
 In implementation, each source is a different connection pool or client that used
 to connect to the database and execute the tool. 
 
-## Available Sources
\ No newline at end of file
+## Kinds of Sources
```
kurtisvg: Can you revert the changes to this page? It should automatically populate once you add the proper header to the page below (you'll be able to validate on the preview)

docs/en/resources/sources/dgraph.md
```diff
@@ -0,0 +1,37 @@
+# Dgraph Source
```
kurtisvg:
```suggestion
---
title: "Dgraph"
type: docs
weight: 1
description: >
  Dgraph is a horizontally scalable and distributed graph database.

---

## About 
```

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
+// Copyright 2024 Google LLC
```
kurtisvg: nit: this needs to be 2025 now for new files

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+func (hc *DgraphClient) ExecuteQuery(query string, paramsMap map[string]interface{},
+	isQuery bool, timeout string) ([]byte, error) {
+	if isQuery {
+		return hc.postDqlQuery(query, paramsMap, timeout)
+	} else {
+		return hc.mutate(query, paramsMap)
+	}
+}
+
+func (hc *DgraphClient) postDqlQuery(query string, paramsMap map[string]interface{}, timeout string) ([]byte, error) {
```
kurtisvg, line 140: Could we make sure to add brief comments for doc, particularly for non-interface methods.

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	type params struct {
+		Query     string                 `json:"query"`
+		Variables map[string]interface{} `json:"variables"`
+	}
+
+	p := params{
+		Query:     query,
+		Variables: paramsMap,
+	}
```
kurtisvg: nit: I think this is the only use. Could this be an anon struct? 
```suggestion
	type params 
	p := struct{
		Query     string                 `json:"query"`
		Variables map[string]interface{} `json:"variables"`
	}{
		Query:     query,
		Variables: paramsMap,
	}
```

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return hc.doReq(req)
+}
+
+func (hc *DgraphClient) doReq(req *http.Request) ([]byte, error) {
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Printf("[WARNING] error closing response body: %v", err)
```
kurtisvg: This log is using Printf instead of the logging library, which means it's avoiding the user's preferences on logging. 

I think we should probably more explicitly close the body before return? (Alternatively we need to refactor to get the logging library down here. )

kurtisvg:
> @kurtisvg is there an account we need to give access to for the fork so that the workflows can check out the code since it's private?

It looks like the [recommended approach here is to use a PAT](https://github.com/actions/checkout?tab=readme-ov-file#checkout-multiple-repos-private) -- so if you give permission to me, I will create a PAT we can use for this.

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+	}()
+
+	if err != nil {
+		if strings.Contains(err.Error(), "Token is expired") {
+			if err = hc.loginUsingToken(); err != nil {
+				if hc.HttpToken != nil {
+					req.Header.Add("X-Dgraph-AccessToken", hc.AccessJwt)
+				}
+				resp, err = client.Do(req)
+				if err != nil {
```
kurtisvg: Any reason to not preemptively add the header as opposed to waiting for a failed response first?

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+					return nil, err
+				}
+			} else {
+				return nil, err
+			}
```
kurtisvg: I'm concerned with the readability of this statement -- namely it looks like we're replacing err with a version that might be nil, and if it is nil we are still returning `err` (but the new err, not the previous one?).

Overall, would prefer if we refactor to use guard clauses (and keep the happy path to the left if possible).

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+		return nil, fmt.Errorf("error reading response body: url: [%v], err: [%v]", req.URL, err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("got non 200 resp: %v", string(respBody))
+	}
+
+	return respBody, nil
+}
+
+func (hc *DgraphClient) login() error {
```
kurtisvg: nit: Should this be `loginWithCredentials`?

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+	credentials := map[string]interface{}{
+		"userid":    hc.UserId,
+		"password":  hc.Password,
+		"namespace": hc.Namespace,
+	}
+	return hc.doLogin(credentials)
+}
+
+func (hc *DgraphClient) loginUsingToken() error {
```
kurtisvg: nit:
```suggestion
func (hc *DgraphClient) loginWiithToken() error {
```

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,381 @@
...
+	p := params{
+		Query:     query,
+		Variables: paramsMap,
+	}
+	body, err := json.Marshal(p)
+	if err != nil {
+		return nil, fmt.Errorf("error marshlling json: %v", err)
+	}
+
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(body))
```
kurtisvg, line 159: Would it be better to use an httpClient here so there's connection pooling on the requests?

docs/en/resources/sources/_index.md
shivaji-kharse: done

internal/sources/dgraph/dgraph.go
shivaji-kharse: done

internal/sources/dgraph/dgraph.go
shivaji-kharse: I've switched to using defer resp.Body.Close() to handle closing the response body. However, I'm not currently using a logging library. Could you please point me to an example of how to log warnings using the logging library?

kurtisvg: The logger is attached to the server, example of a debug statement here: https://github.com/googleapis/genai-toolbox/blob/b0c94946d6b9bbd462ed7c80eb68e42333ea7f07/internal/server/api.go#L81

The logger isn't exposed here because generally we prefer to return errors and log them higher up the stack. If we need to we can revisit that decision.

internal/sources/dgraph/dgraph.go
```diff
@@ -0,0 +1,378 @@
...
+
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const SourceKind string = "dgraph"
+
+// validate interface
+var _ sources.SourceConfig = Config{}
+var httpClient = &http.Client{}
```
kurtisvg: Rather than global state for this, can we please move it into the `DraphClient`?

docs/en/resources/sources/dgraph.md
```diff
@@ -0,0 +1,49 @@
...
+When **connecting to a hosted Dgraph database**, this source uses the API key for access. If you are using a dedicated environment, you will additionally need the namespace and user credentials for that namespace.
+
+For **connecting to a local or self-hosted Dgraph database**, use the namespace and user credentials for that namespace.
+
+## Example
+
+```yaml
+sources:
+    my-dgraph-source:
+        dgraphUrl: "https://xxxx.cloud.dgraph.io"
```
Yuan325, line 39:
```suggestion
        kind: "dgraph"
        dgraphUrl: "https://xxxx.cloud.dgraph.io"
```

docs/en/resources/sources/dgraph.md
```diff
@@ -0,0 +1,49 @@
...
+        apiKey: abc123
+        namepace : 0
+```
+
+## Reference
+
+| **Field**   | **Type** | **Required** | **Description**                                                                                  |
+|-------------|:--------:|:------------:|--------------------------------------------------------------------------------------------------|
+| kind        |  string  |     true     | Must be "dgraph".                                                                                |
+| dgraphurl   |  string  |     true     | Connection URI (e.g. "https://xxx.cloud.dgraph.io", "https://localhost:8080").                   |
```
Yuan325:
```suggestion
| dgraphUrl   |  string  |     true     | Connection URI (e.g. "https://xxx.cloud.dgraph.io", "https://localhost:8080").                   |
```

docs/en/resources/sources/dgraph.md
```diff
@@ -0,0 +1,49 @@
...
+If you're new to Dgraph, the fastest way to get started is to [sign up for Dgraph Cloud][dgraph-login].
+
+[dgraph-docs]: https://dgraph.io/docs
+[dgraph-login]: https://cloud.dgraph.io/login
+
+## Requirements 
+
+### Database User
+
+When **connecting to a hosted Dgraph database**, this source uses the API key for access. If you are using a dedicated environment, you will additionally need the namespace and user credentials for that namespace.
```
Yuan325: nit: please consider wrapping the markdown around ~80 lines.

Can use library such as [ReWrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) (if using VScode)

docs/en/resources/sources/dgraph.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "Dgraph"
+type: docs
```
Yuan325, line 3:
```suggestion
type: docs
weight: 1
```

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Example
+
+### Query:
+
+```yaml
+tools:
+  search_user:
+    kind: dgraph-dql
+    source: dgraph-user-instance
```
Yuan325:
```suggestion
    source: my-dgraph-source
```

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+          "age": 35
+        },
+        {
+          "name": "Bob",
+          "role": "admin",
+          "age": 45
+        }
+      ]
+    parameters:
+      - name: $role
```
Yuan325, line 50:
```suggestion
      - name: role
```

can we handle the param name within `tools/parameters.go`? similar to `AsMapByOrderedKeys()`.

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+        description: admin
+```
+
+### Mutation: 
+
+```yaml
+tools:
+  dgraph-manage-user-instance:
+    kind: dgraph-dql
+    source: dgraph-manage-user-instance
```
Yuan325:
```suggestion
    source: my-dgraph-source
```

both query and mutation will use the same source right?

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+            _:user2 <role> "admin" .
+            _:user2 <age> "45" .
+            }
+           }
+    description: |
+     Use this tool to insert or update user data into the Dgraph database.
+     The mutation adds or updates user details like name, email, role, and age.
+     Example: Add users Alice and Bob as admins with specific ages.
+    parameters:
+      - name: $user1
```
Yuan325:
```suggestion
      - name: user1
```

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+        type: string
+        description: Alice
+      - name: $email1
```
Yuan325:
```suggestion
      - name: email1
```

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+        type: string
+        description: alice@email.com
+      - name: $user2
```
Yuan325:
```suggestion
      - name: user2
```

docs/tools/dgraph-dql.md
```diff
@@ -0,0 +1,105 @@
...
+        type: string
+        description: Bob
+      - name: $email2
```
Yuan325:
```suggestion
      - name: email2
```

tests/dgraph.yaml
```diff
@@ -0,0 +1,31 @@
+sources:
+  dgraph-manage-user-instance:
+    kind: "dgraph"
+    dgraphUrl: "https://green-feather-41401502.ap-south-1.aws.cloud.dgraph.io"
+    apiKey: "OTJmMTc0NDQ4MTJmZDk3MTlmMWY0ZjMzYmE2YjZkNzc="
```
Yuan325: Is this okay to be included here? If you want to rotate this password and send it to me, I can include it in GCP secrets.

tests/dgraph.yaml
```diff
@@ -0,0 +1,31 @@
+sources:
```
Yuan325, line 2: Where are we using this file?

docs/tools/dgraph-dql.md
shivaji-kharse: yes

tests/dgraph.yaml
shivaji-kharse: I just added it for testing purposes; I have removed it now.

kurtisvg:
I was able to get tests to pass on [this commit](https://github.com/googleapis/genai-toolbox/pull/259/commits/a95914acfa67c8922db8cf9663062c99bd9e2000) by pushing directly to repo. Will skip adding a PAT for now since we plan to open repo next week.

PR #244, opened 2025-01-29
feat: add user agent to cloud databases

Yuan325:
Add user agent to cloud databases that provides us anonymized data request count, number of users, number of projects, and other environment settings.

User agent is using the format: `genai-toolbox/$version+metadata`

### DIFF
--- a/internal/sources/alloydbpg/alloydb_pg.go
+++ b/internal/sources/alloydbpg/alloydb_pg.go
@@ -110,7 +110,8 @@ func initAlloyDBPgConnectionPool(ctx context.Context, tracer trace.Tracer, name,
 	}
 
 	// Create a new dialer with options
-	opts, err := getOpts(ipType, ctx.Value(util.UserAgentKey).(string))
+	userAgent := ctx.Value(util.UserAgentKey).(string)
+	opts, err := getOpts(ipType, userAgent)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/sources/cloudsqlmssql/cloud_sql_mssql.go
+++ b/internal/sources/cloudsqlmssql/cloud_sql_mssql.go
@@ -99,7 +99,8 @@ func initCloudSQLMssqlConnection(ctx context.Context, tracer trace.Tracer, name,
 	dsn := fmt.Sprintf("sqlserver://%s:%s@%s?database=%s&cloudsql=%s:%s:%s", user, pass, ipAddress, dbname, project, region, instance)
 
 	// Get dial options
-	opts, err := sources.GetCloudSQLOpts(ipType, ctx.Value(util.UserAgentKey).(string))
+	userAgent := ctx.Value(util.UserAgentKey).(string)
+	opts, err := sources.GetCloudSQLOpts(ipType, userAgent)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


cmd/root.go
```diff
@@ -208,7 +208,7 @@ func run(cmd *Command) error {
 	}
 
 	// start server
-	s, err := server.NewServer(ctx, cmd.cfg, cmd.logger)
+	s, err := server.NewServer(ctx, cmd.cfg, cmd.logger, cmd.cfg.Version)
```
kurtisvg: If we already pass in cmd.cfg, what do we need cmd.cfg.Version as well?

internal/sources/util.go
```diff
@@ -0,0 +1,41 @@
...
+			cloudsqlconn.WithUserAgent(userAgent),
+		}, nil
+	case "public":
+		return []cloudsqlconn.Option{
+			cloudsqlconn.WithDefaultDialOptions(cloudsqlconn.WithPublicIP()),
+			cloudsqlconn.WithUserAgent(userAgent),
+		}, nil
+	default:
+		return nil, fmt.Errorf("invalid ipType %s", ipType)
+	}
```
kurtisvg:
```suggestion
        opts := []cloudsqlconn.DialOption{ cloudsqlconn.WithUserAgent(userAgent) }
	switch strings.ToLower(ipType) {
	case "private":
		opts := append (opts, cloudsqlconn.WithPrivateIP())
	case "public":
		opts := append (opts, cloudsqlconn.WithPublicIP())
	default:
		return nil, fmt.Errorf("invalid ipType %s", ipType)
	}
	return opts, nil
```

internal/server/server_test.go
```diff
@@ -54,8 +54,9 @@ func TestServe(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %s", err)
 	}
+	testUserAgent := "genai-toolbox/test"
```
kurtisvg: nit: should we set an actual version here

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -48,8 +48,8 @@ func (r Config) SourceConfigKind() string {
 	return SourceKind
 }
 
-func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.Source, error) {
-	pool, err := initAlloyDBPgConnectionPool(ctx, tracer, r.Name, r.Project, r.Region, r.Cluster, r.Instance, r.IPType.String(), r.User, r.Password, r.Database)
+func (r Config) Initialize(ctx context.Context, tracer trace.Tracer, userAgent string) (sources.Source, error) {
```
kurtisvg: should we add this to the context instead? It's not being used in most of the tools, wonder if that would be better

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -107,11 +110,11 @@ func initAlloyDBPgConnectionPool(ctx context.Context, tracer trace.Tracer, name,
 	}
 
 	// Create a new dialer with options
-	dialOpts, err := getDialOpts(ipType)
+	opts, err := getOpts(ipType, ctx.Value(util.UserAgentKey).(string))
```
kurtisvg: nit: would do the extraction from context in a separate line

PR #261, opened 2025-02-03
chore: merge two sdk lints

Yuan325:
Merging lint check for sdks with matrix.


Yuan325:
Might not make sense since the two sdks operate as a separate library.

PR #266, opened 2025-02-05
fix: update tool invoke to return json

Yuan325:
Return actual rows as `[]any` that contains `map` of results. Each `map` represent a row, with the key being column name.


tests/common_test.go
```diff
@@ -335,13 +335,14 @@ func RunToolInvocationWithParamsTest(t *testing.T, sourceConfig map[string]any,
...
 			}
-			got, ok := body["result"].(string)
+			got, ok := body["result"].([]any)
 			if !ok {
 				t.Fatalf("unable to find result in response body")
 			}
 
-			if got != tc.want {
-				t.Fatalf("unexpected value: got %q, want %q", got, tc.want)
+            gotS := ConvertResultToString(got)
```
kurtisvg: Can you use json.Marshall here instead of ConvertResultToString()?

internal/tools/mssqlsql/mssqlsql.go
```diff
@@ -123,39 +121,43 @@ func (t Tool) Invoke(params tools.ParamValues) (string, error) {
...
-			return "", fmt.Errorf("unable to parse row: %w", err)
+			return nil, fmt.Errorf("unable to parse row: %w", err)
 		}
-		out.WriteString(fmt.Sprintf("%s", v))
+		vMap := make(map[string]any)
+		for i, name := range cols {
+			vMap[name] = v[i]
+		}
+		out = append(out, vMap)
 	}
```
kurtisvg: I think this can be simplified down:
```
	var out []map[string]any
	for rows.Next() {
                 vals := make([]any, len(cols))
		err = rows.Scan(vals...)
		if err != nil {
			return nil, fmt.Errorf("unable to parse row: %w", err)
		}
		r := make(map[string]any, len(cols))
		for i, name := range cols {
			r[name] = vals[i]
		}
		out = append(out, r)
	}
```

internal/server/api.go
```diff
@@ -227,7 +227,7 @@ var _ render.Renderer = &resultResponse{} // Renderer interface for managing res
 
 // resultResponse is the response sent back when the tool was invocated successfully.
 type resultResponse struct {
-	Result string `json:"result"` // result of tool invocation
+	Result []any `json:"result"` // result of tool invocation
```
kurtisvg: My thought was that this should stay a "string", but we should call `json.Marshall` on the map to convert it to a proper json string?

internal/tools/mssqlsql/mssqlsql.go
```diff
@@ -123,39 +121,43 @@ func (t Tool) Invoke(params tools.ParamValues) (string, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("unable to fetch column types: %w", err)
+		return nil, fmt.Errorf("unable to fetch column types: %w", err)
 	}
-	v := make([]any, len(types))
-	pointers := make([]any, len(types))
-	for i := range types {
+	cl := len(cols)
+	v := make([]any, cl)
+	pointers := make([]any, cl)
```
kurtisvg:
```suggestion
        // create an array of values for each column, which can be re-used to scan each row
	rawValues := make([]any, len(cols))
	values := make([]any, len(cols))
```

PR #267, opened 2025-02-05
fix(langchain-sdk): Fix issue occurring when using a tool with list type.

anubhav756:
Temporarily relaxes type checking in the toolbox manifest handling. Currently, the toolbox manifest does not include `items` information (along with the `type` field) in the `parameters` field of the manifest. This change allows lists containing any type of value (except nested lists) as a workaround. This will be reverted once the toolbox manifest provides type information.

[!NOTE]
This is a temporary fix. It allows tools relying on the manifest to function until the toolbox's manifest includes type information.

### DIFF
--- a/sdks/langchain/tests/test_utils.py
+++ b/sdks/langchain/tests/test_utils.py
@@ -171,7 +171,7 @@ class TestUtils:
             ("integer", int),
             ("number", float),
             ("boolean", bool),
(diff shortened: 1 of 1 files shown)


Yuan325:
Migrate to https://github.com/googleapis/genai-toolbox-langchain-python/pull/33

PR #285, opened 2025-02-07
ci: reorganize integration test files

Yuan325:
Reorganize integration tests, mainly between postgres, mysql, and sql server that shares alot similar code. 
* separate commons_test.go to multiple file:
    * server_test.go: for setting up server etc.
    * tool_test.go: for testing tools endpoint, including testing tools with parameters or auth-invokes
    * source_test.go: for testing source such as connection tests or setting up source-related code
    * auth_test.go: only contain auth set up related code
    * common_test.go: consist of common and general set up such as getting tools config etc.
* Move `switch-case` into individual source's test file so common files only contains code that don't need to be altered when adding new integration test. This will ease future debugging and adding new tests.

The following integration tests are not updated (does not have testing of tool invoke with parameters): 
- dgraph 
- neo4j
- spanner (will require more changes, will update in a separate PR)

This PR also does not re-organize `tests/` folder structure.


tests/alloydb_pg_integration_test.go
```diff
@@ -155,237 +160,22 @@ func TestAlloyDBSimpleToolEndpoints(t *testing.T) {
...
-}
-
-func TestAlloyDBAuthRequiredToolInvocation(t *testing.T) {
-	// create test configs
-	sourceConfig := requireAlloyDBPgVars(t)
-
-	// call generic auth test helper
-	RunAuthRequiredToolInvocationTest(t, sourceConfig, "postgres-sql")
-
+	RunSourceConnectionTest(t, sourceConfig, ALLOYDB_POSTGRES_TOOL_KIND)
```
kurtisvg: nit: would this be better as a table test?

kurtisvg (approved):

PR #293, opened 2025-02-14
fix: add items to parameter manifest

### LINKED ISSUE #292: Array type parameter manifest doesn't include the items type
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Is this a client library issue or a product issue? We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [Support page](https://cloud.google.com/support) to reach the most relevant engineers.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - Programming language:
  - OS:
  - Language runtime version:
  - Package version:

#### Steps to reproduce

  1. Add an array parameter to one of the tool
  2. When running get manifest endpoint, items is not included within the manifest.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

Yuan325:
`items` are not specified in the manifest for parameters, which means the client doesn't parse them correctly when presented to the LLMs. This PR adds `items` as part of manifest.

Example manifest:
```
{
    "serverVersion": "0.1.0",
    "tools": {
        "tool_name": {
            "description": "description of the tool...",
            "parameters": [{
                "name": "foos",
                "type": "array",
                "description": "A list of foo.",
                "authSources": [],
                "items": {"name": "foo", "type": "string", "description": "foobar.", "authSources": []}
            }]
        }
    }
}
```

Fixes #292 🦕

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -234,11 +234,11 @@ func (ps Parameters) Manifest() []ParameterManifest {
 
 // ParameterManifest represents parameters when served as part of a ToolManifest.
 type ParameterManifest struct {
-	Name        string   `json:"name"`
-	Type        string   `json:"type"`
-	Description string   `json:"description"`
-	AuthSources []string `json:"authSources"`
-	// Parameter   *ParameterManifest `json:"parameter,omitempty"`
+	Name        string             `json:"name"`
+	Type        string             `json:"type"`
+	Description string             `json:"description"`
+	AuthSources []string           `json:"authSources"`
+	Items       *ParameterManifest `json:"items,omitempty"`
 }
 
 // CommonParameter are default fields that are emebdding in most Parameter implementations. Embedding this stuct will give the object Name() and Type() functions.
@@ -259,7 +259,7 @@ func (p *CommonParameter) GetType() string {
 	return p.Type
 }
 
-// GetType returns the type specified for the Parameter.
+// Manifest returns the manifest for the Parameter.
 func (p *CommonParameter) Manifest() ParameterManifest {
 	// only list ParamAuthSource names (without fields) in manifest
 	authNames := make([]string, len(p.AuthSources))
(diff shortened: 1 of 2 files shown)


anubhav756 (changes requested):
Should we also update the docs and the examples?

kurtisvg:
> Should we also update the docs and any examples?

This is a protocol only change, so I'm not sure there is anything to update.

anubhav756:
> > Should we also update the docs and any examples?
> 
> This is a protocol only change, so I'm not sure there is anything to update.

I assume that with this PR, users can still pass a `type` as `"array"` and can still run into something like:
```
400 Unable to submit request because ... schema didn't specify the schema type field
```
unless they specify the new `items` field, right?

If my assumption is correct, then I assume we might want to guide the users to add the `items` field if they want to pass in `array` as `type`, right?

*Side note:* If I'm correct, then should we potentially throw an error if the `type` is `"array"` but `items` field is missing?

anubhav756 (commented):
Thinking-out-loud; What would be pros and cons of a simpler approach like:

```json
{
    "serverVersion": "0.1.0",
    "tools": {
        "tool_name": {
            "description": "description of the tool...",
            "parameters": [{
                "name": "foos",
                "type": "array[string]",        <--- Note that we just pass in the type here
                "description": "A list of foo.",
                "authSources": [],
            }]
        }
    }
}
```

Yuan325:
Making the item field might be in a separate PR (I saw there's a bug for this actually, will fix this!)

Regarding to simplifying the config, I'm guessing this might complicate the unmarshalling of a complicated type (perhaps multiple layers of nesting although that's technically unlikely), and possibly make it more restrictive in the case where we want to further enhance the capabilities of Parameter in the future?

kurtisvg (approved):

PR #297, opened 2025-02-18
feat!: deprecate authsource in favor of authservice

Yuan325:
Rename existing `authSource` to `authService` through deprecation. `AuthService` more clearly distinguishes it from `Sources` objects. 

`authSources` will be converted into `authServices` after the unmarshalling process. A warning log is shown if `authSources` are used (for both within tools parameters and defining auth services):
```
2025-02-20T13:57:51.156025-08:00 WARN "`authSources` is deprecated, use `authServices` for parameters instead"
2025-02-20T13:57:51.156569-08:00 WARN "`authSources` is deprecated, use `authServices` instead"
2025-02-20T13:57:52.047584-08:00 INFO "Initialized 1 sources."
...
```

The manifest generated will continue to use `authSources` to keep compatibility with the sdks:
```
{
"serverVersion":"0.1.0",
"tools":{
  "test_tool2":{
    "description":"Use this tool to test\n",
    "parameters":[{
      "name":"user_id",
      "type":"string",
      "description":"Auto-populated from Google login",
      "authSources":["my-google-auth"]
      }]
    }
  }
}
```

Test cases with `authSources` are kept for compatibility. Will be removed when `authSources` are no longer supported.

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -97,12 +97,12 @@ func (p ParamValues) AsMapWithDollarPrefix() map[string]interface{} {
 	return params
 }
 
-func parseFromAuthSource(paramAuthSources []ParamAuthSource, claimsMap map[string]map[string]any) (any, error) {
-	// parse a parameter from claims using its specified auth sources
-	for _, a := range paramAuthSources {
+func parseFromAuthService(paramAuthServices []ParamAuthService, claimsMap map[string]map[string]any) (any, error) {
+	// parse a parameter from claims using its specified auth services
+	for _, a := range paramAuthServices {
 		claims, ok := claimsMap[a.Name]
 		if !ok {
-			// not validated for this authsource, skip to the next one
+			// not validated for this authservice, skip to the next one
 			continue
 		}
 		v, ok := claims[a.Field]
@@ -120,9 +120,9 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
 	params := make([]ParamValue, 0, len(ps))
 	for _, p := range ps {
 		var v any
-		paramAuthSources := p.GetAuthSources()
+		paramAuthServices := p.GetAuthServices()
 		name := p.GetName()
-		if paramAuthSources == nil {
+		if paramAuthServices == nil {
 			// parse non auth-required parameter
 			var ok bool
 			v, ok = data[name]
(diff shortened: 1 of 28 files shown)


internal/tools/parameters.go
```diff
@@ -238,19 +257,20 @@ func (ps Parameters) Manifest() []ParameterManifest {
...
 type ParameterManifest struct {
-	Name        string             `json:"name"`
-	Type        string             `json:"type"`
-	Description string             `json:"description"`
-	AuthSources []string           `json:"authSources"`
-	Items       *ParameterManifest `json:"items,omitempty"`
+	Name         string             `json:"name"`
+	Type         string             `json:"type"`
+	Description  string             `json:"description"`
+	AuthServices []string           `json:"authServices"`
```
kurtisvg: The change in the protocol here means that we'll need to update the SDKs as well -- @anubhav756 @twishabansal FYI

internal/tools/parameters.go
```diff
@@ -199,30 +198,50 @@ func parseParamFromDelayedUnmarshaler(u *util.DelayedUnmarshaler) (Parameter, er
 		if err := dec.Decode(a); err != nil {
 			return nil, fmt.Errorf("unable to parse as %q: %w", t, err)
 		}
+		if a.AuthSources != nil {
+			a.AuthServices = append(a.AuthServices, a.AuthSources...)
+			a.AuthSources = nil
+		}
```
kurtisvg: Should there be a deprecation warning at this level as well?

kurtisvg (changes requested):

internal/util/util.go
```diff
@@ -64,3 +65,16 @@ func NewStrictDecoder(v interface{}) (*yaml.Decoder, error) {
...
 // LoggerKey is the key used to store logger within context
 const LoggerKey contextKey = "logger"
+
+// WithLogger adds a logger into the context as a value
+func WithLogger(ctx context.Context, logger log.Logger) context.Context {
+	return context.WithValue(ctx, LoggerKey, logger)
+}
+
+// LoggerFromContext retreives the logger or return an error
+func LoggerFromContext(ctx context.Context) (log.Logger, error) {
```
kurtisvg: nit: I like this functionality, but I think it should be a separate PR if possible. 

Additionally, do we still need to export LoggerKey if we use this function instead?

kurtisvg (approved):

internal/util/util.go
Yuan325: Oh i dont think we'll have to export it. will update!

PR #305, opened 2025-02-25
ci: Add unit test code coverage check

duwenxin99:
Edited the unit test workflow:
- The unit test coverage calculation excludes the `sources` and `tools` packages (but we still run them without coverage).
- Generate a code coverage analysis file `coverage.out` and fail the unit test if the total coverage is below 40% (target at 90%)


.github/workflows/tests.yaml
```diff
@@ -16,7 +16,7 @@ name: tests
 on:
   push:
     branches:
-    - 'main'
+      - "main"
```
Yuan325: single quote seems to be preferable in yaml unless we need to do escaping. https://symfony.com/doc/current/reference/formats/yaml.html

Yuan325 (changes requested):

.github/workflows/tests.yaml
duwenxin99: updated quotes

Yuan325 (approved):

PR #322, opened 2025-03-10
fix: Use driver for Spanner implementation

duwenxin99:
1. Spanner client library only supports read transactions for the `Query()` API. To support both read and write, we switch to [Google Cloud Spanner Drivier](https://github.com/googleapis/go-sql-spanner/tree/82b7f84c07bf8fb48da364bd3999becfbfab4003)
2. Change Spanner integration test to run generic test suite.


duwenxin99:
database/sql doesn't support user agent implementation

PR #327, opened 2025-03-12
feat: support connection lifecycle initialize phase

Yuan325:
Add support for the `initialize` phase for mcp's connection lifecycle.

After receiving the `initialize` request, the server will respond with it's own capabilities and information. Since there are only one protocol supported, we don't have to do protocol version comparison. Toolbox server will response with the default protocol version.

Initial support for MCP Tools will not include list changed notification.

Test cases included.

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -15,6 +15,10 @@
 package server
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
@@ -39,12 +43,21 @@ func mcpRouter(s *Server) (chi.Router, error) {
 
 // mcpHandler handles all mcp messages.
 func mcpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
+	// Read and returns a body from io.Reader
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		// Generate a new uuid if unable to decode
+		id := uuid.New().String()
+		render.JSON(w, r, newJSONRPCError(id, mcp.PARSE_ERROR, err.Error(), nil))
+	}
+
+	// Generic baseMessage could either be a JSONRPCNotification or JSONRPCRequest
 	var baseMessage struct {
 		Jsonrpc string        `json:"jsonrpc"`
 		Method  string        `json:"method"`
 		Id      mcp.RequestId `json:"id,omitempty"`
 	}
-	if err := decodeJSON(r.Body, &baseMessage); err != nil {
+	if err := decodeJSON(bytes.NewBuffer(body), &baseMessage); err != nil {
 		// Generate a new uuid if unable to decode
 		id := uuid.New().String()
 		render.JSON(w, r, newJSONRPCError(id, mcp.PARSE_ERROR, err.Error(), nil))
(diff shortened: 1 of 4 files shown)


internal/server/api.go
```diff
@@ -266,12 +266,34 @@ func mcpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+
+	var res mcp.JSONRPCMessage
+	switch baseMessage.Method {
+	case "initialize":
+		var req mcp.InitializeRequest
+		if err := json.Unmarshal(message, &req); err != nil {
+			res = newJSONRPCError(baseMessage.Id, mcp.INVALID_REQUEST, fmt.Sprintf("invalid mcp initialize request: %s", err), nil)
+			break
+		}
+		result := mcp.InitializeHandler(s.version)
```
kurtisvg: Name could be improved

internal/server/api_test.go
```diff
@@ -265,16 +267,24 @@ func TestMcpEndpoint(t *testing.T) {
...
 			body: mcp.JSONRPCRequest{
 				Jsonrpc: jsonrpcVersion,
-				Id:      "basic-mcp",
+				Id:      "mcp-initialize",
 				Request: mcp.Request{
-					Method: "foo",
+					Method: "initialize",
+				},
+			},
+			want: fmt.Sprintf(`{"jsonrpc":"2.0","id":"mcp-initialize","result":{"protocolVersion":"%s","capabilities":{"tools":{"listChanged":false}},"serverInfo":{"name":"%s","version":"%s"}}}`, protocolVersion, serverName, fakeVersionString),
```
kurtisvg: nit: worth using a map and decoding to json to compare?

kurtisvg: I think other tests are doing this somewhere

internal/server/mcp/types.go
```diff
@@ -120,7 +123,106 @@ const (
...
+		Subscribe *bool `json:"subscribe,omitempty"`
+		// Whether this server supports notifications for changes to the resource
+		// list.
+		ListChanged *bool `json:"listChanged,omitempty"`
+	} `json:"resources,omitempty"`
+	// Present if the server offers any tools to call.
+	Tools *struct {
+		// Whether this server supports notifications for changes to the tool list.
+		ListChanged *bool `json:"listChanged,omitempty"`
+	} `json:"tools,omitempty"`
```
kurtisvg: avoid the anon structs

kurtisvg (changes requested):

internal/server/mcp.go
```diff
@@ -63,12 +75,34 @@ func mcpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
-		Jsonrpc: mcp.JSONRPC_VERSION,
-		Id:      baseMessage.Id,
-		Result:  mcp.EmptyResult{},
+	// Check if message is a notification
+	if baseMessage.Id == nil {
+		var notification mcp.JSONRPCNotification
+		if err := json.Unmarshal(body, &notification); err != nil {
+			render.JSON(w, r, newJSONRPCError(baseMessage.Id, mcp.PARSE_ERROR, err.Error(), nil))
+		}
+		// Notifications do not expect a response
```
kurtisvg, line 85: nit: add comment that says "toolbox doesn't do anything with notifications yet, so we ignore them"

internal/server/mcp.go
```diff
@@ -39,12 +42,21 @@ func mcpRouter(s *Server) (chi.Router, error) {
...
+	}
+
+	// Generic baseMessage could either be a JSONRPCNotification or JSONRPCRequest
 	var baseMessage struct {
 		Jsonrpc string        `json:"jsonrpc"`
 		Method  string        `json:"method"`
 		Id      mcp.RequestId `json:"id,omitempty"`
 	}
-	if err := decodeJSON(r.Body, &baseMessage); err != nil {
+	if err := json.Unmarshal(body, &baseMessage); err != nil {
```
kurtisvg: nit: should keep using decodeJSON here (just to be safe)

kurtisvg (approved):

PR #330, opened 2025-03-17
docs: add llamaindex sdk info to toolbox docs

twishabansal:


anubhav756 (approved):
nit: Seems to also have some other changes as well.

twishabansal:
> nit: Seems to also have some other changes as well.

Mostly small corrects like using sync interfaces by default and the correct library name for langchain.

Is there anything you see that needs to be reverted?

PR #337, opened 2025-03-19
feat: add tools list method

Yuan325:
Add `tools/list` method. 

Example to call the `tools/list` method:
`curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc": "2.0", "id": 1, "method": "tools/list"}' http://127.0.0.1:5000/mcp`

Toolbox will response. with a valid JSONRPCResponse with a list of all tools available.

### DIFF
--- a/internal/server/mcp_test.go
+++ b/internal/server/mcp_test.go
@@ -121,7 +121,7 @@ func TestMcpEndpoint(t *testing.T) {
 						},
 						map[string]any{
 							"name":        "array_param",
-							"desciprtion": "some description",
+							"description": "some description",
 							"inputSchema": tool3InputSchema,
 						},
 					},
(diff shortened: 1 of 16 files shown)


internal/server/mcp_test.go
```diff
@@ -69,6 +70,54 @@ func TestMcpEndpoint(t *testing.T) {
...
+			want: map[string]any{
+				"jsonrpc": "2.0",
+				"id":      "tools-list",
+				"result": map[string]any{
+					"tools": []any{
+						map[string]any{
+							"name":        "no_params",
+							"inputSchema": map[string]any{"type": "object", "properties": map[string]any{}, "required": []any{}},
+						},
+						map[string]any{
```
kurtisvg, line 118: nit: consider helper function to see they improve readability

kurtisvg (approved):

internal/server/mcp_test.go
Yuan325: will be complicated to implement due to the array parameter. There's no GetItems for the Parameter interface so if we want to implement that we will need to add that function to both the CommonParameter and ArrayParameter.

Instead, I added toolSchema for the 3 tools separately at the top to help readability.

PR #347, opened 2025-03-24
ci: Add renovate config

duwenxin99:
Add renovate config for repo dependency updates.


.github/renovate.yml
```diff
@@ -0,0 +1,31 @@
...
+      matchManagers: [
+        'github-actions',
+      ],
+      pinDigests: true,
+    },
+    {
+      "groupName": "Non-major dependency updates",
+      "matchManagers": ["gomod"],
+      "matchUpdateTypes": ["minor", "patch"],
+    },
```
kurtisvg: I'm not sure I follow this part -- why do we want to group non-major dependency updates? 

What I see in the other repos is that this is [updating the version of Python](https://github.com/googleapis/genai-toolbox-langchain-python/blob/e4bdcb3c43b2a5a505287ec676e53684bcc2cac8/.github/renovate.json5#L37) (but I think we're doing something different here).

kurtisvg (changes requested):
This is a .yml file. Should it be json5?

.github/renovate.yml
duwenxin99: I was thinking this would reduce the PRs needed for dependency update and simplify the process. But I will remove it if we think it is not necessary.

kurtisvg: I prefer individual updates because it's easier to catch that they break tests

kurtisvg (approved):

PR #356, opened 2025-03-28
feat: support sse

Yuan325:
Tested sse implementation with MCP [inspector](https://modelcontextprotocol.io/docs/tools/inspector) tool.

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -16,10 +16,12 @@ package server
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"sync"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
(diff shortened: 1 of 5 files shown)


internal/server/mcp.go
```diff
@@ -36,11 +42,43 @@ func mcpRouter(s *Server) (chi.Router, error) {
...
+// sseHandler handles sse initialization and message.
+func sseHandler(s *Server, w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+
+	flusher, ok := w.(http.Flusher)
+	if !ok {
+		err := fmt.Errorf("error not supported")
```
kurtisvg: Could we add a better error here?

internal/server/mcp.go
```diff
@@ -36,11 +42,43 @@ func mcpRouter(s *Server) (chi.Router, error) {
...
+	}
+	sessionId := uuid.New().String()
+	sseSession := &sseSession{
+		sessionId: sessionId,
+		writer:    w,
+		flusher:   flusher,
+	}
+	s.sseSessions[sessionId] = sseSession
+	defer delete(s.sseSessions, sessionId)
+
```
kurtisvg: This introduces race conditions. We will need to use a mutex here

internal/server/mcp.go
```diff
@@ -164,6 +202,19 @@ func mcpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+	// retrieve sseSession
+	sessionId := r.URL.Query().Get("sessionId")
+	session, ok := s.sseSessions[sessionId]
+	if !ok {
+		fmt.Printf("sse session not available\n")
+	} else {
+		// send response via sse if sse session is available
+		eventData, _ := json.Marshal(res)
+		fmt.Fprintf(session.writer, "event: message\ndata: %s\n\n", eventData)
+		session.flusher.Flush()
```
kurtisvg: This introduces a potential race condition because it has multiple goroutines writing to the same pipe. 

We'll need to restructure so that a single responses are written at once and they are passed to a single point via a channel.

kurtisvg (changes requested):

internal/server/mcp.go
```diff
@@ -36,11 +71,56 @@ func mcpRouter(s *Server) (chi.Router, error) {
...
+		sessionId:  sessionId,
+		writer:     w,
+		flusher:    flusher,
+		done:       make(chan struct{}),
+		eventQueue: make(chan string, 100),
+	}
+	s.sseManager.add(sessionId, session)
+	defer s.sseManager.remove(sessionId)
+
+	// send initil endpoint event
```
kurtisvg:
```suggestion
	// send initial endpoint event
```

internal/server/mcp.go
```diff
@@ -36,11 +71,56 @@ func mcpRouter(s *Server) (chi.Router, error) {
...
+	messageEndpoint := fmt.Sprintf("http://127.0.0.1:5000/mcp?sessionId=%s", sessionId)
```
kurtisvg: need to figure out how to make this more flexible

kurtisvg (changes requested):

kurtisvg (approved):

PR #384, opened 2025-04-02
fix: add `tools-file` flag and deprecate `tools_file` 

### LINKED ISSUE #383: Add support for `--tools-file` alias
I often find myself tripping up and attempting to launch Toolbox with the `--tools-file` flag.

The actual flag is `--tools_file` using an underscore... yet all other flags do not use this pattern.

Adding an alias for `--tools-file` is probably the best path forward to consolidate flag naming pattern without introducing a breaking change.

Yuan325:
Add `tools-file` flag and deprecate `tools_file` flag. This is not a breaking change. The old `tools_file` flag is still usable.

User will see the following message when using `tools_file` flag: 
```
Flag --tools_file has been deprecated, please use --tools-file instead
2025-04-03T10:09:12.803165-07:00 INFO "Initialized 2 sources."
```

Help command will reveal the new flag: 
```
...
      --telemetry-service-name string   Sets the value of the service.name resource attribute for telemetry data. (default "toolbox")
      --tools-file string               File path specifying the tool configuration. (default "tools.yaml")
  -v, --version                         version for toolbox
...
```

Fixes #383


cmd/root.go
```diff
@@ -118,6 +120,16 @@ func NewCommand(opts ...Option) *Command {
 	return cmd
 }
 
+// wordSepNormalizeFunc normalize flag names so that - and _ is the same.
+func wordSepNormalizeFunc(f *pflag.FlagSet, name string) pflag.NormalizedName {
+	from := []string{"_"}
+	to := "-"
+	for _, sep := range from {
+		name = strings.Replace(name, sep, to, -1)
+	}
```
kurtisvg: Why do we iterate over a array with len 1?
```suggestion
	name = strings.ReplaceAll(name, "-", "_")
```

kurtisvg (changes requested):
Are we worried about any future backwards compatibility features here? If we start accepting both hypen and underscore we will have to forever? 

Alternatively, would we be better to add a one-off exception for tools-file? 

We should also update docs to the preffered

kurtisvg (approved):

PR #386, opened 2025-04-02
fix: update required checks name

Yuan325:
renovatebot will skip test with the name `tests / unit tests (pull_request)`. instead of `tests / unit tests (ubuntu-latest) (pull_request)`. Hence it will mark checks as incomplete since GH register two of it as different names. 

Remove ubuntu-latest prevents this bug.


Yuan325:
obsolete.

PR #399, opened 2025-04-04
feat: Add IAM authentication to AlloyDB Source

duwenxin99:
Add IAM support for AlloyDB source connection: https://pkg.go.dev/cloud.google.com/go/alloydbconn#section-readme

### DIFF
--- a/internal/sources/alloydbpg/alloydb_pg.go
+++ b/internal/sources/alloydbpg/alloydb_pg.go
@@ -149,14 +149,15 @@ func getIAMPrincipalEmailFromADC(ctx context.Context) (string, error) {
 func getConnectionConfig(ctx context.Context, user, pass, dbname string) (string, bool, error) {
 	useIAM := true
 
-	// username and password both provided, use password authentication
+	// If username and password both provided, use password authentication
 	if user != "" && pass != "" {
 		dsn := fmt.Sprintf("user=%s password=%s dbname=%s sslmode=disable", user, pass, dbname)
 		useIAM = false
 		return dsn, useIAM, nil
 	}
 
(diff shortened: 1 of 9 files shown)


internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -94,27 +98,96 @@ func getOpts(ipType, userAgent string) ([]alloydbconn.Option, error) {
...
+		return dsn, useIAM, nil
+	}
+	// use IAM authentication if use/pass are not both provided
+	if user == "" {
+		email, err := getIAMPrincipalEmailFromADC(ctx)
+		if err != nil {
+			return "", useIAM, fmt.Errorf("error getting email from ADC: %v", err)
+		}
+		user = email
+	}
```
kurtisvg: This says "use IAM authentication if use/pass are not both provided" (typo: use -> user)

But it's only checking user? 

Can we make it more explicit:
1. If they specify user AND password, use regular authN
2. If they don't specify user AND password, use IAM authN
3. If they specify one but not the other -- throw error?

kurtisvg (changes requested):

internal/sources/alloydbpg/alloydb_pg.go
duwenxin99: Added error for password specified without username. Current logic:
1. If both user AND password specified --> use regular authN
2. If user is specified without password --> use username for IAM authN
3. If password is specified without username --> throws error
4. If both are empty --> fetches email from ADC and use is for  IAM authN

kurtisvg (approved):

Issue #419, opened 2025-04-08
deploy to cloud-run responds with http not https URL in SSE response


jexp:
Version 0.3.0

I followed the instructions to deploy my MCP Toolbox to cloud run (which deploys latest anyhow)
https://googleapis.github.io/genai-toolbox/how-to/deploy_toolbox/

but it wrongly responds with the http instead of https URL of the cloud-run URL on the SSE endpoint

```
https://toolbox-xxx-uc.a.run.app/mcp/sse

event: endpoint
data: http://toolbox-xxx-uc.a.run.app/mcp?sessionId=15874a2a-2487-4cdc-926e-5fe0989da075
```

which then fails with an 405 error when the client is trying to call it
even when you try to call the https URL

From cloudrun logs

<img width="1314" alt="Image" src="https://github.com/user-attachments/assets/ab2942ef-f86f-405d-9b85-3128a5d29074" />

```


{
insertId: "67f5acf80003defd38d3e05d"
labels: {1}
logName: "projects/PROJECT/logs/run.googleapis.com%2Fstdout"
payload: "textPayload"
receiveLocation: "us-central1"
receiveTimestamp: "2025-04-08T23:10:48.266441782Z"
resource: {2}
severity: "DEFAULT"
textPayload: "2025-04-08T23:10:48.254806668Z INFO Response: 405 Client Error service: "httplog" httpRequest: {url: "http://toolbox-xxx-uc.a.run.app/mcp?sessionId=e1acd118-eec0-46a3-9371-01e9ee857526" method: "GET" path: "/mcp" remoteIP: "xxxxxxx:56224" proto: "HTTP/1.1" requestID: "localhost/xrtwL8um0d-000004" scheme: "http" header: {x-forwarded-proto: "https" forwarded: "for=\"xxxxx\";proto=https" x-forwarded-for: "xxxxxxx" request-id: "|85bd8dfe96569651061683e7660ceb24.eb6cf93fa413353e." traceparent: "00-85bd8dfe96569651061683e7660ceb24-21ec919a3088ee31-00" x-cloud-trace-context: "85bd8dfe96569651061683e7660ceb24/2444488789180608049"}} httpResponse: {status: 405 bytes: 0 elapsed: 0.013929} "
timestamp: "2025-04-08T23:10:48.253693Z"
traceSampled: false
}
```

the api/toolset endpoint returns properly the deployed tools.yaml configuration 

```json
[...]

PR #420, opened 2025-04-09
fix: #419 TLS https URL for SSE endpoint

jexp:
Fixes that in SSL deployments, e.g. on k8s or cloud run the SSE endpoint was statically returning `http:` 

not a scheme based on request TLS attribute.

Fixes https://github.com/googleapis/genai-toolbox/issues/419

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -126,8 +126,13 @@ func sseHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	s.sseManager.add(sessionId, session)
 	defer s.sseManager.remove(sessionId)
 
+	// https scheme formatting if request is a TLS request
+	ssl := ""
+	if r.TLS != nil {
+		ssl = "s"
+	} 
 	// send initial endpoint event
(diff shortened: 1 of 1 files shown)


jexp:
Hmm, didn't seem to help after rebuild of this in a new image and redeploy it still seems to return http.
Not sure if I did something wrong in the code or in the deployment :(

jexp:
One of my colleagues said

Hey :slightly_smiling_face: by not working do you mean that the if `r.TLS != nil `code is always false and therefore not appending the 's' to the protocol (always http?).

If so I've seen something like this before in the past in environments with TLS termination - `r.TLS` isn’t reliable when using a reverse proxy (in my case nginx) which might terminate TLS, so the generated URL ended up with the wrong scheme. I ended up inspecting headers like `X-Forwarded-Proto` (or whatever the proxy sets) to correctly determine if the URL should use "http" or "https."

jexp:
Works now

Screenshot 2025-04-09 at 15 11 23

Yuan325 (approved):
Thank you for contributing! :)

PR #421, opened 2025-04-09
docs: make quickstart async-first

twishabansal:


docs/en/getting-started/colab_quickstart.ipynb
```diff
@@ -689,23 +689,26 @@
...
-        "    print(response[\"messages\"][-1].content)"
+        "  # Create a Langraph agent\n",
+        "  agent = create_react_agent(model, tools, checkpointer=MemorySaver())\n",
+        "  config = {\"configurable\": {\"thread_id\": \"thread-1\"}}\n",
+        "  for query in queries:\n",
+        "      inputs = {\"messages\": [(\"user\", prompt + query)]}\n",
+        "      response = agent.invoke(inputs, stream_mode=\"values\", config=config)\n",
+        "      print(response[\"messages\"][-1].content)\n",
+        "\n",
+        "await run_application()"
```
anubhav756, line 716: Should this be something like `asyncio.run(run_application())`

anubhav756 (changes requested):

docs/en/getting-started/colab_quickstart.ipynb
```diff
@@ -815,7 +818,7 @@
         "        print(f\"---- {query} ----\")\n",
         "        print(str(response))\n",
         "\n",
-        "await run_agent()"
+        "await run_application()"
```
anubhav756, line 826: Same as the comment above.

docs/en/getting-started/colab_quickstart.ipynb
twishabansal: In colab, these cells run directly without asyncio.run()

docs/en/getting-started/colab_quickstart.ipynb
twishabansal: Same as above.

docs/en/getting-started/colab_quickstart.ipynb
anubhav756: Makes sense. Optional: Should we write the code directly instead of within run_application?

anubhav756 (approved):
Just an optional thought, but LGTM otherwise.

PR #422, opened 2025-04-09
chore: add toolbox-core to readme

twishabansal:


README.md
```diff
@@ -136,6 +136,31 @@ Once your server is up and running, you can load the tools into your
...
+    ```
+1. Load tools:
+    ```python
+    from toolbox_core import ToolboxClient
+
+    # update the url to point to your server
+    client = ToolboxClient("http://127.0.0.1:5000")
+
+    # these tools can be passed to your application! 
+    tools = await client.load_toolset("toolset_name")
```
anubhav756, line 153: nit: Can we load the default toolset to make it consistent with the code samples of the other frameworks?

anubhav756 (changes requested):
nit change

README.md
twishabansal: Our current release does not allow for loading a default toolset. Will fix this upon the next release.

anubhav756 (approved):

PR #425, opened 2025-04-09
feat: add toolset feature to mcp

### LINKED ISSUE #403: Support Toolsets in MCP
eg via a path in the URL?

`localhost:5000/mcp/<toolset>/sse`

Otherwise all tools will be exposed at once and your carefully curated toolsets are in vain

Yuan325:
Update MCP server to support toolset.

User can now connect to specific toolset's sse via `/mcp/{toolset_name}/sse` url, or POST to `/mcp/{toolset_name}`. If toolset_name is not provided, it will list all tools by default.

Fixes #403


jexp (approved):
LGTM

internal/server/mcp.go
```diff
@@ -128,15 +137,20 @@ func sseHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
 		}
 	}
+
 	// send initial endpoint event
-	messageEndpoint := fmt.Sprintf("%s://%s/mcp?sessionId=%s", proto, r.Host, sessionId)
+	toolsetURL := ""
+	if toolsetName != "" {
+		toolsetURL = fmt.Sprintf("/%s", toolsetName)
+	}
+	messageEndpoint := fmt.Sprintf("%s://%s/mcp%s?sessionId=%s", proto, r.Host, toolsetURL, sessionId)
```
kurtisvg, line 153: Does this need to include toolset parameter?

internal/server/mcp.go
```diff
@@ -85,7 +90,10 @@ func sseHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	r = r.WithContext(ctx)
 
 	sessionId := uuid.New().String()
+	toolsetName := chi.URLParam(r, "toolsetName")
```
kurtisvg, line 93: If toolset name is missing, is this `""`?

kurtisvg (changes requested):

internal/server/mcp.go
Yuan325: Yes to have `tools/list` to work accurately.

internal/server/mcp.go
Yuan325: Yes

kurtisvg (approved):

PR #426, opened 2025-04-09
test(server/mcp): add tls unit test

Yuan325:
Refactor common test so that we can run both non-tls test server and tls test server.
Add unit test for tls server in MCP sse endpoint test cases.


jexp:
I think we need both a test for regular TLS and SSL Termination at an earlier LB that sets `X-Forwarded-Proto`

internal/server/mcp_test.go
```diff
@@ -267,34 +269,60 @@ func TestMcpEndpoint(t *testing.T) {
...
 			name:  "toolset1",
+			tls:   false,
 			url:   "/tool1_only/sse",
 			event: fmt.Sprintf("event: endpoint\ndata: %s/mcp/tool1_only?sessionId=", ts.URL),
 		},
+		{
+			name:  "basic with tls",
+			tls:   true,
+			url:   "/sse",
+			event: fmt.Sprintf("event: endpoint\ndata: %s/mcp?sessionId=", tls.URL),
```
kurtisvg: I think in order to properly regression test we need to have a test that checks the header leads to http or https.

kurtisvg (changes requested):

internal/server/mcp_test.go
Yuan325: Updated.

kurtisvg (approved):

PR #433, opened 2025-04-10
feat: Add `AuthRequired` to Tool Manifest

duwenxin99:
Add `AuthRequired` to Tool Manifest so SDK could throw an error early for unauthorized Tool invocations.
SDK changes: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/72/files

Also added `authRequired` to Neo4j and dgraph tools.


cmd/root_test.go
```diff
@@ -396,6 +397,8 @@ func TestParseToolFileWithAuth(t *testing.T) {
 					description: some description
 					statement: |
 						SELECT * FROM SQL_STATEMENT;
+					authRequired:
+						- my-google-service
```
Yuan325: Can we add a new test case with auth required instead? so we keep the basic test case "basic".

Yuan325 (changes requested):
everything looks good. Left a comment for a quick change. Thank you!

cmd/root_test.go
duwenxin99: added a new test case

Yuan325 (approved):

PR #437, opened 2025-04-14
feat: add support for optional parameters

takryo00:
- Add an optional flag to parameters and modify logic to exclude them from the required list when necessary.
- Add test cases for new optional parameters to verify correct functionality.
- Add a new version/overload to the existing parameter creation function to support specifying optionality.

### DIFF
--- a/internal/server/common_test.go
+++ b/internal/server/common_test.go
@@ -71,7 +71,9 @@ func (t MockTool) McpManifest() tools.McpManifest {
 	for _, p := range t.Params {
 		name := p.GetName()
 		properties[name] = p.McpManifest()
-		required = append(required, name)
+		if !p.IsOptional() {
+			required = append(required, name)
+		}
 	}
 
 	toolsSchema := tools.McpToolsSchema{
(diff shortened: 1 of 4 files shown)


kurtisvg:
Thanks for opening a PR @takryo00! In the future, it might be a good idea to start with an issue so we can try to reach agreement on a solution before we write the code 😄 

We've done some discussion on implementing optional parameters amongst the toolbox team, and had two questions around potential APIs come up that we didn't really have answers for:
1. We've tried to stay close to the OpenAPI schema (although sometimes simplifying or using a subset to simplify implementation). They use `required` instead of `optional` for parameters. Are there any advantages of `required` vs `optional` when specifying an optional parameter?
2. If a parameter is optional, what should we do if the value isn't provided? Should we allow for specifying a default? If so, do we still need a field to specify it as optional?  Should we support `null` as a default option?

kurtisvg:
Hi @takryo00 -- I opened an issue here to gather feedback on some options. If you have any thoughts feel free to comment there: https://github.com/googleapis/genai-toolbox/issues/475

kurtisvg:
Hey @takryo00 -- I think the API proposed in #475 is where we are leaning. Let me know if you are interested in updating your PR or not. Thanks!

Yuan325:
Optional parameters will be added in PR #617. Thank you!

Yuan325:
Updates: Optional parameters are now merged and will be included in the next official release. Thank you again :)

PR #439, opened 2025-04-14
docs: add ADK to local quickstart

fhinkel:
Add ADK to the local quickstart.


fhinkel:
I might have a typo in the docs (not closing a tab or similar) - but this error seems like a problem with the build? `error building site: html/template:_markup/render-heading.html:1:12: no such template "_default/_markup/td-render-heading.html"`

twishabansal:
Hi @fhinkel . Thanks for opening a PR!

I've reviewed the added quickstart and see that while it initializes an agent, it doesn't yet demonstrate the agent using its tools to perform tasks (like booking a hotel, as shown in other examples). Could you please update it to show a complete workflow where the agent processes user queries and calls the appropriate tools?

fhinkel:
Thanks for the feedback @twishabansal. The "ADK way" is to call `adk run`, and then use the interactive session to run the queries, rather than hard code the queries. Do you prefer if I hard code them?

twishabansal:
> Thanks for the feedback @twishabansal. The "ADK way" is to call `adk run`, and then use the interactive session to run the queries, rather than hard code the queries. Do you prefer if I hard code them?

The goal is an end-to-end quickstart needing minimal user intervention. We would like to hard code the queries, and see the results running on the user's system.

Please feel free to add a comment or link to the preferred ADK way.

fhinkel:
Updated, please take another look. Thank you

twishabansal (approved):

twishabansal:
Thank you for the contribution!

PR #444, opened 2025-04-14
chore(ci/cd): pin version of hugo

kurtisvg:
Hugo's recent release seems to have broken docsy: https://github.com/gohugoio/hugo/issues/13599
There's a fix in progress: https://github.com/google/docsy/pull/2215

To unblock CI/CD, I've pinned hugo to 0.145.0 instead.


fhinkel:
Might be worth adding nightly tests that build the docs rather than only build them on docs changes?

fhinkel (approved):

twishabansal (approved):

Issue #431, opened 2025-04-10
External/env parameters


jexp:
Sometimes you need values in statements as parameters that are not part of the tool definition. Think urls or paths or sometimes access keys

It would be great to be able to define them externally either via env variables or declared parameters in the tool or source definition (where the value could again come from an env variable)

duwenxin99:
Fixed by: https://github.com/googleapis/genai-toolbox/pull/462

duwenxin99:
Users are now able to specify environment variables in tools.yaml with the format `${ENV_NAME}`. Thanks for the suggestion!

PR #462, opened 2025-04-21
feat: Support env replacement for tool.yaml

duwenxin99:
Environment variable replacement is needed so that users don't have to hardcode their secrets in configuring `tools.yaml`.
Both formats `$ENV_NAME` and `${ENV_NAME}` are standard ways to declare an environment variable. 
However, some database statement placeholders that are already using the `$ENV_NAME` format. 
Therefore, we only support env var declaration using `${ENV_NAME}` to disambiguate it from other usages.

Fixes issue: https://github.com/googleapis/genai-toolbox/issues/431

### DIFF
--- a/docs/en/getting-started/local_quickstart.md
+++ b/docs/en/getting-started/local_quickstart.md
@@ -131,8 +131,8 @@ In this section, we will download Toolbox, configure our tools in a
    previous step.
 
     {{< notice tip >}}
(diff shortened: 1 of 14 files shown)


kurtisvg (approved):
LGTM, but we need to add some documentation for this. Maybe in configuration for tools.yaml?

PR #479, opened 2025-04-23
ci: add cmd package into unit test

Yuan325:


Yuan325:
Double checked and confirmed cmd was ran. My mistake, closing due to irrelevance. :)

PR #494, opened 2025-04-26
test: add integration tests for postgres-execute-sql tool

Yuan325:
Add `postgres-execute-sql` to compatible source's integration tests.


tests/tool.go
```diff
@@ -301,6 +301,104 @@ func RunToolInvokeTest(t *testing.T, select_1_want, invoke_param_want string) {
...
+		requestBody   io.Reader
+		want          string
+		isErr         bool
+	}{
+		{
+			name:          "invoke my-exec-sql-tool",
+			api:           "http://127.0.0.1:5000/api/tool/my-exec-sql-tool/invoke",
+			requestHeader: map[string]string{},
+			requestBody:   bytes.NewBuffer([]byte(`{"sql":"SELECT 1"}`)),
+			want:          select_1_want,
```
averikitsch, line 325: I am a little hesitant only using Select 1 as our test example. We may want to have a test for ddl queries

averikitsch (approved):
Thank you for adding the text tests!

PR #498, opened 2025-04-28
test: cleanup integration test

Yuan325:
Remove cleaning `\\` or `\` in `got` and `want` strings of test cases. This is to prevent future the need to manually clean strings before comparing. This also allows us to be aware of the amount of backspace to prevent accidental escaping in src code.

Update integration test that is affected. This includes moving the `want` of test cases for `my-param-tool` invocation to each sources since spanner have a different format (return id as strings instead of int).

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -22,8 +22,6 @@ import (
 	"fmt"
 
 	"github.com/googleapis/genai-toolbox/internal/tools"
-
-	bigqueryapi "cloud.google.com/go/bigquery"
 )
 
 // GetToolsConfig returns a mock tools config file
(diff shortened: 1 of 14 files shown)


tests/common.go
```diff
@@ -302,3 +302,30 @@ func GetBigQueryAuthToolInfo(projectID, datasetID, tableName string) (string, st
 	}
 	return createStatement, insertStatement, toolStatement, params
 }
+
+func GetNonSpannerInvokeParamWant() (string, string) {
```
duwenxin99, line 251: Would it be better for us to move the test statement & "want" strings into each source's own file? It feels that right now there are too many abstractions that each source's test file is hardly readable-you have to click back and forth between "common.go" and the test files to understand what each test statement & "want" parameter is set to. For external developers this is extra confusing.

Yuan325: I only moved the ones that have duplications to common (this one specifically is applicable for all databases except for Spanner (x11 databases); postgres (x3 databases); mysql (x2 databases); and mssql (x2 databases)). Others that are only applicable to it's own database are all in the source itself (e.g. bigquery, bigtable, spanner...). It was also for future when there's any update, we will have to make the updates across all databases.

duwenxin99 (approved):

PR #504, opened 2025-04-29
docs: update MCP Toolbox description

AndrewBrook-Google:
Minor edit to description of Toolbox to avoid some sensitivity around production/enterprise suitability as that may imply compliance features or guarantees which aren't suitable to an early stage product.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@
 > MCP Toolbox for Databases is currently in beta, and may see breaking 
 > changes until the first stable release (v1.0).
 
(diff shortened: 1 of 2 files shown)


duwenxin99:
We need to change it in the doc page as well: https://github.com/googleapis/genai-toolbox/blob/a6c17c96f3e7703fa31d4e9727082502057c1517/docs/en/getting-started/introduction/_index.md?plain=1#L9

Yuan325 (approved):

PR #511, opened 2025-04-30
style: add super-linter for markdown lint

Yuan325:
Add super-linter to check markdown lint.


.github/sync-repo-settings.yaml
```diff
@@ -27,6 +27,7 @@ branchProtectionRules:
     requiredStatusCheckContexts:
       - "cla/google"
       - "lint"
+      - "super-linter"
```
kurtisvg, line 30: Can we merge this into the lint workflow instead of a separate one?

kurtisvg (changes requested):
whats the difference between this and #530? It looks like this adds the linter and some linting, and the other one does more linting?

How did we land on `super-lint` as the choice here?

Yuan325:
> whats the difference between this and #530? It looks like this adds the linter and some linting, and the other one does more linting?
> 
> How did we land on `super-lint` as the choice here?

super-lint seeems to be popular for linting. and it covers multiple lints. Eventually I'm planning to use super-lint to run our go linting as well (it will default to using `golangcilint` once enabled).

This is markdownlint (does alot of max line checks etc.) and the other in #530 is Prettier (that does table formatting, indentation consistency etc.).

Yuan325:
Closing this due to the complication added. Revert to manually lint docs during every release.

PR #516, opened 2025-05-01
ci: Enable label-controlled docs preview

duwenxin99:
Allow external contributors to preview doc changes


.github/workflows/docs_preview_deploy.yaml
```diff
@@ -17,6 +17,7 @@ name: "docs"
 permissions:
   contents: write
   pull-requests: write
+  
 
 # This Workflow depends on 'github.event.number',
```
Yuan325:
```suggestion
  pull-requests: write

# This Workflow depends on 'github.event.number',
```

nit

Yuan325 (approved):

.github/workflows/docs_preview_deploy.yaml
duwenxin99: updated

.github/workflows/docs_preview_deploy.yaml
```diff
@@ -17,7 +17,7 @@ name: "docs"
 permissions:
   contents: write
```
averikitsch, line 18: Do we need `contents: write`? We may need this in combo with the pull-request write to add the comment to the PR, but it may be possible to reduce this to read only

.github/labels.yaml
```diff
@@ -69,6 +69,10 @@
   color: 3DED97
   description: Label to trigger Github Action tests.
 
+- name: 'docs: deploy preview'
```
averikitsch: super nit: it may be more common to not use spaces when using a label to trigger an action ie `docs:deploy-preview`. Other previous art is `kokoro:force-run`, `kokoro:run`, `release-please:force-run`

.github/workflows/docs_preview_deploy.yaml
```diff
@@ -27,9 +27,19 @@ on:
       - 'docs/**'
       - 'github/workflows/docs**'
       - '.hugo'
+  pull_request_target:
```
averikitsch, line 30: I'm not convinced this is secure enough as all other repos have removed this functionality. Would this solution work https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks?

.github/workflows/docs_preview_deploy.yaml
duwenxin99: Edit: we will need the `contents: write` permission as stated in the action doc: https://github.com/peaceiris/actions-gh-pages/tree/4f9cc6602d3f66b9c108549d475ec49e8ef4d45e/?tab=readme-ov-file#%EF%B8%8F-first-deployment-with-github_token

.github/workflows/docs_preview_deploy.yaml
duwenxin99: The problem here is that the Github Action that the `GITHUB_TOKEN` we are using does not have access to external repos: https://github.com/peaceiris/actions-gh-pages/tree/4f9cc6602d3f66b9c108549d475ec49e8ef4d45e/?tab=readme-ov-file#%EF%B8%8F-deploy-to-external-repository-external_repository So even if we approve the workflow, it cannot run under the `pull_request` trigger, which runs it on forked repo.

averikitsch: @kurtisvg I have connected with the DevRel team around [recommendations](https://docs.google.com/document/d/1FJTNV2D65mhzgzbMFd22irPeEfk0aAVBatt8MBtUTGA/edit?resourcekey=0-kUgZVjOoBSmQsRQ4MW8ifA&tab=t.dcpc75h3bxsj#heading=h.vs0arb7fw4aa). They have recommended the use of a multi-workflow setup to "start a pull_request job triggers a workflow_run job that runs the tests [(for us the doc preview)] based on the specific commit SHA from the pull_request job that was approved via the button. That way we make sure we run specifically the code that was approved, even if a new commit is immediately pushed (that could be a potential vulnerability with the label gate approach depending on how it's implemented, it's an existing vulnerability with Kokoro for example)."
However, we can justify the use of the pull_request_target since our token does not have access to Google Cloud resources, we do not have automatic label removal steps, we do not cache anything, and using the label gate matches the risk profile of using Cloud Build.

@duwenxin99 will add guidance to the contributing guide to call out the risks and guide maintainers to review the changes before adding the label to trigger the preview.

averikitsch (approved):

averikitsch:
I've added the "Do not merge" label to wait for Kurtis' review

.github/workflows/docs_preview_deploy.yaml
```diff
@@ -26,10 +26,20 @@ on:
     paths:
       - 'docs/**'
       - 'github/workflows/docs**'
-      - '.hugo/**'
+      - '.hugo'
```
kurtisvg: nit: these line needs to be reverted please

.github/workflows/docs_preview_deploy.yaml
```diff
@@ -26,10 +26,20 @@ on:
...
+  pull_request_target:
+    types: [labeled]
+    paths:
+      - 'docs/**'
+      - 'github/workflows/docs**'
+      - '.hugo'
```
kurtisvg: this line as well

kurtisvg (approved):

.github/workflows/docs_preview_deploy.yaml
duwenxin99: added

PR #519, opened 2025-05-01
feat: Add Memorystore for Valkey and Redis Sources and Tool

duwenxin99:


tests/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/googleapis/genai-toolbox/tests"
+	"github.com/redis/go-redis/v9"
+)
+
+var (
+	REDIS_SOURCE_KIND     = "memorystore-redis"
+	REDIS_TOOL_KIND       = "redis"
+	REDIS_ADDRESS         = os.Getenv("MEMORYSTORE_REDIS_ADDRESS")
+	REDIS_PASS            = os.Getenv("MEMORYSTORE_REDIS_PASS")
+	SERVICE_ACCOUNT_EMAIL = os.Getenv("SERVICE_ACCOUNT_EMAIL")
```
Yuan325: I believe we won't need this. it is in `auth.go`.

tests/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,131 @@
...
+	tests.RunMCPToolCallMethod(t, mcpInvokeParamWant, failInvocationWant)
+}
+
+func setupRedisDB(t *testing.T, ctx context.Context, client *redis.Client) func(*testing.T) {
+	keys := []string{"row1", "row2", "row3"}
+	commands := [][]any{
+		{"HSET", keys[0], "name", "Alice", "id", "1"},
+		{"HSET", keys[1], "name", "Jane", "id", "2"},
+		{"HSET", keys[2], "name", "Sid", "id", "3"},
+		{"HSET", SERVICE_ACCOUNT_EMAIL, "name", `{"name":"Alice"}`},
```
Yuan325:
```suggestion
		{"HSET", tests.SERVICE_ACCOUNT_EMAIL, "name", `{"name":"Alice"}`},
```

tests/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,131 @@
...
+var (
+	REDIS_SOURCE_KIND     = "memorystore-redis"
+	REDIS_TOOL_KIND       = "redis"
+	REDIS_ADDRESS         = os.Getenv("MEMORYSTORE_REDIS_ADDRESS")
+	REDIS_PASS            = os.Getenv("MEMORYSTORE_REDIS_PASS")
+	SERVICE_ACCOUNT_EMAIL = os.Getenv("SERVICE_ACCOUNT_EMAIL")
+)
+
+func getRedisVars(t *testing.T) map[string]any {
+	switch "" {
```
Yuan325: we'll want to add check for REDIS_PASS as well.

tests/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,131 @@
...
+	case REDIS_ADDRESS:
+		t.Fatal("'REDIS_ADDRESS' not set")
+	}
+	return map[string]any{
+		"kind":           REDIS_SOURCE_KIND,
+		"address":        REDIS_ADDRESS,
+		"password":       REDIS_PASS,
+		"clusterEnabled": false,
```
Yuan325:
```suggestion
```
if this is a default value, we can just remove this.

tests/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func initMemorystoreRedisClient(ctx context.Context, address, pass string) (*redis.Client, error) {
```
Yuan325: Do we want to add integration test for clustered client as well?

tests/common.go
```diff
@@ -298,3 +301,185 @@ func GetMysqlWants() (string, string) {
...
+	select1Want := "[\"PONG\"]"
+	failInvocationWant := `{"jsonrpc":"2.0","id":"invoke-fail-tool","result":{"content":[{"type":"text","text":"\"error from executing command at index 0: ERR unknown command 'SELEC 1;', with args beginning with: \""}]}}`
+	invokeParamWant := "[\"Alice\",\"Sid\"]"
+	invokeAuthWant := `["{\"name\":\"Alice\"}"]`
+	mcpInvokeParamWant := `{"jsonrpc":"2.0","id":"my-param-tool","result":{"content":[{"type":"text","text":"\"Alice\""},{"type":"text","text":"\"Sid\""}]}}`
+	return select1Want, failInvocationWant, invokeParamWant, invokeAuthWant, mcpInvokeParamWant
+}
+
+// GetPostgresSQLParamToolInfo returns statements and param for my-param-tool postgres-sql kind
+func GetRedisValkeyToolCmds() ([][]string, [][]string) {
```
Yuan325: Can we just add it in the `GetRedisValkeyToolsConfig()` func below? Seems like that's the only place these values are used.

tests/tool.go
```diff
@@ -16,107 +16,16 @@ package tests
...
 	"net/http"
 	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/googleapis/genai-toolbox/internal/server/mcp"
-	"github.com/jackc/pgx/v5/pgxpool"
 )
 
-// SetupPostgresSQLTable creates and inserts data into a table of tool
```
Yuan325: Can we move this in a separate PR to keep things separate?

tests/common.go
```diff
@@ -298,3 +301,185 @@ func GetMysqlWants() (string, string) {
...
+			t.Errorf("Teardown failed: %s", err)
+		}
+	}
+}
+
+// GetRedisWants return the expected wants for redis
+func GetRedisValkeyWants() (string, string, string, string, string) {
+	select1Want := "[\"PONG\"]"
+	failInvocationWant := `{"jsonrpc":"2.0","id":"invoke-fail-tool","result":{"content":[{"type":"text","text":"\"error from executing command at index 0: ERR unknown command 'SELEC 1;', with args beginning with: \""}]}}`
+	invokeParamWant := "[\"Alice\",\"Sid\"]"
```
Yuan325: Is it possible to query so that it returns the `id` field as well? That way it is consistent with the other implementations.

tests/common.go
```diff
@@ -298,3 +301,185 @@ func GetMysqlWants() (string, string) {
...
+	invokeAuthWant := `["{\"name\":\"Alice\"}"]`
```
Yuan325: Seems like the result is similar with the other tools "`[{"name":"Alice"}]`". Redis tool seem to have an additional marshal?

This is equivalent to  `"[\\\"hello\\\":\\\"world\\\"]"`. Might wanna resolve the additional marshaling.

Once this and the `id` field is resolve, i believe we can use the `GetNonSpannerInvokeParamWant()` function instead of having a different ones.

tests/common.go
```diff
@@ -272,10 +274,11 @@ func GetMysqlLAuthToolInfo(tableName string) (string, string, string, []any) {
 	return create_statement, insert_statement, tool_statement, params
 }
 
-func GetNonSpannerInvokeParamWant() (string, string) {
+func GetNonSpannerInvokeParamWant() (string, string, string) {
 	invokeParamWant := "[{\"id\":1,\"name\":\"Alice\"},{\"id\":3,\"name\":\"Sid\"}]"
 	mcpInvokeParamWant := `{"jsonrpc":"2.0","id":"my-param-tool","result":{"content":[{"type":"text","text":"{\"id\":1,\"name\":\"Alice\"}"},{"type":"text","text":"{\"id\":3,\"name\":\"Sid\"}"}]}}`
-	return invokeParamWant, mcpInvokeParamWant
+	invokeAuthWant := `[{"name":"Alice"}]`
```
Yuan325: This is not necessary once we resolve the extra marshaling in Redis. It should be the same.

internal/sources/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,153 @@
...
+			desc: "advanced example",
+			in: `
+			sources:
+				my-redis-instance:
+					kind: memorystore-redis
+					address: 127.0.0.1
+					password: my-pass
+					database: 1
+					useIAM: false
+					clusterEnabled: false
```
Yuan325: If we want to test the non default values, we can have `clusterEnabled: true` instead. Same with useIAM. We can test with `useIAM: true` to make sure that the field is parsed correctly.

internal/sources/memorystoreredis/memorystore_redis_test.go
```diff
@@ -0,0 +1,153 @@
...
+		want server.SourceConfigs
+	}{
+		{
+			desc: "default setting",
+			in: `
+			sources:
+				my-redis-instance:
+					kind: memorystore-redis
+					address: 127.0.0.1
+					clusterEnabled: true
```
Yuan325: default setting for `clusterEnabled` is false? In that case we can remove this row.

internal/sources/util.go
```diff
@@ -85,3 +85,20 @@ func GetIAMPrincipalEmailFromADC(ctx context.Context) (string, error) {
 	email := strings.TrimSuffix(emailValue.(string), ".gserviceaccount.com")
 	return email, nil
 }
+
+func GetIAMAccessToken(ctx context.Context) (string, error) {
+	creds, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		return "", fmt.Errorf("failed to find default credentials (run 'gcloud auth application-default login'?): %w", err)
```
Yuan325, line 92: Is this the only way to get IAM? Will this still work if user didn't run it locally? (e.g. Kubernetes/container/cloud run)

internal/sources/memorystoreredis/memorystore_redis.go
```diff
@@ -0,0 +1,134 @@
...
+			if err != nil {
+				return "", "", err
+			}
+			return "default", token, nil
+		}
+	}
+
+	var client RedisClient
+	var err error
+	if r.ClusterEnabled {
```
Yuan325: what is the difference between clusterclient and regular client?

internal/tools/redis/redis.go
```diff
@@ -0,0 +1,179 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error getting result: %s", err)
+		}
+		out[i] = result
+	}
+
+	return out, nil
+}
+
+// Helper function to replace parameters in the commands
```
Yuan325:
```suggestion
// replaceCommandsParams is a helper function to replace parameters in the commands
```

Yuan325 (dismissed):

tests/memorystoreredis/memorystore_redis_test.go
duwenxin99: The clustered instances cannot be connected from Cloud Build since it uses PSC instead of PSA. I have tested connection manually.

internal/sources/util.go
duwenxin99: Yes it will work on the VMs as long as the ADC is set correctly.

internal/sources/memorystoreredis/memorystore_redis.go
duwenxin99: a cluster client directs transactions to multiple nodes, while a regular client only directs to one node.

kurtisvg (approved):

PR #523, opened 2025-05-03
refactor: move existing mcp version to new package

Yuan325:


internal/server/mcp.go
```diff
@@ -177,6 +177,8 @@ func mcpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	ctx, span := s.instrumentation.Tracer.Start(r.Context(), "toolbox/server/mcp")
 	r = r.WithContext(ctx)
 
+	mcpProtocolVersion := "2024-11-05"
```
kurtisvg: Should this be changed here?

internal/server/mcp.go
```diff
@@ -210,126 +212,98 @@ func mcpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
 	}
 
 	// Generic baseMessage could either be a JSONRPCNotification or JSONRPCRequest
 	var baseMessage struct {
-		Jsonrpc string        `json:"jsonrpc"`
-		Method  string        `json:"method"`
-		Id      mcp.RequestId `json:"id,omitempty"`
+		Jsonrpc string `json:"jsonrpc"`
+		Method  string `json:"method"`
+		Id      any    `json:"id,omitempty"`
```
kurtisvg: hmm why should this be any?

kurtisvg (changes requested):
Quick high level feedback: I don't think your separation of the spec is too low in the implementation. The spec include endpoints like tools/call, which means you probably need to have everything sans the version negotiation happening inside the MCP version for the module.

internal/server/mcp.go
```diff
@@ -348,138 +349,54 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+
 	// Generic baseMessage could either be a JSONRPCNotification or JSONRPCRequest
 	var baseMessage struct {
-		Jsonrpc string        `json:"jsonrpc"`
-		Method  string        `json:"method"`
-		Id      mcp.RequestId `json:"id,omitempty"`
+		Jsonrpc string            `json:"jsonrpc"`
+		Method  string            `json:"method"`
+		Id      mcputil.RequestId `json:"id,omitempty"`
 	}
```
kurtisvg: nit: Any reason should shouldn't be `mcputil.baseMessage`?

internal/server/mcp/mcp.go
```diff
@@ -0,0 +1,94 @@
...
+	v20241105 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20241105"
+	"github.com/googleapis/genai-toolbox/internal/tools"
+)
+
+// LATEST_PROTOCOL_VERSION is the latest version of the MCP protocol supported.
+// Update the version used in InitializeResponse when this value is updated.
+const LATEST_PROTOCOL_VERSION = v20241105.PROTOCOL_VERSION
+
+// SUPPORTED_PROTOCOL_VERSION is the MCP protocol versions that are supported.
+var SUPPORTED_PROTOCOL_VERSION = []string{v20241105.PROTOCOL_VERSION}
```
kurtisvg:
```suggestion
var SUPPORTED_PROTOCOL_VERSIONS = []string{v20241105.PROTOCOL_VERSION}
```

internal/server/mcp.go
```diff
@@ -308,7 +309,7 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 		// Generate a new uuid if unable to decode
 		id := uuid.New().String()
 		s.logger.DebugContext(ctx, err.Error())
-		render.JSON(w, r, newJSONRPCError(id, mcp.PARSE_ERROR, err.Error(), nil))
+		render.JSON(w, r, mcputil.NewError(id, mcputil.PARSE_ERROR, err.Error(), nil))
```
kurtisvg: nit: is this `mcputil` or `jsonrpc`?

internal/server/mcp/util/jsonrpc.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// JSONRPCResponse represents a successful (non-error) response to a request.
+type JSONRPCResponse struct {
+	Jsonrpc string      `json:"jsonrpc"`
+	Id      RequestId   `json:"id"`
+	Result  interface{} `json:"result"`
+}
+
+// McpError represents the error content.
```
kurtisvg: nit: do you need to update this comment to match the struct?

kurtisvg (approved):

internal/server/mcp.go
Yuan325: updated to jsonrpc.baseMessage since this is specific towards jsonrpc request instead of the lifecycle.

PR #532, opened 2025-05-06
feat: Add Valkey Source and Tool

duwenxin99:


tests/common.go
```diff
@@ -392,7 +392,7 @@ func SetupMySQLTable(t *testing.T, ctx context.Context, pool *sql.DB, create_sta
 // GetRedisWants return the expected wants for redis
 func GetRedisValkeyWants() (string, string, string, string, string) {
 	select1Want := "[\"PONG\"]"
-	failInvocationWant := `{"jsonrpc":"2.0","id":"invoke-fail-tool","result":{"content":[{"type":"text","text":"\"error from executing command at index 0: ERR unknown command 'SELEC 1;', with args beginning with:  command length: 1\""}]}}`
+	failInvocationWant := `unknown command 'SELEC 1;'`
```
Yuan325: Ideally, we want to check that the response is correct rather than just partial response.

tests/memorystorevalkey/memorystore_valkey_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"time"
+
+	"github.com/googleapis/genai-toolbox/tests"
+	"github.com/valkey-io/valkey-go"
+)
+
+var (
+	MEMORYSTORE_VALKEY_SOURCE_KIND = "memorystore-valkey"
+	MEMORYSTORE_VALKEY_TOOL_KIND   = "valkey"
+	MEMORYSTORE_VALKEY_DATABASE    = 0
```
Yuan325: we shouldn't have to add the default value.

tests/memorystorevalkey/memorystore_valkey_test.go
```diff
@@ -0,0 +1,143 @@
...
+	switch "" {
+	case MEMORYSTORE_VALKEY_ADDRESS:
+		t.Fatal("'MEMORYSTORE_VALKEY_ADDRESS' not set")
+	}
+	return map[string]any{
+		"kind":         MEMORYSTORE_VALKEY_SOURCE_KIND,
+		"address":      MEMORYSTORE_VALKEY_ADDRESS,
+		"database":     MEMORYSTORE_VALKEY_DATABASE,
+		"useIAM":       false,
+		"disableCache": true,
```
Yuan325: shouldn't have to add the default value.

tests/tool.go
```diff
@@ -91,36 +91,36 @@ func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeAuthWan
...
-			isErr:         true,
-		},
-		{
-			name:          "Invoke my-tool with insufficient parameters",
-			api:           "http://127.0.0.1:5000/api/tool/my-tool/invoke",
-			requestHeader: map[string]string{},
-			requestBody:   bytes.NewBuffer([]byte(`{"id": 1}`)),
-			isErr:         true,
-		},
+		// {
```
Yuan325: might want to uncomment these~

internal/sources/memorystorevalkey/memorystore_valkey.go
```diff
@@ -0,0 +1,106 @@
...
+		SelectDB:          r.Database,
+		AuthCredentialsFn: authFn,
+		DisableCache:      r.DisableCache,
+	})
+
+	if err != nil {
+		log.Fatalf("error creating Valkey client: %v", err)
+	}
+
+	// Ping the server to check connectivity\
```
Yuan325:
```suggestion
	// Ping the server to check connectivity
```

internal/sources/memorystorevalkey/memorystore_valkey_test.go
```diff
@@ -0,0 +1,154 @@
...
+		want server.SourceConfigs
+	}{
+		{
+			desc: "default setting",
+			in: `
+			sources:
+				my-valkey-instance:
+					kind: memorystore-valkey
+					address:
+					  - 127.0.0.1
```
Yuan325:
```suggestion
					address: 127.0.0.1
```

internal/sources/memorystorevalkey/memorystore_valkey_test.go
```diff
@@ -0,0 +1,154 @@
...
+			`,
+			want: map[string]sources.SourceConfig{
+				"my-valkey-instance": memorystorevalkey.Config{
+					Name:    "my-valkey-instance",
+					Kind:    memorystorevalkey.SourceKind,
+					Address: "127.0.0.1",
+					UseIAM:  false,
```
Yuan325: might want to check the other default values as well by adding those fields within the `want`. To make sure that it is set correctly.

internal/sources/memorystorevalkey/memorystore_valkey_test.go
```diff
@@ -0,0 +1,154 @@
...
+			},
+		},
+		{
+			desc: "advanced example",
+			in: `
+			sources:
+				my-valkey-instance:
+					kind: memorystore-valkey
+					address:
+					  - 127.0.0.1
```
Yuan325:
```suggestion
					address: 127.0.0.1
```

internal/tools/redis/redis.go
```diff
@@ -115,7 +115,7 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
 	for i, resp := range responses {
 		if err := resp.Err(); err != nil {
 			// Add error from each command to `errSum`
-			errString := fmt.Sprintf("error from executing command at index %d: %s command length: %d", i, err, len(cmds[i]))
+			errString := fmt.Sprintf("error from executing command at index %d: %s", i, err)
```
Yuan325: Maybe we can make this change at the Redis PR instead?

internal/tools/valkey/valkey.go
```diff
@@ -0,0 +1,189 @@
...
+			continue
+		}
+		out[i] = resp
+	}
+
+	return out, nil
+}
+
+// Helper function to replace parameters in the commands
+func replaceCommandsParams(commands [][]string, params tools.Parameters, paramValues tools.ParamValues) ([][]string, error) {
```
Yuan325, line 158: This seems similar to the redis ones. Is it possible we can dedup this by adding a `util.go` in `/tools` and move this funciton in there?

Yuan325 (changes requested):

tests/common.go
duwenxin99: the valkey and radis failure has a very minor syntax difference, and that's why we have to check partial message

internal/sources/memorystorevalkey/memorystore_valkey.go
duwenxin99: updated

internal/sources/memorystorevalkey/memorystore_valkey_test.go
duwenxin99: updated

internal/tools/valkey/valkey.go
duwenxin99: redis requires []any but valkey commands require []string. It is possible to use the same function but that requires an extra loop to convert the type. It's probably better to keep them separate for now to reduce overhead.

duwenxin99: They require different slice type: []string vs []any. We will have to add an extra step/memory allocation to convert the type. It's probably better to separate them for now,

PR #552, opened 2025-05-09
Mongodb tool v1.2

theshanbhag:
Added MongoDB as source and tools to perform find and vectorsearch on MongoDB


tests/mongodb/mongodb_integration_test.go
```diff
@@ -0,0 +1,139 @@
...
+		{"name": bson.M{"$eq": "Alice"}},
+	}}
+	params := []interface{}{
+		bson.M{"_id": 1, "name": "Alice"},
+	}
+	return query, params
+}
+
+// runMongoDBQueryTest executes a query and validates the results
+func runMongoDBQueryTest(t *testing.T, ctx context.Context, client *mongo.Client, database, collectionName string, query bson.M, expected []interface{}) {
```
duwenxin99, line 122: You don't need to write your own integration test functions. We have 3 generic test suites:
1. `tests.RunToolGetTest()` to test Tool get endpoints
2. `tests.RunToolInvokeTest()` to test Tool invocation endpoints
3. `tests.RunMCPToolCallMethod()` to test MCP Tool invocation endpoints.
Could you make sure to call these 3 test helpers so that we have a good coverage of the MongoDB endpoints? Feel free to add any extra tests if you want to test for any other edge cases.

internal/tools/mongodb/mongodb.go
```diff
@@ -0,0 +1,227 @@
...
+			results = append(results, doc)
+		}
+
+		if err := cursor.Err(); err != nil {
+			return nil, fmt.Errorf("cursor error: %w", err)
+		}
+
+		return results, nil
+
+	case "vectorSearch":
```
duwenxin99, line 128: Should we give the users more freedom to query? For other databases, we allow users to execute arbitrary SQL queries so that they are able to perform all kinds of operations, not limited to read operations. 
In order to make MongoDB an equally flexible source/tool, we should:
1. expand beyond read operations. We should accept write operations such as `insert` as well.
2. Instead of creating a "vectorSearch" abstraction, would it be better to allow the users to use `collection.aggregate()` directly? In this way they have full freedom and control over their queries.

duwenxin99 (changes requested):

PR #558, opened 2025-05-12
refactor: update repo name to mcp-toolbox

Yuan325:


Yuan325:
Decided not to rename repo since this will break existing docs.

PR #563, opened 2025-05-14
feat: Add support for read-only in Spanner tool

### LINKED ISSUE #435: Support declarative read_only: true/false for tools
### Prerequisites

As someone defining tools I want to be able to declare if a tool is read-only or not.

The source implementation should then employ driver level safeguards to apply the correct access mode that cannot be escaped even with injection attacks.

### Suggested Solution(s)

Additional optional config per tool. Ideally we would default to readonly true, but that might break backwards compatibility.

`read_only: true/false`

or
`access_mode: read, write, schema, (could also have append/update)`

### Alternatives Considered

Using top-level read only connections or read-only user accounts, but then all tools would be read-only.

The user's permission would still apply if the statement exceeds their granted permissions.

### Additional Details

We could make it an optional feature that tool providers could opt in implementing. Would be documented per source / tool.

We could also consider adding the same flag on the source level.

trehanshakuntG:
Allowing user to add `readOnly` field in spanner tools.

The existing tool doesn't work for reading schema tables since schema tables can only be accessed through read-only transaction.

This PR also resolve #435 for Spanner tool.


Yuan325 (commented):
LGTM! Let's have @kurtisvg or @averikitsch make a second pass on this :) Thank you!

internal/tools/spanner/spanner.go
```diff
@@ -116,45 +118,68 @@ func getMapParams(params tools.ParamValues, dialect string) (map[string]interfac
 	}
 }
 
+func processRows(iter *spanner.RowIterator) ([]any, error) {
```
kurtisvg, line 122: Can we please add a comment on this function detailing it's purpose? From name alone I'm unable to tell what it's support to do.

internal/tools/spanner/spanner.go
```diff
@@ -116,45 +118,68 @@ func getMapParams(params tools.ParamValues, dialect string) (map[string]interfac
...
-			cols := row.ColumnNames()
-			for i, c := range cols {
-				vMap[c] = row.ColumnValue(i)
+			results, err = processRows(iter)
+			if err != nil {
+				return err
 			}
+			return nil
+		})
+	}
```
kurtisvg: It feels like there's room to reduce this if statement's scope. Any reason we can't just limit to just the transaction execution? 

Something like:

```
stmt := spanner.Statement{
  SQL:    t.Statement,
  Params: mapParams,
}
var results spanner.Results # todo: typing
var err error
if t.ReadOnly {
      // execute read transaction
	} else {
       // execute write transaction
}
// process rows here
```

internal/tools/spanner/spanner_test.go
```diff
@@ -64,6 +64,37 @@ func TestParseFromYamlSpanner(t *testing.T) {
 				},
 			},
 		},
+		{
```
kurtisvg, line 67: should we have an e2e test here as well?

internal/tools/spanner/spanner.go
```diff
@@ -116,45 +118,68 @@ func getMapParams(params tools.ParamValues, dialect string) (map[string]interfac
...
 		return nil, fmt.Errorf("fail to get map params: %w", err)
 	}
 
-	var out []any
+	var results []any
+	var opErr error
 
-	_, err = t.Client.ReadWriteTransaction(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
+	if t.ReadOnly {
+		ro := t.Client.ReadOnlyTransaction()
```
kurtisvg: It looks like we can do `iter := client.Single().Query(ctx, stmt)`?

https://pkg.go.dev/cloud.google.com/go/spanner#hdr-Statements

kurtisvg (changes requested):

internal/tools/spanner/spanner.go
trehanshakuntG: Done.

trehanshakuntG: Done.

internal/tools/spanner/spanner_test.go
trehanshakuntG: The integration test does test for readOnly as false, I will add some tests in the follow-up PRs.

internal/tools/spanner/spanner.go
trehanshakuntG: The process rows will have to be within the write transaction, but yes shifted out the stmt. Thanks for the suggestion.

internal/tools/spanner/spanner.go
trehanshakuntG: Done.

kurtisvg (approved):

internal/tools/spanner/spanner_test.go
kurtisvg: Best practice is to include relevant tests with the PR. I've approved for now to unblock, but let's try to include them up front next time.

PR #572, opened 2025-05-16
fix(server/mcp): add /messages endpoint for mcp

AlexTalreja:
Inspector shows `MCP error -32600: toolset does not exist`  

MCP Typescript SDK made this change to forcefully implement /messages in endpoint: https://github.com/modelcontextprotocol/typescript-sdk/commit/5a25fe3a6ca5f6a159c249a63fe2f358cda8a9c5

To prevent from maintaining a variations of endpoints, this PR also removed the existing `/` endpoint. Since `sse` will return with new endpoint that includes `/messages`, existing sse users WILL NOT be impacted with this update.

Related bug that are submitted to modelcontextprotocol: https://github.com/modelcontextprotocol/typescript-sdk/issues/501

### DIFF
--- a/docs/en/how-to/connect_via_mcp.md
+++ b/docs/en/how-to/connect_via_mcp.md
@@ -54,9 +54,9 @@ Add the following configuration to your MCP client configuration:
 
 If you would like to connect to a specific toolset, replace `url` with `"http://127.0.0.1:5000/mcp/{toolset_name}/sse"`.
 {{% /tab %}} {{% tab header="HTTP POST" lang="en" %}}
-Connect to Toolbox HTTP POST via `http://127.0.0.1:5000/mcp`.
+Connect to Toolbox HTTP POST via `http://127.0.0.1:5000/mcp/messages`.
 
-If you would like to connect to a specific toolset, connect via `http://127.0.0.1:5000/mcp/{toolset_name}`.
+If you would like to connect to a specific toolset, connect via `http://127.0.0.1:5000/mcp/{toolset_name}/messages`.
 {{% /tab %}} {{< /tabpane >}}
 
 ### Using the MCP Inspector with Toolbox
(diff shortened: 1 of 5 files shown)


docs/en/how-to/connect_via_mcp.md
```diff
@@ -54,9 +54,9 @@ Add the following configuration to your MCP client configuration:
 
 If you would like to connect to a specific toolset, replace `url` with `"http://127.0.0.1:5000/mcp/{toolset_name}/sse"`.
 {{% /tab %}} {{% tab header="HTTP POST" lang="en" %}}
-Connect to Toolbox HTTP POST via `http://127.0.0.1:5000/mcp`.
+Connect to Toolbox HTTP POST via `http://127.0.0.1:5000/mcp/messages`.
```
averikitsch, line 57: I'm not sure this is correct. Based on my experience and [docs](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) this may be simplified to `http://127.0.0.1:5000/mcp`

Yuan325: For this fix, we update the POST endpoint with `/messages` since MCP strip the **original** url by removing it's `/sse` and replacing it with `/messages` instead of using the endpoint that is actually returned to the client.

To reduce the need to maintain multiple endpoints, we'll remove the `/` POST endpoint which means user will need to connect via HTTP POST with url that has the `/messages` path.


But, it seems like MCP [rolled-back](https://github.com/modelcontextprotocol/typescript-sdk/pull/506) the PR on their end. So hopefully it should resolve soon (they will respond with updates in the [bug](https://github.com/modelcontextprotocol/typescript-sdk/issues/501) we raised). If that's the case (and once we verified that it worked without this PR's fix), we can close this PR without merging.

AlexTalreja:
Fixed in newer version of MCP Inspector.

PR #573, opened 2025-05-17
chore: implement mcp manager

Yuan325:
Implement mcp manager to manage the different session. This is needed to keep track of the protocol version of each sessions.

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -35,35 +35,44 @@ import (
 )
 
 type sseSession struct {
-	sessionId  string
 	writer     http.ResponseWriter
 	flusher    http.Flusher
 	done       chan struct{}
 	eventQueue chan string
 }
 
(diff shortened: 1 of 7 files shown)


internal/server/mcp.go
```diff
@@ -347,53 +387,53 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 }
 
 // processMcpMessage process the messages received from clients
-func processMcpMessage(ctx context.Context, body []byte, s *Server, toolsetName string) (any, error) {
+func processMcpMessage(ctx context.Context, protocolVersion string, body []byte, s *Server, toolsetName string) (string, any, error) {
```
kurtisvg, line 391: nit: Maybe update comment with the meaning of the returns here?

kurtisvg (approved):

PR #591, opened 2025-05-20
feat: add a new flag to indicate if stdio is supported

Yuan325:
Toolbox sends `debug` and `info` logs to `stdout`. However, in order to stupport stdio, Toolbox **should not** log messages to `stdout` as this will interfere with protocol operations. `stderr` are fine.

If `--stdio`, Toolbox will not send any `debug` and `info` logs, `warn` and `error` logs will still work as usual.


cmd/root.go
```diff
@@ -115,6 +115,7 @@ func NewCommand(opts ...Option) *Command {
 	flags.BoolVar(&cmd.cfg.TelemetryGCP, "telemetry-gcp", false, "Enable exporting directly to Google Cloud Monitoring.")
 	flags.StringVar(&cmd.cfg.TelemetryOTLP, "telemetry-otlp", "", "Enable exporting using OpenTelemetry Protocol (OTLP) to the specified endpoint (e.g. 'http://127.0.0.1:4318')")
 	flags.StringVar(&cmd.cfg.TelemetryServiceName, "telemetry-service-name", "toolbox", "Sets the value of the service.name resource attribute for telemetry data.")
+	flags.BoolVar(&cmd.cfg.Stdio, "stdio", false, "Support MCP stdio transport protocol.")
```
kurtisvg: Can we clarify that this "Listens via MCP STDIO instead of acting as a remote HTTP server".

kurtisvg (approved):

cmd/root.go
Yuan325: @kurtisvg  it will still work as http server, just that the logger will only log warn/error logs.

PR #593, opened 2025-05-21
feat: implement stdio transport protocol

Yuan325:


cmd/root.go
```diff
@@ -280,6 +284,12 @@ func run(cmd *Command) error {
 	// run server in background
 	srvErr := make(chan error)
 	go func() {
+		if cmd.cfg.Stdio {
+			err = s.ServeStdio(ctx, cmd.inStream, cmd.outStream)
+			if err != nil {
+				srvErr <- err
+			}
+		}
```
kurtisvg: 2 things:
1. the defer should stay at the top
2. We should update this to reflect the stdio should mean there isn't an http serve. My read on this if after stdio stops the http server might start?

kurtisvg (approved):

kurtisvg (approved):

kurtisvg (approved):

PR #604, opened 2025-05-22
feat: Adding support for the --prebuilt flag

dishaprakash:
Introduces a new --prebuilt <source_type> flag. This flag is mutually exclusive with the existing --tools-file flag.

Added a new directory cmd/prebuiltconfigs/ to store the prebuilt tools.yaml files (e.g., alloydb.yaml, postgres.yaml, etc.).
These YAML files are embedded into the Go binary using the //go:embed directive.


cmd/prebuiltconfigs/alloydb.yaml
```diff
@@ -0,0 +1,100 @@
+sources:
```
kurtisvg: nit: this package should probably be under /internal

cmd/root.go
```diff
@@ -40,10 +42,16 @@ var (
 	versionString string
 	// metadataString indicates additional build or distribution metadata.
 	metadataString string
+
+	//go:embed prebuiltconfigs/*.yaml
+	prebuiltConfigsFS embed.FS
+
+	prebuiltToolYAMLs map[string][]byte
```
kurtisvg: nit: I would expose this in the /internal/prebuiltconfigs module. That way you can add tests for specific modules. 

I would also expose it using a `.Get` function so you can prevent writes from outside the pacakge

cmd/root.go
```diff
@@ -222,14 +272,47 @@ func run(cmd *Command) error {
...
-	// Read tool file contents
-	buf, err := os.ReadFile(cmd.tools_file)
-	if err != nil {
-		errMsg := fmt.Errorf("unable to read tool file at %q: %w", cmd.tools_file, err)
+	var yamlFileContent []byte
+
+	if cmd.prebuiltSourceType != "" && cmd.tools_file != "" {
+		errMsg := fmt.Errorf("--prebuilt and --tools-file flags cannot be used simultaneously")
 		cmd.logger.ErrorContext(ctx, errMsg.Error())
 		return errMsg
```
kurtisvg: nit: should you move this check under the next isstatement instead?

cmd/root_test.go
```diff
@@ -858,3 +858,35 @@ func TestEnvVarReplacement(t *testing.T) {
 	}
 
 }
+
+func TestPrebuiltTools(t *testing.T) {
```
kurtisvg, line 900: Part of this test should be making sure that these parse correctly

cmd/root.go
```diff
@@ -222,14 +272,47 @@ func run(cmd *Command) error {
...
+			prebuiltHelpSuffix := "No prebuilt configurations found."
+			if len(availablePrebuiltKeys) > 0 {
+				prebuiltHelpSuffix = fmt.Sprintf("Available: %s", strings.Join(availablePrebuiltKeys, ", "))
+			}
+			errMsg := fmt.Errorf("prebuilt source type '%s' not found. %s", cmd.prebuiltSourceType, prebuiltHelpSuffix)
+			cmd.logger.ErrorContext(ctx, errMsg.Error())
+			return errMsg
+		}
+		yamlFileContent = content
+		cmd.logger.InfoContext(ctx, "Using prebuilt tool configuration -", "source_type", cmd.prebuiltSourceType)
```
kurtisvg: does source-type need to be here?

cmd/root.go
```diff
@@ -222,14 +272,47 @@ func run(cmd *Command) error {
...
+			for k := range prebuiltToolYAMLs {
+				availablePrebuiltKeys = append(availablePrebuiltKeys, k)
+			}
+			prebuiltHelpSuffix := "No prebuilt configurations found."
+			if len(availablePrebuiltKeys) > 0 {
+				prebuiltHelpSuffix = fmt.Sprintf("Available: %s", strings.Join(availablePrebuiltKeys, ", "))
+			}
+			errMsg := fmt.Errorf("prebuilt source type '%s' not found. %s", cmd.prebuiltSourceType, prebuiltHelpSuffix)
+			cmd.logger.ErrorContext(ctx, errMsg.Error())
+			return errMsg
```
kurtisvg, line 258: Do we need to log this error message and return it?

cmd/root.go
```diff
@@ -222,14 +272,47 @@ func run(cmd *Command) error {
...
+			cmd.tools_file = "tools.yaml"
+		}
+		// Read tool file contents
+		buf, err := os.ReadFile(cmd.tools_file)
+		if err != nil {
+			errMsg := fmt.Errorf("unable to read tool file at %q: %w", cmd.tools_file, err)
+			cmd.logger.ErrorContext(ctx, errMsg.Error())
+			return errMsg
+		}
+		yamlFileContent = buf
```
kurtisvg: nit: this is kind of the 'default' use case -- should we push it first to represent that?

kurtisvg (changes requested):

cmd/prebuiltconfigs/alloydb.yaml
dishaprakash: Added a new package under /internal

cmd/root.go
dishaprakash: Added a new package with a get function to load all YAML files

cmd/root.go
dishaprakash: Done

cmd/root.go
dishaprakash: Changed it into format string for better readability

cmd/root_test.go
dishaprakash: I've added tests to check for the toolset under each source configuration now

cmd/root.go
```diff
@@ -68,11 +69,12 @@ func Execute() {
...
 
-	cfg        server.ServerConfig
-	logger     log.Logger
-	tools_file string
-	outStream  io.Writer
-	errStream  io.Writer
+	cfg                server.ServerConfig
+	logger             log.Logger
+	tools_file         string
+	prebuiltSourceType string
```
kurtisvg: super-nit:
```suggestion
	prebuiltConfig string
```

internal/prebuiltconfigs/prebuiltconfigs.go
```diff
@@ -0,0 +1,58 @@
...
+
+			prebuiltToolYAMLs[sourceTypeKey] = content
+		}
+	}
+	if len(prebuiltToolYAMLs) == 0 {
+		errMsg := fmt.Errorf("no prebuilt tool configurations were loaded.%w", err)
+		return nil, errMsg
+	}
+
+	return prebuiltToolYAMLs, nil
```
kurtisvg: This logic specifically can be moved into an `init` method, so that this becomes "compile time" logic: https://go.dev/doc/effective_go#init

cmd/root.go
```diff
@@ -222,14 +225,57 @@ func run(cmd *Command) error {
...
+				availablePrebuiltKeys = append(availablePrebuiltKeys, k)
+			}
+			prebuiltHelpSuffix := "No prebuilt configurations found."
+			if len(availablePrebuiltKeys) > 0 {
+				prebuiltHelpSuffix = fmt.Sprintf("Available: %s", strings.Join(availablePrebuiltKeys, ", "))
+			}
+			errMsg := fmt.Errorf("prebuilt source type '%s' not found. %s", cmd.prebuiltSourceType, prebuiltHelpSuffix)
+			cmd.logger.ErrorContext(ctx, errMsg.Error())
+			return errMsg
+		}
```
kurtisvg: We can push this logic down so that the ToolsYaml map is read-only. 

```suggestion
		prebuiltConfig, ok := prebuiltconfigs.Get(cmd.prebuildSourceType)
		if !ok {
			errMsg := prebuiltconfigs.NewNotFoundError(cmd.prebuildSourceType)
			return err
		}
```

internal/prebuiltconfigs/prebuiltconfigs.go
dishaprakash: Created an init method and call this function from the init. This is so we can test this load method as well.

cmd/root.go
dishaprakash: Moved the logic into a `.Get` method in the prebuiltconfigs package.

internal/prebuiltconfigs/prebuiltconfigs.go
```diff
@@ -0,0 +1,86 @@
...
+
+	prebuiltToolYAMLs    map[string][]byte
+	prebuiltToolsSources []string
+)
+
+func init() {
+	var err error
+	prebuiltToolYAMLs, prebuiltToolsSources, err = loadPrebuiltToolYAMLs()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unexpected Error: %v\n", err)
```
kurtisvg: nit: this should panic

internal/prebuiltconfigs/prebuiltconfigs.go
```diff
@@ -0,0 +1,86 @@
...
+	if !ok {
+		prebuiltHelpSuffix := "no prebuilt configurations found."
+		if len(prebuiltToolsSources) > 0 {
+			prebuiltHelpSuffix = fmt.Sprintf("available: %s", strings.Join(prebuiltToolsSources, ", "))
+		}
+		errMsg := fmt.Errorf("prebuilt source tool for '%s' not found. %s", prebuiltSourceConfig, prebuiltHelpSuffix)
+		return nil, errMsg
+	}
+	return content, nil
+
```
kurtisvg:
```suggestion
```

internal/prebuiltconfigs/prebuiltconfigs.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func loadPrebuiltToolYAMLs() (map[string][]byte, []string, error) {
+
```
kurtisvg:
```suggestion
```

cmd/root_test.go
```diff
@@ -866,6 +897,119 @@ func TestEnvVarReplacement(t *testing.T) {
...
+				},
+			},
+		},
+		{
+			name: "bigquery prebuilt tools",
+			in:   bigquery_config,
+			wantToolset: server.ToolsetConfigs{
+				"bigquery-postgres-database-tools": tools.ToolsetConfig{
+					Name:      "bigquery-postgres-database-tools",
+					ToolNames: []string{"execute_sql", "list_tables"},
```
dishaprakash: This test seems to be failing as all the tools aren't mentioned

averikitsch: Thank you!

averikitsch (approved):

PR #612, opened 2025-05-26
feat: Add MongoDB as Source

theshanbhag:
Added files required to add MongoDB as datasource for the MCP toolbox.

### DIFF
--- a/internal/sources/mongodb/mongodb.go
+++ b/internal/sources/mongodb/mongodb.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/sources"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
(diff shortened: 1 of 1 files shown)


internal/sources/mongodb/mongodb.go
```diff
@@ -0,0 +1,98 @@
...
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const SourceKind string = "mongodb"
+
+// validate interface
+var _ sources.SourceConfig = Config{}
+
```
duwenxin99, line 32:
```suggestion

func init() {
	if !sources.Register(SourceKind, newConfig) {
		panic(fmt.Sprintf("source kind %q already registered", SourceKind))
	}
}

func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources.SourceConfig, error) {
	actual := Config{Name: name, IPType: "public"} // Default IPType
	if err := decoder.DecodeContext(ctx, &actual); err != nil {
		return nil, err
	}
	return actual, nil
}
```

docs/en/resources/sources/mongodb.md
```diff
@@ -0,0 +1,34 @@
...
+type: docs
+weight: 1
+description: >
+  MongoDB is a no-sql data platform that can not only serve general purpose data requirements also perform VectorSearch where both operational data and embeddings used of search can reside in same document.
+
+---
+
+## About
+
+[MongoDB][mongodb-docs] leading nosql database that can not only cater your operational data needs but also perform vector search.
```
duwenxin99:
```suggestion
[MongoDB][mongodb-docs] is a leading nosql database that can not only cater your operational data needs but also perform vector search.
```

duwenxin99 (commented):
Add mongodb source as dependency here: https://github.com/googleapis/genai-toolbox/blob/518a0e4c70f0d82bf37bf326806931e4ba7693f6/cmd/root.go#L75

theshanbhag:
> Add mongodb source as dependency here:
> 
> https://github.com/googleapis/genai-toolbox/blob/518a0e4c70f0d82bf37bf326806931e4ba7693f6/cmd/root.go#L75

added and commited

PR #614, opened 2025-05-26
refactor: implement dynamic source registration

kurtisvg:
This commit refactors the source configuration and loading mechanism to use a dynamic registration pattern. Each source package now registers itself with a central registry via its init() function.

The server configuration code uses this registry to decode and initialize sources, decoupling it from specific source implementations and simplifying the addition of new sources.

Key changes:
- Introduced `sources.Register()` and `newConfig()` constructor in each source package.
- Moved source package imports to `cmd/root.go` as blank imports to trigger `init()` functions for self-registration.
- Removed direct imports of specific source packages from `internal/server/config.go`.
- Renamed `SourceKind` constants to `Kind` within each source package.
- Updated tests to use the new `Kind` constants and reflect registration changes.

### DIFF
--- a/internal/sources/alloydbpg/alloydb_pg.go
+++ b/internal/sources/alloydbpg/alloydb_pg.go
@@ -42,7 +42,7 @@ func init() {
 func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources.SourceConfig, error) {
 	actual := Config{Name: name, IPType: "public"} // Default IPType
 	if err := decoder.DecodeContext(ctx, &actual); err != nil {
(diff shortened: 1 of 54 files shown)


Yuan325 (approved):

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -46,7 +61,7 @@ type Config struct {
 }
 
 func (r Config) SourceConfigKind() string {
-	return SourceKind
+	return Kind
```
kurtisvg: nit: I think we should revert the naming to SourceKind until we have a better reason to change it

PR #615, opened 2025-05-27
feat(tools/postgressql): add templateParameters field

### LINKED ISSUE #535: Support  usage of `parameters` fields with non-filtering parameters
### Prerequisites

### Toolbox version

0.4.0+binary.linux.amd64.0.4.0.4ed16ccd18768ac02402356f9660526b3452418b

### Environment

1. Linux aaronpc2025 6.8.0-59-generic #61-Ubuntu SMP PREEMPT_DYNAMIC Fri Apr 11 23:16:11 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
2. Downloaded from https://github.com/googleapis/genai-toolbox/releases

### Client

_No response_

### Expected Behavior

When running a mysql query tool through MCP Inspector, it should return the rows inside the table.

### Current Behavior

But when trying to run this mysql tool through MCP Inspector, the following error occurs: "unable to execute query: Error 1146 (42S02): Table 'test.$2' doesn't exist"
This means that the statement inside the tool in tools.yaml isn't placing the parameters inside the statement. This is the yaml code:
```
yaml
sources:
  source1:
    kind: mysql
    host: 127.0.0.1
    port: 3306
    database: test
    user: ${MYSQL_USER}
    password: ${MYSQL_PASS}
tools:
  select_mcp_mysql:
    kind: mysql-sql
    source: source1
    description: Read from database.
    parameters:
      - name: column_names
        type: string
        description: The name of the columns, separated by commas. If you want to select all columns, use '*'.
      - name: table
        type: string
        description: The name of the table.
      - name: instruction
        type: string
        description: The instruction given by the user, including where, order by, group by statements.
    statement: SELECT $1 FROM $2 $3;

```
In the docs found here https://googleapis.github.io/genai-toolbox/resources/tools/ this is the format used to insert parameters inside the query, so it should work. The example found here https://googleapis.github.io/genai-toolbox/resources/tools/mysql-sql/ uses ? instead, which returns syntax errors too, even when using simple queries like SELECT ? FROM ?

### Steps to reproduce?

1. Create tools.yaml with the code above
2. Have a functional MySQL server and user & password envs
3. ./toolbox --tools-file tools.yaml
4. In another terminal: npx @modelcontextprotocol/inspector (following https://googleapis.github.io/genai-toolbox/getting-started/mcp_quickstart/)
5. Try running this tool

AlexTalreja:
Add new tool field, templateParameters, to support non-filter parameters and DDL statements

Fix #535 for postgressql tool.


internal/tools/parameters.go
```diff
@@ -117,7 +119,9 @@ func parseFromAuthService(paramAuthServices []ParamAuthService, claimsMap map[st
 
 // ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
 func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[string]any) (ParamValues, error) {
+
 	params := make([]ParamValue, 0, len(ps))
+
 	for _, p := range ps {
```
Yuan325:
```suggestion
func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[string]any) (ParamValues, error) {
	params := make([]ParamValue, 0, len(ps))
	for _, p := range ps {
```

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
 	return params, nil
 }
 
+// TODO: is it ok that this only works if user entered array of strings ["id", "name"] in yaml
+func convertArrayParamToString(param any) (string, error) {
```
Yuan325: Please add unit tests for this (in `parameters_test.go`).

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+	switch v := param.(type) {
+	case []any:
+		var stringValues []string
+		for _, item := range v {
+			stringVal := fmt.Sprint(item)
```
Yuan325:
```suggestion
			stringVal, ok := item.(string)
			if !ok {
			    return "", fmt.Errorf("only support string arrays")
			}
```

`fmt.Sprint()` will technically convert bool / int / float etc. into the string value. In this case since we are only supporting string arrays, we might want to do type assertion. if the type assertion to string fails, we can throw an error.

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
 	return params, nil
 }
 
+// TODO: is it ok that this only works if user entered array of strings ["id", "name"] in yaml
```
Yuan325: yes, seems like we will only support string arrays for now. feel free to remove this comment and add a comment to describe this helper function instead.

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+			stringVal := fmt.Sprint(item)
+			stringValues = append(stringValues, stringVal)
+		}
+		return strings.Join(stringValues, ", "), nil
+	default:
+		return "", fmt.Errorf("invalid parameter type, expected array of type any")
+	}
+}
+
+func ExtractTemplateParams(templateParams Parameters, originalStatement string, params ParamValues) (ParamValues, string, error) {
```
Yuan325:
```suggestion
func ResolveTemplateParams(templateParams Parameters, originalStatement string, params ParamValues) (ParamValues, string, error) {
```

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+		if !ok {
+			return params, "", fmt.Errorf("error extracting template parameter %s", k)
+		}
+		templateParamsMap[k] = v
+	}
+
+	funcMap := template.FuncMap{
+		"array": convertArrayParamToString,
+	}
+	t, tError := template.New("stTemplate").Funcs(funcMap).Parse(originalStatement)
```
Yuan325:
```suggestion
	t, err := template.New("statement").Funcs(funcMap).Parse(originalStatement)
```

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+	if tError != nil {
+		return params, "", fmt.Errorf("error creating go template %s", tError)
```
Yuan325:
```suggestion
	if err != nil {
		return params, "", fmt.Errorf("error creating go template %s", err)
```

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+	}
+
+	var result bytes.Buffer
+	err := t.Execute(&result, templateParamsMap)
```
Yuan325:
```suggestion
	err = t.Execute(&result, templateParamsMap)
```
if err is defined above, we wont need `:`

internal/tools/parameters.go
```diff
@@ -146,6 +150,58 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+		return params, "", fmt.Errorf("error executing go template %s", err)
+	}
+
+	leftoverParams := make(ParamValues, 0)
+	for _, p := range params {
+		if _, usedParameter := templateParamsMap[p.Name]; !usedParameter {
+			leftoverParams = append(leftoverParams, p)
+		}
+	}
+
```
Yuan325: Instead of looping through the list and cross check with templateParamsMap, can we make a new function: 
```
// GetParams return the ParamValues that are associated with the Parameters.
func GetParams(params Parameters, paramValues ParamValues) (ParamValues, error) {
	paramValuesMap := paramValues.AsMap()
	paramsMap := make(map[string]any)
	for _, p := range paramValues {
		k := p.GetName()
		v, ok := paramValuesMap[k]
		if !ok {
			return nil, fmt.Errorf("missing parameter %s", k)
		}
		paramsMap[k] = v
	}
	return paramsMap, nil
}
```

This function can be used to get paramValues for both the templateParameter and the regular parameters.

We can also update this function to just return `(string, error)` instead of `(ParamValues, string, error)`.

internal/tools/postgressql/postgressql.go
```diff
@@ -105,8 +149,15 @@ type Tool struct {
 }
 
 func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
-	sliceParams := params.AsSlice()
-	results, err := t.Pool.Query(ctx, t.Statement, sliceParams...)
+
+	newParams, newStatement, err := tools.ExtractTemplateParams(t.TemplateParameters, t.Statement, params)
```
Yuan325:
```suggestion
	newParams, newStatement, err := tools.ExtractTemplateParams(t.TemplateParameters, t.Statement, params)
```

Yuan325: This will have to be updated when the `parameters.go` is updated.

internal/tools/postgressql/postgressql.go
```diff
@@ -105,8 +149,15 @@ type Tool struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to extract template params %w", err)
+	}
+
+	sliceParams := newParams.AsSlice()
+
+	results, err := t.Pool.Query(ctx, newStatement, sliceParams...)
```
Yuan325:
```suggestion
	sliceParams := newParams.AsSlice()
	results, err := t.Pool.Query(ctx, newStatement, sliceParams...)
```

internal/tools/postgressql/postgressql_test.go
```diff
@@ -25,6 +25,78 @@ import (
 	"github.com/googleapis/genai-toolbox/internal/tools/postgressql"
 )
 
+func TestParseFromYamlWithTemplateParamsPostgres(t *testing.T) {
```
Yuan325: nit: move this to the bottom (under the other test function)

internal/tools/postgressql/postgressql_test.go
```diff
@@ -25,6 +25,78 @@ import (
...
+	tcs := []struct {
+		desc string
+		in   string
+		want server.ToolConfigs
+	}{
+		{
+			desc: "basic example",
+			in: `
+			tools:
+              example_tool:
```
Yuan325: please fix indentation here :) (make it similar to the ones from the other function).

internal/tools/parameters.go
Yuan325: Please add unit test for this function as well.

internal/tools/parameters.go
Yuan325: Once this change is updated, make sure we have a unit test to test this.

tests/common.go
```diff
@@ -127,6 +127,85 @@ func AddPgExecuteSqlConfig(t *testing.T, config map[string]any) map[string]any {
...
+		"templateParameters": []tools.Parameter{
+			tools.NewStringParameter("tableName", "some description"),
+			tools.NewArrayParameter("columns", "The columns to create", tools.NewStringParameter("column", "A column name that will be created")),
+		},
+	}
+	toolsMap["insert-table-templateParams-tool"] = map[string]any{
+		"kind":        "postgres-sql",
+		"source":      "my-instance",
+		"description": "Insert tool with template parameters",
+		"statement":   "INSERT INTO {{.tableName}} ({{array .columns}}) VALUES ({{array .values}})",
```
Yuan325: Left a comment at the design doc regarding this. We might not want to support multiple-kind arrays. In the Parameter, user will have to set the `items`'s type. In that case, if it is set as string, we shouldn't take ints.

tests/tool.go
```diff
@@ -165,6 +165,130 @@ func RunToolGetTest(t *testing.T) {
 	}
 }
 
+func RunToolInvokeWithTemplateParameters(t *testing.T, tableName string) {
```
Yuan325: nit: ideally place this function under `RunToolInvoke()`.

tests/tool.go
```diff
@@ -165,6 +165,130 @@ func RunToolGetTest(t *testing.T) {
...
+
+	select_all_want := "[{\"age\":21,\"id\":1,\"name\":\"Alex\"},{\"age\":100,\"id\":2,\"name\":\"Alice\"}]"
```
Yuan325:
```suggestion
	select_all_want := "[{\"age\":21,\"id\":1,\"name\":\"Alex\"},{\"age\":100,\"id\":2,\"name\":\"Alice\"}]"
```
Removing extra empty line.

Yuan325 (commented):
Great work! LGTM :) Let's have @kurtisvg  and @duwenxin99  do another pass before merging

averikitsch (commented):
lgtm, but I would prefer Kurtis to do the final approval.

kurtisvg (approved):

Yuan325 (approved):

PR #617, opened 2025-05-27
feat: add support for optional parameters

### LINKED ISSUE #475: Add support for optional parameters
### Prerequisites

Currently, all parameters in Toolbox are required. The value for a parameter must be provided -- either by the LLM, bound at the SDK level, or bound as an authenticated parameter. 

However, this can mean introduce several tools that all do roughly the same thing, with only slightly variations. Sometimes this is a reasonable solution -- our testing shows that models handle multiple tools better than optional parameters. However as models continue to get better, this advantage will likely fade. 

There are a couple of implementation details to figure out, specifically:

1. If we mark a parameter as optional, what is the default value? 
2. Should we allow for optional parameters to default to null? 

### Suggested Solution(s)

Option 1: Add a `default` field to parameters, that enables users to specify a default value.

e.g.
```yaml
  parameters:
    - name: name
      type: string
      default: "some-default-value"
      description: The name of the hotel.
```

If this parameter is invoked without specifying `name`, the parameter would default to "some-default-value"

### Alternatives Considered

Option 2a: Add a `optional` field to parameters, that defaults to `false`:
e.g.
```yaml
  parameters:
    - name: name
      type: string
      optional: true
      description: The name of the hotel.
```

This could default to either `null` or some zero value for different types e.g.` 0` for ints.

Option 2b: Add a `required` field to parameters, that defaults to `true`:
e.g.
```yaml
  parameters:
    - name: name
      type: string
      required: true
      description: The name of the hotel.
```

This could default to either `null` or some zero value for different types e.g.` 0` for ints.

Option 3:  Add both `required` field and `default` field to parameters, that defaults to `true`:
```yaml
  parameters:
    - name: name
      type: string
      required: true
      default: "some-value"
      description: The name of the hotel.
```

This could provide "null" if default is unset, and the value of "default" otherwise. 

### Additional Details

_No response_

Yuan325:
Add a `default` field to parameters, that enables users to specify a default value.

e.g.

```
  parameters:
    - name: name
      type: string
      default: "some-default-value"
      description: The name of the hotel.
```
if this parameter is invoked without specifying `name`, the parameter would default to "some-default-value"

For parameter manifest, there will be an additional `Required` field. The default `Required` field is true. If a `default` value is presented, `Required: false`. Array parameter's item's `Required` field will inherit the array's `Required` field.

Fixes #475


internal/tools/parameters.go
```diff
@@ -366,23 +379,25 @@ type ParamAuthService struct {
 }
 
 // NewStringParameter is a convenience function for initializing a StringParameter.
-func NewStringParameter(name, desc string) *StringParameter {
+func NewStringParameter(name string, defaultV any, desc string) *StringParameter {
```
duwenxin99: Why do we make defaultV of type `any` instead of the corresponding param value type?

Yuan325: The current implementation is to enforce default value's type in ParseParams function. In this implementation, the `Default` field is added at the `CommonParameter` struct, and `CommonParameter` includes all different types so it will have to be any.

If we want it to be specific towards the parameter type, it's possible, but we'll have to remove the `Default` value from `CommonParameter` --> add `Default` to the each individual parameter types (`StringParameter`, `IntParameter`, etc.) --> run a type assertion in the parameter unmarshaler. We'll also need a `func (p *StringParameter) Manifest() ParameterManifest` for each type of parameters. since its currently running on `func (p *CommonParameter Manifest()`. Not sure if whats the best implementation here. @kurtisvg any thoughts?

Yuan325: I was trying to implement this but I think it will add much more complications. E.g. if we have the value as string, Go's string nil value will be "" instead of `nil`. This will make it more complicated to check for nil-ness when we want to replace it within `ParseParam()`. Personal preference to keep it as is within `CommonParameter`.

internal/server/common_test.go
```diff
@@ -95,16 +95,16 @@ var tool1 = MockTool{
 var tool2 = MockTool{
 	Name: "some_params",
 	Params: tools.Parameters{
-		tools.NewIntParameter("param1", "This is the first parameter."),
```
kurtisvg: nit: rather than changing all these invocations, should there be a NewIntParameterWithDefault?

internal/tools/parameters.go
kurtisvg: It looks like defaultV here (for these functions) could still be string and Default field (in common parameter) could still be `Any`?

internal/server/common_test.go
kurtisvg: If we feel like that's getting too cumbersome, we can switch to an options pattern here (but maybe in a separate PR)

internal/tools/parameters.go
```diff
@@ -295,6 +301,7 @@ func (ps Parameters) McpManifest() McpToolsSchema {
 type ParameterManifest struct {
 	Name         string             `json:"name"`
 	Type         string             `json:"type"`
+	Default      any                `json:"default"`
```
kurtisvg: Do we need to supply the default to the tool just just mark it as required? 

I'm not sure the Tool needs to know the default?

Yuan325: oh true, the parameter manifest might not need to know the default value. I need to check with @anubhav756  to see if we want some sort of "required" or "optional" field though. cause the LLM (or langchain tools) probably needs to know that its an optional field.

internal/server/common_test.go
Yuan325: Will set a new invocation functions WithDefault

internal/tools/parameters.go
Yuan325: if its `""`, it will count as default value presented. Switch to a new invocation function -- `WithDefault` -- instead.

PR #625, opened 2025-05-29
docs: highlight dev assist workflow

averikitsch:


AndrewBrook-Google (approved):
LGTM.  The management tasks well expand as we add control plane support but there's enough there for now to keep that statement.  The rest seems reasonable.

PR #629, opened 2025-05-29
feat: add v20250326 schema

Yuan325:
Add v2025-03-26 version protocol [schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts) to Toolbox.

This PR also added: 
* tests for the v2025-03-26 protocol version
* update `internal/server/mcp/mcp.go` to include v20250326
* initialization response has to include "Mcp-Session-Id"

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -32,6 +32,7 @@ import (
 	"github.com/googleapis/genai-toolbox/internal/server/mcp/jsonrpc"
 	mcputil "github.com/googleapis/genai-toolbox/internal/server/mcp/util"
 	v20241105 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20241105"
+	v20250326 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20250326"
 	"github.com/googleapis/genai-toolbox/internal/util"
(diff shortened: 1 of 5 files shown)


internal/server/mcp.go
```diff
@@ -299,6 +300,12 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// check if client have `Mcp-Session-Id` header
```
kurtisvg, line 304: Can we clarify this comment a bit better?

"Assume that if 'Mcp-Session-Id' header is set, we are using 20250326. This is because the previous version doesn't use this header."

internal/server/mcp.go
```diff
@@ -343,6 +350,12 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 		s.logger.DebugContext(ctx, err.Error())
 	}
 
+	// for v20240326, add the `Mcp-Session-Id` header
```
kurtisvg:
```suggestion
	// for v20250326, add the `Mcp-Session-Id` header
```

PR #630, opened 2025-05-30
chore: migrate golangci-lint to v2

Yuan325:
gosimple had been deprecated in favor of staticcheck: https://github.com/golangci/golangci-lint/issues/357

Other requirements are all migrated.

`std-error-handling` exclusions is included because without that, it will ask to check all error returns from (`Close()`, or `os.Setenv`s, or `fmt.Fprint`s...

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -21,16 +21,8 @@ linters:
     - staticcheck
     - unused
   exclusions:
-    generated: lax
     presets:
-      - comments
-      - common-false-positives
-      - legacy
       - std-error-handling
-    paths:
-      - third_party$
-      - builtin$
-      - examples$
 issues:
   fix: true
 formatters:
@@ -43,9 +35,3 @@ formatters:
           replacement: any
         - pattern: a[b:len(a)]
           replacement: a[b:]
-  exclusions:
-    generated: lax
-    paths:
-      - third_party$
-      - builtin$
-      - examples$
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,21 +12,40 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+version: "2"
 linters:
   enable:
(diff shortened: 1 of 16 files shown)


.golangci.yaml
```diff
@@ -12,21 +12,40 @@
...
     - unused
-linters-settings:
-  gofmt:
-    rewrite-rules:
-      - pattern: 'interface{}'
-        replacement: 'any'
-      - pattern: 'a[b:len(a)]'
-        replacement: 'a[b:]'
+  exclusions:
+    generated: lax
```
averikitsch: Do we need this if we don't have generated files?

cmd/root.go
```diff
@@ -228,13 +228,13 @@ func run(cmd *Command) error {
...
 	default:
-		return fmt.Errorf("logging format invalid.")
+		return fmt.Errorf("logging format invalid")
 	}
 
 	ctx = util.WithLogger(ctx, cmd.logger)
 
 	// Set up OpenTelemetry
-	otelShutdown, err := telemetry.SetupOTel(ctx, cmd.Command.Version, cmd.cfg.TelemetryOTLP, cmd.cfg.TelemetryGCP, cmd.cfg.TelemetryServiceName)
+	otelShutdown, err := telemetry.SetupOTel(ctx, cmd.Version, cmd.cfg.TelemetryOTLP, cmd.cfg.TelemetryGCP, cmd.cfg.TelemetryServiceName)
```
averikitsch, line 237: Should this be cmd.cfg.Version?

averikitsch (approved):

cmd/root.go
Yuan325: here: https://github.com/googleapis/genai-toolbox/blob/ad97578fdfa39d92ae3650793cb2939bb032970c/cmd/root.go#L89

Both of it works~

PR #635, opened 2025-06-02
docs: add bigquery prebuilt tools documentation.

Genesis929:


docs/en/samples/bigquery/prebuilt_tools.md
averikitsch, line 1: Hi Huan, Thank you for your pull request. What is the goal of this additional documentation? There is quite a lot of overlap with the current documentation https://googleapis.github.io/genai-toolbox/how-to/connect-ide/bigquery_mcp/. There is also a gap for connecting the MCP server to an AI-enable IDE or other tool. We previously discussed an E2E sample. If that is the goal, we should ensure we capture more steps to ensure a good experience. If that is not the goal, we could add the sample prompts to the current docs to better highlight the value.

Genesis929: Yes, this may not needed. Thanks.

PR #637, opened 2025-06-02
chore: terminate mcp session after timeout

Yuan325:
In addition to allow client to send HTTP DELETE request, MCP Toolbox will terminate the session if it is inactive for >10 minutes to prevent memory bloat. However, this rule only applies to clients that uses protocol version v2025-03-26.


internal/server/mcp.go
```diff
@@ -63,16 +65,27 @@ type mcpManager struct {
 	mcpSessions map[string]*mcpSession
 }
 
+func newMcpManager() *mcpManager {
+	mcpM := &mcpManager{
+		mu:          sync.RWMutex{},
+		mcpSessions: make(map[string]*mcpSession),
+	}
+	go mcpM.cleanupRoutine()
```
kurtisvg: There should be a way to close this so the gorountine doesn't leak (if the MCP manager shutdown for any reason).

Can you pass a context with cancel in here?

internal/server/mcp.go
```diff
@@ -63,16 +65,27 @@ type mcpManager struct {
...
+	return mcpM
+}
+
 func (m *mcpManager) get(id string) (*mcpSession, bool) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 	session, ok := m.mcpSessions[id]
+	session.lastActive = time.Now()
```
kurtisvg: this is a write, so you need to use the write mutex?

internal/server/mcp.go
```diff
@@ -63,16 +65,27 @@ type mcpManager struct {
...
 	return session, ok
 }
 
 func (m *mcpManager) add(id string, session *mcpSession) {
 	m.mu.Lock()
 	m.mcpSessions[id] = session
+	session.lastActive = time.Now()
 	m.mu.Unlock()
```
kurtisvg: nit: prefer defer

internal/server/mcp.go
```diff
@@ -71,6 +84,30 @@ func (m *sseManager) remove(id string) {
...
+
+	for range ticker.C {
+		m.mu.Lock()
+		now := time.Now()
+		for id, sess := range m.sseSessions {
+			if now.Sub(sess.lastActive) > timeout {
+				delete(m.sseSessions, id)
+			}
+		}
+		m.mu.Unlock()
```
kurtisvg: nit: pro-tip for these situations is you can still use defer in an anon function to ensure the mutex is unlocked even if a panic occurs
```suggestion
                 // use a function here in order to user 'defer' to unlock this mutex when complete
		func() {
		  m.mu.Lock()
		  defer m.mu.Unlock()
		  now := time.Now()
		  for id, sess := range m.sseSessions {
			  if now.Sub(sess.lastActive) > timeout {
				  delete(m.sseSessions, id)
			  }
		  }
		 }()
		
```

internal/server/mcp.go
```diff
@@ -71,6 +84,30 @@ func (m *sseManager) remove(id string) {
...
+	}
+
+	// to shutdown goroutine if ctx is cancelled
+	select {
+	case <-ctx.Done():
+		return
+	default:
+	}
```
kurtisvg: I don't think this is right. you will only get here once the `range ticker.C` is closed. 

Shouldn't you listen for this in between ticks?

kurtisvg (changes requested):

internal/server/mcp.go
```diff
@@ -313,7 +350,7 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check if client have `Mcp-Session-Id` header
-	// if `Mcp-Session-Id` header is set, we are using v2025-03-26 since 
+	// if `Mcp-Session-Id` header is set, we are using v2025-03-26 since
```
kurtisvg, line 353: nit: since what?

kurtisvg (approved):

internal/server/mcp.go
Yuan325: there's more comment in the next line

PR #638, opened 2025-06-02
fix(server/stdio): notifications should not return a response

Yuan325:
Currently the `stdio` transport protocol will throw a `ZodError` during initialization. This is due to Toolbox writing `null` to stdout when it received a notification. This is not expected hence the `ZodError` occurs. Per the MCP protocol, notifications do not expect any response.

This fix added a condition to check if the responses is `nil` before writing to stdout.


averikitsch (approved):
Thank you! Tested this locally as well and the error is gone.

PR #641, opened 2025-06-03
fix(tools/mysqlsql): handled the null value for string case in mysqlsql tools

iammujtaba:
@Yuan325  please review and merge this if all looks ok.
In existing code, for the character field with null values it will panic with error:
"panic: interface conversion: interface {} is nil, not []uint8"


Yuan325:
Thank you!

PR #643, opened 2025-06-03
docs: Simplify repetitive auth login command in bigquery quickstart

anubhav756:


Yuan325:
One of it is to authorize access to GCP with Google user credential, and another is for ADC. Do we only need one of them? alternatively, we might be able to use `gcloud auth login --update-adc` if both are needed.

duwenxin99:
> One of it is to authorize access to GCP with Google user credential, and another is for ADC. Do we only need one of them? alternatively, we might be able to use `gcloud auth login --update-adc` if both are needed.

Let's simplify it to gcloud auth login --update-adc

PR #644, opened 2025-06-03
Fix: panic: interface conversion: interface {} is nil, not []uint8

kym6464:
Fixes a bug with the mysql-execute-sql tool which causes a panic: interface conversion: interface {} is nil, not []uint8

I added an integration test case but did not go through the effort of setting it up locally since an admin can run the test suite against the PR.

FYI I didn't realize https://github.com/googleapis/genai-toolbox/pull/641 fixes the same issue so only of these should be merged

### DIFF
--- a/internal/tools/mysqlexecutesql/mysqlexecutesql.go
+++ b/internal/tools/mysqlexecutesql/mysqlexecutesql.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"slices"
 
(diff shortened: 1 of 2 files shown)


kym6464:
Just signed the CLA. @Yuan325 would you mind reviewing & running the integration tests?

Yuan325:
Hi! Thank you for submitting this! I've merged the other PR that solves the same issue. Will close this PR. Also, thanks for catching the typo! I created a separate PR to [fix the typo](https://github.com/googleapis/genai-toolbox/pull/645).

PR #658, opened 2025-06-04
fix: improve versionString

kurtisvg:
Reduces complexity of the version string calculation and fixes the ci/cd pipeline to pass the correct parameters.


.ci/versioned.release.cloudbuild.yaml
```diff
@@ -56,7 +56,7 @@ steps:
         #!/usr/bin/env bash
         export VERSION=$(cat ./cmd/version.txt)
         CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
-          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.metadataString=binary.linux.amd64.$VERSION.$(git rev-parse HEAD)" -o toolbox.linux.amd64
+          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.buildType=binary -X github.com/googleapis/genai-toolbox/cmd.commitSha=$(git rev-parse HEAD)" -o toolbox.linux.amd64
```
Yuan325, line 59: Do we not want to include the OS in buildType metadata?

Yuan325: oh nvm just saw that we are extracting it directing when adding metadata in cmd/root.

PR #670, opened 2025-06-05
feat: detect edits in tools file

AlexTalreja:
Use fsnotify to detect edits to the tools config file (will only trigger on save of file).

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -173,7 +173,12 @@ func parse(ctx context.Context, buf []byte, logger log.Logger) {
 }
 
 // watchFile checks for changes in the provided yaml tools file.
-func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
+func watchFile(ctx context.Context, toolsFileName string) {
+	logger, err := util.LoggerFromContext(ctx)
+	if err != nil {
+		return
+	}
+
 	w, err := fsnotify.NewWatcher()
 	if err != nil {
 		logger.WarnContext(ctx, "error setting up new watcher %s", err)
@@ -187,33 +192,22 @@ func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
 		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
 	}
 
-	logger.DebugContext(ctx, fmt.Sprintf("Now watching tools file %s", toolsFileName))
+	cleanedFilename := filepath.Clean(toolsFileName)
+	logger.DebugContext(ctx, fmt.Sprintf("Now watching tools file %s", cleanedFilename))
 
-	// when the user saves a file, we wait for 100ms of no new events (due to how many editors perform writes)
+	// debounce timer is used to prevent multiple writes triggering multiple reloads
 	debounceDelay := 100 * time.Millisecond
-
-	var timerC <-chan time.Time
-	var debounceTimer *time.Timer
-
-	defer func() {
-		if debounceTimer != nil {
-			debounceTimer.Stop()
-		}
-	}()
+	debounce := time.NewTimer(1 * time.Minute)
+	debounce.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
-			// potentially unnecessary due to defer above?
-			if debounceTimer != nil {
-				debounceTimer.Stop()
-			}
 			logger.DebugContext(ctx, "file watcher context cancelled")
 			return
-
 		case err, ok := <-w.Errors:
 			if !ok {
-				logger.WarnContext(ctx, "file watcher alredy closed")
+				logger.WarnContext(ctx, "file watcher was closed unexpectedly")
 				return
 			}
 			if err != nil {
@@ -227,35 +221,13 @@ func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
 				return
 			}
 
-			if filepath.Clean(e.Name) == filepath.Clean(toolsFileName) && e.Op == fsnotify.Write {
+			if e.Op == fsnotify.Write && filepath.Clean(e.Name) == cleanedFilename {
 				logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, e.Name))
-				// if a debounceTimer is already running, stop and drain to reset and wait 100ms again
-				if debounceTimer != nil {
-					if !debounceTimer.Stop() {
-						/*
-							NOTE: supposedly this is the proper way, due to chance that Stop() has returned false
-							(meaning timer already expired) so it should have sent on the channel, but if at the same
-							instant the case <-timer.C reads it, then this would deadlock without the select
-
-							Also, is draining needed after 1.23? https://pkg.go.dev/time#Timer.Reset
-						*/
-						select {
-						case <-debounceTimer.C:
-						default:
-						}
-					}
-				}
-				// either way (repeat event or new) create a new debounceTimer and wait 100ms before allowing parse
-				debounceTimer = time.NewTimer(debounceDelay)
-				timerC = debounceTimer.C
+				debounce.Reset(debounceDelay)
 			}
-		case <-timerC:
-			debounceTimer.Stop()
-
-			// remove references
-			debounceTimer = nil
-			timerC = nil
-
+		case <-debounce.C:
+			debounce.Stop()
+			logger.DebugContext(ctx, "re-reading tools file")
 			buf, err := os.ReadFile(toolsFileName)
 			if err != nil {
 				logger.WarnContext(ctx, "error reading reloaded file", err)
@@ -372,7 +344,7 @@ func run(cmd *Command) error {
 	}()
 
 	// start watching for file changes to trigger dynamic reloading
-	go watchFile(cmd.tools_file, ctx, cmd.logger)
+	go watchFile(ctx, cmd.tools_file)
 
 	// wait for either the server to error out or the command's context to be canceled
 	select {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -184,7 +184,7 @@ func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
 		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
 	}
 
-	logger.InfoContext(ctx, "Now watching tools file %s", toolsFileName)
+	logger.InfoContext(ctx, fmt.Sprintf("Now watching tools file %s", toolsFileName))
 	var debounceTimer *time.Timer
 	debounceDelay := 100 * time.Millisecond
 	for {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -190,19 +190,20 @@ func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
 	for {
 		select {
 		case <-ctx.Done():
-			logger.WarnContext(ctx, "watcher context cancelled")
+			logger.WarnContext(ctx, "file watcher context cancelled")
 			return
 		case err, ok := <-w.Errors:
 			if !ok {
-				logger.WarnContext(ctx, "error watcher alredy closed %s", err)
+				logger.WarnContext(ctx, "file watcher alredy closed")
 			}
 
 			if err != nil {
-				logger.WarnContext(ctx, "error watching file %s", err)
+				logger.WarnContext(ctx, "file watcher error %s", err)
 			}
 		case e, ok := <-w.Events:
 			if !ok {
-				logger.WarnContext(ctx, "error with event %s", err)
+				logger.WarnContext(ctx, "file watcher already closed")
+				return
 			}
 			if strings.HasSuffix(e.Name, toolsFileName) && e.Op == fsnotify.Write {
 				if debounceTimer == nil {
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -268,6 +326,9 @@ func run(cmd *Command) error {
 		}
 	}()
 
+	// start watching for file changes to trigger dynamic reloading
+	go watchFile(string(cmd.tools_file), ctx, cmd.logger)
```
Yuan325: `cmd.tools_file` should already be string? do we need the additional parsing here?

cmd/root.go
```diff
@@ -163,6 +165,62 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+func parse(ctx context.Context, logger log.Logger) {
+	logger.DebugContext(ctx, "Attempting to parse updated tools file.")
+	// TODO: add logic for parsing
+}
+
+// watchFile checks for changes in the provided yaml tools file.
+func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.WarnContext(ctx, "error setting up new watcher %s", err)
```
Yuan325, line 184: We might wanna crash Toolbox if it fails to setup dynamic reloading capabilities. @kurtisvg WDYT?

cmd/root.go
```diff
@@ -163,6 +165,62 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+	var debounceTimer *time.Timer
+	debounceDelay := 100 * time.Millisecond
+	for {
+		select {
+		case <-ctx.Done():
+			logger.WarnContext(ctx, "watcher context cancelled")
+			return
+		case err, ok := <-w.Errors:
+			if !ok {
+				logger.WarnContext(ctx, "error watcher alredy closed %s", err)
```
```suggestion
				logger.WarnContext(ctx, "file watcher already closed")
```

Seems like we are logging the error separately under this?

cmd/root.go
```diff
@@ -163,6 +165,62 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
+	}
+
+	logger.InfoContext(ctx, fmt.Sprintf("Now watching tools file %s", toolsFileName))
+	var debounceTimer *time.Timer
+	debounceDelay := 100 * time.Millisecond
+	for {
+		select {
+		case <-ctx.Done():
+			logger.WarnContext(ctx, "watcher context cancelled")
```
```suggestion
			logger.WarnContext(ctx, "file watcher context cancelled")
```

cmd/root.go
```diff
@@ -163,6 +165,62 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+			return
+		case err, ok := <-w.Errors:
+			if !ok {
+				logger.WarnContext(ctx, "error watcher alredy closed %s", err)
+			}
+
+			if err != nil {
+				logger.WarnContext(ctx, "error watching file %s", err)
```
```suggestion
			if err != nil {
				logger.WarnContext(ctx, "file watcher error: %s", err)
```

cmd/root.go
```diff
@@ -163,6 +165,62 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+			}
+		case e, ok := <-w.Events:
+			if !ok {
+				logger.WarnContext(ctx, "error with event %s", err)
```
Yuan325: Will this exit if there's error?
```suggestion
				logger.WarnContext(ctx, "file watcher error with event: %s", err)
```

AlexTalreja: Originally not always, added a return and updated the message to be more accurate.

cmd/root.go
Yuan325: Resolve this for now. We'll need to think more about this in the future (e.g. if we add a flag to trigger dynamic reloading then we'll probably want to crash it).

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+	}
+
+	defer w.Close()
+
+	err = w.Add(filepath.Dir(toolsFileName))
+	if err != nil {
+		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
+	}
+
+	logger.InfoContext(ctx, fmt.Sprintf("Now watching tools file %s", toolsFileName))
```
kurtisvg: nit: should be debug

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+func parse(ctx context.Context, logger log.Logger) {
+	logger.DebugContext(ctx, "Attempting to parse updated tools file.")
+	// TODO: add logic for parsing
+}
+
+// watchFile checks for changes in the provided yaml tools file.
+func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.WarnContext(ctx, "error setting up new watcher %s", err)
```
kurtisvg, line 184: should stop at this point? Can't do l182 if the watcher has failed.

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
+	}
+
+	logger.InfoContext(ctx, fmt.Sprintf("Now watching tools file %s", toolsFileName))
+	var debounceTimer *time.Timer
+	debounceDelay := 100 * time.Millisecond
+	for {
+		select {
+		case <-ctx.Done():
+			logger.WarnContext(ctx, "file watcher context cancelled")
```
kurtisvg: Should be debug. Will happen every-time the server shutdown.

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+
+			if err != nil {
+				logger.WarnContext(ctx, "file watcher error %s", err)
+			}
+		case e, ok := <-w.Events:
+			if !ok {
+				logger.WarnContext(ctx, "file watcher already closed")
+				return
+			}
+			if strings.HasSuffix(e.Name, toolsFileName) && e.Op == fsnotify.Write {
```
kurtisvg: do we get notified for other events? Is this just a precaution?

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+				logger.WarnContext(ctx, "file watcher alredy closed")
+			}
+
+			if err != nil {
+				logger.WarnContext(ctx, "file watcher error %s", err)
+			}
+		case e, ok := <-w.Events:
+			if !ok {
+				logger.WarnContext(ctx, "file watcher already closed")
+				return
```
kurtisvg, line 221: Why do we return here but not at Errors?

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+					debounceTimer = time.NewTimer(debounceDelay)
+					go func() {
+						<-debounceTimer.C
+						logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, e.Name))
+						parse(ctx, logger)
+						debounceTimer = nil
+					}()
+				} else {
+					debounceTimer.Reset(debounceDelay)
+				}
```
kurtisvg: Could we get a comment describing this functionality?

cmd/root.go
```diff
@@ -163,6 +165,63 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+				return
+			}
+			if strings.HasSuffix(e.Name, toolsFileName) && e.Op == fsnotify.Write {
+				if debounceTimer == nil {
+					debounceTimer = time.NewTimer(debounceDelay)
+					go func() {
+						<-debounceTimer.C
+						logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, e.Name))
+						parse(ctx, logger)
+						debounceTimer = nil
```
kurtisvg: I think there is a race condition introduced here since this is happening in a seperate goroutine. What happens if this timer is set to nil and then reset ?

kurtisvg (changes requested):

cmd/root.go
AlexTalreja: We do get notified of other events, so this filters just for `Write`

cmd/root.go
```diff
@@ -163,6 +165,107 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
 
+func parse(ctx context.Context, buf []byte, logger log.Logger) {
+	logger.DebugContext(ctx, "Attempting to parse updated tools file.")
+	// time.Sleep(5 * time.Second)
+	// logger.DebugContext(ctx, "Parse finished.")
+	// TODO: add logic for parsing
+}
+
+// watchFile checks for changes in the provided yaml tools file.
+func watchFile(toolsFileName string, ctx context.Context, logger log.Logger) {
```
kurtisvg: nit: ctx should be first argument.

Can we extract logger from context? I think we might be doing this in other places.

cmd/root.go
```diff
@@ -163,6 +165,107 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+				return
+			}
+
+		case e, ok := <-w.Events:
+			if !ok {
+				logger.WarnContext(ctx, "file watcher already closed")
+				return
+			}
+
+			if filepath.Clean(e.Name) == filepath.Clean(toolsFileName) && e.Op == fsnotify.Write {
```
kurtisvg: nit: should save the output from `filepath.Clean(toolsFileName)` somewhere so it doesn't need to run every time

kurtisvg: nit: flip conditions here -- e.Op == fsnotify.Write is trivial compared to calling Clean()

cmd/root.go
```diff
@@ -163,6 +165,107 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+			timerC = nil
+
+			buf, err := os.ReadFile(toolsFileName)
+			if err != nil {
+				logger.WarnContext(ctx, "error reading reloaded file", err)
+				return
+			}
+			parse(ctx, buf, logger)
+		}
+	}
```
```suggestion
 # debounce timer is used to prevent multiple writes triggering multiple reloads
        debounceDelay := 100 * time.Millisecond
        debounce := NewTimer(1 * time.Minute)
        debounce.Stop()

        for {
                select {
                case <-ctx.Done():
                        logger.DebugContext(ctx, "file watcher context cancelled")
                        return
                case err, ok := <-w.Errors:
                        if !ok {
                                logger.WarnContext(ctx, "file watcher was closed unexpectedly")
                                return
                        }
                        if err != nil {
                                logger.WarnContext(ctx, "file watcher error %s", err)
                                return
                        }

                case e, ok := <-w.Events:
                        if !ok {
                                logger.WarnContext(ctx, "file watcher already closed")
                                return
                        }

                        if filepath.Clean(e.Name) == filepath.Clean(toolsFileName) && e.Op == fsnotify.Write {
                                logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, e.Name))
                                debouce.Reset(debounceDelay)
                        }
                case <-timerC:
                        debouce.Stop()
                        logger.DebugContext(ctx, fmt.Sprintf("re-reading file at: %q", e.Name))                 
                        buf, err := os.ReadFile(toolsFileName)
                        if err != nil {
                                logger.WarnContext(ctx, "error reading reloaded file", err)
                                return
                        }
                        parse(ctx, buf, logger)
                }
```

kurtisvg (changes requested):

cmd/root.go
AlexTalreja: Yes, modified to use `util.LoggerFromContext(ctx)`. However I'm not sure how best to handle the error case where we fail to extract the logger (if that could happen at this stage). 

Since watchFile() doesn't return or write to a channel, would we need to use fmt or built-in log to display the error?

kurtisvg: I don't think it should happen (if the logger is already created it should be in the context) but if it does we probably need to panic

cmd/root_test.go
```diff
@@ -858,3 +866,125 @@ func TestEnvVarReplacement(t *testing.T) {
...
+	}
+	if err := f.Close(); err != nil {
+		cleanup()
+		return "", nil, err
+	}
+	return f.Name(), cleanup, err
+}
+
+// matches server logs a single line that matches the provided regex.
+// adapted from tests/server.go
```
kurtisvg: nit: should this test just be in tests/server.go instead?

kurtisvg (approved):

cmd/root_test.go
AlexTalreja: Initially `WaitForString()` couldn't be reused from tests/server.go since it was a method connected to CmdExec and anything from tests/ would also cause an import cycle, but it could be moved into internal/testutils and deduped with the server.go version in another PR.

PR #507, opened 2025-04-30
Add Helm chart for Kubernetes deployment

brandon-powers:
# Overview

This PR adds a Helm chart to the MCP toolbox, related to https://github.com/googleapis/genai-toolbox/issues/486. It's under a top-level directory named `helm/`, containing the chart & configurable `values.yaml`. To name a few,
- HPA auto-scaling & pod topology constraint / [anti-]affinity
- Parity with CLI options
- Auto service account creation with annotations (for auth)

It also includes an update to the GKE documentation, with a new page for Helm installation. Currently working on testing it fully & docs, though wanted to raise it here as a draft.

PR #673, opened 2025-06-06
feat: helm support - rewritten

alessandro-festa:
feat: helm Support - Rewritten


alessandro-festa:
I've completly rewritten #507  that was stale and based on @kurtisvg suggestion I've pick the work. Unfortunately I had to rewrite everything from scratch and this is why I'm opening up a new PR. So #507 could probably be closed in favor of this one.

alessandro-festa:
ping to see if this is still of interest...

kurtisvg:
Hi @alessandro-festa -- sorry this slipped through the cracks. I'll take a look by EOW. Sorry for the delay!

alessandro-festa:
> Hi @alessandro-festa -- sorry this slipped through the cracks. I'll take a look by EOW. Sorry for the delay!

no worries at all! Just thought maybe you guys already have something even better in the roadmap :)

helm/genai-toolbox/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: genai-toolbox
```
kurtisvg:
```suggestion
name: mcp-toolbox
```

helm/genai-toolbox/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: A Helm chart for Google GenAI Toolbox (formerly MCP Toolbox for Databases)
```
kurtisvg:
```suggestion
description: A Helm chart for MCP Toolbox for Databases, an open source MCP server for databases.
```

helm/genai-toolbox/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# MCP Toolbox for Databases is an open source MCP server for databases. 
+# It enables you to develop tools easier, faster, and more securely by handling the complexities
+# such as connection pooling, authentication, and more.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
kurtisvg, line 13: nit: Should we just match the repo version with this?

helm/genai-toolbox/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.6.0"
```
kurtisvg: currently 0.7.0

We also need to add this file here: https://github.com/googleapis/genai-toolbox/blob/main/.github/release-please.yml#L19 so that our automation bumps this automatically

helm/genai-toolbox/README.md
```diff
@@ -0,0 +1,72 @@
+# MCP Toolbox for Databases Helm Chart
+
+This chart installs the [MCP Toolbox for Databases](https://googleapis.github.io/genai-toolbox/getting-started/introduction/) (formerly "Gen AI Toolbox for Databases") on [Kubernetes](https://kubernetes.io) via the [Helm](https://helm.sh) package manager.
```
kurtisvg:
```suggestion
This chart installs the [MCP Toolbox for Databases](https://googleapis.github.io/genai-toolbox/getting-started/introduction/) on [Kubernetes](https://kubernetes.io) via the [Helm](https://helm.sh) package manager.
```

helm/genai-toolbox/README.md
```diff
@@ -0,0 +1,72 @@
...
+- Kubernetes 1.22+
+- kubectl 1.22+ 
+- Helm 3.8.0+ 
+
+## Installation
+
+# Add the Helm repository
+
+```shell
+helm repo add genai-toolbox https://googleapis.github.io/genai-toolbox
```
kurtisvg:
```suggestion
helm repo add mcp-toolbox https://googleapis.github.io/genai-toolbox
```

helm/genai-toolbox/templates/_helpers.tpl
```diff
@@ -0,0 +1,93 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "genai-toolbox.name" -}}
```
kurtisvg: Can we use `toolbox` over `genai-toolbox` please?

helm/genai-toolbox/values.yaml
```diff
@@ -0,0 +1,197 @@
+# Copyright 2024 Google LLC
```
kurtisvg: nit: 2025

helm/genai-toolbox/values.yaml
```diff
@@ -0,0 +1,197 @@
...
+      items:
+      - key: tools.yaml
+        path: tools.yaml
+
+# Additional volumeMounts on the output Deployment definition.
+volumeMounts: 
+  - name: toolbox-config
+    mountPath: "/app/tools.yaml"
+    subPath: tools.yaml
+    readOnly: true
```
kurtisvg, line 151: I'm a little unclear on why we need this and the config section below. 

Would it be better to tell the user to mount the file a secret and give them the option to specify the secret for tools.yaml instead?

helm/genai-toolbox/values.yaml
```diff
@@ -0,0 +1,197 @@
...
+  telemetry_gcp: false
+  telemetry_otlp: ""
+  telemetry_service_name: "toolbox"
+
+config: |
+# config: |
+#   authServices:
+#   #   my_auth_app_1:
+#   #     kind: google
+#   #     clientId: ${YOUR_CLIENT_ID_1}
```
kurtisvg, line 173: I don't see any way to specify env variables or secrets to mount to the pod. Can we add something that handles that for the user?

alessandro-festa:
> Any thoughts on how we could best test this?

I have a micro cluster based on k3s I can add some steps in the helm documentation and use it to deploy the helm chart but is manual steps ..I was wondering if we could use a github action to spin up a cluster  and launch a script to install the helm?

helm/genai-toolbox/Chart.yaml
alessandro-festa: humm not sure I mean I think the version should be related to the helm chart not necessarily the repo version, reason is that while the repo keep releasing we not necessarily update the chart.

helm/genai-toolbox/values.yaml
alessandro-festa: Agreed, actually the secret itself is a bit of a limitation due the fact that cannot be changed once created, I was tempted to use a configmap but on reverse while is easy to update has the problem the information is not encrypted so any authentication token or username/password will be exposed.

alessandro-festa: So actually I reviewed the current documentation and in the actual example for K8s it is used the same way. I'm not sure is possible to pass a file as reference in env file outside of the method I've used.

helm/genai-toolbox/values.yaml
alessandro-festa: As above I think the only possible way is to stick with the same steps (9 & 10) in the actual documentation https://googleapis.github.io/genai-toolbox/how-to/deploy_gke/

kurtisvg:
@duwenxin99 could you take a look at this from a testing perspective please?

helm/genai-toolbox/Chart.yaml
kurtisvg: My knowledge of helm charts is slightly dated, but I thought they used to be linked to the version of the repo they were in (or at least the commit). How does a user select an older version if it's not through some tag on github?

helm/genai-toolbox/values.yaml
kurtisvg: I'm not sure I follow. If the helm chart is deploying toolbox, they env vars need to be passed in where toolbox is running in order to be substituted. Doesn't that mean we need an `env` section or something that allows the user to provide env variables that we mount in the toolbox deployment?

kurtisvg: (or maybe more securely, a map of secrets that should be mounted as env variables?)

helm/genai-toolbox/templates/deployment.yaml
```diff
@@ -0,0 +1,110 @@
...
+            "{{ .Values.options.telemetry_service_name }}",
+            {{- end }}
+          ]
+          ports:
+            - containerPort: {{ .Values.options.port }}
+          # volumeMounts:
+          #   - name: {{ .Values.name }}-config
+          #     mountPath: "/app/tools.yaml"
+          #     subPath: tools.yaml
+          #     readOnly: true
```
kurtisvg, line 73: nit: should this be closer to where volumeMounts is below?

helm/genai-toolbox/values.yaml
kurtisvg: I'm not sure I follow. Can't we do something like:

```
# values.yaml

# The name of the existing Kubernetes secret to use for this deployment.
config-secret: my-tools-file-secret
```

```
# deployment.yaml
  containers:
        -  # ... other container configuration ...
          volumeMounts:
            - name: secret-volume
              mountPath: "/etc/tools.yaml"
              readOnly: true
      volumes:
        - name: secret-volume
          secret:
            secretName: {{ .Values.config-secret }}
```

helm/genai-toolbox/values.yaml
kurtisvg: Something like:

```
# values.yaml
secrets:
    var1: secret1
    var2: secret2
```

then:
```
# templates/deployment.yaml

# ... (inside the container spec)
          env:
          {{- range $key, $value := .Values.secrets }}
            - name: {{ $key | quote }}
              valueFrom:
                secretKeyRef:
                  name: {{ $value | quote }}
                  key: {{ $key | quote }}
          {{- end }}
```

Issue #661, opened 2025-06-05
[http] allow http tools to severs with self-signed certs 


alessandro-festa:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### Question

I did not find anything in the documentation, so I'm posting here the question:
Let say I have an HTTP source that has a self-signed certifcate (i.e.: for developing purposes) and I want to use that source as tool.
Is there a way to pass any parameter to ingnore the selfsigned certificate ? Something similar to:
```
sources:
    my-http-source:
        kind: http
        baseUrl: https://api.mydomain.com
        insecureHTTP: true
```

duwenxin99:
Hi @alessandro-festa, you can now disable SSL verification by setting the `disableSslVerification` field to `true`:

```
sources:
    my-http-source:
        kind: http
        baseUrl: https://api.mydomain.com
        disableSslVerification: true
```

Let me know if you get further questions. Thanks for your input!

PR #674, opened 2025-06-06
feat: Support disable SSL verification for HTTP Source

duwenxin99:
Fix issue: https://github.com/googleapis/genai-toolbox/issues/661

### DIFF
--- a/internal/sources/http/http.go
+++ b/internal/sources/http/http.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/sources"
+	"github.com/googleapis/genai-toolbox/internal/util"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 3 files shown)


Yuan325 (approved):

internal/sources/http/http.go
```diff
@@ -63,8 +65,19 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
 		return nil, fmt.Errorf("unable to parse Timeout string as time.Duration: %s", err)
 	}
+
+	tr := &http.Transport{}
+
+	if r.DisableSslVerification {
+		tr.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true,
+		}
+		fmt.Printf("WARNING: InsecureHTTP is enabled for HTTP source %s. TLS certificate verification is skipped.\n", r.Name)
```
Yuan325: iirc, we should be able to extract the logger from ctx using https://github.com/googleapis/genai-toolbox/blob/29aa0a70da3c2eb409a38993b3782da8bec7cb85/internal/util/util.go#L90. The logger is added to the ctx in `cmd/root.go`. Instead of printing, we can do a warning log~

PR #677, opened 2025-06-06
Feature/databricks sql driver

jeffreyrubi:
Add databricks SQL Source and Tools to the Toolbox.


kurtisvg:
Hey @jeffreyrubi -- looks like you are working on support for databricks. Can I recommend starting by opening an issue? That'll help us collect interest and you can signal to other folks you are working on something to avoid potential duplication.

PR #683, opened 2025-06-09
refactor: move config initialization out of NewServer

AlexTalreja:
Move config initialization (sources, authServices, tools, and toolsets) from within NewServer to a separate function, to allow reuse within dynamic reloading.


Yuan325 (commented):
LGTM.

PR #691, opened 2025-06-10
feat(tools/spanner): add templateParameters field for spanner

Yuan325:
Add `templateParameters` to support non-filter parameters and DDL statements.

Added a new argument `ignoreDdl` at integration test. Admin client is needed to execute ddl statement in spanner. Toolbox does not use admin client.

Part of #535


mquinteiro:
@Yuan325 thanks Yuan, what is the placeholder for tamplateParameter? is it **SQL_STATEMETN** as I can see in the samples/test?
this this definition:
  describe-table:
    kind: mysql-sql
    source: my-mysql-source
    description: Comand to describe a database table, it will show the db fields and types it is like mysql command desc table_name
    templateParameters:
      - name: table_name
        type: string
        description: The name of the table to describe
    statement: desc SQL_STATEMENT;

I got:
"unable to execute query: Error 1146 (42S02): Table 'mydatabase.SQL_STATEMENT' doesn't exist"

toolbox version 0.7.0+binary.linux.amd64.714d990c34ee990e268fac1aa6b89c4883ae5023

Yuan325:
@mquinteiro Thanks for the question! templateParameter docs is [here](https://googleapis.github.io/genai-toolbox/resources/tools/#template-parameters). It is used to replace parameters that is not part of the regular `Parameters`.

For example, 
```
tools:
 select_columns_from_table:
    kind: postgres-sql
    source: my-pg-instance
    statement: |
      SELECT * FROM {{.tableName}}
    description: |
      Use this tool to list all information from a specific table.
    templateParameters:
      - name: tableName
        type: string
        description: Table to select from
```

If you would like to use sql_statement as an input, please refer to the `mysql-execute-sql` tool ([more information here](https://googleapis.github.io/genai-toolbox/resources/tools/mysql-execute-sql/)). We also have similar tool for other databases such as: `postgres-execute-sql`, `bigquery-execute-sql`, `mssql-execute-sql`, `spanner-execute-sql`.

mquinteiro:
Great! I have to say that it have more sense than $1 or much more than a just a ?  in the mysql case.
Thanks for your time.

averikitsch (approved):

PR #692, opened 2025-06-10
feat(tools/bigtable): add templateParameters field for bigtable

Yuan325:
Add templateParameters to support non-filter parameters and DDL statements.

Added a new argument `ignoreInsert` at integration test. Bigtable only allow `SELECT` statement. This is used to filter insert statement for bigtable.

Part of #535


internal/tools/spanner/spanner.go
averikitsch, line 1: Looks like this PR needs to be rebased on main

PR #693, opened 2025-06-10
feat: validate reloaded edits to tools file

AlexTalreja:
Re-parse and initialize configs after an edit is detected in the tools file.

Any parsing/initialization errors are logged as warnings to prevent crashing the existing toolbox server.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -226,17 +226,35 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
 	return toolsFile, nil
 }
 
-func validateReloadEdits(ctx context.Context, toolsFile ToolsFile, logger log.Logger) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error) {
-	logger.DebugContext(ctx, "Attempting to parse and validate reloaded tools file.")
-
-	instrumentation, err := server.CreateTelemetryInstrumentation(versionString)
+// validateReloadEdits checks that the reloaded tools file configs can initialized without failing
+func validateReloadEdits(
+	ctx context.Context, toolsFile ToolsFile,
+) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error,
+) {
+	logger, err := util.LoggerFromContext(ctx)
 	if err != nil {
-		errMsg := fmt.Errorf("unable to create telemetry instrumentation for reload: %w", err)
-		logger.WarnContext(ctx, errMsg.Error())
-		return nil, nil, nil, nil, err
+		panic(err)
 	}
 
-	sourcesMap, authServicesMap, toolsMap, toolsetsMap, err := server.InitializeConfigs(ctx, versionString, toolsFile.Sources, toolsFile.AuthServices, toolsFile.Tools, toolsFile.Toolsets, logger, instrumentation)
+	instrumentation, err := util.InstrumentationFromContext(ctx)
+	if err != nil {
+		panic(err)
+	}
+
+	logger.DebugContext(ctx, "Attempting to parse and validate reloaded tools file.")
+
+	ctx, span := instrumentation.Tracer.Start(ctx, "toolbox/server/reload")
+	defer span.End()
+
+	reloadedConfig := server.ServerConfig{
+		Version:            versionString,
+		SourceConfigs:      toolsFile.Sources,
+		AuthServiceConfigs: toolsFile.AuthServices,
+		ToolConfigs:        toolsFile.Tools,
+		ToolsetConfigs:     toolsFile.Toolsets,
+	}
+
+	sourcesMap, authServicesMap, toolsMap, toolsetsMap, err := server.InitializeConfigs(ctx, reloadedConfig)
 	if err != nil {
 		errMsg := fmt.Errorf("unable to initialize reloaded configs: %w", err)
 		logger.WarnContext(ctx, errMsg.Error())
@@ -246,18 +264,11 @@ func validateReloadEdits(ctx context.Context, toolsFile ToolsFile, logger log.Lo
 	return sourcesMap, authServicesMap, toolsMap, toolsetsMap, nil
 }
 
-func updateServer(ctx context.Context, l log.Logger, sourcesMap map[string]sources.Source, authServicesMap map[string]auth.AuthService, toolsMap map[string]tools.Tool, toolsetsMap map[string]tools.Toolset) error {
-	l.DebugContext(ctx, "Attempting to update the server with reloaded configs")
-
-	// TODO: handle updating logic
-	return nil
-}
-
 // watchFile checks for changes in the provided yaml tools file.
 func watchFile(ctx context.Context, toolsFileName string) {
 	logger, err := util.LoggerFromContext(ctx)
 	if err != nil {
-		panic(fmt.Errorf("unable to extract logger from context %w", err))
+		panic(err)
 	}
 
 	w, err := fsnotify.NewWatcher()
@@ -303,7 +314,7 @@ func watchFile(ctx context.Context, toolsFileName string) {
 			}
 
 			if e.Op == fsnotify.Write && filepath.Clean(e.Name) == cleanedFilename {
-				logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, e.Name))
+				logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, cleanedFilename))
 				debounce.Reset(debounceDelay)
 			}
 		case <-debounce.C:
@@ -313,28 +324,22 @@ func watchFile(ctx context.Context, toolsFileName string) {
 			if err != nil {
 				errMsg := fmt.Errorf("unable to read reloaded tools file at %q: %w", toolsFileName, err)
 				logger.WarnContext(ctx, errMsg.Error())
-				return
+				continue
 			}
 
 			toolsFile, err := parseToolsFile(ctx, buf)
 			if err != nil {
 				errMsg := fmt.Errorf("unable to parse reloaded tools file at %q: %w", toolsFileName, err)
 				logger.WarnContext(ctx, errMsg.Error())
-				return
+				continue
 			}
 
-			sourcesMap, authServicesMap, toolsMap, toolsetsMap, err := validateReloadEdits(ctx, toolsFile, logger)
+			// TODO: will update when updateServer() function is added to use return values
+			_, _, _, _, err = validateReloadEdits(ctx, toolsFile)
 			if err != nil {
 				errMsg := fmt.Errorf("unable to validate reloaded edits: %w", err)
 				logger.WarnContext(ctx, errMsg.Error())
-				return
-			}
-
-			err = updateServer(ctx, logger, sourcesMap, authServicesMap, toolsMap, toolsetsMap)
-			if err != nil {
-				errMsg := fmt.Errorf("unable to update server after reload: %w", err)
-				logger.WarnContext(ctx, errMsg.Error())
-				return
+				continue
 			}
 		}
 	}
@@ -463,8 +468,17 @@ func run(cmd *Command) error {
 		return errMsg
 	}
 
+	instrumentation, err := telemetry.CreateTelemetryInstrumentation(versionString)
+	if err != nil {
+		errMsg := fmt.Errorf("unable to create telemetry instrumentation: %w", err)
+		cmd.logger.ErrorContext(ctx, errMsg.Error())
+		return errMsg
+	}
+
+	ctx = util.WithInstrumentation(ctx, instrumentation)
+
 	// start server
-	s, err := server.NewServer(ctx, cmd.cfg, cmd.logger)
+	s, err := server.NewServer(ctx, cmd.cfg)
 	if err != nil {
 		errMsg := fmt.Errorf("toolbox failed to initialize: %w", err)
 		cmd.logger.ErrorContext(ctx, errMsg.Error())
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -223,11 +226,31 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
 
-func parse(ctx context.Context, buf []byte, logger log.Logger) {
-	logger.DebugContext(ctx, "Attempting to parse updated tools file.")
-	// time.Sleep(5 * time.Second)
-	// logger.DebugContext(ctx, "Parse finished.")
-	// TODO: add logic for parsing
+func validateReloadEdits(ctx context.Context, toolsFile ToolsFile, logger log.Logger) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error) {
+	logger.DebugContext(ctx, "Attempting to parse and validate reloaded tools file.")
+
+	instrumentation, err := server.CreateTelemetryInstrumentation(versionString)
```
Yuan325, line 232: ideally we wanna use the existing telemetry instrumentation.

We set this in Server. https://github.com/googleapis/genai-toolbox/blob/d51dbc759ba493021d3ec6f5417fc04c21f7044f/internal/server/server.go#L46

Let's add the instrumentation into the context. In `internal/util/util.go`, we can create a new `const instrumentationKey contextKey = "instrumentation"` https://github.com/googleapis/genai-toolbox/blob/d51dbc759ba493021d3ec6f5417fc04c21f7044f/internal/util/util.go#L82 and add the `WithInstrumentation` and `InstrumentationFromContext` functions.

cmd/root.go
```diff
@@ -223,11 +226,31 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+	if err != nil {
+		errMsg := fmt.Errorf("unable to create telemetry instrumentation for reload: %w", err)
+		logger.WarnContext(ctx, errMsg.Error())
+		return nil, nil, nil, nil, err
+	}
+
```
Yuan325, line 243: Once we have the instrumentation, we want to add a parent trace before parsing into initialize config https://github.com/googleapis/genai-toolbox/blob/d51dbc759ba493021d3ec6f5417fc04c21f7044f/internal/server/server.go#L62

cmd/root.go
```diff
@@ -288,10 +311,31 @@ func watchFile(ctx context.Context, toolsFileName string) {
...
+			if err != nil {
+				errMsg := fmt.Errorf("unable to validate reloaded edits: %w", err)
+				logger.WarnContext(ctx, errMsg.Error())
+				return
+			}
+
+			err = updateServer(ctx, logger, sourcesMap, authServicesMap, toolsMap, toolsetsMap)
+			if err != nil {
+				errMsg := fmt.Errorf("unable to update server after reload: %w", err)
+				logger.WarnContext(ctx, errMsg.Error())
```
Yuan325: let's remove this from here since this is technically added in the next PR.

cmd/root.go
```diff
@@ -223,11 +226,31 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+
+	return sourcesMap, authServicesMap, toolsMap, toolsetsMap, nil
+}
+
+func updateServer(ctx context.Context, l log.Logger, sourcesMap map[string]sources.Source, authServicesMap map[string]auth.AuthService, toolsMap map[string]tools.Tool, toolsetsMap map[string]tools.Toolset) error {
+	l.DebugContext(ctx, "Attempting to update the server with reloaded configs")
+
+	// TODO: handle updating logic
+	return nil
 }
```
Yuan325: let's remove this from here since this is technically added in the next PR.

Yuan325 (changes requested):

cmd/root.go
```diff
@@ -288,10 +300,25 @@ func watchFile(ctx context.Context, toolsFileName string) {
 			logger.DebugContext(ctx, "re-reading tools file: %s", cleanedFilename)
 			buf, err := os.ReadFile(toolsFileName)
 			if err != nil {
-				logger.WarnContext(ctx, "error reading reloaded file", err)
+				errMsg := fmt.Errorf("unable to read reloaded tools file at %q: %w", toolsFileName, err)
+				logger.WarnContext(ctx, errMsg.Error())
+				return
```
Yuan325: Within `case <- debounce.C`, if it hits an error, we might not want to return? If return, it means that the `watchFile()` goroutine will stop. In this case, we might just want to ignore the changes. we might wanna just do `break` instead?

Same with the other 2 returns within `if err!= nil`

internal/server/server.go
```diff
@@ -52,10 +52,12 @@ type Server struct {
 	toolsets     map[string]tools.Toolset
 }
 
-func InitializeConfigs(ctx context.Context, cfg ServerConfig, l log.Logger, instrumentation *Instrumentation) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error) {
+func InitializeConfigs(ctx context.Context, VersionIn string, SourceConfigsIn SourceConfigs, AuthServiceConfigsIn AuthServiceConfigs, ToolConfigsIn ToolConfigs, ToolsetConfigsIn ToolsetConfigs, l log.Logger, instrumentation *Instrumentation) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error) {
```
kurtisvg: nit: this line is past readable -- suggest splitting it up over a few lines. 

It's probably also time to use a struct to pass them in as well.

internal/server/server.go
```diff
@@ -162,17 +164,10 @@ func InitializeConfigs(ctx context.Context, cfg ServerConfig, l log.Logger, inst
 }
 
 // NewServer returns a Server object based on provided Config.
-func NewServer(ctx context.Context, cfg ServerConfig, l log.Logger) (*Server, error) {
-	instrumentation, err := CreateTelemetryInstrumentation(cfg.Version)
-	if err != nil {
-		return nil, fmt.Errorf("unable to create telemetry instrumentation: %w", err)
-	}
-
+func NewServer(ctx context.Context, cfg ServerConfig, l log.Logger, instrumentation *Instrumentation) (*Server, error) {
```
kurtisvg: Why does instrumentation need to be passed in here?

Should  logger / instrumentation be moved into the context?

kurtisvg (changes requested):

internal/util/util.go
```diff
@@ -93,3 +94,18 @@ func LoggerFromContext(ctx context.Context) (log.Logger, error) {
...
+func WithInstrumentation(ctx context.Context, instrumentation *telemetry.Instrumentation) context.Context {
+	return context.WithValue(ctx, instrumentationKey, instrumentation)
+}
+
+// InstrumentationFromContext retreives the instrumentation or return an error
+func InstrumentationFromContext(ctx context.Context) (*telemetry.Instrumentation, error) {
+	if instrumentation, ok := ctx.Value(instrumentationKey).(*telemetry.Instrumentation); ok {
+		return instrumentation, nil
+	}
+	return nil, fmt.Errorf("unable to retrieve logger")
```
Yuan325:
```suggestion
	return nil, fmt.Errorf("unable to retrieve instrumentation")
```

internal/util/util.go
```diff
@@ -93,3 +94,18 @@ func LoggerFromContext(ctx context.Context) (log.Logger, error) {
...
 }
+
+const instrumentationKey contextKey = "instrumentation"
+
+// WithInstrumentation adds an instrumentation into the context as a value
+func WithInstrumentation(ctx context.Context, instrumentation *telemetry.Instrumentation) context.Context {
+	return context.WithValue(ctx, instrumentationKey, instrumentation)
+}
+
+// InstrumentationFromContext retreives the instrumentation or return an error
```
Yuan325:
```suggestion
// InstrumentationFromContext retrieves the instrumentation or return an error
```

cmd/root.go
```diff
@@ -223,15 +226,45 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
-	// time.Sleep(5 * time.Second)
-	// logger.DebugContext(ctx, "Parse finished.")
-	// TODO: add logic for parsing
+func validateReloadEdits(
+	ctx context.Context, toolsFile ToolsFile,
+) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error,
+) {
+	logger, err := util.LoggerFromContext(ctx)
+	if err != nil {
+		panic("attempted to parse logger when not in context")
```
Yuan325: Is there any reason to throw panic? Will it work to return an error instead? (`return nil, nil, nil, nil, err`)

cmd/root.go
```diff
@@ -223,15 +226,45 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+	}
+
+	instrumentation, err := util.InstrumentationFromContext(ctx)
+	if err != nil {
+		panic("attempted to parse instrumentation when not in cotexet")
```
Yuan325: same with this. return err instead?

cmd/root.go
```diff
@@ -223,15 +226,45 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+		logger.WarnContext(ctx, errMsg.Error())
+		return nil, nil, nil, nil, err
+	}
+
+	return sourcesMap, authServicesMap, toolsMap, toolsetsMap, nil
 }
 
 // watchFile checks for changes in the provided yaml tools file.
-func watchFile(ctx context.Context, toolsFileName string) {
+func watchFile(ctx context.Context, toolsFileName string, instrumentation *telemetry.Instrumentation) {
```
```suggestion
func watchFile(ctx context.Context, toolsFileName string) {
```

seems like we won't need instrumentation anymore? since we can grab it from ctx.

cmd/root.go
```diff
@@ -223,15 +226,45 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
 	return toolsFile, nil
 }
 
-func parse(ctx context.Context, buf []byte, logger log.Logger) {
-	logger.DebugContext(ctx, "Attempting to parse updated tools file.")
-	// time.Sleep(5 * time.Second)
-	// logger.DebugContext(ctx, "Parse finished.")
-	// TODO: add logic for parsing
+func validateReloadEdits(
```
Yuan325, line 230: can we add a comment for this function?

cmd/root.go
```diff
@@ -456,7 +513,7 @@ func run(cmd *Command) error {
 	}
 
 	// start watching for file changes to trigger dynamic reloading
-	go watchFile(ctx, cmd.tools_file)
+	go watchFile(ctx, cmd.tools_file, instrumentation)
```
```suggestion
	go watchFile(ctx, cmd.tools_file)
```

cmd/root_test.go
```diff
@@ -944,12 +946,20 @@ func TestSingleEdit(t *testing.T) {
...
-	go watchFile(ctx, fileToWatch)
+	instrumentation, err := telemetry.CreateTelemetryInstrumentation(versionString)
+	if err != nil {
+		t.Fatalf("failed to setup instrumentation %s", err)
+	}
+
+	go watchFile(ctx, fileToWatch, instrumentation)
+
+	// escape backslash so regex doesn't fail on windows filepaths
+	regexEscapedPath := strings.ReplaceAll(fileToWatch, `\`, `\\\\*\\`)
```
Yuan325, line 953: What's the expected path for `fileToWatch`?

internal/server/server.go
```diff
@@ -52,10 +53,27 @@ type Server struct {
 	toolsets     map[string]tools.Toolset
 }
 
-func InitializeConfigs(ctx context.Context, cfg ServerConfig, l log.Logger, instrumentation *Instrumentation) (map[string]sources.Source, map[string]auth.AuthService, map[string]tools.Tool, map[string]tools.Toolset, error) {
+func InitializeConfigs(ctx context.Context, serverConfig ServerConfig) (
```
Yuan325: why do we want to change the name to serverConfig? can we keep `cfg`?

cmd/root.go
AlexTalreja: I think earlier we discussed it might be preferred to panic() if we fail to get the logger (https://github.com/googleapis/genai-toolbox/pull/670#discussion_r2145991739) since that would indicate more problems.

Although that was during execution of watchFile(), would we want a different result here?

cmd/root_test.go
AlexTalreja: It varies based on the OS and env, so for example
- gLinux: `/tmp/243880375`
- windows: `C:\\Users\\alextalreja\\AppData\\Local\\Temp\\4259873817`

The actual file creation is just from the existing helper func `tmpFileWithCleanup()`.

cmd/root.go
```suggestion
		panic(err)
```
since the error that was return is pretty straight forward, we can just use it.

cmd/root.go
```suggestion
		panic(err)
```

cmd/root.go
```diff
@@ -223,15 +226,45 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+
+	return sourcesMap, authServicesMap, toolsMap, toolsetsMap, nil
 }
 
 // watchFile checks for changes in the provided yaml tools file.
-func watchFile(ctx context.Context, toolsFileName string) {
+func watchFile(ctx context.Context, toolsFileName string, instrumentation *telemetry.Instrumentation) {
 	logger, err := util.LoggerFromContext(ctx)
 	if err != nil {
 		panic(fmt.Errorf("unable to extract logger from context %w", err))
```
```suggestion
		panic(err)
```

Yuan325 (approved):

PR #699, opened 2025-06-11
feat(tools/bigquery): add templateParameters field for bigquery

Yuan325:


tests/tool.go
```diff
@@ -214,17 +214,22 @@ func RunToolInvokeTest(t *testing.T, select_1_want, invoke_param_want string) {
 // RunToolInvokeWithTemplateParameters runs tool invoke test cases with template parameters.
 // ignoreDdl is used for sources that does not support DDL statement.
 // replaceNameFieldArray and replaceNameColFilter is used for bigtable since it have a different formatting for sql statement.
+// createTableCol is used for bigquery since it have a different type for columns.
 // ignoreInsert is used for bigtable since it does not support other DML statement other than `SELECT`.
-func RunToolInvokeWithTemplateParameters(t *testing.T, tableName, select_all_want, select_only_1_want, replaceNameFieldArray, replaceNameColFilter string, ignoreDdl, ignoreInsert bool) {
+func RunToolInvokeWithTemplateParameters(t *testing.T, tableName, select_all_want, select_only_1_want, replaceNameFieldArray, replaceNameColFilter, createTableCol string, ignoreDdl, ignoreInsert bool) {
```
averikitsch: I like that we can extend this function to allow for customization, but it is going to be harder to use with more customizations and passing in empty strings. I don't have a solution right now, but something we may want to think about in the future.

Yuan325: Sounds good, yeah probably something I'll have to clean up~

kurtisvg: Time to add options?

Yuan325: will update to use options in the upcoming PR~

PR #719, opened 2025-06-16
fix(server/mcp): do not listen from port for stdio

### LINKED ISSUE #711: Do not start server for stdio
### Prerequisites

### Toolbox version

0.7.0

### Environment

1. OS type and version: Darwin ARM64
  - As a downloaded binary 

### Client

_No response_

### Expected Behavior

When I use --stdio no port is opened

### Current Behavior

When I use --stdio, it also starts Toolbox on 127.0.0.1:5000

### Steps to reproduce?

1. `./toolbox --prebuilt alloydb-postgres --stdio`
2. `sudo lsof -i -n -P | grep 5000`

### Additional Details

This also makes it difficult to run multiple instances of Toolbox.

Yuan325:
Fix #711

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -352,7 +352,17 @@ func run(cmd *Command) error {
 		return errMsg
 	}
 
-	if !cmd.cfg.Stdio {
+	// run server in background
+	srvErr := make(chan error)
+	if cmd.cfg.Stdio {
+		go func() {
+			defer close(srvErr)
+			err = s.ServeStdio(ctx, cmd.inStream, cmd.outStream)
+			if err != nil {
+				srvErr <- err
+			}
+		}()
+	} else {
 		err = s.Listen(ctx)
 		if err != nil {
 			errMsg := fmt.Errorf("toolbox failed to start listener: %w", err)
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -352,13 +352,15 @@ func run(cmd *Command) error {
 		return errMsg
 	}
 
-	err = s.Listen(ctx)
-	if err != nil {
-		errMsg := fmt.Errorf("toolbox failed to start listener: %w", err)
-		cmd.logger.ErrorContext(ctx, errMsg.Error())
-		return errMsg
+	if !cmd.cfg.Stdio {
```
kurtisvg: Can we combine this and the serve in background block below? 

Specifically I think it's good to try and have a single "if std else server http" structure to simplify layout

cmd/root.go
```diff
@@ -352,30 +352,33 @@ func run(cmd *Command) error {
...
-	}
-	cmd.logger.InfoContext(ctx, "Server ready to serve!")
-
 	// run server in background
 	srvErr := make(chan error)
-	go func() {
-		defer close(srvErr)
-		if cmd.cfg.Stdio {
-			err = s.ServeStdio(ctx, cmd.inStream, cmd.outStream)
+	if !cmd.cfg.Stdio {
```
kurtisvg: nit: Should have the "default" path at the end -- so put stdio first and HTTP second. 

This is a small readability improvement -- it lets you read the code as "if stdio" as opposed to "if not stdio", which doesn't really tell you what the block is then (e.g. you need to know that "not stdio" is http).

PR #721, opened 2025-06-16
feat: update toolbox server with reloaded configs

AlexTalreja:
Dynamically updates both the `server` and `ServerConfig` with the reloaded yaml configuration after successful parsing and validation.

Also modified `server` access patterns by adding a manager with RWMutex to ensure thread safety. However, did not update `serverConfig` as it's not used elsewhere after initialization.


internal/server/server.go
```diff
@@ -45,13 +45,52 @@ type Server struct {
...
 
+type ConfigManager struct {
+	mu           sync.RWMutex
 	sources      map[string]sources.Source
 	authServices map[string]auth.AuthService
 	tools        map[string]tools.Tool
 	toolsets     map[string]tools.Toolset
 }
 
+func (m *ConfigManager) getSource(sourceName string) sources.Source {
```
AlexTalreja: I added `getSource` when making the config manager so all getters were included, though currently it's not used anywhere. Should I just remove it or is it good practice to leave it in case it's needed in the future?

cmd/root_test.go
```diff
@@ -1131,3 +1141,230 @@ func TestUpdateLogLevel(t *testing.T) {
...
+			},
+		},
+		{
+			desc: "add to few configs",
+			cfg:  server.ServerConfig{},
+			toolsFile: ToolsFile{
+				Sources:      server.SourceConfigs{},
+				AuthServices: server.AuthServiceConfigs{},
+				Tools:        server.ToolConfigs{},
+				Toolsets: server.ToolsetConfigs{
```
Yuan325:
```suggestion
				Toolsets: server.ToolsetConfigs{
```
Should be able to work with nil values? since these aren't required.

cmd/root_test.go
```diff
@@ -1131,3 +1141,230 @@ func TestUpdateLogLevel(t *testing.T) {
...
+			toolsFile: ToolsFile{
+				Sources: server.SourceConfigs{
+					"test-source": cloudsqlpgsrc.Config{
+						Name: "test-source",
+					},
+				},
+				AuthServices: server.AuthServiceConfigs{},
+				Tools:        server.ToolConfigs{},
+				Toolsets:     server.ToolsetConfigs{},
+			},
```
Yuan325:
```suggestion
			},
```
Same with these~ and the other test cases as well.

internal/server/mcp.go
```diff
@@ -428,8 +428,8 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, toolsetName
 		}
 		toolName := req.Params.Name
 		toolArgument := req.Params.Arguments
-		logger.DebugContext(ctx, fmt.Sprintf("tool name: %s", toolName))
-		tool, ok := s.tools[toolName]
+		s.logger.DebugContext(ctx, fmt.Sprintf("tool name: %s", toolName))
```
Yuan325:
```suggestion
		logger.DebugContext(ctx, fmt.Sprintf("tool name: %s", toolName))
```

internal/server/server.go
```diff
@@ -46,13 +46,60 @@ type Server struct {
...
+	return toolset, ok
+}
+
+func (m *ConfigManager) SetConfigs(sourcesMap map[string]sources.Source, authServicesMap map[string]auth.AuthService, toolsMap map[string]tools.Tool, toolsetsMap map[string]tools.Toolset) {
+	m.mu.Lock()
+	m.sources = sourcesMap
+	m.authServices = authServicesMap
+	m.tools = toolsMap
+	m.toolsets = toolsetsMap
+	m.mu.Unlock()
```
```suggestion
	m.mu.Lock()
	defer m.mu.Unlock()
	m.sources = sourcesMap
	m.authServices = authServicesMap
	m.tools = toolsMap
	m.toolsets = toolsetsMap
```

internal/server/server.go
```diff
@@ -273,6 +324,18 @@ func NewServer(ctx context.Context, cfg ServerConfig) (*Server, error) {
 	return s, nil
 }
 
+func UpdateServer(ctx context.Context, s *Server, sourcesMap map[string]sources.Source, authServicesMap map[string]auth.AuthService, toolsMap map[string]tools.Tool, toolsetsMap map[string]tools.Toolset) error {
```
Yuan325: why do we need a separate function? can we just call `s.ConfigManager.SetConfigs()` directly?

cmd/root.go
```diff
@@ -226,6 +226,38 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+	}
+
+	err = updateCfg(ctx, toolsFile, cfg)
+	if err != nil {
+		errMsg := fmt.Errorf("unable to update server after reload: %w", err)
+		logger.WarnContext(ctx, errMsg.Error())
+		return err
+	}
+
+	s.ConfigManager.SetConfigs(sourcesMap, authServicesMap, toolsMap, toolsetsMap)
```
kurtisvg: It's a little confusing that we are setting configs in two different places. Can we instead just set it in one place instead?

Ideally we want to have a single source of truth -- we don't want folks mutating something in cfg and it not getting updated here.

internal/server/server_test.go
```diff
@@ -102,3 +108,67 @@ func TestServe(t *testing.T) {
...
+		t.Errorf("error updating server, got: %s, want: %s", gotSource, newSources["example-auth"])
+	}
+
+	gotTool, _ := s.ResourceManager.GetTool("example-tool")
+	if gotTool != newTools["example-tool"] {
+		t.Errorf("error updating server, got: %s, want: %s", gotSource, newSources["example-tool"])
+	}
+
+	gotToolset, _ := s.ResourceManager.GetToolset("example-toolset")
+	if diff := cmp.Diff(gotToolset, newToolsets["example-toolset"]); diff != "" {
```
Yuan325, line 171: Why do we use `cmp.Diff` for this but direct comparison for others? Might be better to keep it consistent.

Yuan325 (approved):
nit, otherwise LGTM.

internal/server/server_test.go
AlexTalreja: Fixed. Originally `cmp.Diff` was only used for toolset struct since it couldn't be compared via != while the others could (nil and pointer), but I updated them all for consistency, plus it seems `cmp.Diff` is standard to use throughout other tests (ex: root_test.go)

cmd/root.go
AlexTalreja: Note: As discussed, we have updated the code so that Server handles it's own state through `resourceManager`, which acts as the single source of truth.

Cfg is only during creation of the new server, and will not updated or referenced later.

Issue #701, opened 2025-06-12
Error when using parameters of type array


ngocquang2797:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

toolbox version 0.7.0

### Environment

1. OS=darwin/arm64

### Client

1. Client: <name and link to the client are you using>
2. Version: <what exact version of the client are you using> 
3. Example: If possible, please include your code of configuration:

```python
# Code goes here! 
```

### Expected Behavior

get "MCP error -32001: Error POSTing to endpoint (HTTP 500): "

### Current Behavior

```
panic: runtime error: hash of unhashable type []interface {}

 -> github.com/googleapis/genai-toolbox/internal/tools.ParamValues.AsReversedMap
 ->   /workspace/internal/tools/parameters.go:69

    github.com/googleapis/genai-toolbox/internal/tools/bigquery.Tool.Invoke
      /workspace/internal/tools/bigquery/bigquery.go:122
    github.com/googleapis/genai-toolbox/internal/server/mcp.ToolCall
      /workspace/internal/server/mcp/method.go:54
    github.com/googleapis/genai-toolbox/internal/server.processMcpMessage
      /workspace/internal/server/mcp.go:466
    github.com/googleapis/genai-toolbox/internal/server.httpHandler
      /workspace/internal/server/mcp.go:318
    github.com/googleapis/genai-toolbox/internal/server.mcpRouter.func2
      /workspace/internal/server/mcp.go:179
    net/http.HandlerFunc.ServeHTTP
      /usr/local/go/src/net/http/server.go:2294
    github.com/go-chi/chi/v5.(*Mux).routeHTTP
      /gopath/pkg/mod/github.com/go-chi/chi/v5@v5.2.1/mux.go:480
    net/http.HandlerFunc.ServeHTTP
      /usr/local/go/src/net/http/server.go:2294
    github.com/googleapis/genai-toolbox/internal/server.mcpRouter.SetContentType.func4.1
      /gopath/pkg/mod/github.com/go-chi/render@v1.0.3/content_type.go:52
    net/http.HandlerFunc.ServeHTTP
      /usr/local/go/src/net/http/server.go:2294
    github.com/go-chi/chi/v5/middleware.StripSlashes.func1
      /gopath/pkg/mod/github.com/go-chi/chi/v5@v5.2.1/middleware/strip.go:30
    net/http.HandlerFunc.ServeHTTP
[...]

PR #722, opened 2025-06-16
fix: Fix array param in BigQuery and MSSQL

duwenxin99:
Fix: https://github.com/googleapis/genai-toolbox/issues/701

Things done:
1. Replace the `AsReversedMap()` helper with `AsNameAndValueSlices()` because arrays cannot be hashed as keys
2. BigQuery's QueryParameter only accepts typed slices as input, but our arrays are passed in as []any. Therefore, add a logic to convert []any to a typed array based on the item type.

Tested on MCP inspector:
Screenshot 2025-06-16 at 5 15 55 PM

### DIFF
--- a/internal/tools/bigquery/bigquery.go
+++ b/internal/tools/bigquery/bigquery.go
@@ -120,8 +120,65 @@ type Tool struct {
 func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
 	namedArgs := make([]bigqueryapi.QueryParameter, 0, len(params))
 	paramNames, paramValues := params.AsNameAndValueSlices()
+
 	for i, name := range paramNames {
 		value := paramValues[i]
+		switch vType := value.(type) {
+
+		// BigQuery's QueryParameter only accepts typed slices as input
+		// This checks if the param is an array.
+		// If yes, convert []any to typed slice (e.g []string, []int)
+		case []any:
+			var itemType string
+			for _, p := range t.Parameters {
+				if name == p.GetName() {
+					itemType = p.McpManifest().Items.Type
+				}
+			}
+			switch itemType {
+			case "string":
+				typedSlice := make([]string, len(vType))
+				for j, item := range value.([]any) {
+					if s, ok := item.(string); ok {
+						typedSlice[j] = s
+					} else {
+						return nil, fmt.Errorf("parameter '%s': expected item at index %d to be string, got %T", name, j, item)
+					}
+				}
+				value = typedSlice
+			case "integer":
+				typedSlice := make([]int64, len(vType))
+				for j, item := range value.([]any) {
+					if i, ok := item.(int); ok {
+						typedSlice[j] = int64(i)
+					} else {
+						return nil, fmt.Errorf("parameter '%s': expected item at index %d to be integer, got %T", name, j, item)
+					}
+				}
+				value = typedSlice
+			case "float":
+				typedSlice := make([]float64, len(vType))
+				for j, item := range value.([]any) {
+					if f, ok := item.(float64); ok {
+						typedSlice[j] = f
+					} else {
+						return nil, fmt.Errorf("parameter '%s': expected item at index %d to be float, got %T", name, j, item)
+					}
+				}
+				value = typedSlice
+			case "boolean":
+				typedSlice := make([]bool, len(vType))
+				for j, item := range value.([]any) {
+					if b, ok := item.(bool); ok {
+						typedSlice[j] = b
+					} else {
+						return nil, fmt.Errorf("parameter '%s': expected item at index %d to be boolean, got %T", name, j, item)
+					}
+				}
+				value = typedSlice
+			}
+		}
+
 		if strings.Contains(t.Statement, "@"+name) {
 			namedArgs = append(namedArgs, bigqueryapi.QueryParameter{
 				Name:  name,
--- a/internal/tools/bigquery/bigquery.go
+++ b/internal/tools/bigquery/bigquery.go
@@ -119,17 +119,17 @@ type Tool struct {
 
 func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
 	namedArgs := make([]bigqueryapi.QueryParameter, 0, len(params))
-	paramsMap := params.AsReversedMap()
-	for _, v := range params.AsSlice() {
-		paramName := paramsMap[v]
-		if strings.Contains(t.Statement, "@"+paramName) {
+	paramNames, paramValues := params.AsNameAndValueSlices()
+	for i, name := range paramNames {
+		value := paramValues[i]
+		if strings.Contains(t.Statement, "@"+name) {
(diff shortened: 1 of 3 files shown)


internal/tools/mssqlsql/mssqlsql.go
```diff
@@ -138,16 +138,19 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
...
+	for i, name := range paramNames {
+		value := paramValues[i]
+		if strings.Contains(newStatement, "@"+name) {
+			namedArgs = append(namedArgs, sql.Named(name, value))
 		} else {
-			namedArgs = append(namedArgs, v)
+			namedArgs = append(namedArgs, value)
 		}
+
 	}
```
Yuan325:
```suggestion
		}
	}
```

internal/tools/mssqlsql/mssqlsql.go
```diff
@@ -138,16 +138,19 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
...
 	namedArgs := make([]any, 0, len(newParams))
-	newParamsMap := newParams.AsReversedMap()
-	// To support both named args (e.g @id) and positional args (e.g @p1), check if arg name is contained in the statement.
-	for _, v := range newParams.AsSlice() {
-		paramName := newParamsMap[v]
-		if strings.Contains(newStatement, "@"+paramName) {
-			namedArgs = append(namedArgs, sql.Named(paramName, v))
+	// To support both named args (e.g @id) and positional args (e.g @p1), check
+	// if arg name is contained in the statement.
+	paramNames, paramValues := newParams.AsNameAndValueSlices()
```
Yuan325: curious why don't we just use `newParams.AsMap()`, and loop through the map? Seems like that's simpler and it works as well.

internal/tools/bigquery/bigquery.go
```diff
@@ -119,17 +119,75 @@ type Tool struct {
 
 func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
 	namedArgs := make([]bigqueryapi.QueryParameter, 0, len(params))
-	paramsMap := params.AsReversedMap()
-	for _, v := range params.AsSlice() {
-		paramName := paramsMap[v]
-		if strings.Contains(t.Statement, "@"+paramName) {
+	paramNames, paramValues := params.AsNameAndValueSlices()
```
Yuan325: If we use `params.AsMap()` here.

we can do the for loop on `t.Parameter`. That way, we also can easily get the type by using `GetType()`.
```
paramsMap := params.AsMap()

for _, p := range t.Parameter {
    name := p.GetName()
    value := paramsMap[name]
    if p.GetType() == "array" {
        typedSlice := getTypedSlice(p.GetItem().GetType())
    }
}
```

Wondering if it's worth to add a `GetItem()` to `type Parameter interface{}` with other common parameter returning `nil` for that except for array parameter. Might be better than depending on the manifest?

Yuan325 (changes requested):

internal/tools/mssqlsql/mssqlsql.go
duwenxin99: We want to preserve the order of the parameters because the user can use the [order notations](https://github.com/googleapis/genai-toolbox/blob/main/docs/en/resources/tools/mssql-sql.md#about) instead of the names (e.g @p1, @p2).

internal/tools/bigquery/bigquery.go
duwenxin99: I thought about it, but so far this is the only place that will be using item type so I'm thinking it maybe not be necessary, considering this method will return nil for all other param types.

Yuan325: Got it, maybe worth to at least move the whole typed slice to a separate function to simplify this function a little.

duwenxin99: That I tried as well, until I realized that helper function has to output a type []any, which diminished the whole point of this code...

Yuan325: if you output to any, it should retain the typing. https://go.dev/play/p/K2-y8cOPa0b

duwenxin99: True! Updated accordingly

Yuan325 (approved):
Please add test in parameter_test.go before merging, otherwise LGTM.

internal/tools/bigquery/bigquery.go
```diff
@@ -119,17 +119,38 @@ type Tool struct {
...
+
+	for i, name := range paramNames {
+		value := paramValues[i]
+
+		// BigQuery's QueryParameter only accepts typed slices as input
+		// This checks if the param is an array.
+		// If yes, convert []any to typed slice (e.g []string, []int)
+		switch arrayParam := value.(type) {
+		case []any:
```
kurtisvg, line 143: nit: why not just if value.(type) == []any?

internal/tools/bigquery/bigquery.go
```diff
@@ -119,17 +119,38 @@ type Tool struct {
...
+			var itemType string
+			var err error
+			for _, p := range t.Parameters {
+				// iterate through parameters to get array's item type
+				if name == p.GetName() {
+					itemType = p.McpManifest().Items.Type
+				}
+			}
```
kurtisvg: Why do we need to iterate through the array? Can't we just grab the first one?

internal/tools/bigquery/bigquery.go
```diff
@@ -178,3 +199,47 @@ func (t Tool) McpManifest() tools.McpManifest {
...
+		}
+	case "integer":
+		typedSlice := make([]int64, len(s))
+		for j, item := range s {
+			if i, ok := item.(int); ok {
+				typedSlice[j] = int64(i)
+			} else {
+				return nil, fmt.Errorf("parameter '%s': expected item at index %d to be integer, got %T", paramName, j, item)
+			}
```
kurtisvg: nit: this would be better formatter as a sentry clause:

```suggestion
			i, ok := item.(int);
			if  !ok {
			    return nil, fmt.Errorf("parameter '%s': expected item at index %d to be integer, got %T", paramName, j, item)
			} 
			typedSlice[j] = int64(i)
```

internal/tools/bigquery/bigquery.go
duwenxin99: it needs to find the parameter that matches the `name` we are looking for. We are looping through the parameters list of the tool, not the specific array value itself

internal/tools/bigquery/bigquery.go
duwenxin99: Go only allows type to be checked using a switch: https://go.dev/ref/spec#Type_switches

ngocquang2797:
Screenshot 2025-07-03 at 14 29 47
I face a different issue when using parrameter array

kurtisvg:
@ngocquang2797 can you please open an issue for us? We'll need more information so we can reproduce and fix. Thank you!

PR #747, opened 2025-06-24
chore: error response when client sent batch requests

Yuan325:
Error with "not supporting batch requests" message when client send batch request.

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -409,7 +409,7 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
 		unmarshalErr := json.Unmarshal(body, &a)
 		if unmarshalErr == nil {
 			err = fmt.Errorf("not supporting batch requests")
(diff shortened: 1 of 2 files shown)


internal/server/mcp.go
```diff
@@ -403,6 +403,15 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
 	if err = util.DecodeJSON(bytes.NewBuffer(body), &baseMessage); err != nil {
 		// Generate a new uuid if unable to decode
 		id := uuid.New().String()
+
+		// check if user is sending a batch request
+		var a []any
+		unmarshalErr := json.Unmarshal(body, &a)
+		if unmarshalErr == nil {
+			err = fmt.Errorf("not supporting batch requests")
+			return "", jsonrpc.NewError(id, jsonrpc.PARSE_ERROR, err.Error(), nil), err
```
kurtisvg: PARSE_ERROR is really the right response here? I would expect "Invalid Request" or "Invalid params"?

Yuan325: Updated to invalid requests since its more accurate than invalid params.

PR #757, opened 2025-06-26
fix(tools/mysqlexecutesql): Handle nil panic and connection leak in Invoke

megatron0000:
The Invoke function had two bugs:

1. A panic would occur when scanning a row containing a NULL value in a TEXT or VARCHAR column. The code did not check for nil before attempting a type assertion on the scanned value.

2. The *sql.Rows result was not being closed on all code paths, leading to connection leaks that could exhaust the database connection pool.

This change corrects both issues. A guard clause now checks for nil values before processing, and rows.Close() is deferred to guarantee the connection is released.


Yuan325 (approved):
Thank you! I submitted another PR for mysqlsql tools in #758.

PR #758, opened 2025-06-26
fix(tools/mysqlsql): Handle nil panic and connection leak in invoke

Yuan325:
Copy fix from #757. 

The Invoke function had two bugs:

A panic would occur when scanning a row containing a NULL value in a TEXT or VARCHAR column. The code did not check for nil before attempting a type assertion on the scanned value.

The *sql.Rows result was not being closed on all code paths, leading to connection leaks that could exhaust the database connection pool.

This change corrects both issues. A guard clause now checks for nil values before processing, and rows.Close() is deferred to guarantee the connection is released.


internal/tools/mysqlsql/mysqlsql.go
```diff
@@ -167,29 +168,30 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
 		}
 		vMap := make(map[string]any)
 		for i, name := range cols {
+            val := rawValues[i]
```
kurtisvg: nit: indentation is off

internal/tools/mysqlsql/mysqlsql.go
```diff
@@ -167,29 +168,30 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
...
 			// mysql driver return []uint8 type for "TEXT", "VARCHAR", and "NVARCHAR"
 			// we'll need to cast it back to string
 			switch colTypes[i].DatabaseTypeName() {
 			case "TEXT", "VARCHAR", "NVARCHAR":
 				if rawValues[i] != nil {
-					vMap[name] = string(rawValues[i].([]byte))
+					vMap[name] = string(val.([]byte))
 				} else {
 					vMap[name] = nil
 				}
```
kurtisvg: Do we still need this nil check if we check on L172?

kurtisvg (changes requested):

kurtisvg (approved):

PR #760, opened 2025-06-26
feat: add support for multiple YAML configuration files

iammujtaba:
- **Fix: Handled the null value in case of string case for mysqlgo tools.**
- **feat: add support for multiple YAML configuration files**


cmd/root.go
```diff
@@ -221,6 +227,126 @@ func parseToolsFile(ctx context.Context, raw []byte) (ToolsFile, error) {
...
+// Toolsets are merged by combining all tools from toolsets with the same name.
+func mergeToolsFiles(files ...ToolsFile) ToolsFile {
+	merged := ToolsFile{
+		Sources:      make(server.SourceConfigs),
+		AuthServices: make(server.AuthServiceConfigs),
+		Tools:        make(server.ToolConfigs),
+		Toolsets:     make(server.ToolsetConfigs),
+	}
+
+	for _, file := range files {
```
duwenxin99: Can we raise error in case of resource conflicts instead of overwriting here? We think it's better to warn the users instead of quietly overwriting for them. Thanks!

iammujtaba: Thanks for pointing out this, yes this may cause issue when multiple people are working on the same tools.
Fixed: Now it will raise error in case of conflict of the source,tools or toolset.

duwenxin99: Thanks! I will update the docs for this change.

iammujtaba: Alright!, I am waiting for PR to be reviewed and then merged. Thanks.

iammujtaba:
@Yuan325 @duwenxin99 Can you please review the changes?

duwenxin99:
> @Yuan325 @duwenxin99 Can you please review the changes?

Hi @iammujtaba , the changes look good to me but we are missing tests. Could you add unit test for the new flags (e.g tests for [--tools-file](https://github.com/googleapis/genai-toolbox/blob/a1b60100c279d17f63e54d246f6018bed96084ea/cmd/root_test.go#L192))? We wanna make sure they work correctly.

iammujtaba:
Hi @duwenxin99 ,
I've added the test cases as well.
Thanks!

duwenxin99 (approved):

duwenxin99:
Thanks for the contribution, @iammujtaba! The PR has been merged and will be included in the release today.

PR #764, opened 2025-06-27
refactor: introduce resourceManager to handle server resources

AlexTalreja:
Introduce the `resourceManager` struct to handle how the server stores resources (sources, authServices, tools, toolsets)

### DIFF
--- a/internal/server/common_test.go
+++ b/internal/server/common_test.go
@@ -158,12 +158,16 @@ func setUpServer(t *testing.T, router string, tools map[string]tools.Tool, tools
 		sseSessions: make(map[string]*sseSession),
 	}
 
-	resourceManager := &ResourceManager{
-		tools:    tools,
-		toolsets: toolsets,
+	resourceManager := NewResourceManager(nil, nil, tools, toolsets)
+
+	server := Server{
+		version:         fakeVersionString,
+		logger:          testLogger,
+		instrumentation: instrumentation,
+		sseManager:      sseManager,
+		resourceMgr:     resourceManager,
 	}
 
-	server := Server{version: fakeVersionString, logger: testLogger, instrumentation: instrumentation, sseManager: sseManager, ResourceManager: resourceManager}
 	var r chi.Router
 	switch router {
 	case "api":
--- a/internal/server/common_test.go
+++ b/internal/server/common_test.go
@@ -158,7 +158,12 @@ func setUpServer(t *testing.T, router string, tools map[string]tools.Tool, tools
 		sseSessions: make(map[string]*sseSession),
 	}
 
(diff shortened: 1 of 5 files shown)


internal/server/server.go
```diff
@@ -46,13 +46,60 @@ type Server struct {
 	logger          log.Logger
 	instrumentation *telemetry.Instrumentation
 	sseManager      *sseManager
+	ResourceManager *ResourceManager
```
```suggestion
	resourceManager *ResourceManager
```

By default it's always good to not make it exported. Can update it when we need it to be exported :)

Yuan325 (approved):

internal/server/server.go
```diff
@@ -46,13 +46,60 @@ type Server struct {
 	logger          log.Logger
 	instrumentation *telemetry.Instrumentation
 	sseManager      *sseManager
+	resourceManager *ResourceManager
```
kurtisvg: nit: go likes the short names 
```suggestion
	resourceMgr *ResourceManager
```

internal/server/common_test.go
```diff
@@ -158,7 +158,12 @@ func setUpServer(t *testing.T, router string, tools map[string]tools.Tool, tools
 		sseSessions: make(map[string]*sseSession),
 	}
 
-	server := Server{version: fakeVersionString, logger: testLogger, instrumentation: instrumentation, sseManager: sseManager, tools: tools, toolsets: toolsets}
+	resourceManager := &ResourceManager{
+		tools:    tools,
+		toolsets: toolsets,
+	}
+
+	server := Server{version: fakeVersionString, logger: testLogger, instrumentation: instrumentation, sseManager: sseManager, resourceManager: resourceManager}
```
kurtisvg: nit: line is a bit long

internal/server/server.go
```diff
@@ -46,13 +46,60 @@ type Server struct {
...
 
+type ResourceManager struct {
+	mu           sync.RWMutex
 	sources      map[string]sources.Source
 	authServices map[string]auth.AuthService
 	tools        map[string]tools.Tool
 	toolsets     map[string]tools.Toolset
 }
 
+func (m *ResourceManager) GetSource(sourceName string) (sources.Source, bool) {
```
kurtisvg: nit: (apply elsewhere)
```suggestion
func (r *ResourceManager) GetSource(sourceName string) (sources.Source, bool) {
```

internal/server/server.go
```diff
@@ -46,13 +46,60 @@ type Server struct {
 	logger          log.Logger
 	instrumentation *telemetry.Instrumentation
 	sseManager      *sseManager
+	resourceManager *ResourceManager
+}
 
+type ResourceManager struct {
+	mu           sync.RWMutex
```
kurtisvg, line 54: nit: this should probably be a pointer to prevent accidental copy. If you leave it as direct you should provide a NewResourceMethod that returns a pointer (instead of the struct itself) .

PR #768, opened 2025-06-30
test: add null column test case

Yuan325:
Add integration tests to check for `null` columns. ref #757

### DIFF
--- a/tests/tool.go
+++ b/tests/tool.go
@@ -76,7 +76,7 @@ func RunToolGetTest(t *testing.T) {
 }
 
 // RunToolInvoke runs the tool invoke endpoint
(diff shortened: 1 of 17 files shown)


tests/tool.go
```diff
@@ -76,7 +76,7 @@ func RunToolGetTest(t *testing.T) {
 }
 
 // RunToolInvoke runs the tool invoke endpoint
-func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant string) {
+func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeParamWant2 string) {
```
duwenxin99: Can we rename `invokeParamWant2` to be something more descriptive such as `invokeParamWantNull`?

PR #774, opened 2025-07-01
fix: set default value to field's type during unmarshalling

### LINKED ISSUE #771: Error with non-string parameter's default values
### Prerequisites

### Toolbox version

main

### Environment

1. OS type and version: `Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:29 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6030 arm64`
  - As a downloaded binary `curl -O https://storage.googleapis.com/genai-toolbox/main/linux/amd64/toolbox`

### Client

1. Client: `curl`
2. Version: 
3. Example: If possible, please include your code of configuration:

Tools file:
```yaml
...
tools:
  search-rows:
    ...
    parameters:
      - name: "email"
        type: "string"
        description: "..."
      - name: "data"
        type: "string"
        description: "..."
        default: "row2"
      - name: "id"
        type: "integer"
        description: "..."
        default: 0
    ...
```

```bash
curl --location 'http://127.0.0.1:5000/api/tool/search-rows/invoke' \
--header 'Content-Type: application/json' \
--data-raw '{
    "email": "..."
}'
```

This throws an error:
```
{"status":"Bad Request","error":"provided parameters were invalid: unable to parse value for \"id\": '\\x00' not type \"integer\""}
```

This is the same for more types as well eg. `"float"`. Seems like the `default` value provided in the `tools.yaml` is being parsed as a generic `Any` type, which ultimately makes the integer values (like `0` in this case) be parsed as `uint64`, which ultimately leads to this error.

### Expected Behavior

A success response with the provided `default` value of `0` being interpreted as an integer `0`.

### Current Behavior

Currently the integer value of `0` is being incorrectly casted as a `uint64` `0x0`.

### Steps to reproduce?

1. Write a tool with a parameter with `type` as `integer` and `default` as `0`
2. Invoke this tool without passing that optional parameter.

### Additional Details

I also tried by passing the `default` as `int!! 0` instead of just `0`, but it led to the same error.

Yuan325:
When go-yaml decode into CommonParameter with Default being an any type, int will be converted into []uint64.
It will fail the Parse() when the value is being used since it does not belong to either of the int types.

Unmarshal `default` value into each field's type directly.

Fixes #771

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ toolchain go1.24.4
 require (
 	cloud.google.com/go/alloydbconn v1.15.3
 	cloud.google.com/go/bigquery v1.69.0
-	cloud.google.com/go/bigtable v1.37.0
+	cloud.google.com/go/bigtable v1.38.0
 	cloud.google.com/go/cloudsqlconn v1.17.2
 	cloud.google.com/go/spanner v1.83.0
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.53.0
@@ -17,7 +17,7 @@ require (
 	github.com/go-chi/chi/v5 v5.2.2
 	github.com/go-chi/httplog/v2 v2.1.1
 	github.com/go-chi/render v1.0.3
-	github.com/go-playground/validator/v10 v10.26.0
+	github.com/go-playground/validator/v10 v10.27.0
 	github.com/go-sql-driver/mysql v1.9.3
 	github.com/goccy/go-yaml v1.18.0
 	github.com/google/go-cmp v0.7.0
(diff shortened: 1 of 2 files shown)


internal/tools/parameters.go
```diff
@@ -527,10 +503,38 @@ func (p *StringParameter) Parse(v any) (any, error) {
...
+func (p *StringParameter) Manifest() ParameterManifest {
+	// only list ParamAuthService names (without fields) in manifest
+	authNames := make([]string, len(p.AuthServices))
+	for i, a := range p.AuthServices {
+		authNames[i] = a.Name
+	}
+	var required bool
+	if p.Default == nil {
+		required = true
+	}
```
kurtisvg: nit: Could this be `required := p.Default == nil` instead?

kurtisvg (approved):

PR #780, opened 2025-07-02
feat: launch web server with --ui flag

AlexTalreja:
Add a flag `--ui` which will launch the Toolbox UI web server (and currently a skeleton HTML page).

The web server is on the same port as Toolbox and will serve HTML pages from `static/`


internal/server/web.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/chi/v5/middleware"
+)
+
+//go:embed all:static
+var embedFS embed.FS
+
+// webRouter creates a router that represents the routes under /web
+func webRouter() (http.Handler, error) {
```
Yuan325:
```suggestion
func webRouter() (chi.Router, error) {
```

internal/server/web.go
```diff
@@ -0,0 +1,32 @@
...
+	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+		const targetPath = "static/index.html"
+		htmlContent, err := embedFS.ReadFile(targetPath)
+		if err != nil {
+			http.Error(w, "Internal Server Error: Could not load page.", http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "text/html; charset=utf-8")
+		_, _ = w.Write(htmlContent)
```
Yuan325: can we move these codes into a separate function instead to keep the `webRouter()` cleaner? e.g. `func homePage()`

internal/server/web_test.go
```diff
@@ -0,0 +1,69 @@
...
+		wantContentType string
+		wantBodySubstr  string
+	}{
+		{
+			name:            "web index page GET",
+			method:          http.MethodGet,
+			path:            "/",
+			wantStatus:      http.StatusOK,
+			wantContentType: "text/html; charset=utf-8",
+			wantBodySubstr:  "<title>Toolbox UI</title>",
```
Yuan325: To what extend do we want to check the body substring? (especially when we add the other endpoints. Are we just checking the page's title?

Wondering if it'll be better to keep this `testWebEndpoint` to specifically just checking it's status; and have a separate individual test for each "page" to check the content instead?

Yuan325 (changes requested):

internal/server/web_test.go
AlexTalreja: That's something I wasn't too sure of, initially I thought comparing the entire HTML was unnecessary as long as each page has a distinct title, but if that's preferred maybe we could just compare the response to the actual files?

internal/server/web.go
AlexTalreja: Moved out of `webRouter()`, and I was thinking we could even go further, maybe instead of having one new func per page, could we just have a single `serveHTML(filepath string)` and then call that for different routes/pages. 

Do you think that could work? (see newest commit for example implementation)

internal/server/web_test.go
Yuan325: Got it, if we want to compare the title, lets update the `wantBodySubstr` to `wantPageTitle`. If we need to check for the HTML in the future we can add new tests.

internal/server/web.go
Yuan325: Will all the different pages uses the same codes in serveHTML?

I'm thinking in cases like the tools page, will we be calling the apis in `web.go`, or will it all be done within he js file?

AlexTalreja: I was thinking it can just be done within the JS

internal/server/web_test.go
AlexTalreja: NOTE: ended up switching to `wantPageTitle` for now, and also used goquery for parsing the HTML to get the `<title>` field. This ensures consistency as opposed to just string matching the title.

Yuan325 (approved):

kurtisvg (approved):

PR #781, opened 2025-07-02
feat: support multiple files for dynamic reloading

AlexTalreja:
Updates dynamic reloading to work for multiple forms of YAML config files.

This adjusts file-watching logic to account for the newly introduced changes in https://github.com/googleapis/genai-toolbox/pull/760.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -440,8 +440,16 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 	// if watchedFiles is empty, indicates that user passed entire folder instead
 	if len(watchedFiles) == 0 {
 		watchingFolder = true
-		for k := range watchDirs {
-			folderToWatch = k
+
+		// validate that watchDirs only has single element
+		if len(watchDirs) > 1 {
+			logger.WarnContext(ctx, "error setting watcher, expected single tools folder if no file(s) are defined.")
+			return
+		}
+
+		for onlyKey := range watchDirs {
+			folderToWatch = onlyKey
+			break
 		}
 	}
 
@@ -449,9 +457,9 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 		err := w.Add(dir)
 		if err != nil {
 			logger.WarnContext(ctx, fmt.Sprintf("Error adding path %s to watcher: %s", dir, err))
-		} else {
-			logger.DebugContext(ctx, fmt.Sprintf("Added directory %s to watcher.", dir))
+			break
 		}
+		logger.DebugContext(ctx, fmt.Sprintf("Added directory %s to watcher.", dir))
 	}
 
 	// debounce timer is used to prevent multiple writes triggering multiple reloads
@@ -480,6 +488,7 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 				return
 			}
 
+			// only check for write events which indicate user saved a new tools file
 			if e.Op != fsnotify.Write {
 				continue
 			}
@@ -487,11 +496,11 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 			cleanedFilename := filepath.Clean(e.Name)
 			logger.DebugContext(ctx, fmt.Sprintf("%s event detected in %s", e.Op, cleanedFilename))
 
-			if watchingFolder && (strings.HasSuffix(cleanedFilename, "yaml") || strings.HasSuffix(cleanedFilename, "yml")) {
-				// case where we watch an entire folder
-				debounce.Reset(debounceDelay)
-			} else if watchedFiles[cleanedFilename] {
-				// case where we watch specific files
+			folderChanged := watchingFolder &&
+				(strings.HasSuffix(cleanedFilename, ".yaml") || strings.HasSuffix(cleanedFilename, ".yml"))
+
+			if folderChanged || watchedFiles[cleanedFilename] {
+				// indicates the write event is on a relevant file
 				debounce.Reset(debounceDelay)
 			}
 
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -439,13 +434,25 @@ func watchFile(ctx context.Context, toolsFileName string, s *server.Server) {
...
 
-	cleanedFilename := filepath.Clean(toolsFileName)
-	logger.DebugContext(ctx, fmt.Sprintf("Now watching tools file %s", cleanedFilename))
+	for dir := range watchDirs {
+		err := w.Add(dir)
+		if err != nil {
+			logger.WarnContext(ctx, fmt.Sprintf("Error adding path %s to watcher: %s", dir, err))
+		} else {
+			logger.DebugContext(ctx, fmt.Sprintf("Added directory %s to watcher.", dir))
+		}
```
kurtisvg: nit: happy path to the left
```suggestion
		if err != nil {
			logger.WarnContext(ctx, fmt.Sprintf("Error adding path %s to watcher: %s", dir, err))
		}
		logger.DebugContext(ctx, fmt.Sprintf("Added directory %s to watcher.", dir))

```

cmd/root.go
```diff
@@ -473,23 +480,44 @@ func watchFile(ctx context.Context, toolsFileName string, s *server.Server) {
 				return
 			}
 
-			if e.Op == fsnotify.Write && filepath.Clean(e.Name) == cleanedFilename {
-				logger.DebugContext(ctx, fmt.Sprintf("%s event detected in tools file: %s", e.Op, cleanedFilename))
+			if e.Op != fsnotify.Write {
```
kurtisvg, line 492: nit: worth a comment here I think

cmd/root.go
```diff
@@ -439,13 +434,25 @@ func watchFile(ctx context.Context, toolsFileName string, s *server.Server) {
...
-		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
+	watchingFolder := false
+	var folderToWatch string
+
+	// if watchedFiles is empty, indicates that user passed entire folder instead
+	if len(watchedFiles) == 0 {
+		watchingFolder = true
+		for k := range watchDirs {
+			folderToWatch = k
+		}
```
kurtisvg: I don't follow this loop. What's the point?

cmd/root.go
```diff
@@ -473,23 +480,44 @@ func watchFile(ctx context.Context, toolsFileName string, s *server.Server) {
...
+			cleanedFilename := filepath.Clean(e.Name)
+			logger.DebugContext(ctx, fmt.Sprintf("%s event detected in %s", e.Op, cleanedFilename))
+
+			if watchingFolder && (strings.HasSuffix(cleanedFilename, "yaml") || strings.HasSuffix(cleanedFilename, "yml")) {
+				// case where we watch an entire folder
+				debounce.Reset(debounceDelay)
+			} else if watchedFiles[cleanedFilename] {
+				// case where we watch specific files
 				debounce.Reset(debounceDelay)
 			}
```
kurtisvg: This looks like the same thing in each statement. Should we simplify?

kurtisvg: Also consider -- what is the "default" case here?

cmd/root.go
```diff
@@ -709,8 +753,16 @@ func run(cmd *Command) error {
...
 	if !cmd.cfg.DisableReload {
-		// start watching for file changes to trigger dynamic reloading
-		go watchFile(ctx, cmd.tools_file, s)
+
+		// extract parent dir for relevant files and dedup
+		for _, f := range relevantFiles {
+			cleanFile := filepath.Clean(f)
+			watchedFiles[cleanFile] = true
+			watchDirs[filepath.Dir(cleanFile)] = true
+		}
```
kurtisvg: nit: currently there is a vary long gap between initialization here and usage (especially for watchedFiles and watchDirs). Can we move things closer together?

One reasonable thing might be to move this into a function that returns those things and then call `go watchcChanges` inside section above.

kurtisvg (changes requested):

cmd/root.go
AlexTalreja: This is just getting the only element from the map since within this case (tools folder) there will only be 1 key/value pair in watchDirs, which is the folder to watch. From what I found this seemed the simplest way to extract the only key from a map in go.

Would renaming + comment + break statement make this more readable?

```go
// in this case, watchDirs will only have the single folder to watch
for onlyKey := range watchDirs {
	folderToWatch = onlyKey
        break
}
```

cmd/root.go
AlexTalreja: Added a new function `resolveWatcherInputs()` which returns the watchedFiles and watchDirs.

cmd/root.go
```diff
@@ -439,13 +434,26 @@ func watchFile(ctx context.Context, toolsFileName string, s *server.Server) {
...
-	err = w.Add(filepath.Dir(toolsFileName))
-	if err != nil {
-		logger.WarnContext(ctx, "error adding the tools file to watcher %s", err)
+	watchingFolder := false
+	var folderToWatch string
+
+	// if watchedFiles is empty, indicates that user passed entire folder instead
+	if len(watchedFiles) == 0 {
+		watchingFolder = true
+		for onlyKey := range watchDirs {
```
kurtisvg, line 450: nit: it would be good to validate the assumption that watchDirs only has 1 entry in it

cmd/root.go
```diff
@@ -473,23 +481,44 @@ func watchFile(ctx context.Context, toolsFileName string, s *server.Server) {
...
+			if e.Op != fsnotify.Write {
+				continue
+			}
+
+			cleanedFilename := filepath.Clean(e.Name)
+			logger.DebugContext(ctx, fmt.Sprintf("%s event detected in %s", e.Op, cleanedFilename))
+
+			isYaml := strings.HasSuffix(cleanedFilename, "yaml") || strings.HasSuffix(cleanedFilename, "yml")
+
+			if (watchingFolder && isYaml) || watchedFiles[cleanedFilename] {
```
kurtisvg: nit: slight optimization to allow for string search to be short circuited
```suggestion
			folderChanged := watchingFolder && ( strings.HasSuffix(cleanedFilename, "yaml") || strings.HasSuffix(cleanedFilename, "yml") )

			if folderChanged || watchedFiles[cleanedFilename] {
```

PR #786, opened 2025-07-03
feat: Add Firestore as Source

trehanshakuntG:
Firestore is a NoSQL document database built for automatic scaling, high performance, and ease of application development. It's a fully managed, serverless database that supports mobile, web, and server development.

This change adds Firestore as a source in toolbox

### DIFF
--- a/internal/tools/firestore/firestorequerycollection/firestorequerycollection.go
+++ b/internal/tools/firestore/firestorequerycollection/firestorequerycollection.go
@@ -267,7 +267,7 @@ type QueryResponse struct {
 }
 
 // Invoke executes the Firestore query based on the provided parameters
-func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
+func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error) {
 	// Parse parameters
 	queryParams, err := t.parseQueryParameters(params)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/sources/firestore/firestore.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+var _ sources.Source = &Source{}
+
+type Source struct {
+	// Firestore struct with client
+	Name     string `yaml:"name"`
+	Kind     string `yaml:"kind"`
+	Client   *firestore.Client
+	Database string `yaml:"database"`
```
duwenxin99: Why do we need to save the `Database` in the source struct? I assume we don't need it later for tool invocations?

duwenxin99:
Can we merge it into a feature branch, so that we can merge the source, tools, tests, and docs all in one PR (e.g, "feat: Add Firestore Source and Tool")? Thanks

internal/sources/firestore/firestore.go
trehanshakuntG: This is required for "get-rules" and "validate-rules" tools.

averikitsch: Is this because we will have to use a different client to make these requests?

duwenxin99: what are these tools for?

trehanshakuntG: @averikitsch yes, a different client is needed for firebase rules

duwenxin99: If a different client is needed, it is probably better to create it in source initialization and store both clients in the source if applicable.

trehanshakuntG: Noted, added!

duwenxin99 (approved):

PR #800, opened 2025-07-07
feat: dynamic reloading for toolbox config

AlexTalreja:
Allow Toolbox to automatically update when users modify their tool configuration file(s), instead of requiring a restart.

This feature is automatically enabled, but can be turned off with the flag `--disable-reload`.


Yuan325 (approved):
Yay!

jexp:
Should probably add to the docs here https://googleapis.github.io/genai-toolbox/getting-started/configure/

Issue #793, opened 2025-07-04
Error when using parameters of type array


ngocquang2797:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

toolbox version 0.8.0+binary.darwin.arm64.f13e9635ba64425b57805c8d64a5968ba569292c

### Environment

1. darwin/arm64
2. How are you running Toolbox: 
  - As a downloaded binary (e.g. from `curl -O https://storage.googleapis.com/genai-toolbox/v$VERSION/linux/amd64/toolbox`)

### Client

1. Client: MCP Inspector v0.15.0

### Expected Behavior

```
  get-shipments-by-ids:
    kind: bigquery-sql
    source: data-source
    description: Retrieve tracking codes for given shipment IDs
    parameters:
      - name: shipment_ids
        type: array
        description: A list of shipment IDs to retrieve tracking codes for.
        items:
          name: shipment_id
          type: integer
          description: A shipment ID value
    statement: |
      SELECT *
      FROM `${RAW_PROJECT_ID}.${RAW_POSTGRES_DATASET}.shipments`
      WHERE id IN UNNEST(@shipment_ids)
```

### Current Behavior

<img width="1124" height="962" alt="Image" src="https://github.com/user-attachments/assets/ba1d665f-37c4-46c2-b22b-9168dbffb186" />

### Steps to reproduce?

1.

duwenxin99:
Hi @ngocquang2797, we have fixed this issue and it will be included in the next release. Thank you!

PR #801, opened 2025-07-07
fix: nil parameter error when arrays are used

duwenxin99:
- Spanner: convert array to typed before querying, as Spanner does not accept non-typed array []any
- BigQuery: fix https://github.com/googleapis/genai-toolbox/issues/793
- Bigtable: add the required `ElemType` for array-type params 
- Redis/Valkey: change indexing to append to avoid extra spaces

Add integration tests for array parameters, skipped for the sources not supporting arrays:
- SQLite
- Cloud SQL MSSQL
- Cloud SQL MySQL
- MSSQL
- MySQL


internal/tools/redis/redis.go
```diff
@@ -176,7 +176,7 @@ func (t Tool) Authorized(verifiedAuthServices []string) bool {
 	return tools.IsAuthorized(t.AuthRequired, verifiedAuthServices)
 }
 
-// replaceCommandsParams is a helper function to replace parameters in the commands
+// Helper function to replace parameters in the commands
```
Yuan325:
```suggestion
// replaceCommandsParams  is a helper function to replace parameters in the commands
```

Golang's doc comment usually start with the name~

internal/tools/spanner/spannersql/spannersql.go
```diff
@@ -175,6 +175,26 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
...
+		name := p.GetName()
+		value := newParams[i].Value
+
+		// Spanner only accepts typed slices as input
+		// This checks if the param is an array.
+		// If yes, convert []any to typed slice (e.g []string, []int)
+		switch arrayParam := value.(type) {
+		case []any:
+			var err error
+			itemType := p.McpManifest().Items.Type
```
Yuan325: for future ref: If there are more cases of using this, might be worth to add a GetItems() in the `Parameter` interface... unsure if we want to rely on the McpManifest to get array's item.

internal/tools/common.go
```diff
@@ -23,3 +24,50 @@ var validName = regexp.MustCompile(`^[a-zA-Z0-9_-]*$`)
 func IsValidName(s string) bool {
 	return validName.MatchString(s)
 }
+
+func ConvertAnySliceToTyped(s []any, itemType, paramName string) (any, error) {
```
Yuan325: Is it worth adding an argument for `paramName` just for the error message? I think ideally here we will return `expected item at index %d...` and on the upper layer, we can do something like `("error with converting parameter %s: %s", name, err)`

internal/tools/spanner/spannersql/spannersql.go
duwenxin99: added GetItems() method to Parameter struct

internal/tools/common.go
duwenxin99: good point. Removed.

internal/tools/valkey/valkey.go
```diff
@@ -154,22 +154,24 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, erro
 	return out, nil
 }
 
-// Helper function to replace parameters in the commands
+// replaceCommandsParams is a helper function to replace parameters in the commands
+
```
Yuan325:
```suggestion
```

Yuan325 (approved):

PR #802, opened 2025-07-08
feat: add support for null optional parameter

### LINKED ISSUE #736: Optional Null value types for parameters
### Prerequisites

Hello,

Is it possible to have null types when specifying tool parameters? For example in tools.yaml:
```
parameters: 
    - name: name 
      type: ['string', 'null']
      description: The name of the hotel.
```

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

Yuan325:
Added an option for user to indicate if the parameter is required. 

Example:
```
parameters:
  - name: foo
    type: string
    description: foo
    required: false
```

If the `required` field is not provided, it will be defaulted to `true`. 

If a `default` value is provided, `required = false` regardless if the field is indicated.
```
parameters:
  - name: foo
    type: string
    description: foo
    default: hello world
```

Fixes #736

### DIFF
--- a/internal/tools/parameters.go
+++ b/internal/tools/parameters.go
@@ -109,11 +109,17 @@ func parseFromAuthService(paramAuthServices []ParamAuthService, claimsMap map[st
 	return nil, fmt.Errorf("missing or invalid authentication header")
 }
 
+// CheckParamRequired checks if a parameter is required based on the required and default field.
+func CheckParamRequired(required bool, defaultV any) bool {
+	return required && defaultV == nil
+}
+
 // ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
 func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[string]any) (ParamValues, error) {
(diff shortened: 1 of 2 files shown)


internal/tools/parameters.go
```diff
@@ -109,11 +109,17 @@ func parseFromAuthService(paramAuthServices []ParamAuthService, claimsMap map[st
 	return nil, fmt.Errorf("missing or invalid authentication header")
 }
 
+// CheckParamRequired checks if a parameter is required based on the required and default field.
+func CheckParamRequired(required bool, defaultV any) bool {
+	return required && defaultV == nil
+}
```
kurtisvg, line 115: nit: does this really need a function? If you think it's more clear you can leave it, but it looks like it's only being used in 1 place

kurtisvg (approved):

internal/tools/parameters.go
Yuan325: it's used in each Parameter's Manifest() as well.

PR #811, opened 2025-07-08
chore: add schema for mcp v2025-06-18

Yuan325:
Add schema for mcp version `2025-06-18`.

### DIFF
--- a/internal/server/mcp/util/lifecycle.go
+++ b/internal/server/mcp/util/lifecycle.go
@@ -89,8 +89,22 @@ type ServerCapabilities struct {
 	Tools *ListChanged `json:"tools,omitempty"`
 }
 
+// Base interface for metadata with name (identifier) and title (display name) properties.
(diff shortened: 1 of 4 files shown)


internal/server/mcp/util/lifecycle.go
```diff
@@ -89,8 +89,22 @@ type ServerCapabilities struct {
...
+	// If not provided, the name should be used for display (except for Tool,
+	// where `annotations.title` should be given precedence over using `name`,
+	// if present).
+	Title string `json:"title,omitempty"`
+}
+
 // Implementation describes the name and version of an MCP implementation.
 type Implementation struct {
-	Name    string `json:"name"`
+	BaseMetadata
```
kurtisvg, line 108: Why is this nested instead of two separate fields?

Yuan325: It's how the MCP schema is defined.

Issue #715, opened 2025-06-15
Proposal: Enabling MCP Toolbox for Databases as MCP Server


tuannvm:
### Prerequisites
- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### Overview
Currently, the MCP Toolbox for Databases acts solely as an MCP server, exposing tools to clients (such as IDEs or agent frameworks) via the MCP protocol. To maximize flexibility and composability, we propose extending the Toolbox to also function as an MCP client. This will allow it to dynamically connect to other MCP servers, aggregate their tools, and expose them as a unified toolset to its own clients. This dual role enables federated tool management, cross-organization sharing, and dynamic tool composition.

#### 1. MCP Client Mode
- **Configurable MCP Servers**: Administrators can specify a list of remote MCP servers (with endpoints and credentials) that the Toolbox should connect to as a client.
- **Dynamic Tool Loading**: The Toolbox will periodically (or on-demand) fetch tool definitions from these remote MCP servers and merge them into its own tool registry.
- **Conflict Resolution**: If tool names or IDs collide, configurable strategies (e.g., namespacing, override, or ignore) can be applied.
 
---

### Sequence Diagram
Below is a sequence diagram illustrating the interaction when the Toolbox acts as both MCP client and server.

```mermaid
sequenceDiagram
participant Admin
participant Toolbox
participant MCP_A
participant MCP_B
participant Client
%% Admin configures Toolbox as MCP client
Admin->>Toolbox: Configure remote MCP servers (MCP_A, MCP_B)
%% Toolbox fetches tool lists from remote MCP servers
Toolbox->>MCP_A: Fetch tool list
Toolbox->>MCP_B: Fetch tool list
MCP_A-->>Toolbox: Return tool definitions (A1, A2)
MCP_B-->>Toolbox: Return tool definitions (B1, B2)
%% Toolbox merges local and remote tools
Toolbox->>Toolbox: Merge local and remote tools
%% Client requests available tools from Toolbox
Client->>Toolbox: List available tools
Toolbox-->>Client: Return unified tool list (Local, A1, A2, B1, B2)
%% Client invokes a remote tool via Toolbox
Client->>Toolbox: Invoke remote tool (A1)
Toolbox->>MCP_A: Proxy tool invocation (A1)
MCP_A-->>Toolbox: Return result
Toolbox-->>Client: Return result
```

---
### Example Configuration
```yaml
# tools.yaml
...
mcp_clients:
[...]

Issue #817, opened 2025-07-09
Toolbox Control Plane


averikitsch:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

Currently, managing resources within the MCP Toolbox primarily involves direct YAML manipulations that may become toilsome with a large number of sources and tools. 

### Suggested Solution(s)

We propose the addition of a Control Plane to the MCP Toolbox. This control plane would serve as a unified interface via the CLI that enables users to: Add / Edit / Delete / List Resources.

PR #821, opened 2025-07-09
feat: MCP Server Source Type

wagnerjt:
Supports #715, where MCP servers are the source.

# Implementation

1. Added the official MCP go-sdk and currently is using the `go get https://github.com/modelcontextprotocol/go-sdk@main` so the version is not pinned.
2. Created the `mcp-server` source type
3. Opted for a new Interface `DynamicToolSource` where the source can dynamically create tools. This requires only the mcp source definition to be present without the need to have additional `Tool` definitions in the yaml. Likewise, there is no additional need to perform the `Register` flow.

```go
// DynamicToolSource is a source type that can dynamically create tools
type DynamicToolSource interface {
	// The tools array that are dynamically created
	GetTools(context.Context) ([]Tool, error)
}
```

When adding the tools to the server definition, we follow a few rules with the main functions:

```go
// IsDynamicToolSource returns true if the source dynamically adds tools.
func IsDynamicToolSource(source sources.Source) bool {
	_, ok := source.(tools.DynamicToolSource)

	return ok
}

// MergeTools merges the locally defines tools and the dynamic tools maps in place. The operation is as follows:
// (a) conflict between local Toolbox's tools vs imported server's tools: take precedence on local tool.
// (b) conflict between two imported server: last imported tool will be selected.
// We expect the dynamicTools to already have (b) satisfied.
func MergeTools(localToolsMap map[string]tools.Tool, dynamicToolsMap map[string]tools.Tool) {
	for toolName, tool := range dynamicToolsMap {
		_, ok := localToolsMap[toolName]
		if !ok {
			localToolsMap[toolName] = tool
		}
	}
}
```

## Disclaimer

This is a rough draft of implementation and I am quite new to go, so really looking for feedback on this interface and flow -- specifically the location and if we wanted to compose the `DynamicToolSource` along with the `Source` interface. There is much more iterations that can be done:
1. Fetch tools on a timer
2. Hook into server the `notifications/tools/list_changed` to rebuild tool list
3. Reuse client sessions per mcp server
4. More auth

I am trying to keep these things out of scope for this PR.

# Example

```yaml
sources:
  my-local-mcp-server:
    kind: mcp-server
    endpoint: http://127.0.0.1:8080/sse
    transport: sse
    specVersion: 2025-03-26
    authMethod: bearer
    authSecret: ${MCP_PAT}
```

## Reference

| **field**   | **type** | **required** | **description**                                    |
| ----------- | :------: | :----------: | -------------------------------------------------- |
| kind        |  string  |     true     | Must be "mcp-server".                              |
| endpoint    |  string  |     true     | Connect Uri - `http://127.0.0.1/mcp`               |
| specVersion |  string  |    false     | One of the supported mcp specification versions.   |
| transport   |  string  |    false     | One of the supported mcp transport types.          |
| authMethod  |  string  |    false     | One of the supported auth method types.            |
| authSecret  |  string  |    false     | The secret value used along with the `authMethod`. |

For more specifics and limitations, refer to the `mcpservers.md`.

I tested this with a SSE server without auth, since [this is still open](https://github.com/modelcontextprotocol/go-sdk/pull/105).

### DIFF
--- a/internal/sources/mcpserver/mcp_server.go
+++ b/internal/sources/mcpserver/mcp_server.go
@@ -22,6 +22,7 @@ import (
 	"github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/sources"
 	"github.com/googleapis/genai-toolbox/internal/tools"
+	"github.com/modelcontextprotocol/go-sdk/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"go.opentelemetry.io/otel/trace"
 )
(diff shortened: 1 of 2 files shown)


internal/sources/mcpserver/mcp_server.go
```diff
@@ -0,0 +1,354 @@
...
+// GetMyInterfaces returns a slice of MyInterface from a slice of MyStruct.
+func toToolboxTools(structs []MCPServerTool) []tools.Tool {
+	interfaces := make([]tools.Tool, len(structs))
+	for i, s := range structs {
+		interfaces[i] = s // Assigning a concrete struct to an interface element.
+	}
+	return interfaces
+}
+
+var _ tools.Tool = MCPServerTool{}
```
Yuan325, line 241: Thanks for submitting this PR!! 

Wondering if there's any justifications to put these in here instead of having it in https://github.com/googleapis/genai-toolbox/tree/main/internal/tools  along with the other tools?

wagnerjt: I was initially placing this definition in there, however it didn't feel right since we weren't following the `init` and `register` pattern so I kept it closer to the source.
After adopting the new `DynamicToolSource` interface, maybe it makes sense to move all of this in a new module?
Very much open to feedback and design on this! 

Another thought is how the design of dynamic merging of sources and tools from the [control plane](https://github.com/googleapis/genai-toolbox/issues/817) is going to work

averikitsch:
Closing until further development

Issue #791, opened 2025-07-03
UI: Copy button on code blocks is not visible in light theme in the MCP Toolbox for Databases Documentation


AnmolShukla2002:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

toolbox version 0.8.0

### Environment

1. OS type and version: Linux penguin 6.6.76-08024-gb30cb4a129c2
2. How are you running Toolbox: 
  - As a downloaded binary (e.g. from `curl -O https://storage.googleapis.com/genai-toolbox/v0.8.0/$OS/toolbox`)

### Expected Behavior

The button "Copy to Clipboard" should be clearly visible against the code block background irrespective of the theme.
Link to Documentation  : https://googleapis.github.io/genai-toolbox/getting-started/introduction/

### Current Behavior

The 'Copy to Clipboard' button on all code blocks is not visible when the site is in light theme but is visible when theme is changed to dark.

### Steps to reproduce?

1. Open the link : https://googleapis.github.io/genai-toolbox/getting-started/introduction/
2. See the code block in light theme and then dark theme. The 'Copy to Clipboard button is visible in the dark theme but not in the light theme'.

### Additional Details

![Image](https://github.com/user-attachments/assets/167daebc-e68b-4125-8555-ea3b3799ea5c)
As you can see, in this light theme Copy to Clipboard button is not visible, which is supposed to be there on the top right corner of the code block.

AnmolShukla2002:
I  am looking into it currently.

PR #826, opened 2025-07-09
docs: fix copy to clipboard button visibility in light mode

AnmolShukla2002:
This PR fixes the issue #791 and updated the info box color so that tags are visible in dark mode as well in docsite.

### DIFF
--- a/docs/en/getting-started/local_quickstart.md
+++ b/docs/en/getting-started/local_quickstart.md
@@ -35,7 +35,7 @@ accessed by our agent, and create a database user for Toolbox to connect with.
 
     Here, `postgres` denotes the default postgres superuser.
 
-    {{% pageinfo color="warning" %}}
+    {{< notice info >}}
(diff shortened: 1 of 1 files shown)


docs/en/getting-started/local_quickstart.md
```diff
@@ -35,7 +35,7 @@ accessed by our agent, and create a database user for Toolbox to connect with.
 
     Here, `postgres` denotes the default postgres superuser.
 
-    {{< notice info >}}
+    {{% pageinfo color="warning" %}}
```
twishabansal: Let's move this change to another PR. This is a style change that might be needed at other places too.

AnmolShukla2002: Okay sure!

AnmolShukla2002: Done

PR #827, opened 2025-07-09
info box styling change

AnmolShukla2002:
Updated the info box color from this:
image

To this:
image

because the highlighted tags were not visible in the dark mode.


kurtisvg:
Hugo Notice (the extension we are using) has support for light/dark mode: https://github.com/martignoni/hugo-notice?tab=readme-ov-file#enabling-dark-mode

Would it be better to just use this when users have dark mode enabled instead?

twishabansal:
> Hugo Notice (the extension we are using) has support for light/dark mode: https://github.com/martignoni/hugo-notice?tab=readme-ov-file#enabling-dark-mode
> 
> Would it be better to just use this when users have dark mode enabled instead?

@AnmolShukla2002 did you get a chance to try this out?

AnmolShukla2002:
@twishabansal I tried but was not able to resolve the `--disable-flag' issue.  It's still not clearly visible in dark mode. Let me know if this works:
image

AnmolShukla2002:
We can review this later.

PR #829, opened 2025-07-09
feat: support tidb in data source, sql tool, and execute sql tool

Icemap:
This PR supports TiDB in:

1. sources - tidb: As a data source;
2. tools - tidb - tidbsql: As a prepared SQL running tool;
3. tools - tidb - tidbexecutesql: As an arbitrary SQL running tool (for development purposes).

And its corresponding docs.


Icemap:
CLA was signed now. How can I make the CLA checking bot recheck the status?

kurtisvg:
@Icemap there's a little button in the check details that says "click here to update CLA check". I've triggered it for this PR and looks like it's green

Icemap:
Thanks @kurtisvg, if you find any issues or have some suggestions during the code review, please let me know. I'll fix it ASAP.

kurtisvg:
Thanks @Icemap! We've had a lot of interest in the repo over the last two days, which is making us run behind. Please give us a few days and we'll get back to you!

docs/en/resources/sources/tidb.md
```diff
@@ -0,0 +1,81 @@
...
+{{< notice tip >}}
+Use environment variable replacement with the format ${ENV_NAME}
+instead of hardcoding your secrets into the configuration file.
+{{< /notice >}}
+
+## Reference
+
+| **field** | **type** | **required** | **description**                                                                             |
+|-----------|:--------:|:------------:|---------------------------------------------------------------------------------------------|
+| kind      |  string  |     true     | Must be "tidb".                                                                             |
```
Yuan325:
```suggestion
| **field** | **type** | **required** | **description**                                                                            |
|-----------|:--------:|:------------:|--------------------------------------------------------------------------------------------|
| kind      |  string  |     true     | Must be "tidb".                                                                            |
```

docs/en/resources/sources/tidb.md
```diff
@@ -0,0 +1,81 @@
...
+    ```yaml
+    sources:
+        my-tidb-source:
+            kind: tidb
+            host: 127.0.0.1
+            port: 4000
+            database: my_db
+            user: ${TIDB_USERNAME}
+            password: ${TIDB_PASSWORD}
+            use_ssl: false  # Optional: enable SSL for secure connections
```
Yuan325:
```suggestion
            # use_ssl: true  # Optional: enable SSL for secure connections
```

It'll be best to comment this out since it is optional. And the default value is false right? User will only need to add the configuration if they want it to be true?

internal/sources/tidb/tidb.go
```diff
@@ -0,0 +1,133 @@
...
+
+type Config struct {
+	Name     string `yaml:"name" validate:"required"`
+	Kind     string `yaml:"kind" validate:"required"`
+	Host     string `yaml:"host" validate:"required"`
+	Port     string `yaml:"port" validate:"required"`
+	User     string `yaml:"user" validate:"required"`
+	Password string `yaml:"password" validate:"required"`
+	Database string `yaml:"database" validate:"required"`
+	UseSSL   bool   `yaml:"use_ssl"`
```
Yuan325:
```suggestion
	UseSSL   bool   `yaml:"useSsl"`
```

To keep consistency with our other field, camel case is preferred. We can also just do `ssl`.

internal/sources/tidb/tidb.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (s *Source) SourceKind() string {
+	return SourceKind
+}
+
+func (s *Source) TiDBPool() *sql.DB {
+	return s.Pool
+}
+
+func isTiDBCloudHost(host string) bool {
```
Yuan325: can we have a unit test in `tidb_test.go` that specifically test this function?

internal/sources/tidb/tidb_test.go
```diff
@@ -0,0 +1,182 @@
...
+			in: `
+			sources:
+				my-tidb-instance:
+					kind: tidb
+					host: 0.0.0.0
+					port: my-port
+					database: my_db
+					user: my_user
+					password: my_pass
+					use_ssl: false
```
Yuan325:
```suggestion
```

Basic example test doesn't include optional parameter. By default `useSsl` is false, so this should work.

internal/sources/tidb/tidb_test.go
```diff
@@ -0,0 +1,182 @@
...
+			want: server.SourceConfigs{
+				"my-tidb-cloud": tidb.Config{
+					Name:     "my-tidb-cloud",
+					Kind:     tidb.SourceKind,
+					Host:     "gateway01.us-west-2.prod.aws.tidbcloud.com",
+					Port:     "4000",
+					Database: "test_db",
+					User:     "cloud_user",
+					Password: "cloud_pass",
+					UseSSL:   true,
```
Yuan325, line 105: This might fail the test since this test only check the parsing. It doesn't check the initialization of the source. `initialization` should be checked in the integration tests instead :).

internal/tools/tidb/tidbexecutesql/tidbexecutesql.go
```diff
@@ -0,0 +1,194 @@
+// Copyright 2025 Google LLC
```
Yuan325, line 1: The `tidb-execute-sql` and `tidb-sql` tool seems to be the exact same as `mysql-execute-sql` and `mysql-sql` tool. Instead of creating new tools for tidb, can we just use the `mysql-execute-sql` and `mysql-sql` tool instead? We can add `tidb` source as a compatible source for those 2 tools.

Yuan325 (changes requested):
Hi @Icemap ! Sorry for the late response. I left some comments on this PR. While you make the changes, I'll also start setting up TiDB vm in our integration test.

Please let me know if you require more clarification on any of the feedbacks! :) Thank you again for contributing!!

tests/tidb/tidb_integration_test.go
```diff
@@ -0,0 +1,135 @@
...
+		"user":     TiDBUser,
+		"password": TiDBPass,
+	}
+}
+
+// Copied over from tidb.go
+func initTiDBConnectionPool(host, port, user, pass, dbname string, useSSL bool) (*sql.DB, error) {
+	dsn := fmt.Sprintf("%s:%s@tcp(%s:%s)/%s?parseTime=true", user, pass, host, port, dbname)
+	if useSSL {
+		dsn += "&tls=true"
```
Yuan325: QQ on this, do we always want this to be true? Our testing instance won't be in tidb cloud. It will be running locally with a vm.

Correct me if I'm wrong -- For setting up the tidb instance locally, I'll just have to run the following instructions in my vm: https://docs.pingcap.com/tidb/stable/quick-start-with-tidb/#simulate-production-deployment-on-a-single-machine , and then connect with 127.0.0.1, port 4000?

c4pt0r:
> Thanks @Icemap! We've had a lot of interest in the repo over the last two days, which is making us run behind. Please give us a few days and we'll get back to you! 

any updates on this? really appreciate it

Icemap:
Great, @Yuan325. Thanks for the feedback. I'll apply those suggestions ASAP.

tests/tidb/tidb_integration_test.go
Icemap: This part means if we set `useSSL` to `true`, then the connection string will add a `&tls=true` at the end. For the `dsn`, the default value of `tls` is `false` as well. But let me make it the same as `tidb.go`.

Icemap: And for setting up a tidb instance in the VM, it's the correct way to start it. After installing `tiup`, you can use `tiup playground` to start a test-purpose TiDB easily.

docs/en/resources/sources/tidb.md
Icemap: You're right. The default value is `false`. If they switch on the SSL in TiDB, they need to set `true` here as well.

internal/sources/tidb/tidb.go
Icemap: Got it, I changed it to the line like:

```go
UseSSL   bool   `yaml:"ssl"`
```

internal/sources/tidb/tidb_test.go
Icemap: OK

internal/sources/tidb/tidb_test.go
Icemap: It doesn't lead to an error, because I checked if it is a TiDB Cloud URL in the `newConfig` function like:

```go
func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources.SourceConfig, error) {
	actual := Config{Name: name}
	if err := decoder.DecodeContext(ctx, &actual); err != nil {
		return nil, err
	}

	// If the host is a TiDB Cloud instance, force to use SSL
	if IsTiDBCloudHost(actual.Host) {
		actual.UseSSL = true
	}

	return actual, nil
}
```

Icemap: But if you want me to get rid of the parameter change in the `newConfig`, please let me know, and I'll delete this logic.

internal/tools/tidb/tidbexecutesql/tidbexecutesql.go
Icemap: Hi @Yuan325. For now, I think they are similar. I just added some convenient features for TiDB's users, like supporting `ssl` parameter and auto-setting it is `true` for the TiDB Cloud URL, etc. 
However, since TiDB has more interesting features like `vector` data type, `auto_random` for randomizing the primary key, and auto-embedding, etc. For the next step, we will add those features into our `tidb-execute-sql` and `tidb-sql`, this PR is only a basis rather than just using `mysql-execute-sql` and `mysql-sql`.
Thanks for understanding.

Icemap:
Hi @Yuan325 . I've applied all your suggestions. You are such a great reviewer! Thanks for your collaboration. If I missed something, please feel free to leave more suggestions. Great work!

docs/en/resources/sources/tidb.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "TiDB"
+type: docs
+weight: 2
```
Yuan325:
```suggestion
weight: 1
```

docs/en/resources/tools/tidb/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "TiDB"
+type: docs
+weight: 2
```
Yuan325:
```suggestion
weight: 1
```

docs/en/resources/tools/tidb/tidb-execute-sql.md
```diff
@@ -0,0 +1,41 @@
+---
+title: "tidb-execute-sql"
+type: docs
+weight: 2
```
Yuan325:
```suggestion
weight: 1
```

internal/tools/tidb/tidbexecutesql/tidbexecutesql.go
Yuan325: Sounds good

internal/sources/tidb/tidb_test.go
Yuan325: just saw that, we might want to check/change it only once. It seems like we are checking&changing it twice (in `newConfig` and another time in `initTiDBConnectionPool`). Do u wanna remove the ones in `initTiDBConnectionPool` in that case?

tests/common.go
```diff
@@ -315,6 +315,29 @@ func AddMSSQLExecuteSqlConfig(t *testing.T, config map[string]any) map[string]an
 	return config
 }
 
+// AddTiDBExecuteSqlConfig gets the tools config for `tidb-execute-sql`
```
Yuan325: Please move this `AddTiDBExecuteSqlConfig` (and rename to non-export `addTiDBExecuteSqlConfig`) and the `GetTiDBWants` (rename as well to `getTiDBWants`) into `tidb_integration_test.go` since these are only used for tidb. :) Thanks!

tests/tidb/tidb_integration_test.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+var (
+	TiDBSourceKind = "tidb"
+	TiDBToolKind   = "tidb-sql"
+	TiDBDatabase   = os.Getenv("TIDB_DATABASE")
+	TiDBHost       = os.Getenv("TIDB_HOST")
+	TiDBPort       = os.Getenv("TIDB_PORT")
+	TiDBUser       = os.Getenv("TIDB_USERNAME")
+	TiDBPass       = os.Getenv("TIDB_PASSWORD")
```
Yuan325:
```suggestion
	TiDBUser       = os.Getenv("TIDB_USER")
	TiDBPass       = os.Getenv("TIDB_PASS")
```

Yuan325:
hi @Icemap Is there a way to host tidb in a vm without having to run `tiup playground`? We will be running the integration test with a trigger, so ideally, not be running `tidb playground` everytime when the integration test is run.

Also, not sure if you are aware of this, I created a vm with google cloud and installed `tiup` in it, however, every time I attempt to run `tiup playground`, the VM would freeze.

tests/common.go
Icemap: Sure. 
1. Deleted the logic to change `UseSSL` to true for TiDB Cloud in `initTiDBConnectionPool`
2. Moved `AddTiDBExecuteSqlConfig` and `GetTiDBWants` to `tidb_integration_test.go`
3. Renamed function `AddTiDBExecuteSqlConfig` to `addTiDBExecuteSqlConfig`
4. Renamed function `GetTiDBWants` to `getTiDBWants`

Icemap:
> hi @Icemap Is there a way to host tidb in a vm without having to run `tiup playground`? We will be running the integration test with a trigger, so ideally, not be running `tidb playground` everytime when the integration test is run.
> 
> Also, not sure if you are aware of this, I created a vm with google cloud and installed `tiup` in it, however, every time I attempt to run `tiup playground`, the VM would freeze.

Hi, @Yuan325. `tiup playground` is a foreground command. You can use `nohup tidb playground  > tidb.log 2>&1 &` to change it to a background service.
However, as you can see, it is a ***playground*** literally, which is not a good idea to run for a long time. You can follow the second half of the [previous document](https://docs.pingcap.com/tidb/stable/quick-start-with-tidb/#simulate-production-deployment-on-a-single-machine) that you saw, which is the "Simulate production deployment on a single machine" section. 

Another thing that I want to bring out is that the default password of TiDB is `""`, but it is not allowed for our `tidb-execute-sql` and `tidb-sql` because of security purposes. After the creation of TiDB, you might need to run:

```bash
mysql -u root -h 127.0.0.1 -P 4000 -e "CREATE USER 'google-genai-toolbox'@'%' IDENTIFIED BY 'google-genai-toolbox'; GRANT ALL PRIVILEGES ON *.* TO 'google-genai-toolbox'@'%';"
```

And then, set the new username and password to run the test cases:

```bash
export TIDB_USER='google-genai-toolbox'
export TIDB_PASS='google-genai-toolbox'
```

Yuan325:
Hi @Icemap I tried to set up the topo.yaml in my VM, and replaced all the host with my VM's internal IP. However, when I try to execute the cluster deploymnet, I get the following error: `Error: failed to fetch cpu-arch or kernel-name: executor.ssh.execute_failed: Failed to execute command over SSH for 'root@10.128.0.41:22' {ssh_stderr: , ssh_stdout: , ssh_command: export LANG=C; PATH=$PATH:/bin:/sbin:/usr/bin:/usr/sbin; uname -m}, cause: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain`. Are there anything else I need to update in topo.yaml?

^ I updated my ssh passcode and used that to authenticate the deployment.

When I try `tiup cluster check ./topo.yaml`, it says `none of ssh password, identity file, SSH_AUTH_SOCK specified (tui.id_read_failed)` but I don't see that in the example template provided in the docs.

Icemap:
Hi @Yuan325 . I think there is an authentication error. You can follow this [document](https://docs.pingcap.com/tidb/stable/production-deployment-using-tiup/#step-4-run-the-deployment-command). And maybe try different ways, like using secret keys.

Another thing is, for the default `topo.yaml`, it uses `"tidb"` this user to deploy the cluster. You can use `--user root` to specify the username. It looks like:

```
tiup cluster check ./topo.yaml --user root -p
tiup cluster deploy tidb-test v8.5.0 ./topo.yaml --user root -p
```

BTW, we have a useful tool, [tidb.ai](https://tidb.ai/), by using LLM and GraphRAG to solve TiDB's questions. You can have a try.

Yuan325:
@Icemap Unfortunately I am still unable to deploy it in a GCP VM. Wondering if you and your team had tried to simulate production deployment on a Google Cloud VM before and was successful in deploying it?

Icemap:
Hi @Yuan325, I got your point. I will deploy a simulated production deployment of a TiDB cluster on a Google Cloud VM and make a detailed document for you.

Icemap:
BTW, we tested GCP for sure. And I think we deployed a dozen clusters for our customers on GCP.

Yuan325:
@Icemap Got it, will keep you posted!

internal/tools/tidb/tidbsql/tidbsql.go
```diff
@@ -0,0 +1,202 @@
...
+	TemplateParameters tools.Parameters `yaml:"templateParameters"`
+	AllParams          tools.Parameters `yaml:"allParams"`
+
+	Pool        *sql.DB
+	Statement   string
+	manifest    tools.Manifest
+	mcpManifest tools.McpManifest
+}
+
+func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error) {
```
Yuan325, line 126: This will have to be updated to match `/tools/mysql/mysqlsql/mysqlsql.go` instead of the `mysqlexecutesql`. Thank you!

Yuan325 (changes requested):
Hi @Icemap I was able to set up the instance and run the integration test! seems like it is failing right now. Left a comment :)

internal/tools/tidb/tidbsql/tidbsql.go
Yuan325: Let me know when this is updated! :) I'll re-run the integration tests.

Icemap: Got it. Updated. Many thanks for the detailed prompts!

Icemap:
That's cool! @Yuan325 It's a great move, and I'll add more features to enrich these tools later.

Yuan325:
@Icemap Hi! We found that the TiDB's log has been taking a big space from the VM and often hit disk full state. 

I've tried to add the following config in my `topo.yaml`:
```
server_configs:
  tidb:
    log.file.max-days: 1
    log.file.max-backups: 1

monitoring_servers:
  - host: 10.128.0.2
    storage_retention: 1d
```

However, my `~/data/tidb-deploy/prometheus-9090/data/tidb-data` still consist of datas generated from 6 days ago. I'm wondering if that's not the right configuration to set prometheus storage retention? 

Thank you!

Icemap:
Got it. Let me loop in our o11y engineers in this PR.

Icemap:
Here is a solution that you can use to delete the extra data immediately, which is restarting Prometheus with an extra parameter. @Yuan325 https://ask.pingcap.com/t/how-to-adjust-the-data-retention-period-for-monitoring-servers/9439
And then, we can check why the TiUP's config didn't sync up to Prometheus.

Yuan325:
@Icemap Got it. Yeah, when I check `ps -ef|grep prometheus`, it shows `--storage.tsdb.retention=30d`, which is different from what I set in topo.yaml (2d).

My prometheus service only start with tiup, how do I restarting just that with the extra parameter (`tiup cluster restart <cluster-name> -R prometheus` won't allow to add prometheus specific flags)? I'm assuming we should not be restarting prometheus with the `systemctl`. If you have the command line that'll be great.

PR #830, opened 2025-07-09
feat: add queryTimeout support to MySQL source

kym6464:
## Summary
- Added configurable query timeout to MySQL source configuration
- Updated connection DSN to include readTimeout parameter  
- Added documentation and example usage
- Added test coverage for queryTimeout configuration

## Test plan
- [x] Added unit test for queryTimeout configuration parsing
- [x] Updated documentation with queryTimeout field description
- [x] Verified timeout parameter is correctly added to DSN when specified

## Caveat

When queryTimeout is exceeded, we get an obscure error message ([screenshot](https://github.com/user-attachments/assets/fd292f91-328d-4ebc-9a87-2d92e9887300)): 
```
unable to execute query: invalid connection
```

This seems to be a problem with the mysql-go library: https://stackoverflow.com/q/65253798/10720618

I tried to use [MAX_EXECUTION_TIME](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-execution-time) but it didn't work as expected (my `sleep(MAX_EXECUTION_TIME+3)` query finished successfully after MAX_EXECUTION_TIME milliseconds)

Any ideas on what can be done here? The error message is very misleading. My goal with adding timeouts is to communicate to the LLM when it has issued a slow query and force it to adjust (e.g. query indexes and write a more optimized query) but this defeats the purpose.


duwenxin99:
Hi @kym6464, thanks for putting in a PR. The timeout field looks good to me. For your usage of `MAX_EXECUTION_TIME`, have you tried executing an actual long query instead of using `sleep()`? I'm guessing that MySQL only interrupts query processes that are actively consuming I/O as suggested in their error message: `Query execution was interrupted, maximum statement execution time exceeded`. Could you try using a SELECT statement with `MAX_EXECUTION_TIME` and see it if works?

duwenxin99 (approved):

kym6464:
@duwenxin99 Looks like your guess is correct! I issued a more realistic query and received the following error:
```
SQL Error [3024] [HY000]: Query execution was interrupted, maximum statement execution time exceeded
```

Should I change the approach to use MAX_EXECUTION_TIME instead? Since this would only work for SELECT statements, I think renaming `queryTimeout` to `selectTimeout` would make more sense as well

duwenxin99:
> @duwenxin99 Looks like your guess is correct! I issued a more realistic query and received the following error:
> 
> ```
> SQL Error [3024] [HY000]: Query execution was interrupted, maximum statement execution time exceeded
> ```
> 
> Should I change the approach to use MAX_EXECUTION_TIME instead? Since this would only work for SELECT statements, I think renaming `queryTimeout` to `selectTimeout` would make more sense as well

Oh I was thinking that you needed `MAX_EXECUTION_TIME` for your application. For our MySQL source, it's definitely better to have a timeout parameter that applies to all queries instead of just the SELECT statements. Let's keep it this way. If you don't have other changes to make, I will merge this PR for you. Thank you!

PR #847, opened 2025-07-10
feat: add Dataplex source and tool

### LINKED ISSUE #831: [source] Add support for Dataplex source/tools
### Prerequisites

[Google Cloud's Dataplex](https://cloud.google.com/dataplex) is an intelligent data fabric that allows organizations to centrally discover, manage, and govern their data across various storage systems, such as data lakes, data warehouses, and databases. It provides a unified interface to control data without the need for data movement, enabling businesses to unlock the value of their data for analytics and artificial intelligence.

### Suggested Solution(s)

Add source / tools for connecting to Dataplex. 

### Alternatives Considered

_No response_

### Additional Details

_No response_

AnujJhunjhunwala:
- Users have the preference to choose their clients. Below example is using Gemini CLI.

- Users can use the pre-built Dataplex tools by creating a settings.json file under .gemini directory. The contents of settings.json would be as follows:-

```
{
  "mcpServers": {
    "dataplex": {
      "command": "./toolbox",
      "args": ["--prebuilt","dataplex","--stdio"],
      "env": {
          "DATAPLEX_PROJECT": "test-project"
      }
    }
  }
}
```

Fixes #831


docs/en/resources/tools/dataplex/dataplex-search-entries.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Example
+
+```yaml
+tools:
+  dataplex-search-entries:
+    kind: dataplex-search-entries
+    source: my-dataplex-source
+    description: Use this tool to get all the entries based on user query.
+    parameters:
```
duwenxin99: If the `query` parameter is required, we can initialize this param in Tool initialization so that users don't need to configure it. An example implementation is our `postgres-execute-sql` tool that takes in a SQL string, so we initialize a `sql` param here: https://github.com/googleapis/genai-toolbox/blob/2b69700c5e40ad37eafd94f378a2e1174879402b/internal/tools/postgres/postgresexecutesql/postgresexecutesql.go#L85

docs/en/resources/tools/dataplex/dataplex-search-entries.md
```diff
@@ -0,0 +1,41 @@
...
+title: "dataplex-search-entries"
+type: docs
+weight: 1
+description: > 
+  A "dataplex-search-entries" tool returns all entries in Dataplex Catalog.
+aliases:
+- /resources/tools/dataplex-search-entries
+---
+
+## About
```
duwenxin99, line 11: Can we have a section that mentions dataplex's authentication requirements? Example: [AlloyDB doc](https://github.com/googleapis/genai-toolbox/blob/main/docs/en/resources/sources/alloydb-pg.md#requirements). Since the client uses ADC to connect, we want to make sure these requirements clear to the users.

duwenxin99 (changes requested):
Hi @AnujJhunjhunwala, I left a few comments for improvements. Everything else looks good. Could you also add integration tests following this [guide](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md#3-add-integration-tests)? Thank you!

docs/en/resources/tools/dataplex/dataplex-search-entries.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Reference
+
+| **field**   |                  **type**                  | **required** | **description**                                                                                  |
+|-------------|:------------------------------------------:|:------------:|--------------------------------------------------------------------------------------------------|
+| kind        |                   string                   |     true     | Must be "dataplex-search-entries".                                                               |
+| source      |                   string                   |     true     | Name of the source the tool should execute on.                                                   |
+| description |                   string                   |     true     | Description of the tool that is passed to the LLM.                                               |
+| parameters  | [parameters](_index#specifying-parameters) |     true     | List of [parameters](_index#specifying-parameters) that will be passed                           |
```
duwenxin99:
```suggestion
```

docs/en/resources/tools/dataplex/dataplex-search-entries.md
AnujJhunjhunwala: This parameter "query" is a required parameter for the API being called in the tool invocation. But "query" should be optional from users perspective. If its not defined by the user in tools.yaml file explicitly, LLM would be responsible to restructure user prompt and pass it to query.

Maybe I am missing some code to make it optional?

```
sources:
 my-dataplex-source:
   kind: dataplex
   project: YOUR_PROJECT_ID

tools:
 find-recently-created-tables:
   kind: dataplex-search-entries
   source: my-dataplex-source
   parameters:
      query: createdAt>2025-07-01 type=table <- Optional; if user doesn't provides it, LLM will try to formulate the query using description of query parameter
   

toolsets:
 my_dataplex_toolset:
   - find-recently-created-tables
```

docs/en/resources/tools/dataplex/dataplex-search-entries.md
AnujJhunjhunwala: Fixed [bdb5bac](https://github.com/googleapis/genai-toolbox/pull/847/commits/bdb5bac1e8260439b5c1f922281083d5b82e60eb)

docs/en/resources/tools/dataplex/dataplex-search-entries.md
AnujJhunjhunwala: Added [bdb5bac](https://github.com/googleapis/genai-toolbox/pull/847/commits/bdb5bac1e8260439b5c1f922281083d5b82e60eb)

docs/en/resources/tools/dataplex/dataplex-search-entries.md
AnujJhunjhunwala: Initialized here [2131908](https://github.com/googleapis/genai-toolbox/pull/847/commits/21319087da3695118c3148d6542d9dba19f6689b)
Doc update [a65aa47](https://github.com/googleapis/genai-toolbox/pull/847/commits/a65aa4757142fb28cabe6fd09b25b819e5790e40)

docs/en/resources/sources/dataplex.md
```diff
@@ -0,0 +1,31 @@
...
+    kind: "dataplex"
+    project: "my-project-id"
+```
+
+## Reference
+
+| **field** | **type** | **required** | **description**                                                               |
+|-----------|:--------:|:------------:|-------------------------------------------------------------------------------|
+| kind      |  string  |     true     | Must be "dataplex".                                                           |
+| project   |  string  |     true     | Id of the GCP project (e.g. "my-project-id").                                 |
```
estshy: What exactly is this project being used for? Dataplex search tool should work across all projects by default. 

Is it a project used for quota and billing? Can we be more specific?

estshy: If the project is being used as search scope parameter then we should also allow orgId.

https://cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations/searchEntries

.gitignore
```diff
@@ -14,3 +14,8 @@ node_modules
 
 # coverage
 .coverage
+
```
AnujJhunjhunwala, line 20: These are specific. It landed here due to my directory structure on local. I would remove these. Thanks for pointing out. [99fc5ce](https://github.com/googleapis/genai-toolbox/pull/847/commits/99fc5ce796e57b00b0bdccc99ae63c7ff1dbf668)

docs/en/resources/sources/dataplex.md
AnujJhunjhunwala: Project id here is for quota and billing. I will update it. [c6b4c92](https://github.com/googleapis/genai-toolbox/pull/847/commits/c6b4c92920cbab0de864ebbf04353f9d3d067b56)

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (t *SearchTool) Authorized(verifiedAuthServices []string) bool {
+	return tools.IsAuthorized(t.AuthRequired, verifiedAuthServices)
+}
+
+func (t *SearchTool) Invoke(ctx context.Context, params tools.ParamValues) (any, error) {
+	paramsMap := params.AsMap()
+	query, _ := paramsMap["query"].(string)
+
+	req := &dataplexpb.SearchEntriesRequest{
```
estshy, line 137: This doesn't work as today. It misses scope param.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -0,0 +1,160 @@
...
+		Query: query,
+		Name:  fmt.Sprintf("projects/%s/locations/global", t.ProjectID),
+	}
+
+	it := t.CatalogClient.SearchEntries(ctx, req)
+	var results []*dataplexpb.SearchEntriesResult
+	for {
+		entry, err := it.Next()
```
estshy, line 153: When there are no search results then MCP inspector timeouts. I have observed similar problem in github copilot.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
AnujJhunjhunwala: As discussed, scope param defaults to the organization where the project provided in "name" is located. So, it should not be a part of parameters.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -0,0 +1,171 @@
...
+	paramsMap := params.AsMap()
+	query, _ := paramsMap["query"].(string)
+	name, _ := paramsMap["name"].(string)
+	pageSize, _ := paramsMap["pageSize"].(int32)
+	pageToken, _ := paramsMap["pageToken"].(string)
+	orderBy, _ := paramsMap["orderBy"].(string)
+
+	req := &dataplexpb.SearchEntriesRequest{
+		Query:     query,
+		Name:      name,
```
duwenxin99: If the user doesn't provide a `Name`, we should use the source's `ProjectID` and `Location` to construct the Name path?

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
duwenxin99: How do we plan on adding the scope param?

docs/en/resources/tools/dataplex/dataplex-search-entries.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+## About
+
+A `dataplex-search-entries` tool returns all entries in Dataplex Catalog (e.g. tables, views, models) that matches given user query.
+It's compatible with the following sources:
+
+- [dataplex](../sources/dataplex.md)
+
+dataplex-search-entries requires a `query` parameter as input based on which entries are filtered and returned to the user.
```
duwenxin99: We should also document the other parameters, though they are optional.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -0,0 +1,171 @@
...
+		PageSize:  pageSize,
+		PageToken: pageToken,
+		OrderBy:   orderBy,
```
duwenxin99: Not a must but what's our thoughts on supporting the `SemanticSearch` field here?

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
duwenxin99: Does it happen all the time? We didn't seem to have that issue previously with other DBs.

tests/dataplex/dataplex_integration_test.go
```diff
@@ -0,0 +1,177 @@
...
+				break
+			}
+		}
+		if !found {
+			t.Fatalf("expected parameter %q not found in tool parameters", want)
+		}
+	}
+}
+
+func runDataplexSearchEntriesToolInvokeTest(t *testing.T) {
```
duwenxin99: We should use a test table to test for different input/output.

tests/dataplex/dataplex_integration_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	runDataplexSearchEntriesToolGetTest(t)
```
duwenxin99, line 105: It's great to have a dedicated Get test for dataplex-specific fields. Let's also call the generic [ToolGetTest](https://github.com/googleapis/genai-toolbox/blob/2a650349cbda5ed24efffd87898b3947a06f595d/tests/tool.go#L31) here just to make it more secure.

duwenxin99 (changes requested):

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
AnujJhunjhunwala: We will not be adding scope param here. It defaults to the organization of the project id provided in the name parameter.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
AnujJhunjhunwala: If user doesn't provide a `Name`, we are using source's `ProjectID` to construct Name path. It's initialized as default value.

docs/en/resources/tools/dataplex/dataplex-search-entries.md
AnujJhunjhunwala: Updated [cc2383c](https://github.com/googleapis/genai-toolbox/pull/847/commits/cc2383c077aec96df1665b4bbe196db4f68f2e61)

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
AnujJhunjhunwala: `SemanticSearch` is not enabled in client library so we can not use it yet.

tests/dataplex/dataplex_integration_test.go
AnujJhunjhunwala: I understand this but search entries tool is a tool which lists "entries" (tables, views, datasets, etc) in json format. The query parameter it accepts is not a regular SQL query. It has particular [search syntax](https://cloud.google.com/dataplex/docs/search-syntax). Therefore it seems that a test table does not fit here.

tests/dataplex/dataplex_integration_test.go
AnujJhunjhunwala: Since we are setting default parameters and don't allow additional parameters, it won't work with the search entries tool.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
duwenxin99: It was included in the [latest release](https://pkg.go.dev/cloud.google.com/go/dataplex/apiv1/dataplexpb#SearchEntriesRequest)

AnujJhunjhunwala: `SemanticSearch` available in v1.26.0. Added [ef6c739](https://github.com/googleapis/genai-toolbox/pull/847/commits/ef6c739bb5f70abfdc200754e99620675a31cd1e)

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -84,7 +84,8 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 	pageSize := tools.NewIntParameterWithDefault("pageSize", 100, "Number of results in the search page.")
 	pageToken := tools.NewStringParameterWithDefault("pageToken", "", "Page token received from a previous locations.searchEntries call. Provide this to retrieve the subsequent page.")
 	orderBy := tools.NewStringParameterWithDefault("orderBy", "relevance", "Specifies the ordering of results. Supported values are: relevance, last_modified_timestamp, last_modified_timestamp asc")
-	parameters := tools.Parameters{query, name, pageSize, pageToken, orderBy}
+	semanticSearch := tools.NewBooleanParameterWithDefault("semanticSearch", true, "Whether to use semantic search for the query. If true, the query will be processed using semantic search capabilities.")
```
duwenxin99, line 87: Should we default it to false?

AnujJhunjhunwala: I would imagine this to be true in future. But, it would be a good idea to stay consistent for now. Setting default as false. [6a26b2a](https://github.com/googleapis/genai-toolbox/pull/847/commits/6a26b2a4eb91345616cfeeda7ab668788d268bc6)

.gitignore
estshy: Ok

docs/en/resources/sources/dataplex.md
estshy: @AnujJhunjhunwala Please correct me if I am wrong but I think we use organization of this project as search scope. Can we write it down?

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
estshy: I cannot reproduce anymore.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
estshy: We should have it default as true. Keyword search is default one to avoid introducing a breaking change. We are working on new functionally so we do not have to maintain backward compatibility. 

Semantic search is way much better fit for our use case.

docs/en/resources/sources/dataplex.md
AnujJhunjhunwala: I added details about the search scope in search entries doc. [08f61fd](https://github.com/googleapis/genai-toolbox/pull/847/commits/08f61fd9e9f2920510e8aa84dd184738bd16da80)

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
AnujJhunjhunwala: Fixed [eaa825c](https://github.com/googleapis/genai-toolbox/pull/847/commits/eaa825c41446f0f37a17f0c28e09fd90e7cea267)

tests/dataplex/dataplex_integration_test.go
AnujJhunjhunwala: Added test table [39cc428](https://github.com/googleapis/genai-toolbox/pull/847/commits/39cc428a2b9791269f6df4d947bf02efde5b21dd)

docs/en/resources/tools/dataplex/dataplex-search-entries.md
```diff
@@ -0,0 +1,63 @@
...
+It's compatible with the following sources:
+
+- [dataplex](../sources/dataplex.md)
+
+`dataplex-search-entries` takes a required `query` parameter based on which entries are 
+filtered and returned to the user and a required `name` parameter which is constructed using source's project if user does not provide it explicitly and has the following format: projects/{project}/locations/global. It also optionally accepts following parameters:
+  - `pageSize` - Number of results in the search page.
+  - `pageToken` - Page token received from a previous locations.searchEntries call.
+  - `orderBy` - Specifies the ordering of results. Supported values are: relevance (default), last_modified_timestamp, last_modified_timestamp asc
+  - `semanticSearch` - Specifies whether the search should understand the meaning and intent behind the query, rather than just matching keywords.
```
duwenxin99:
```suggestion
  - `semanticSearch` - Specifies whether the search should understand the meaning and intent behind the query, rather than just matching keywords. Defaults to `true`.
```

docs/en/resources/tools/dataplex/dataplex-search-entries.md
```diff
@@ -0,0 +1,63 @@
...
+## About
+
+A `dataplex-search-entries` tool returns all entries in Dataplex Catalog (e.g. tables, views, models) that matches given user query.
+It's compatible with the following sources:
+
+- [dataplex](../sources/dataplex.md)
+
+`dataplex-search-entries` takes a required `query` parameter based on which entries are 
+filtered and returned to the user and a required `name` parameter which is constructed using source's project if user does not provide it explicitly and has the following format: projects/{project}/locations/global. It also optionally accepts following parameters:
+  - `pageSize` - Number of results in the search page.
```
duwenxin99:
```suggestion
  - `pageSize` - Number of results in the search page. Defaults to `100`.
```

AnujJhunjhunwala: Added [8ea3516](https://github.com/googleapis/genai-toolbox/pull/847/commits/8ea35166bdb5ad4de245822f67f4526e254e98dc)

docs/en/resources/tools/dataplex/dataplex-search-entries.md
AnujJhunjhunwala: Added [8ea3516](https://github.com/googleapis/genai-toolbox/pull/847/commits/8ea35166bdb5ad4de245822f67f4526e254e98dc)

tests/dataplex/dataplex_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+				break
+			}
+		}
+		if !found {
+			t.Fatalf("expected parameter %q not found in tool parameters", want)
+		}
+	}
+}
+
+func runDataplexSearchEntriesToolInvokeTest(t *testing.T, tableName string, datasetName string) {
```
duwenxin99, line 236: Can we do something like this so that we also check for the content of the entry?
```func TestDataplexSearchEntriesToolInvoke(t *testing.T) {
	testCases := []struct {
		name           string
		tableName      string
		wantStatusCode int
		expectResult   bool
		wantContentKey string
	}{
		{
			name:           "Success - Entry Found",
			tableName:      "my_existing_table",
			wantStatusCode: http.StatusOK,
			expectResult:   true,
			wantContentKey: "dataplex_entry",
		},
		{
			name:           "Failure - Entry Not Found",
			tableName:      "non_existent_table",
			wantStatusCode: http.StatusOK,
			expectResult:   false,
			wantContentKey: "",
		},
	}

	for _, tt := range testCases {
		t.Run(tt.name, func(t *testing.T) {
			query := fmt.Sprintf(`displayname=%s SYSTEM=bigquery`, tt.tableName)
			reqBodyMap := map[string]string{"query": query}
			reqBodyBytes, err := json.Marshal(reqBodyMap)
			if err != nil {
				t.Fatalf("failed to marshal request body: %v", err)
			}

			resp, err := http.Post("http://127.0.0.1:5000/api/tool/my-search-entries-tool/invoke", "application/json", bytes.NewBuffer(reqBodyBytes))
			if err != nil {
				t.Fatalf("error making POST request: %s", err)
			}
			defer resp.Body.Close()

			if resp.StatusCode != tt.wantStatusCode {
				t.Fatalf("want status code %d, got %d", tt.wantStatusCode, resp.StatusCode)
			}

			var result map[string]interface{}
			if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
				t.Fatalf("error parsing response body: %v", err)
			}

			resultStr, ok := result["result"].(string)
			if !ok {
				if result["result"] == nil && !tt.expectResult {
					return
				}
				t.Fatalf("expected 'result' field to be a string, got %T", result["result"])
			}

			if !tt.expectResult && (resultStr == "" || resultStr == "[]") {
				return
			}

			var entries []interface{}
			if err := json.Unmarshal([]byte(resultStr), &entries); err != nil {
				t.Fatalf("error unmarshalling result string: %v", err)
			}

			if tt.expectResult {
				if len(entries) == 0 {
					t.Fatal("expected at least one entry, but got 0")
				}
				entry, ok := entries[0].(map[string]interface{})
				if !ok {
					t.Fatalf("expected first entry to be a map, got %T", entries[0])
				}
				if _, ok := entry[tt.wantContentKey]; !ok {
					t.Fatalf("expected entry to have key '%s', but it was not found in %v", tt.wantContentKey, entry)
				}
			} else {
				if len(entries) != 0 {
					t.Fatalf("expected 0 entries, but got %d", len(entries))
				}
			}
		})
	}
}```

Comment (AnujJhunjhunwala) [reply tests/dataplex/dataplex_integration_test.go:236]: Done [74c5bcf](https://github.com/googleapis/genai-toolbox/pull/847/commits/74c5bcfe4e41e7fe8bcbbafad1ccbaf89e8f51a6)

File: tests/dataplex/dataplex_integration_test.go [hunk 17]
```diff
@@ -0,0 +1,318 @@
...
+	}
+	return map[string]any{
+		"kind":    DataplexSourceKind,
+		"project": DataplexProject,
+	}
+}
+
+// Copied over from bigquery.go
+func initBigQueryConnection(project string) (*bigqueryapi.Client, error) {
+	ctx := context.Background()
```
kurtisvg: should this take in a context as an argument?

kurtisvg (approved):

tests/dataplex/dataplex_integration_test.go
AnujJhunjhunwala: context is initialized as context.Background(). Seems to be the recommended way in tests https://pkg.go.dev/context#Background

bq integration tests had similar implementation - https://github.com/googleapis/genai-toolbox/blob/c600c30374443b6106c1f10b60cd334fd202789b/tests/bigquery/bigquery_integration_test.go#L59

Moreover this client was just used to create the table and dataset in runtime.

kurtisvg: The point of context is that it should terminate if the test does -- since this is being called from other tests that use context we should just borrow theirs so if the test is canceled or timed out for whatever reason it terminates correctly.

AnujJhunjhunwala: I get your point now, added [7fac35e](https://github.com/googleapis/genai-toolbox/pull/847/commits/7fac35e6cb38d6204881b6994e0631c4b4657e96)

PR #864, opened 2025-07-11
feat: Add intial set of firestore tools

trehanshakuntG:
This PR adds the following tools for firestore source:

1. Firestore get documents
2. Firestore delete documents
3. Firestore list collections

Please note : the integration tests would be added in the coming PRs, this PR is being merged to the feature branch.


internal/tools/firestore/firestoredeletedocuments/firestoredeletedocuments.go
```diff
@@ -0,0 +1,192 @@
...
+	Parameters   tools.Parameters `yaml:"parameters"`
+
+	Client      *firestoreapi.Client
+	manifest    tools.Manifest
+	mcpManifest tools.McpManifest
+}
+
+func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
+	mapParams := params.AsMap()
+	documentPathsRaw, ok := mapParams[documentPathsKey].([]any)
```
duwenxin99, line 120: Can we use a type switch here to check if `documentPathsRaw` is a []string instead, so that we don't need line 121-136

internal/tools/firestore/firestoredeletedocuments/firestoredeletedocuments.go
```diff
@@ -0,0 +1,192 @@
...
+	bulkWriter.End()
+
+	// Collect results
+	results := make([]any, len(documentPaths))
+	for i, job := range jobs {
+		docData := make(map[string]any)
+		docData["path"] = documentPaths[i]
+
+		// Wait for the job to complete and get the result
+		_, err := job.Results()
```
duwenxin99, line 166: Why do we construct our own result map instead of returning the original results here?

internal/tools/firestore/firestoregetdocuments/firestoregetdocuments.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	// Convert snapshots to response data
+	results := make([]any, len(snapshots))
+	for i, snapshot := range snapshots {
+		docData := make(map[string]any)
+		docData["path"] = documentPaths[i]
+		docData["exists"] = snapshot.Exists()
+
+		if snapshot.Exists() {
```
duwenxin99, line 159: If snapshot doesn't exist for a document, should we add a line that says "Document at path .. doesn't exist" in the result?

docs/en/resources/tools/firestore/firestore-list-collections.md
```diff
@@ -0,0 +1,38 @@
...
+---
+
+## About
+
+A `firestore-list-collections` tool lists collections in Firestore, either at the root level or as subcollections of a specific document.
+It's compatible with the following sources:
+
+- [firestore](../sources/firestore.md)
+
+`firestore-list-collections` takes an optional `parentPath` parameter to specify a document
```
duwenxin99, line 18: should we rename it to "firestore-get-doc-subcollections" for clarity? Open to discuss.

internal/tools/firestore/firestorelistcollections/firestorelistcollections.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) ([]any, error) {
+	mapParams := params.AsMap()
+
+	var collectionRefs []*firestoreapi.CollectionRef
+	var err error
+
+	// Check if parentPath is provided
+	parentPath, hasParent := mapParams[parentPathKey].(string)
```
duwenxin99, line 126: Let's stick to Golang's [happy path to the left edge](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) here.
```
if !hasParent{
    // list the root collection
}
// List the user input doc's subcollections

File: docs/en/resources/tools/firestore/firestore-list-collections.md [hunk 6]
```diff
@@ -0,0 +1,38 @@
...
+path. If provided, it lists all subcollections of that document. If not provided, it lists
```
duwenxin99, line 19: Let's add a hyperlink to "subcollections" here: https://firebase.google.com/docs/firestore/data-model

duwenxin99 (changes requested):
Could you add integration test for each tool? Thanks!

internal/tools/firestore/firestoredeletedocuments/firestoredeletedocuments.go
trehanshakuntG: For now it's used to mark if the operation was a success or failure, in the future, we might add more verbose fields for the agent to understand the error better.

internal/tools/firestore/firestoregetdocuments/firestoregetdocuments.go
trehanshakuntG: The above two lines handles it, docData has the path and exists as false.

internal/tools/firestore/firestorelistcollections/firestorelistcollections.go
trehanshakuntG: In this case both the cases are happy paths, it's either root collection or subcollection, inverting the if, would still need an else block.

docs/en/resources/tools/firestore/firestore-list-collections.md
trehanshakuntG: I believe 'firestore-list-collections' is preferable for its generality. Since a subcollection is fundamentally a type of collection, using the broader term 'collection' ensures the name remains accurate for both root level collections and subcollections.

docs/en/resources/tools/firestore/firestore-list-collections.md
trehanshakuntG: Done.

internal/tools/firestore/firestoredeletedocuments/firestoredeletedocuments.go
trehanshakuntG: I did try this, but Go doesn't allow direct conversion of any[] to string[].

trehanshakuntG: Using `ConvertAnySliceToTyped` as suggested, thanks!

internal/tools/firestore/firestoredeletedocuments/firestoredeletedocuments.go
trehanshakuntG: job.Results only returns update time if the operation was successful, else throws and error, hence we need to create our own result map to store the success/failure for the given path.

docs/en/resources/tools/firestore/firestore-list-collections.md
duwenxin99: I think the problem here is that it doesn't really list collections with any paths. It only list the sub-collections for a document. If the user tries to provide a collection path, it will raise an error. That's why I think we need to make it clear in the name that it only gets collections for a document.

trehanshakuntG: This intends to fetch root level collection or subcollections. A collection/subcollection can't have a collection within it directly, only a doc or the root level can have a collection. I want to keep this name since it is consistent with what firebase provides in it's tools list.

duwenxin99: Ok that makes more sense to me. I had a misunderstanding that you could have a collection within another collection. Thanks for clarifying that.

duwenxin99 (approved):

PR #868, opened 2025-07-12
docs: update mcp docs to include v2025-03-26

Yuan325:


Yuan325:
Yes, will be in a separate PR that updates all the default SSE url to streamable http.

PR #874, opened 2025-07-13
feat(sources/mssql): add support for encrypt connection parameter

ShanQincheng:
## 1. Why do we need to support the `encrypt` parameter?
MSSQL databases that `genai-toolbox` attempts to connect to may have their encryption levels set differently. For example, a testing/demo purpose MSSQL database may not require encryption at all. However, `genai-toolbox` currently uses the default encryption parameter (`encrypt=false`) to connect to this type of database and will throw an error:
```
ERROR "toolbox failed to initialize: unable to initialize configs: unable to initialize source "my-mssql-source": unable to connect successfully: TLS Handshake failed: cannot read handshake packet: EOF"
```
> In this case, the encryption parameter should be set to `encrypt=disable`.

## 2. Is this a necessary feature?
`genai-toolbox` uses the `github.com/microsoft/go-mssqldb` package as a dependency to connect to MSSQL databases. According to the [README](https://github.com/microsoft/go-mssqldb?tab=readme-ov-file#common-parameters) of the `github.com/microsoft/go-mssqldb` package, `encrypt` is one of the common parameters. Therefore, I believe supporting the `encrypt` parameter in `genai-toolbox` is necessary.

### 3.1 Use this `docker-compose.yaml` file to start a demo MSSQL instance
```
services:
  demo-mssql-database:
    image: mcr.microsoft.com/mssql/server:2017-CU1-ubuntu
    ports:
      - "20256:1433"
    environment:
      ACCEPT_EULA: "Y"
      MSSQL_SA_PASSWORD: "hellopassword!"
    restart: unless-stopped
    healthcheck:
      test: ["CMD", "/opt/mssql-tools/bin/sqlcmd", "-S", "localhost", "-U", "sa", "-P", "hellopassword!", "-Q", "SELECT 1"]
      interval: 5s
      retries: 10

  demo-mssql-database-init:
    image: mcr.microsoft.com/mssql/server:2017-CU1-ubuntu
    network_mode: service:demo-mssql-database
    command: >
      /bin/bash -c "/opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P hellopassword! -d master -Q 'CREATE DATABASE DemoDatabase;'"
    depends_on:
      demo-mssql-database:
        condition: service_healthy
```

### 3.2 Use `genai-toolbox` to connect to the above demo MSSQL database with this `tools.yaml` configuration file:
```
sources:
       my-mssql-source:
                kind: mssql
                host: localhost
                port: 20256
                database: master
                user: sa
                password: 'hellopassword!'
```

### 3.3 We shall see the error:
```
ERROR "toolbox failed to initialize: unable to initialize configs: unable to initialize source "my-mssql-source": unable to connect successfully: TLS Handshake failed: cannot read handshake packet: EOF"
```


docs/en/resources/sources/mssql.md
```diff
@@ -35,6 +35,7 @@ sources:
         database: my_db
         user: ${USER_NAME}
         password: ${PASSWORD}
+        encrypt: strict
```
Yuan325:
```suggestion
        # encrypt: strict
```
Comment out since this is an optional field.

internal/sources/mssql/mssql.go
```diff
@@ -54,6 +54,9 @@ type Config struct {
 	User     string `yaml:"user" validate:"required"`
 	Password string `yaml:"password" validate:"required"`
 	Database string `yaml:"database" validate:"required"`
+
+	// Encrypt allowed values can reference the README of the github.com/microsoft/go-mssqldb package
+	Encrypt string `yaml:"encrypt"`
```
Yuan325:
```suggestion
	Encrypt string `yaml:"encrypt"`
```

internal/sources/mssql/mssql_test.go
```diff
@@ -124,3 +124,75 @@ func TestFailParseFromYaml(t *testing.T) {
 		})
 	}
 }
+
+// TestParseEncryptField checks that the encrypt field is parsed correctly from YAML.
+func TestParseEncryptField(t *testing.T) {
```
Yuan325: For testing parsing, can we just add a new case in `TestParseFromYamlMssql()` above that includes setting the encrypt value? We will only need one test case to make sure that the value is parsed (instead of testing all 3 values).

internal/sources/mssql/mssql_test.go
```diff
@@ -124,3 +124,75 @@ func TestFailParseFromYaml(t *testing.T) {
...
+				t.Fatalf("unexpected config type")
+			}
+			if cfg.Encrypt != tc.want {
+				t.Errorf("Encrypt: got %q, want %q", cfg.Encrypt, tc.want)
+			}
+		})
+	}
+}
+
+// TestMissingEncryptField checks that the encrypt field is optional and defaults to empty.
```
Yuan325: Do we need to test this? This should be covered in the first test case in `TestParseFromYamlMssql()` where the `encrypt` field is not set.

Yuan325 (changes requested):
@ShanQincheng Thanks for submitting this PR! :) Added some feedback. Please re-request reviews once the updates are made.

docs/en/resources/sources/mssql.md
ShanQincheng: Commented out this field.

internal/sources/mssql/mssql.go
ShanQincheng: Removed code comment

internal/sources/mssql/mssql_test.go
ShanQincheng: Added a new test case in `TestParseFromYamlMssql()` for the `Encrypt` field. And removed the redundant `TestParseEncryptField()`.

internal/sources/mssql/mssql_test.go
ShanQincheng: Removed the `TestMissingEncryptField()` 

> I was thinking that explicitly adding unit tests for new features wouldn’t add much overhead (aside from increased test runtime) but could save time when understanding or modifying existing unit tests. Just my personal thought.

docs/en/resources/sources/mssql.md
```diff
@@ -44,11 +45,12 @@ instead of hardcoding your secrets into the configuration file.
...
+| port      |  string  |     true     | Port to connect to (e.g. "1433").                                                                                                                                                          |
+| database  |  string  |     true     | Name of the SQL Server database to connect to (e.g. "my_db").                                                                                                                              |
+| user      |  string  |     true     | Name of the SQL Server user to connect as (e.g. "my-user").                                                                                                                                |
+| password  |  string  |     true     | Password of the SQL Server user (e.g. "my-password").                                                                                                                                      |
+| encrypt   |  string  |    false     | Encryption level for data transmitted between the client and server (e.g., "strict"). If not specified, defaults to the `github.com/microsoft/go-mssqldb` package's default encrypt value. |
```
kurtisvg: Should we list the potential options here?

internal/sources/mssql/mssql.go
```diff
@@ -101,14 +102,25 @@ func (s *Source) MSSQLDB() *sql.DB {
...
 	//nolint:all // Reassigned ctx
 	ctx, span := sources.InitConnectionSpan(ctx, tracer, SourceKind, name)
 	defer span.End()
 
 	// Create dsn
 	query := url.Values{}
 	query.Add("database", dbname)
+	if encrypt != "" {
+		query.Add("encrypt", encrypt)
+	}
```
kurtisvg, line 122: Can we validate the encrypt field is correct?

docs/en/resources/sources/mssql.md
ShanQincheng: The potential options are listed in the README of the `github.com/microsoft/go-mssqldb` dependency and controlled by that project. I suppose we shouldn’t list them here because their README should be the Single Source of Truth and it’s hard to keep our documentation in sync with the dependency .

internal/sources/mssql/mssql.go
ShanQincheng: The encrypt option is passed to the github.com/microsoft/go-mssqldb dependency for handling and validation. Whether an encrypt value is valid is controlled by that dependency, not by our code. I suppose we can’t validate the encrypt field here, as doing so would force us to keep our validation in sync with the dependency.

Yuan325 (approved):
Thank you @ShanQincheng !

PR #879, opened 2025-07-13
feat: add support for DuckDB

### LINKED ISSUE #861: <brief summary of the proposed feature>
### Prerequisites

The current mcp toolbox has no integration for my duckdb database. It would be very handy if it has support for duckdb.

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

bpranava:
Fixes #861 
This PR adds support for DuckDB which is a free, open-source, embedded, in-process, relational database management system (RDBMS) designed for analytical processing (OLAP)


averikitsch:
Hi @bpranava, Thanks for the PR. I will be reviewing it shortly!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/googleapis/genai-toolbox
 
-go 1.23.8
+go 1.24
```
averikitsch, line 3: Please use "1.23.8"

docs/en/resources/sources/duckdb.md
```diff
@@ -0,0 +1,69 @@
...
+
+- In-process, serverless database engine
+- Supports complex SQL queries for analytical processing
+- Can operate on in-memory or persistent storage
+- Zero-configuration - no external dependencies or server setup required
+- Highly optimized for columnar data storage and query execution
+
+For more details, refer to the [DuckDB Documentation](https://duckdb.org/).
+
+## Requirements
```
averikitsch:
```suggestion


## Available Tools

- [`duckdb-sql`](../tools/duckdb/duckdb-sql.md)  
  Execute pre-defined prepared SQL queries in DuckDB.
  
## Requirements
```

docs/en/resources/tools/duckdb/duckdb.md
```diff
@@ -0,0 +1,80 @@
+---
+title: DuckDB SQL Tools
```
averikitsch:
```suggestion
title: duckdb-sql
```

docs/en/resources/tools/duckdb/duckdb.md
averikitsch, line 1: please update this file name to be duckdb-sql to match the tool name

internal/sources/duckdb/duckdb.go
```diff
@@ -0,0 +1,132 @@
...
+
+func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources.SourceConfig, error) {
+	actual := Config{Name: name}
+	if err := decoder.DecodeContext(ctx, &actual); err != nil {
+		return nil, err
+	}
+	return actual, nil
+}
+
+type DuckDbSource struct {
```
averikitsch:
```suggestion
type Source struct {
```

averikitsch: Please update everywhere

internal/sources/duckdb/duckdb.go
```diff
@@ -0,0 +1,132 @@
...
+	for key, value := range duckdbConfiguration {
+		if !first {
+			configStr.WriteString("&")
+		}
+		configStr.WriteString(url.QueryEscape(key))
+		configStr.WriteString("=")
+		configStr.WriteString(url.QueryEscape(value))
+		first = false
+	}
+	return configStr.String()
```
averikitsch: You could potentially simplify this code and use the built in url.Values to handle escaping:
```
	// If there are no parameters, just return the path.
	// This also fixes a bug in the original where it would return "path?".
	if len(duckdbConfiguration) == 0 {
		return dbFilePath
	}

	params := url.Values{}
	for key, value := range duckdbConfiguration {
		params.Set(key, value)
	}

	// A strings.Builder is still efficient for the final assembly.
	var b strings.Builder
	b.WriteString(dbFilePath)
	b.WriteString("?")
	b.WriteString(params.Encode()) // .Encode() handles all formatting and escaping.

	return b.String()
```

go.mod
bpranava: @averikitsch , the latest version of duckdb seems to be using 1.24. I have downgraded the duckdb go client version to 1.8.4

internal/sources/duckdb/duckdb.go
bpranava: Thanks for the review @averikitsch!

docs/en/resources/sources/duckdb.md
bpranava: Just a suggestion...since this is a document for the duckdb source, would it be better to add the "available tools" section at the end?

internal/sources/duckdb/duckdb.go
bpranava: updated

docs/en/resources/tools/duckdb/duckdb.md
bpranava: updated

go.mod
bpranava: @averikitsch, I just found out that the integration test for [array parameters](https://github.com/googleapis/genai-toolbox/pull/879/files#diff-950cc182085ed10599df700058f9f2fd837e556eb5c9925df02a32faa35fcbcfR123) fails when we use lower versions of duckdb go client.
The error message is as below
```
unable to execute query: sql: converting argument $1 type: unsupported type []interface {}, a slice of interface"}
```

On looking deeper I found that the support for array parameters was introduced only after go-client version [2.3.1 ](https://github.com/marcboeker/go-duckdb/releases/tag/v2.3.1). Ref - https://github.com/marcboeker/go-duckdb/pull/423

To reproduce the error, try:
```
package main

import (
	"context"
	"database/sql"
	"fmt"
	"log"

	_ "github.com/marcboeker/go-duckdb"
)

func main() {
	// Create a context
	ctx := context.Background()

	db, err := sql.Open("duckdb", "")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	// Create table using QueryContext
	rows, err := db.QueryContext(ctx, `CREATE TABLE people (id INTEGER, name VARCHAR)`)
	if err != nil {
		log.Fatal(err)
	}
	rows.Close()

	// Insert multiple data rows for demonstration
	_, err = db.ExecContext(ctx, `INSERT INTO people VALUES (42, 'John'), (43, 'Jane'), (44, 'Bob')`)
	if err != nil {
		log.Fatal(err)
	}

	// Prepare parameterized query with ANY clause
	// Using a parameterized subquery with an array of values
	validIDs := []int{42, 44} // Example input values
	query := `SELECT id, name FROM people WHERE id = ANY ($1)`
	rows, err = db.QueryContext(ctx, query, validIDs)
	if err != nil {
		log.Fatal(err)
	}
	defer rows.Close()

	// Get column names
	cols, err := rows.Columns()
	if err != nil {
		log.Fatal(err)
	}

	values := make([]any, len(cols))
	valuePtrs := make([]any, len(cols))
	for i := range values {
		valuePtrs[i] = &values[i]
	}

	// Prepare the result slice
	var result []map[string]interface{}
	// Iterate through the rows
	for rows.Next() {
		// Scan the row into the value pointers
		if err := rows.Scan(valuePtrs...); err != nil {
			log.Fatal(err)
		}

		// Create a map for this row
		rowMap := make(map[string]interface{})
		for i, col := range cols {
			val := values[i]
			// Handle nil values
			if val == nil {
				rowMap[col] = nil
				continue
			}
			// Store the value in the map
			rowMap[col] = val
		}
		result = append(result, rowMap)
	}

	// Print the results from ANY query
	fmt.Println("Results from SELECT with parameterized ANY clause:")
	for _, row := range result {
		fmt.Printf("id: %v, name: %v\n", row["id"], row["name"])
	}
}
```

The fix would be to go with the latest version of go-client v.2.3.3

But it looks like the [latest version uses go version 1.24](https://github.com/marcboeker/go-duckdb/blob/bfe08b2e9006a5de8b2f83e6ad740dd5f5480a48/go.mod#L3)

bpranava: Let me know your thoughts on the same @averikitsch

docs/en/resources/sources/duckdb.md
averikitsch: This is the current pattern to be consistent with our other docs.

bpranava: okay

docs/en/resources/sources/duckdb.md
```diff
@@ -0,0 +1,73 @@
...
+## Example
+
+For a persistent DuckDB database:
+
+```yaml
+sources:
+    my-duckdb:
+        kind: "duckdb"
+        dbFilePath: "/path/to/database.db"
+        configurations:
```
averikitsch: nit: What do you think about using `config:` to make this less verbose?

tests/duckdb/duckdb_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+	DuckDbKind = "duckdb-sql"
+	dbPath     = "/tmp/users.db"
+)
+
+func getDuckDbVars() map[string]any {
+	return map[string]any{
+		"kind":       "duckdb",
+		"dbFilePath": dbPath,
+		"configurations": map[string]any{
+			"access_mode": "READ_WRITE",
```
averikitsch, line 42: Is access_mode a required setting?

tests/duckdb/duckdb_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	select1Want, failInvocationWant, _ := GetDuckDbWants()
+
+	invokeParamWant, invokeParamWantNull, nullWant, mcpInvokeParamWant := GetDuckDbInvokeParamWant()
+	tests.RunToolInvokeTest(t, select1Want, invokeParamWant, invokeParamWantNull, nullWant, true, true)
+	tests.RunMCPToolCallMethod(t, mcpInvokeParamWant, failInvocationWant)
+	tests.RunToolInvokeWithTemplateParameters(t, tableNameTemplateParam, tests.DuckDBTemplateParameterTestConfig())
+}
+
+func GetDuckDbParamToolInfo(tableName string) (string, string, string, string, string, string, []any) {
```
averikitsch, line 127: If this is similar to Postgres, can we reuse the functions "GetPostgresSQLParamToolInfo" and "GetPostgresSQLAuthToolInfo"?

tests/duckdb/duckdb_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	tests.RunToolGetTest(t)
+
+	select1Want, failInvocationWant, _ := GetDuckDbWants()
+
+	invokeParamWant, invokeParamWantNull, nullWant, mcpInvokeParamWant := GetDuckDbInvokeParamWant()
+	tests.RunToolInvokeTest(t, select1Want, invokeParamWant, invokeParamWantNull, nullWant, true, true)
+	tests.RunMCPToolCallMethod(t, mcpInvokeParamWant, failInvocationWant)
+	tests.RunToolInvokeWithTemplateParameters(t, tableNameTemplateParam, tests.DuckDBTemplateParameterTestConfig())
```
averikitsch: Instead of creating a new TestConfig, see https://github.com/googleapis/genai-toolbox/blob/92845c943a6b1b1a5f2da216cea6cc7da1b313dc/tests/spanner/spanner_integration_test.go#L166-L170 for how to customize tests.NewTemplateParameterTestConfig().

.ci/integration.cloudbuild.yaml
```diff
@@ -487,6 +487,24 @@ steps:
...
+    name: golang:1
+    waitFor: ["compile-test-binary"]
+    entrypoint: /bin/bash
+    env:
+      - "GOPATH=/gopath"
+      - "SERVICE_ACCOUNT_EMAIL=$SERVICE_ACCOUNT_EMAIL"
+    volumes:
+      - name: "go"
+        path: "/gopath"
+    secretEnv: ["CLIENT_ID"]
```
averikitsch, line 500: Client_id and service account shouldn't be required here.

internal/tools/duckdb/duckdb.go
averikitsch, line 1: Please update this filename and packagename to be duckdbsql to match the tool name.

internal/tools/duckdb/duckdb.go
```diff
@@ -0,0 +1,210 @@
...
+	mcpManifest tools.McpManifest
+}
+
+// Authorized implements tools.Tool.
+func (d Tool) Authorized(verifiedAuthSources []string) bool {
+	return tools.IsAuthorized(d.AuthRequired, verifiedAuthSources)
+}
+
+// Invoke implements tools.Tool.
+func (d Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error) {
```
averikitsch: for codebase consistency, please use `t Tool`.

averikitsch (commented):
I apologize, I had to do my review in 2 parts. I am consulting with my team on the 1.24 go version update and I will get back to you shortly.

docs/en/resources/sources/duckdb.md
bpranava: I am ok with both.

tests/duckdb/duckdb_integration_test.go
bpranava: No it's optional. All configurations are optional. Access mode by default is READ_ONLY. Since I am reading as well as writing in my test cases I need to set it to READ_WRITE

.ci/integration.cloudbuild.yaml
bpranava: I have 2 auth tool tests. Isn't this required for them to run?

docs/en/resources/sources/duckdb.md
bpranava: https://github.com/googleapis/genai-toolbox/blob/669a8731db2a431ad16b9e742d5360b19c507e2e/internal/sources/duckdb/duckdb.go#L64-L68

should I call it DuckDbConfig considering the struct is already named Config?

tests/duckdb/duckdb_integration_test.go
bpranava: For GetPostgresSQLParamToolInfo  I wanted to add some DuckDb specific function tests. Looks like I have added only one unique case - EXCLUDE . I can add one or 2 more. That way it would be unique test for duckdb. What do you feel?

bpranava:
@averikitsch  apologies, I force pushed after a rebase.. will avoid it in the future

internal/tools/duckdb/duckdb.go
bpranava: updated

go.mod
averikitsch: Thanks for your patience and the context! We can use 1.24.

tests/duckdb/duckdb_integration_test.go
bpranava: thank you!

docs/en/resources/sources/duckdb.md
bpranava: https://duckdb.org/docs/stable/configuration/overview.html#local-configuration-options
The duckdb documentation calls it Configuration. Hence I went with the same.

tests/duckdb/duckdb_integration_test.go
bpranava: https://duckdb.org/docs/stable/sql/functions/overview - has the list of unique functions supported by duckdb. I will update the PR with few of them

bpranava: GetPostgresSQLAuthToolInfo cannot be reused for DuckDB since it does not support the SERIAL keyword. So I had to use different queries.
https://github.com/googleapis/genai-toolbox/blob/18392b49ddae2d4b7e784fd4c829663982240c22/tests/duckdb/duckdb_integration_test.go#L129-L131
https://github.com/googleapis/genai-toolbox/blob/18392b49ddae2d4b7e784fd4c829663982240c22/tests/common.go#L331-L334

go.mod
bpranava: Thanks for confirming

tests/duckdb/duckdb_integration_test.go
bpranava: updated the PR to reuse GetNonSpannerInvokeParamWant and also updated GetDuckDbParamToolInfo

.ci/integration.cloudbuild.yaml
averikitsch: Apologies, you are correct

tests/tool.go
```diff
@@ -951,3 +959,24 @@ func runRequest(t *testing.T, method, url string, body io.Reader, header map[str
...
+		insert1Want:    "[{\"Count\":1}]",
+	}
+
+	// Apply provided options
+	for _, option := range options {
+		option(templateParamTestOption)
+	}
+
+	return templateParamTestOption
+}
```
averikitsch: I don't believe this is being used anymore. Please remove.

tests/duckdb/duckdb_integration_test.go
averikitsch: This is great. Thank you!

docs/en/resources/sources/duckdb.md
averikitsch: (probably a bit of bike shedding) Can we name this "configuration" with no "s"? Thanks!

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -0,0 +1,80 @@
+---
+title: "dgraph-dql"
```
averikitsch:
```suggestion
title: "duckdb-sql"
```

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -0,0 +1,80 @@
...
+| source             | string                         |     true     | Name of the DuckDB source configuration (see [DuckDB source documentation](../sources/duckdb.md)).                                         |
+| description        | string                         |     true     | Description of the tool that is passed to the LLM.                                                                                         |
+| statement          | string                         |     true     | The SQL statement to execute.                                                                                                              |
+| authRequired       | []string                       |    false     | List of authentication requirements for the tool (if any).                                                                                 |
+| parameters         | [parameters](_index#specifying-parameters)       |    false     | List of parameters that will be inserted into the SQL statement                      |
```
averikitsch:
```suggestion
| parameters         | [parameters](../_index#specifying-parameters)       |    false     | List of parameters that will be inserted into the SQL statement                      |
```

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -0,0 +1,80 @@
...
+| templateParameters | [templateParameters](_index#template-parameters) |    false     | List of template parameters that will be inserted into the SQL statement before executing the prepared statement.                           |
```
averikitsch:
```suggestion
| templateParameters | [templateParameters](../_index#template-parameters) |    false     | List of template parameters that will be inserted into the SQL statement before executing the prepared statement.                           |
```

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -0,0 +1,80 @@
...
+        type: string
+        description: The name to search for
+      - name: min_age
+        type: integer
+        description: Minimum age
+```
+
+## Example with Template Parameters
+
+> **Note:** Template parameters allow direct modifications to the SQL statement, including identifiers, column names, and table names, which makes them more vulnerable to SQL injections. Using basic parameters (see above) is recommended for performance and safety. For more details, see the [templateParameters](_index#template-parameters) section.
```
averikitsch:
```suggestion
> **Note:** Template parameters allow direct modifications to the SQL statement, including identifiers, column names, and table names, which makes them more vulnerable to SQL injections. Using basic parameters (see above) is recommended for performance and safety. For more details, see the [templateParameters](../_index#template-parameters) section.
```

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -0,0 +1,80 @@
...
+weight: 1
+description: >
+  Execute SQL statements against a DuckDB database using the DuckDB SQL tools configuration.
+aliases:
+- /resources/tools/duckdb-sql
+---
+
+## About
+
+A `duckdb-sql` tool executes a pre-defined SQL statement against a [DuckDB](https://duckdb.org/) database. It is compatible with any DuckDB source configuration as defined in the [DuckDB source documentation](../sources/duckdb.md).
```
averikitsch:
```suggestion
A `duckdb-sql` tool executes a pre-defined SQL statement against a [DuckDB](https://duckdb.org/) database. It is compatible with any DuckDB source configuration as defined in the [DuckDB source documentation](../../sources/duckdb.md).
```

docs/en/resources/sources/duckdb.md
bpranava: makes sense. I have updated

tests/tool.go
bpranava: thanks! I have removed it

docs/en/resources/tools/duckdb/duckdb-sql.md
bpranava: thank you for catching this

bpranava:
@averikitsch there was one integration test failing, I have fixed that

bpranava:
Screenshot 2025-07-29 at 10 36 01 PM
I was able to fix the ../../sources/duckdb.md broken link. But I am not able to figure out the others...

Also one request..can we make it such that the lint check runs without /gcbrun?

bpranava:
@averikitsch I just saw your previous [PR](https://github.com/googleapis/genai-toolbox/pull/1014/files) where you have removed _index from links. I have followed the same now

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -44,7 +44,7 @@ tools:
 
 ## Example with Template Parameters
 
-> **Note:** Template parameters allow direct modifications to the SQL statement, including identifiers, column names, and table names, which makes them more vulnerable to SQL injections. Using basic parameters (see above) is recommended for performance and safety. For more details, see the [templateParameters](_index#template-parameters) section.
+> **Note:** Template parameters allow direct modifications to the SQL statement, including identifiers, column names, and table names, which makes them more vulnerable to SQL injections. Using basic parameters (see above) is recommended for performance and safety. For more details, see the [templateParameters](#template-parameters) section.
```
averikitsch:
```suggestion
> **Note:** Template parameters allow direct modifications to the SQL statement, including identifiers, column names, and table names, which makes them more vulnerable to SQL injections. Using basic parameters (see above) is recommended for performance and safety. For more details, see the [templateParameters](../#template-parameters) section.
```

docs/en/resources/tools/duckdb/duckdb-sql.md
```diff
@@ -72,9 +72,9 @@ tools:
...
 | authRequired       | []string                       |    false     | List of authentication requirements for the tool (if any).                                                                                 |
-| parameters         | [parameters](_index#specifying-parameters)       |    false     | List of parameters that will be inserted into the SQL statement                      |
-| templateParameters | [templateParameters](_index#template-parameters) |    false     | List of template parameters that will be inserted into the SQL statement before executing the prepared statement.                           |
+| parameters         | [parameters](../#specifying-parameters)       |    false     | List of parameters that will be inserted into the SQL statement                      |
+| templateParameters | [templateParameters](#template-parameters) |    false     | List of template parameters that will be inserted into the SQL statement before executing the prepared statement.                           |
```
averikitsch:
```suggestion
| templateParameters | [templateParameters](../#template-parameters) |    false     | List of template parameters that will be inserted into the SQL statement before executing the prepared statement.                           |
```

PR #881, opened 2025-07-14
feat: Add firestore query collection tool

trehanshakuntG:


duwenxin99 (approved):

docs/en/resources/tools/firestore/firestore-query-collection.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Example Usage
+
+### Query with filters
+
+```json
+{
+  "collectionPath": "users",
+  "filters": [
+    "{\"field\": \"age\", \"op\": \">\", \"value\": 18}",
```
duwenxin99, line 87: Do we really need to escape the quotes here?

trehanshakuntG: Yes, it is the exact payload that will be sent from the agent to the tool.

PR #885, opened 2025-07-14
feat: Create wait for tool

prernakakkar-google:
`
tools:
  wait_for_tool:
    kind: wait-for
    description: "Tool puts chat to sleep for given timeout example 3m,10s, 5m etc. For tasks such as cluster creation no need to wait. It takes longer time such as 8m for instance creation. Use timeout value as default if nothing is provided"
    timeout: 10s
`

Output: 
`
prernakakkar@prernakakkar:~/senseai/genai-toolbox$ curl -X POST -H "Content-Type: application/json" -d '{"duration": "5m"}' http://127.0.0.1:5000/api/tool/wa
it_for_tool/invoke
{"result":"[\"Wait for 5m0s completed successfully.\"]"}
`

### DIFF
--- a/docs/en/resources/tools/utility/wait-for.md
+++ b/docs/en/resources/tools/utility/wait-for.md
@@ -1,0 +1,36 @@
+---
+title: "wait-for"
+type: docs
+weight: 1
+description: > 
+  A "wait-for" tool pauses execution for a specified duration.
+aliases:
+- /resources/tools/utility/wait-for
+---
+
+## About
+
+A `wait-for` tool pauses execution for a specified duration. This can be useful in workflows where a delay is needed between steps.
+
+`wait-for` takes one input parameter `duration` which is a string representing the time to wait (e.g., "10s", "2m", "1h").
+
+> **Note:** This tool is intended for developer assistant workflows with
+> human-in-the-loop and shouldn't be used for production agents.
+
+## Example
+
+```yaml
+tools:
+  wait_for_tool:
+    kind: wait-for
+    description: Use this tool to pause execution for a specified duration.
+    timeout: 30s
+```
+
+## Reference
+
+| **field**   |                  **type**                  | **required** | **description**                                                                                  |
+|-------------|:------------------------------------------:|:------------:|--------------------------------------------------------------------------------------------------|
+| kind        |                   string                   |     true     | Must be "wait-for".                                                                     |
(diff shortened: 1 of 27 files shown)


internal/tools/utility/wait-for/waitfor.go
```diff
@@ -0,0 +1,120 @@
...
+	AuthRequired []string `yaml:"authRequired"`
+}
+
+var _ tools.ToolConfig = Config{}
+
+func (cfg Config) ToolConfigKind() string {
+	return kind
+}
+
+func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
```
trehanshakuntG: `srcs` can be replaced with `_` since it's not being used

docs/en/resources/tools/utility/wait-for.md
```diff
@@ -0,0 +1,36 @@
...
+- /resources/tools/utility/wait-for
+---
+
+## About
+
+A `wait-for` tool pauses execution for a specified duration. This can be useful in workflows where a delay is needed between steps.
+
+`wait-for` takes one input parameter `duration` which is a string representing the time to wait (e.g., "10s", "2m", "1h").
+
+> **Note:** This tool is intended for developer assistant workflows with
```
averikitsch: Use `{{% notice info %}}` and `{{% /notice %}}` to make this more visible. https://learn.netlify.app/en/shortcodes/notice/

docs/en/resources/tools/utility/wait-for.md
```diff
@@ -0,0 +1,36 @@
...
+> human-in-the-loop and shouldn't be used for production agents.
+
+## Example
+
+```yaml
+tools:
+  wait_for_tool:
+    kind: wait-for
+    description: Use this tool to pause execution for a specified duration.
+    timeout: 30s
```
averikitsch: In the future, we may want to make the timeout an optional input from the LLM with a default value

docs/en/resources/tools/utility/wait-for.md
```diff
@@ -0,0 +1,36 @@
+---
+title: "wait-for"
```
averikitsch: nit: maybe this should just be "wait". "wait-for" may have some additional connotations like being able to wait for events or other actions.

PR #889, opened 2025-07-14
test: integration tests for null optional parameters

Yuan325:
Note: null optional parameters currently doesn't work with BigQuery.

### DIFF
--- a/tests/tool.go
+++ b/tests/tool.go
@@ -76,7 +76,7 @@ func RunToolGetTest(t *testing.T) {
 }
 
 // RunToolInvoke runs the tool invoke endpoint
-func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeIdParamWant, nullString string, supportNullParam, supportsArray bool) {
+func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeIdNullWant, nullString string, supportNullParam, supportsArray bool) {
 	// Get ID token
 	idToken, err := GetGoogleIdToken(ClientId)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


tests/tool.go
```diff
@@ -109,13 +109,21 @@ func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeParamWa
 			isErr:         false,
 		},
 		{
-			name:          "invoke my-param-tool2 with nil response",
-			api:           "http://127.0.0.1:5000/api/tool/my-param-tool2/invoke",
+			name:          "invoke my-param-tool-id with nil response",
```
duwenxin99: Renaming `my-param-tool-id` to `my-get-id-tool`, `my-param-tool-name` to `my-name-tool`?
Feeling like these test names get too long and don't make sense anymore...

tests/tool.go
```diff
@@ -76,7 +76,7 @@ func RunToolGetTest(t *testing.T) {
 }
 
 // RunToolInvoke runs the tool invoke endpoint
-func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeParamWantNull string, supportsArray bool) {
+func RunToolInvokeTest(t *testing.T, select1Want, invokeParamWant, invokeIdParamWant, nullString string, supportNullParam, supportsArray bool) {
```
duwenxin99: If we are renaming `invokeParamWantNull` we should probably also rename it in each test files, otherwise the readability is decreased. Also, is it better to rename it "invokeIdWantNull" since it is testing null results as well? I think it is more descriptive, and also feeling the `Param` part isn't needed.

duwenxin99 (changes requested):

tests/tool.go
Yuan325: Updated `my-param-tool` --> `my-tool` , `my-param-tool-id` --> `my-tool-by-id`, `my-param-tool-name` --> `my-tool-by-name`. Since technically all 3 of them is accessing the same table. just with different parameters.

tests/tool.go
Yuan325: Updated to `invokeIdNullWant` since `Want` is the suffix for the variable.

PR #891, opened 2025-07-15
fix: Initialize logger in context for the traditional api just like mcp

drstrangelooker:


Yuan325 (approved):
Thank you! Seems like we have some tests to fix.

PR #895, opened 2025-07-15
feat(tools/oceanbase): Add oceanbase Source and Tool 

### LINKED ISSUE #894: [source] Add Oceanbase Source and Tool
### Prerequisites

Oceanbase database is an open source database of Ant and has some excellent features, and we should support it

### Suggested Solution(s)

Oceanbase database is an open source database of Ant and has some excellent features, and we should support it

### Alternatives Considered

In the future, you can enrich the support of vectors and OBKV on the basis of the Oceanbase plug-in

### Additional Details

Please refer to https://en.oceanbase.com/ for more detailed informatio

xxsc0529:
1.Added oceanbase Source and Tool
2.Add tests related to OceanBase
3.Add the documentation for Oceanbase
Close #894


cmd/root.go
```diff
@@ -58,6 +58,8 @@ import (
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqlexecutesql"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqlsql"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/neo4j"
+	_ "github.com/googleapis/genai-toolbox/internal/tools/oceanbase/oceanbaseexecutesql"
+	_ "github.com/googleapis/genai-toolbox/internal/tools/oceanbase/oceanbasesql"
```
duwenxin99, line 97: Could you also add `oceanbase` to the prebuilt flag description [here](https://github.com/googleapis/genai-toolbox/blob/bc3b31c59f7afac010ff24abaf9a20bc238156d1/cmd/root.go#L189). Thanks!

xxsc0529: hi @duwenxin99,I have added it, please review it, thanks

xxsc0529:
> Hi @xxsc0529, thanks for the contribution! Sorry for the delayed review as we got many integration requests lately. The PR LGTM overall with one minor change. Could you rebase on the latest main branch and then I'll run the integration test. Thanks!

Ok, I will resolve this conflict, thank you for reviewing it, oceanbase is an excellent database, support OBKV, vector storage, AP scenario and TP scenario, I will continue to maintain it, official website address: https://en.oceanbase.com

duwenxin99 (changes requested):
@xxsc0529 Could you remove .DS_Store and workspace.xml please? Thanks!

xxsc0529:
> @xxsc0529 Could you remove .DS_Store and workspace.xml please? Thanks!

hi @duwenxin99 ,I've deleted them

duwenxin99:
Hi @xxsc0529, could you elaborate on how OceanBase source/tools defer from the existing MySQL source/tools? Do you plan on adding more features in the future or maintaining it in the long term?

> LGTM, but we should confirm that we need this in addition to the MySQL source/tool since they are pretty much the same implementation.
> 
> What are we gaining by having ocean source if the existing MySQL source/tool is compatible?

xxsc0529:
> Hi @xxsc0529, could you elaborate on how OceanBase source/tools defer from the existing MySQL source/tools? Do you plan on adding more features in the future or maintaining it in the long term?
> 
> > LGTM, but we should confirm that we need this in addition to the MySQL source/tool since they are pretty much the same implementation.
> > What are we gaining by having ocean source if the existing MySQL source/tool is compatible?

Hi~@kurtisvg,@duwenxin99 , OceanBase Database is an enterprise-level, native distributed database independently developed by the OceanBase team.OceanBase Database has also set new records in the TPC-C benchmark test, supporting more than 1,500 nodes in one cluster. Our website: https://en.oceanbase.com/docs/oceanbase-database

OceanBase Database supports both MySQL tenant mode and Oracle tenant mode. The current submitted changes primarily enable MySQL tenant mode, which is why it appears very similar to MySQL's implementation. Support for Oracle tenant mode will be added in the future. However, even in MySQL tenant mode, there are differences from MySQL, as the underlying implementations are completely different. When connecting to an OceanBase MySQL tenant, a variety of annotations are supported, such as parallel reads, partitioned reads, and more. Additionally, support for advanced data types like Array, Map, and Vector is also provided.

duwenxin99:
> > Hi @xxsc0529, could you elaborate on how OceanBase source/tools defer from the existing MySQL source/tools? Do you plan on adding more features in the future or maintaining it in the long term?
> > > LGTM, but we should confirm that we need this in addition to the MySQL source/tool since they are pretty much the same implementation.
> > > What are we gaining by having ocean source if the existing MySQL source/tool is compatible?
> 
> Hi~@kurtisvg,@duwenxin99 , OceanBase Database is an enterprise-level, native distributed database independently developed by the OceanBase team.OceanBase Database has also set new records in the TPC-C benchmark test, supporting more than 1,500 nodes in one cluster. Our website: https://en.oceanbase.com/docs/oceanbase-database
> 
> OceanBase Database supports both MySQL tenant mode and Oracle tenant mode. The current submitted changes primarily enable MySQL tenant mode, which is why it appears very similar to MySQL's implementation. Support for Oracle tenant mode will be added in the future. However, even in MySQL tenant mode, there are differences from MySQL, as the underlying implementations are completely different. When connecting to an OceanBase MySQL tenant, a variety of annotations are supported, such as parallel reads, partitioned reads, and more. Additionally, support for advanced data types like Array, Map, and Vector is also provided.

That's good to hear. Are you willing to maintain the OceanBase source/tools in the long term? Our team will be responsible for the horizontal MCP features. However, if a user requests a new feature for the OceanBase source/tools specifically (adding a new OceanBase tool, adding new config fields etc), we do hope you will be responsible for implementing them.

kurtisvg:
@xxsc0529 Just to clarify my concern a bit more explicitly -- since OceanBase is MySQL compatible, what is the advantage of adding a new Source for OceanBase vs using the existing MySQL Source to connect? 

From a maintenance perspective, there's additional overhead in supporting what is essentially 2 copies of very similar code -- every fix we make to the MySQL source needs to be re-copied onto the OceanBase one to ensure it's working correctly. If the OceanBase stuff is really just the MySQL stuff, can we just tell users to use MySQL instead?

xxsc0529:
> @xxsc0529 Just to clarify my concern a bit more explicitly -- since OceanBase is MySQL compatible, what is the advantage of adding a new Source for OceanBase vs using the existing MySQL Source to connect?
> 
> From a maintenance perspective, there's additional overhead in supporting what is essentially 2 copies of very similar code -- every fix we make to the MySQL source needs to be re-copied onto the OceanBase one to ensure it's working correctly. If the OceanBase stuff is really just the MySQL stuff, can we just tell users to use MySQL instead?

Thank you for your response. I understand your concern. OceanBase's MySQL tenant mode is compatible with native MySQL, but it is not 100% compatible. At the same time, OceanBase is also a superset of MySQL. If users are only performing CRUD operations, using native MySQL is sufficient. However, when utilizing more advanced capabilities, the connection methods of native MySQL can no longer meet the requirements and need to be extended.

duwenxin99:
Hi @xxsc0529, thanks for the clarification. We are good with merging it as long as you are willing to maintain it in the long term. Could you fix the test errors so we can merge it? Thanks!

duwenxin99 (approved):

xxsc0529:
> Hi @xxsc0529, thanks for the clarification. We are good with merging it as long as you are willing to maintain it in the long term. Could you fix the test errors so we can merge it? Thanks!

Thank you, I've fixed this

PR #911, opened 2025-07-16
feat: invoke tools from Toolbox UI

AlexTalreja:
Adds `Run` button and functionality to modify parameters + test tools.

This currently supports both standard and optional params, but not auth.

### DIFF
--- a/internal/server/static/index.html
+++ b/internal/server/static/index.html
@@ -17,7 +17,7 @@
             const navbarContainer = document.getElementById('navbar-container');
             const activeNav = navbarContainer.getAttribute('data-active-nav');
             renderNavbar('navbar-container', activeNav);
-            renderMainContent('main-content-container', 'homepage-display-area')
+            renderMainContent('main-content-container', '')
         });
     </script>
 </body>
--- a/internal/server/static/js/mainContent.js
+++ b/internal/server/static/js/mainContent.js
@@ -24,11 +24,12 @@ function renderMainContent(containerId, idString) {
         return;
     }
 
+    const idAttribute = idString ? `id="${idString}"` : '';
     const contentHTML = `
         <div class="main-content-area">
         <div class="top-bar">
         </div>
-        <main class="content" id="${idString}">
+        <main class="content" ${idAttribute}">
             <h1>Welcome to MCP Toolbox UI</h1>
             <p>This is the main content area. Click a tab on the left to navigate.</p>
         </main>
--- a/internal/server/static/js/runTool.js
+++ b/internal/server/static/js/runTool.js
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { isParamIncluded } from "./toolDisplay.js";
+
 /**
  * Runs a specific tool using the /api/tools/toolName/invoke endpoint
  * @param {string} toolId The unique identifier for the tool.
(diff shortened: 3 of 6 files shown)


internal/server/static/index.html
```diff
@@ -17,7 +17,7 @@
             const navbarContainer = document.getElementById('navbar-container');
             const activeNav = navbarContainer.getAttribute('data-active-nav');
             renderNavbar('navbar-container', activeNav);
-            renderMainContent('main-content-container', '')
+            renderMainContent('main-content-container', 'homepage-display-area')
```
Yuan325: Do we need this? doesn't seem like we are using the idString anywhere else.

internal/server/static/js/runTool.js
```diff
@@ -0,0 +1,160 @@
...
+                console.debug(`Optional parameter ${NAME} is missing, skipping.`);
+                continue;
+            }
+
+            // process remaining types
+            if (VALUE_TYPE && VALUE_TYPE.startsWith('array<')) {
+                typedParams[NAME] = parseArrayParameter(RAW_VALUE, VALUE_TYPE, NAME);
+            } else {
+                switch (VALUE_TYPE) {
+                    case 'number':
```
Yuan325, line 56: Does it check against the manifest's `type` field? If that's the case, it will return string / integer / float / boolean / array. 

Seems like float type doesn't work in the ui. It will send the request with string input instead of a float.

Yuan325 (changes requested):

internal/server/static/js/runTool.js
AlexTalreja: It checked against a different field which represented the input type, but was essentially based on the manifest type. However previously `float` was not properly translated to the correct `number` type. Fixed.

internal/server/static/index.html
AlexTalreja: Originally it was just to prevent setting an `id=''` within the main content, however I've reverted this and instead updated the logic in `mainContent.js` so it handles an empty string appropriately.

internal/server/static/js/mainContent.js
```diff
@@ -24,11 +24,13 @@ function renderMainContent(containerId, idString) {
         return;
     }
 
+    const idAttribute = idString ? `id="${idString}"` : '';
+
```
Yuan325:
```suggestion
```

internal/server/static/js/runTool.js
```diff
@@ -0,0 +1,160 @@
...
+ * @param {!HTMLTextAreaElement} responseArea The textarea to display results or errors.
+ * @param {!Array<!Object>} parameters An array of parameter definition objects
+ * @param {!HTMLInputElement} prettifyCheckbox The checkbox to control JSON formatting.
+ * @param {function(?Object): void} updateLastResults Callback to store the last results.
+ */
+export async function handleRunTool(toolId, form, responseArea, parameters, prettifyCheckbox, updateLastResults) {
+    responseArea.value = 'Running tool...';
+    updateLastResults(null);
+    const formData = new FormData(form);
+    const typedParams = {};
```
Yuan325:
```suggestion
    const formData = new FormData(form);
    const typedParams = {};
    responseArea.value = 'Running tool...';
    updateLastResults(null);
```

internal/server/static/js/runTool.js
```diff
@@ -0,0 +1,160 @@
...
+
+        try {
+            if (VALUE_TYPE === 'boolean') {
+                typedParams[NAME] = RAW_VALUE !== null;
+                console.debug(`Parameter ${NAME} (boolean) set to: ${typedParams[NAME]}`);
+                continue; 
+            }
+
+            // handle missing values for non-boolean types
+            if (RAW_VALUE === null || RAW_VALUE === undefined || RAW_VALUE === '') {
```
Yuan325: What about cases where the parameter input is empty string? ""

internal/server/static/js/toolDisplay.js
```diff
@@ -79,6 +81,13 @@ export function renderToolInterface(tool, containerElement) {
     const TOOL_ID = tool.id;
     containerElement.innerHTML = '';
 
+    let lastResults = null;
+
+    // function to update lastResults so we can toggle json
+    const updateLastResults = (newResults) => {
```
Yuan325, line 155: Unsure of the usage for this function. Is this for the `Prettify JSON`? Seems like the toggle is still available even when we set `updateLastResults(null)`.

AlexTalreja: Yes, it's for the Prettify JSON functionality and so that within `toolDisplay.js` we can update and cache the result from the `handleRunTool` so it can be reformatted back and forth without any further calls.

internal/server/static/js/runTool.js
AlexTalreja: NOTE: Discussed offline with team due to challenges with deciding how to handle `null` vs `'""` which also affects optional parameters. 

Decided to add a checkbox for each param to indicate whether or not the param will be sent (default to checked), so users can clarify intentions of `null` vs `""`. An unchecked text field will send null, while a checked but blank text field will send the empty string.

Yuan325 (approved):

PR #920, opened 2025-07-17
feat: Add wait for operation tool with exponential backoff

prernakakkar-google:
Example:

```
  alloydb-operations-get:
    kind: wait-for-operation
    source: alloydb-api-source
    method: GET
    path: /v1/projects/{{.projectId}}/locations/{{.locationId}}/operations/{{.operationId}}
    description: "Makes API call to check whether operation is done or not. This tool is run first then wait for tool. if its still in create phase trigger it after 3 minutes.  Print a message saying still not done. We will notify once its done."
    pathParams:
      - name: projectId
        type: string
        description: The dynamic path parameter
      - name: locationId
        type: string
        description: The dynamic path parameter
        default: us-central1
      - name: operationId
        type: string
        description: Operation status check for previous task

```


docs/en/resources/tools/utility/waitforoperation.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Wait for Operation"
+type: docs
+weight: 10
+description: >
+  Wait for a long-running operation to complete.
+---
+
+The `wait-for-operation` tool is a utility tool that waits for a long-running operation to complete. It does this by polling an operation status endpoint until the operation is finished using exponential backoff.
```
averikitsch: Please add the same info block that we expect users to only use this for dev assist workflows.

docs/en/resources/tools/utility/waitforoperation.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Example
+
+```yaml
+tools:
+  alloydb-operations-get:
+    kind: wait-for-operation
+    source: alloydb-api-source
```
averikitsch: Are you adding a new source or can this be removed?

docs/en/resources/tools/utility/waitforoperation.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Wait for Operation"
```
averikitsch:
```suggestion
title: "wait-for-operation"
```

docs/en/resources/tools/utility/waitforoperation.md
```diff
@@ -0,0 +1,43 @@
...
+## Example
+
+```yaml
+tools:
+  alloydb-operations-get:
+    kind: wait-for-operation
+    source: alloydb-api-source
+    method: GET
+    path: /v1/projects/{{.projectId}}/locations/{{.locationId}}/operations/{{.operationId}}
+    description: "Makes API call to check whether operation is done or not using exponential backoff. if its still in create phase trigger it after 10 minutes timeout.  Print a message saying still not done. "
```
averikitsch: To improve the doc rendering, please update this to be a multi-line description

docs/en/resources/tools/utility/waitforoperation.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Example
+
+```yaml
+tools:
+  alloydb-operations-get:
+    kind: wait-for-operation
+    source: alloydb-api-source
+    method: GET
+    path: /v1/projects/{{.projectId}}/locations/{{.locationId}}/operations/{{.operationId}}
```
averikitsch: nit: we may want to provide more information about this path and potentially link to CGC documentation.

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,323 @@
...
+		return nil, fmt.Errorf("error populating path parameters: %s", err)
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
+	defer cancel()
+
+	delay := 3 * time.Second    // Initial delay
+	maxDelay := 4 * time.Minute // Maximum delay
+	multiplier := 2.0           // Exponential backoff multiplier
+	maxRetries := 10            // Maximum number of retries
```
averikitsch: Do we want these value to be configurable by the developer?

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,323 @@
...
+	retries := 0                // Current number of retries
+
+	client := *t.Client
```
averikitsch: Can we use t.client directly?

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,323 @@
...
+		resp, err := client.Do(req)
+		if err != nil {
+			fmt.Printf("error making HTTP request during polling: %s, retrying in %v\n", err, delay)
+			time.Sleep(delay)
+			delay = time.Duration(float64(delay) * multiplier)
+			if delay > maxDelay {
+				delay = maxDelay
+			}
+			retries++
+			continue
```
averikitsch: Can we group this retry logic with the same logic below?

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("unexpected status code during polling: %d, response body: %s", resp.StatusCode, string(body))
+		}
+
+		var data map[string]any
+		if err := json.Unmarshal(body, &data); err != nil {
+			// If not a JSON object, we can't check the condition.
+			// We'll just keep polling.
```
averikitsch: Can we remove this empty block by changing the conditional?

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// Initialize initializes the tool from the configuration.
+func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
+	rawS, ok := srcs[cfg.Source]
+	if !ok {
+		return nil, fmt.Errorf("no source named %q configured", cfg.Source)
+	}
+
+	s, ok := rawS.(*httpsrc.Source)
```
averikitsch: Will the user need to set up an HTTP source?

docs/en/resources/tools/utility/waitforoperation.md
prernakakkar-google: Added http source to add clarity.

docs/en/resources/tools/utility/waitforoperation.md
prernakakkar-google: This will differ from case to case and each database will have different documentation. I wanted to keep it generic, this one is example call. Let me know example you have in mind.

internal/tools/utility/wait-for-operation/waitforoperation.go
prernakakkar-google: yes the tool will work with http source. Added in documentation example for clarity.

internal/tools/utility/wait-for-operation/waitforoperation.go
prernakakkar-google: As per my understanding this may not be needed from user plus it will increase parameters and questions for user by agent

kurtisvg: They could be config options on the tool itself right? They don't need to be parameters on the tool

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,396 @@
...
+	region = parts[3]
+	cluster = parts[5]
+
+	if len(parts) >= 8 && parts[6] == "instances" {
+		instance = parts[7]
+	} else {
+		return "", false
+	}
+
+	var b strings.Builder
```
kurtisvg: As opposed to writing this out line by line, I recommend putting the output you want in a string constant and using template for format it: https://pkg.go.dev/text/template

internal/tools/utility/wait-for-operation/waitforoperation.go
```diff
@@ -0,0 +1,396 @@
...
+	Name         string            `yaml:"name" validate:"required"`
+	Kind         string            `yaml:"kind" validate:"required"`
+	Source       string            `yaml:"source" validate:"required"`
+	Description  string            `yaml:"description" validate:"required"`
+	AuthRequired []string          `yaml:"authRequired"`
+	Path         string            `yaml:"path" validate:"required"`
+	Method       tools.HTTPMethod  `yaml:"method" validate:"required"`
+	Headers      map[string]string `yaml:"headers"`
+	PathParams   tools.Parameters  `yaml:"pathParams"`
+	HeaderParams tools.Parameters  `yaml:"headerParams"`
```
kurtisvg: I don't think a generic tool for this makes sense -- it requires the Toolbox user to understand the path configure everything appropriately. Additionally, the output is hard coded for AlloyDB.

I think we should just have a "wait-for-instance-create" operation that has hardcoded path/params for now. If we need to abstract (e.g. for Cloud SQL) we can do so on the code level instead of the config level.

internal/tools/utility/wait-for-operation/waitforoperation.go
prernakakkar-google: If we are packaging them as part of prebuilt tools it would be hardcoded by us. Let me know if you feel otherwise.

kurtisvg: I want to both have the ability to package them as prebuilt tools as well as leave the door open for a user to be able to specify themselves. The ability to customize what tools are provided is a differentiator for Toolbox.

internal/tools/utility/wait-for-operation/waitforoperation.go
prernakakkar-google: Done

internal/tools/utility/wait-for-operation/waitforoperation.go
prernakakkar-google: Done

docs/en/resources/tools/utility/alloydbwaitforoperation.md
```diff
@@ -0,0 +1,47 @@
...
+    maxRetries: 10
+```
+
+## Reference
+
+| **field**   | **type** | **required** | **description**                                                                                                  |
+| ----------- | :------: | :----------: | ---------------------------------------------------------------------------------------------------------------- |
+| kind        |  string  |     true     | Must be "alloydb-wait-for-operation".                                                                            |
+| source      |  string  |     true     | Name of the source the HTTP request should be sent to.                                                           |
+| description |  string  |    false     | A description of the tool.                                                                                       |
```
averikitsch:
```suggestion
| description |  string  |    required     | A description of the tool.                                                                                       |
```

docs/en/resources/tools/utility/alloydbwaitforoperation.md
```diff
@@ -0,0 +1,47 @@
...
+## Example
+
+```yaml
+sources:
+  alloydb-api-source:
+    kind: http
+    baseUrl: https://alloydb.googleapis.com
+    headers:
+      Authorization: Bearer ${API_KEY}
+      Content-Type: application/json
```
averikitsch, line 24: Requiring users to properly set up this source may be a  point of friction. If we abstract away the path and parameters, we should so abstract away the source. I am fine with keeping the source so we can release quicker but since we are using an HTTP source the API key is required. I would like us to support ADC ie generate this token for them on their behalf.

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
```diff
@@ -0,0 +1,368 @@
...
+          "ALLOYDB_POSTGRES_USER": "<your-user>",
+          "ALLOYDB_POSTGRES_PASSWORD": "<your-password>"
+      }
+    }
+  }
+}
+` + "```" + `
+
+**If running remotely:**
+For remote deployments, you will need to set the following environment variables in your deployment configuration. It is recommended to manage these as secrets:
```
averikitsch: What does manage these as secrets mean here?

tests/utility/alloydb_wait_for_operation_test.go
averikitsch, line 1: For consistency, please add integration to the filename. To run these tests, please update https://github.com/googleapis/genai-toolbox/blob/33306b03a9eb2e6afef9ebcaa42577ea30fac780/.ci/integration.cloudbuild.yaml

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
```diff
@@ -0,0 +1,368 @@
...
+
+	maxRetries := cfg.MaxRetries
+	if maxRetries == 0 {
+		maxRetries = 10
+	}
+
+	baseURL := s.BaseURL
+	if baseURL == "" {
+		baseURL = "https://alloydb.googleapis.com"
+	}
```
averikitsch: BaseURL is a required, so this should never be empty. Do we want to do a check here?

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
```diff
@@ -0,0 +1,368 @@
...
+	if !ok {
+		return nil, fmt.Errorf("invalid or missing source for %q tool: source kind must be `http`", kind)
+	}
+
+	combinedHeaders := make(map[string]string)
+	maps.Copy(combinedHeaders, s.DefaultHeaders)
+	maps.Copy(combinedHeaders, cfg.Headers)
+
+	allParameters := tools.Parameters{
+		&tools.StringParameter{CommonParameter: tools.CommonParameter{Name: "project", Desc: "The project ID"}},
```
averikitsch: I believe these should use NewStringParameter not CommonParamter https://github.com/googleapis/genai-toolbox/blob/fa3e9ac04b52d2b05b90b092da5292788b77385f/internal/tools/parameters.go#L487

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
```diff
@@ -0,0 +1,368 @@
...
+	inputSchema.Required = []string{"project", "location", "operation_id"}
+
+	mcpManifest := tools.McpManifest{
+		Name:        cfg.Name,
+		Description: cfg.Description,
+		InputSchema: inputSchema,
+	}
+
+	delay, err := time.ParseDuration(cfg.Delay)
+	if err != nil {
```
averikitsch: If the user tries to set this but sets it wrong it will silently change the value. We should distinguish between setting the default and a parsing issue

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
```diff
@@ -0,0 +1,368 @@
...
+	retries := 0
+
+	for retries < maxRetries {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("timed out waiting for operation: %w", ctx.Err())
+		default:
+		}
+
+		req, _ := http.NewRequest(http.MethodGet, urlString, nil)
```
averikitsch: here or in the source we should set the user agent to track these requests

docs/en/resources/tools/utility/alloydbwaitforoperation.md
prernakakkar-google: Since we are already using http tools for create cluster and instance I think as first pass we would still need API KEY.

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
prernakakkar-google: Removed this to avoid confusion it ws suggested for created password.

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
prernakakkar-google: Done.

prernakakkar-google: Added for extra check since its not required for tool but required by source, so removing now.

docs/en/resources/tools/utility/alloydbwaitforoperation.md
prernakakkar-google: Made it true, Thanks for pointing out.

tests/utility/alloydb_wait_for_operation_test.go
prernakakkar-google: Done.

internal/tools/http/http.go
```diff
@@ -330,6 +331,12 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
 		req.Header.Set(k, v)
 	}
 
+	if ua, err := util.UserAgentFromContext(ctx); err == nil {
+		req.Header.Set("User-Agent", ua)
+	} else if req.Header.Get("User-Agent") == "" {
+		req.Header.Set("User-Agent", "http-tool")
+	}
```
averikitsch: We have to use the UserAgentFromContext for the user agent to be picked up in the pipeline. There is no need for the back up user agent.

internal/tools/utility/alloydb-wait-for-operation/alloydbwaitforoperation.go
```diff
@@ -248,6 +256,12 @@ func (t *Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error
 			req.Header.Set(k, v)
 		}
 
+		if ua, err := util.UserAgentFromContext(ctx); err == nil {
+			req.Header.Set("User-Agent", ua)
+		} else if req.Header.Get("User-Agent") == "" {
+			req.Header.Set("User-Agent", "alloydb-wait-for-operation")
+		}
```
averikitsch: Same comment here on user agent

.ci/integration.cloudbuild.yaml
```diff
@@ -445,8 +445,25 @@ steps:
...
+    secretEnv: ["CLIENT_ID"]
+    volumes:
+      - name: "go"
+        path: "/gopath"
+    args:
+      - -c
+      - |
+        .ci/test_with_coverage.sh \
+          "Alloydb Wait for Operation" \
+          alloydbwaitforoperation \
```
averikitsch:
```suggestion
          utility \
```

averikitsch: @Yuan325 can you confirm this is the right set up (https://github.com/googleapis/genai-toolbox/blob/main/.ci/test_with_coverage.sh)

docs/en/resources/tools/utility/alloydbwaitforoperation.md
```diff
@@ -0,0 +1,47 @@
...
+title: "alloydb-wait-for-operation"
+type: docs
+weight: 10
+description: >
+  Wait for a long-running AlloyDB operation to complete.
+---
+
+The `alloydb-wait-for-operation` tool is a utility tool that waits for a long-running AlloyDB operation to complete. It does this by polling the AlloyDB Admin API operation status endpoint until the operation is finished, using exponential backoff.
+
+{{% notice info %}}
```
averikitsch:
```suggestion
{{< notice info >}}
```

docs/en/resources/tools/utility/alloydbwaitforoperation.md
```diff
@@ -0,0 +1,47 @@
...
+This tool is intended for developer assistant workflows with human-in-the-loop and shouldn't be used for production agents.
+{{% /notice %}}
```
averikitsch:
```suggestion
{{< /notice >}}
```

averikitsch (approved):

Issue #915, opened 2025-07-17
BQ MCP toolbox - error returned even when the job is performed


Kylekr:
### Prerequisites

- [x] I've searched the current open issues
- [ ] I've updated to the latest version of Toolbox

### Toolbox version

toolbox version 0.8.0+binary.darwin.arm64.f13e9635ba64425b57805c8d64a5968ba569292c

### Environment

1. OS type and version: (output of `uname -a`)
: Mac, Win

3. How are you running Toolbox:   - As a downloaded binary (e.g. from `curl -O https://storage.googleapis.com/genai-toolbox/v$VERSION/linux/amd64/toolbox`)
  - As a container (e.g. from `us-central1-docker.pkg.dev/database-toolbox/toolbox/toolbox:$VERSION`)
  - Compiled from source (include the command used to build)
: Following this GCP docs for Cursor - MCP
: https://cloud.google.com/bigquery/docs/pre-built-tools-with-mcp-toolbox#mcp-configure-your-mcp-client-cursor

### Client

1. Client: <name and link to the client are you using>
2. Version: <what exact version of the client are you using> 
3. Example: If possible, please include your code of configuration:

```python
# Code goes here! 
```

### Expected Behavior

return appropriate value OR success message when the job is performed.

### Current Behavior

after connecting BQ MCP toolbox, It can perform `bigquery-execute-sql`, but some jobs such as DELETE, SELECT * from empty table returns error msg(Error: no result from tool. The user likely interrupted the tool call to send you a message.) but actually it worked when I checked it the table in BQ

### Steps to reproduce?

1. setup MCP Toolbox for BQ in Cursor based on this doc(https://cloud.google.com/bigquery/docs/pre-built-tools-with-mcp-toolbox#mcp-configure-your-mcp-client-cursor)
2. 
3. Asking Cursor to perform table row DELETE OR SELECT * to the empty table

Genesis929:
Hi @Kylekr , thank you for the detailed feedback! We are working on a fix and will update soon.

Genesis929:
I was unable to reproduce the issue during local testing. However, I noticed that we are returning null for non-SELECT queries. I suspect the LLM struggles to process null values, so we should modify this to always return a proper result, like a result string, instead. 

It's likely the problem can be fixed with this change.

PR #925, opened 2025-07-17
fix(tools/bigquery-execute-sql): ensure invoke always returns a non-null value

Genesis929:
- Added a dry run step to identify the query type (e.g., SELECT, DML), which allows the tool to correctly handle the query's output.
- The recommended high-level client, cloud.google.com/go/bigquery, does not expose the statement type from a dry run. To circumvent this limitation, the low-level BigQuery REST API client (google.golang.org/api/bigquery/v2) was added to gain access to these necessary details.

fixes: #915

### DIFF
--- a/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
+++ b/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
@@ -177,6 +177,9 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
 		}
 		out = append(out, vMap)
 	}
+	if len(out) == 0 {
+		return "The query returned 0 rows.", nil
+	}
 	return out, nil
 }
 
--- a/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
+++ b/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
@@ -177,7 +177,7 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
 		}
 		out = append(out, vMap)
 	}
-	if len(out) == 0 {
+	if out == nil {
 		return "The query returned 0 rows.", nil
 	}
 	return out, nil
--- a/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
+++ b/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
@@ -23,6 +23,7 @@ import (
 	"github.com/googleapis/genai-toolbox/internal/sources"
 	bigqueryds "github.com/googleapis/genai-toolbox/internal/sources/bigquery"
 	"github.com/googleapis/genai-toolbox/internal/tools"
+	bigqueryrestapi "google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/iterator"
 )
 
@@ -44,6 +45,7 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (tools.T
 
 type compatibleSource interface {
 	BigQueryClient() *bigqueryapi.Client
+	BigQueryRestService() *bigqueryrestapi.Service
 }
 
 // validate compatible sources are still compatible
@@ -95,6 +97,7 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 		Parameters:   parameters,
 		AuthRequired: cfg.AuthRequired,
 		Client:       s.BigQueryClient(),
+		RestService:  s.BigQueryRestService(),
 		manifest:     tools.Manifest{Description: cfg.Description, Parameters: parameters.Manifest(), AuthRequired: cfg.AuthRequired},
 		mcpManifest:  mcpManifest,
 	}
@@ -110,6 +113,7 @@ type Tool struct {
 	AuthRequired []string         `yaml:"authRequired"`
 	Parameters   tools.Parameters `yaml:"parameters"`
 	Client       *bigqueryapi.Client
+	RestService  *bigqueryrestapi.Service
 	manifest     tools.Manifest
 	mcpManifest  tools.McpManifest
 }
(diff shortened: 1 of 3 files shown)


internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -164,3 +195,22 @@ func (t Tool) McpManifest() tools.McpManifest {
 func (t Tool) Authorized(verifiedAuthServices []string) bool {
 	return tools.IsAuthorized(t.AuthRequired, verifiedAuthServices)
 }
+
+func dryRunQuery(ctx context.Context, restService *bigqueryrestapi.Service, projectID string, sql string) (*bigqueryrestapi.Job, error) {
```
Yuan325:
```suggestion

// dryRunQuery .... please add comment on what this function does
func dryRunQuery(ctx context.Context, restService *bigqueryrestapi.Service, projectID string, sql string) (*bigqueryrestapi.Job, error) {
```

tests/bigquery/bigquery_integration_test.go
```diff
@@ -380,6 +380,7 @@ func runBigQueryExecuteSqlToolInvokeTest(t *testing.T, select1Want, invokeParamW
 			api:           "http://127.0.0.1:5000/api/tool/my-exec-sql-tool/invoke",
 			requestHeader: map[string]string{},
 			requestBody:   bytes.NewBuffer([]byte(`{"sql":"CREATE TABLE t (id SERIAL PRIMARY KEY, name TEXT)"}`)),
+			want:          "\"Operation completed successfully.\"",
```
Yuan325:
```suggestion
			want:          `"Operation completed successfully."`,
```

tests/bigquery/bigquery_integration_test.go
```diff
@@ -395,14 +396,15 @@ func runBigQueryExecuteSqlToolInvokeTest(t *testing.T, select1Want, invokeParamW
 			api:           "http://127.0.0.1:5000/api/tool/my-exec-sql-tool/invoke",
 			requestHeader: map[string]string{},
 			requestBody:   bytes.NewBuffer([]byte(`{"sql":"DROP TABLE t"}`)),
+			want:          "\"Operation completed successfully.\"",
```
Yuan325:
```suggestion
			want:          `"Operation completed successfully."`,
```

tests/bigquery/bigquery_integration_test.go
```diff
@@ -395,14 +396,15 @@ func runBigQueryExecuteSqlToolInvokeTest(t *testing.T, select1Want, invokeParamW
...
 			isErr:         true,
 		},
 		{
 			name:          "invoke my-exec-sql-tool insert entry",
 			api:           "http://127.0.0.1:5000/api/tool/my-exec-sql-tool/invoke",
 			requestHeader: map[string]string{},
 			requestBody:   bytes.NewBuffer([]byte(fmt.Sprintf("{\"sql\":\"INSERT INTO %s (id, name) VALUES (4, 'test_name')\"}", tableNameParam))),
-			want:          "null",
+			want:          "\"Operation completed successfully.\"",
```
Yuan325:
```suggestion
			want:          `"Operation completed successfully."`,
```

Yuan325 (commented):
Thank you! Some nit suggestions. The issuer also mentioned that `SELECT * from empty table` returns error message as well? Do we want to add a check for nil before returning results from `SELECT` statement?

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
Genesis929: Updated.

Genesis929:
> Thank you! Some nit suggestions. The issuer also mentioned that `SELECT * from empty table` returns error message as well? Do we want to add a check for nil before returning results from `SELECT` statement?

A checker added.

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -145,7 +177,9 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
 		}
 		out = append(out, vMap)
 	}
-
+	if len(out) == 0 {
```
Yuan325:
```suggestion
	if out == nil {
```

Yuan325 (approved):
Quick nit, otherwise LGTM, thank you again!

kurtisvg (approved):

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -121,18 +125,46 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
 		return nil, fmt.Errorf("unable to get cast %s", sliceParams[0])
 	}
 
+	dryRunJob, err := dryRunQuery(ctx, t.RestService, t.Client.Project(), sql)
```
slilichenko, line 128: It doesn't look like you pass the parameters to the dry run. Wouldn't a parameterized query fail in this case? 

It also seems pretty costly to do dry run for every sql statement. Dry runs against external tables without metadata caching can have latencies of minutes. Suggest reconsidering this approach and trying to find out the type of the statement from the actual job execution. If that's not feasible - consider caching. 

Also, test how it would work with "EXECUTE IMMEDIATE" statement - the dry run might return an unexpected result.

Genesis929: This is for bigquery-execute-sql so there is no parameter input. I'm also working on a dry-run  function for bigquery-sql, which need a dry-run that supports parameters. I already have a local branch for that, but due to rest api requires us to handle parameter types manually, I'm still working on the tests to make sure it works as expected for all data types we support.

slilichenko: OK, sorry - I missed that this is not for the bigquery-sql tool. I would still consider not using dry-run if at all possible.

Genesis929: Thanks! The "EXECUTE IMMEDIATE" statement seems to be an issue, it's type is something else.  Will try figure a way to make this work.

PR #928, opened 2025-07-18
feat: Add `Map` parameters support

duwenxin99:
support both generic and typed map. Config example:
```
 parameters:
      - name: user_scores
        type: map
        description: A map of user IDs to their scores. All scores must be integers.
        valueType: integer # This enforces the value type for all entries. Leave it blank for generic map

```
Represented as `Object` with `additionalProperties` in manifests.

Added a util function to convert json.Number (string type) to int/float types to address the problem where int/float values are converted to strings for the generic map.


docs/en/resources/tools/_index.md
```diff
@@ -115,6 +115,40 @@ in the list using the items field:
 Items in array should not have a default value. If provided, it will be ignored.
 {{< /notice >}}
 
+Of course! Here is the documentation for the map parameter type, following the format you provided.
```
kurtisvg:
```suggestion
```

internal/tools/parameters.go
```diff
@@ -401,19 +413,21 @@ func (ps Parameters) McpManifest() McpToolsSchema {
...
-	Description  string             `json:"description"`
-	AuthServices []string           `json:"authSources"`
-	Items        *ParameterManifest `json:"items,omitempty"`
+	Name                 string             `json:"name"`
+	Type                 string             `json:"type"`
+	Required             bool               `json:"required"`
+	Description          string             `json:"description"`
+	AuthServices         []string           `json:"authSources"`
+	Items                *ParameterManifest `json:"items,omitempty"`
+	AdditionalProperties any                `json:"AdditionalProperties,omitempty"`
```
kurtisvg, line 422: is additional properties the best way to send this for Toolbox protocol? Or should it just be "valueType"?

internal/tools/parameters.go
```diff
@@ -1022,3 +1036,206 @@ func (p *ArrayParameter) McpManifest() ParameterMcpManifest {
...
+}
+
+// Parse validates and parses an incoming value for the map parameter.
+func (p *MapParameter) Parse(v any) (any, error) {
+	convertedData, err := util.ConvertNumbers(v)
+	if err != nil {
+		return nil, fmt.Errorf("error converting json.Number: %s", err)
+	}
+
+	convertedMap := convertedData.(map[string]any)
```
kurtisvg: do we need to check to see if this type cast was succesfull?

kurtisvg: May want this to be the first step to validate it is infact a map

internal/tools/parameters.go
```diff
@@ -1022,3 +1036,206 @@ func (p *ArrayParameter) McpManifest() ParameterMcpManifest {
...
+	case "float":
+		return NewFloatParameter("", ""), nil
+	default:
+		return nil, fmt.Errorf("unsupported valueType %q for map parameter", typeName)
+	}
+}
+
+// Parse validates and parses an incoming value for the map parameter.
+func (p *MapParameter) Parse(v any) (any, error) {
+	convertedData, err := util.ConvertNumbers(v)
```
kurtisvg: it seems like we only need to do this if it's of type "any". Can we refactor so that this only gets called if that's the value type?

internal/tools/parameters.go
```diff
@@ -1022,3 +1036,206 @@ func (p *ArrayParameter) McpManifest() ParameterMcpManifest {
...
+	if err != nil {
+		return nil, fmt.Errorf("error converting json.Number: %s", err)
```
kurtisvg: can we rephrase this error message into something that might make sense for an end user / agent?

internal/tools/parameters.go
```diff
@@ -1022,3 +1036,206 @@ func (p *ArrayParameter) McpManifest() ParameterMcpManifest {
...
+
+	convertedMap := convertedData.(map[string]any)
+
+	// If no valueType is specified, accept any values and return the map directly.
+	if p.ValueType == "" {
+		return convertedMap, nil
+	}
+
+	// Otherwise, get a prototype and parse each value in the map.
+	prototype, err := getPrototypeParameter(p.ValueType)
```
kurtisvg, line 1164: why do we need a prototype parameter?

duwenxin99: I was thinking this allows us to return to the LLM which item is causing the parsing error. But we can also just do a general map cast if this is not needed.

internal/tools/parameters.go
duwenxin99: I think it would make the implementation on the client side easier if we want to add a `object` type in the future. By using the object json schema for both, `map` and `object` could basically share the same code.

internal/tools/parameters.go
duwenxin99: To convert map[any]any into typed map in golang, we'll have to iterate through the parameters and cast each individual ones. I reuse each typed parameter's parsing function for simplicity and also reuse their existing int/float [conversion on json.Numbers](https://github.com/googleapis/genai-toolbox/blob/9a55b804821a6ccfcd157bcfaee7e599c4a5cb63/internal/tools/parameters.go#L741).

internal/tools/parameters.go
duwenxin99: updated

internal/tools/parameters.go
kurtisvg: https://go.dev/play/p/WCOtCD5z6Z5

internal/tools/parameters.go
kurtisvg: We aren't really using JSONSchema for the other formats, it seems odd to bring it in only for this "AdditionalProperties" field?

internal/tools/parameters.go
```diff
@@ -1022,3 +1036,211 @@ func (p *ArrayParameter) McpManifest() ParameterMcpManifest {
...
 	}
 }
+
+// MapParameter is a parameter representing a map with string keys. If ValueType is
+// specified (e.g., "string"), values are validated against that type. If ValueType
+// is empty, it is treated as a generic map[string]any.
+type MapParameter struct {
+	CommonParameter `yaml:",inline"`
+	Default         *map[string]any `yaml:"default,omitempty"`
+	ValueType       string          `yaml:"valueType,omitempty"`
```
kurtisvg, line 1046: Should we validate ValueType should only be one a a few values?

internal/tools/parameters.go
kurtisvg: nvm, I think I see what you were saying. We've already parsed it so it's not the right type to begin with.

internal/tools/parameters.go
duwenxin99: We are using it for array's `items`. We could also come up with our own format. Using the same JSON scheme just makes things easier on  our end to maintain.

internal/tools/parameters.go
duwenxin99: we are validating it [here](https://github.com/googleapis/genai-toolbox/blob/74127ae44ed64a2fbabbf2840a2e254deb45fb2b/internal/tools/parameters.go#L1116)

PR #929, opened 2025-07-18
feat: allow users to edit headers when running tools

AlexTalreja:
Allows users to edit request header to support `AuthServices` for Toolbox UI.

### DIFF
--- a/internal/server/static/css/style.css
+++ b/internal/server/static/css/style.css
@@ -1,3 +1,11 @@
+:root {
+    --toolbox-blue: #4285f4;
+    --text-primary-gray: #444444;
+    --text-secondary-gray: #6e6e6e;
+    --button-primary: var(--toolbox-blue);
+    --button-secondary: #616161;
+}
+
 body {
     display: flex;
     height: 100vh;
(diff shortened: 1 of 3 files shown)


internal/server/static/css/style.css
```diff
@@ -356,6 +358,92 @@ body {
...
+                font-weight: bolder;
+                cursor: pointer;
+                border-radius: 30px;
+                align-items: center;
+                color: white;
+                transition: background-color 0.3s ease;
+            }
+
+            .save-headers-btn {
+                background-color: #000000;
```
Yuan325: why do we have different color for the `save` and `close` and `edit header` button? Can we have a general `btn` class with all the same font/radius/padding etc. and use that for all of these buttons? Seems like those 3 buttons share alot of similar styling.

Can also potentially share it with the `run tool` button, but overwrite the color to blue for the run tool specifically.

And if we need specialize styling for a specific button, we can do it: `class="btn save-header-btn"` <-- iirc this will apply both the `btn` and `save-header-btn` 's style to the button.

internal/server/static/js/toolDisplay.js
```diff
@@ -150,12 +241,22 @@ export function renderToolInterface(tool, containerElement) {
     containerElement.innerHTML = '';
 
     let lastResults = null;
+    let currentHeaders = {
+        "Content-Type": "application/json",
+        "Accept": "application/json"
```
Yuan325: Do we need the `Accept` header? I thought we only need the `Content-Type` header?

Yuan325 (changes requested):

internal/server/static/css/style.css
AlexTalreja: I was thinking of having the `save` and `close` buttons as different to distinguish actions clearly, originally I had the standard red/green pattern but that didn't seem to fit with the rest of the design so I chose both gray/black. I also read that another way to give visual clues would be making the primary/confirm action bolder, so the `save` is black to stand out more.

Although this is kind of a subjective design, and my thoughts so let me know what you think. I could definitely switch them to be the same color or go back to red/green if you think that would be better.

AlexTalreja: Also yes, I'll work on reformatting the css to dedup styles

internal/server/static/js/toolDisplay.js
AlexTalreja: Oh true, since tool  invoke will always send back JSON I think it would be fine to leave this out.

internal/server/static/css/style.css
```diff
@@ -1,3 +1,9 @@
+:root {
+    --toolbox-blue: #4285f4;
+    --text-primary-gray: #444444;
+    --text-secondary-gray: #6e6e6e;
```
Yuan325, line 4: are these for text or buttons?

Yuan325: can we at least have the `edit-header` and `close` button to be the same color?

AlexTalreja: Fixed. Also updated `save` and `run` button to match, keeping the gray/blue color theme and preventing `save` and `close` from looking too similar (previously gray and black).

PR #932, opened 2025-07-18
feat(tools/bigquery-ask-data-insights): add bigquery ask-data-insights tool

Genesis929:
1. Add ask-data-insights tool based on conversational analytic API.
2. Add tokenSource for ask-data-insights tool, it uses access token instead of client or restService.
3. Add a max row count to source, currently fixed to 50 and used only for ask-data-insights tool. Later we may make it available for user to make change and apply to bigquery-execute-sql and bigquery-sql to avoid return too many data by accident.


docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
```diff
@@ -0,0 +1,53 @@
...
+  A "bigquery-ask-data-insights" tool allows conversational interaction with a BigQuery source.
+aliases:
+- /resources/tools/bigquery-ask-data-insights
+---
+
+## About
+
+A `bigquery-ask-data-insights` tool allows you to ask questions about your data in natural language. 
+
+This function takes a user's  question (which can include conversational history for context)
```
averikitsch:
```suggestion
This function takes a user's question (which can include conversational history for context) 
```

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
```diff
@@ -0,0 +1,53 @@
...
+**Note**: This tool requires additional setup in your project. Please refer to the 
+official [Conversational Analytics API documentation](https://cloud.google.com/gemini/docs/conversational-analytics-api/overview)
+for instructions.
+
+It's compatible with the following sources:
+
+- [bigquery](../sources/bigquery.md)
+
+
+The tool takes one input parameter `question`, which is the natural language question
```
averikitsch: Do we need to update this to mention it takes another param for the table name?

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
```diff
@@ -0,0 +1,53 @@
...
+weight: 1
+description: > 
+  A "bigquery-ask-data-insights" tool allows conversational interaction with a BigQuery source.
+aliases:
+- /resources/tools/bigquery-ask-data-insights
+---
+
+## About
+
+A `bigquery-ask-data-insights` tool allows you to ask questions about your data in natural language.
```
averikitsch: discussion point: For AlloyDB's NL 2 SQL capable tool type is named after the product name "alloydb-ai-nl". For custom tools and prebuilt tools, any tool name can be chosen to use this type of tool. For example,
```
  ask_data_insights:
    kind: bigquery-ask-data-insights
    source: bigquery-source
```
Do we to be more consistent with tool type and use something like `bigquery-data-analytics` or `bigquery-converse-analytics`?

```
  ask_data_insights:
    kind: bigquery-data-analytics
    source: bigquery-source
```

CC @kurtisvg @Yuan325 @duwenxin99

kurtisvg: I lean towards preferring that we more closely align with the API for discoverability/consistency w/ product messaging, but I don't think it's a hard requirement if we have a good rationale for not doing so.

internal/tools/bigquery/bigqueryaskdatainsights/bigqueryaskdatainsights.go
```diff
@@ -0,0 +1,484 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("API returned non-200 status: %d %s", resp.StatusCode, string(body))
+	}
+
+	var acc strings.Builder
+	var messages []map[string]any
+	scanner := bufio.NewScanner(resp.Body)
```
averikitsch: Since we are not streaming info the users, can we just read the response body and use json.marshal with a defined struct?

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
Genesis929: If we use different tool name  ask_data_insights and tool kind bigquery-data-analytics it may be confusing to user, I think. Meanwhile I'll discuss the naming issue with @shobsi and @jiaxunwu and update later.

internal/tools/bigquery/bigqueryaskdatainsights/bigqueryaskdatainsights.go
Genesis929: There is one major issue: the CA API is streaming everything locally. In our tests, this stream can be extremely long. Even when it's not, it sometimes causes issues because the token length is too great for the LLM to handle.

Therefore, it would be better to select what the user needs to see and remove the unnecessary, token-heavy information beforehand. For this reason, I don't think simply reading the entire response body will work.

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
Genesis929: Yes, I forgot that, thanks!

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
shobsi: Chiming in with my 2 cents, I prefer consistency (e.g. `bigquery-execute-sql` and `spanner-execute-sql`) but concur with @kurtisvg that if a producer would like to name more closely with their well known product names then we should be open for it.

Reg `bigquery-ask-data-insights` in particular, we actually spent a fair amount of time deciding the name that adheres to our tool development guidelines go/bq-agent-tools-dev. Another name we considered was `bigquery-ask-nl-question`, but when @Genesis929 tested it alongside `bigquery-execute-sql` tool, LLM often picked the latter while we expected the former. So we agreed upon `bigquery-ask-data-insights` as it meets the following criteria:

* It is a verb, consistent with other bigquery tool names
* Reflects the behavior that you can "ask" in natural language.
* Present to LLM the best differentiating behavior when used alongside other BigQuery tools

kurtisvg: Just to point out -- the name of the tool and the name of the tool kind don't need to be the same. You can surface the tool kind as "bigquery-converse-analytics" and users can name it whatever they want. The tool kind is more about discovery of the tool for the user.

internal/tools/bigquery/bigqueryaskdatainsights/bigqueryaskdatainsights.go
averikitsch: It would be best to not manually parse the JSON. You can use JSON.Decoder and a struct to do this more securely like
```

// User represents a user object from the JSON stream.
type User struct {
	ID   int    `json:"id"`
	Name string `json:"name"`
}

func main() {
	// Simulate an HTTP GET request to a streaming API.
	// In a real application, you would use http.Get().
	// For this example, we'll use a mock reader.
	jsonStream := `[
		{"id": 1, "name": "Alice"},
		{"id": 2, "name": "Bob"},
		{"id": 3, "name": "Charlie"}
	]`

	// The http.Response.Body is an io.ReadCloser, so this is how you would use it.
	respBody := strings.NewReader(jsonStream)
	decoder := json.NewDecoder(respBody)

	// Read the opening bracket of the JSON array.
	if _, err := decoder.Token(); err != nil {
		log.Fatal(err)
	}

	// Loop to decode each user object.
	var user User
	for decoder.More() {
		// Decode each object in the stream.
		err := decoder.Decode(&user)
		if err != nil {
			if err == io.EOF {
				break // End of stream
			}
			log.Fatal(err)
		}
		// Process the decoded user object.
		fmt.Printf("Decoded User: %+v\n", user)
	}

	// Read the closing bracket of the JSON array.
	if _, err := decoder.Token(); err != nil {
		log.Fatal(err)
	}
}
```

Genesis929: Updated, now the code uses decoder

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
Genesis929: Since this is a prebuilt tool, having the tool name and kind be different could make the documentation confusing. When a user writes a custom YAML file, they need to know a specific name for the kind, but when they call the prebuilt tool, they have to use a very different name. This approach doesn't seem intuitive.

kurtisvg: @Genesis929 Toolbox doesn't have a concept of prebuilt tool. This is a kind of tool, which might be added to a prebuilt config. We have no way to stop a user from naming the tool whatever they want. 

The user is very unlikely to actually call the tool -- the LLM will call the tool, and the LLM has no concept of the tool kind. 

Also -- you've already has asks to constrain the tools (limit to specific datasets, etc). What happens when you have an agent that needs 2 versions of a tool with different constraints? The user will need to choose two different names.

Genesis929: To clarify my point, when a user sees the tool name ask-data-insights in the prebuilt config, they then need to remember a very different kind name, bigquery-data-analytics, when writing their own custom config.

I'm not suggesting users have to keep the same tool name we provide. My main concern is that this could be confusing for users trying to understand the relationship between the two names, especially when they switch between writing a custom config and using the prebuilt one. They might wonder, "Why do all the other prebuilt tools have related kind and tool names, except for this one?"

Of course, using a different kind and tool name will still work functionally. If you don't think this difference in naming logic is a significant issue, then I'm fine with making the change.

drstrangelooker:
Watching this. Looker also supports the conversational analytics API and we will implement something similar soon.

docs/en/resources/tools/bigquery/bigquery-ask-data-insights.md
shobsi: Either way is not the end of the world, but from the developer's point of view I concur with @Genesis929 on leaning towards the coherent naming between the tool kind and the tool name. It's very similar to naming instances of a class in OOP, where

```python
  ask_data_insights = BigQueryAskDataInsights()
``` 
would be much more intuitive than

```python
  ask_data_insights = BigQueryDataAnalytics()
``` 

If two versions of the same tool kind are needed then we/developer can use more differentiating but still coherent names, e.g.:

```yaml
  ask_data_insights_sales:
    kind: bigquery-ask-data-insights
    source: bigquery-source-sales
  ask_data_insights_operations:
    kind: bigquery-ask-data-insights
    source: bigquery-source-operations
```

kurtisvg: >  They might wonder, "Why do all the other prebuilt tools have related kind and tool names, except for this one?"

This is a BigQuery specific trait, and most of the other prebuilt configs don't follow this pattern. The majority of docs and samples do not as well. 

> leaning towards the coherent naming between the tool kind and the tool name.

The big concern here is are they able to discover the tool in the first place? Are they going to be looking for the name of the product or are the going to be looking for the name of the tool in the prebuilt config? The name of the product is how the docs refer to it, which why I strongly prefer we remain consistent with how the other tools handle this.

Genesis929: Name updated.

averikitsch (approved):

PR #937, opened 2025-07-21
feat: Add alloydb control plane as a prebuilt config

prernakakkar-google:


cmd/root.go
```diff
@@ -190,7 +190,7 @@ func NewCommand(opts ...Option) *Command {
 	flags.BoolVar(&cmd.cfg.TelemetryGCP, "telemetry-gcp", false, "Enable exporting directly to Google Cloud Monitoring.")
 	flags.StringVar(&cmd.cfg.TelemetryOTLP, "telemetry-otlp", "", "Enable exporting using OpenTelemetry Protocol (OTLP) to the specified endpoint (e.g. 'http://127.0.0.1:4318')")
 	flags.StringVar(&cmd.cfg.TelemetryServiceName, "telemetry-service-name", "toolbox", "Sets the value of the service.name resource attribute for telemetry data.")
-	flags.StringVar(&cmd.prebuiltConfig, "prebuilt", "", "Use a prebuilt tool configuration by source type. Cannot be used with --tools-file. Allowed: 'alloydb-postgres', 'bigquery', 'cloud-sql-mysql', 'cloud-sql-postgres', 'cloud-sql-mssql', 'postgres', 'spanner', 'spanner-postgres'.")
+	flags.StringVar(&cmd.prebuiltConfig, "prebuilt", "", "Use a prebuilt tool configuration by source type. Cannot be used with --tools-file. Allowed: 'alloydb-postgres-control-plane', alloydb-postgres', 'bigquery', 'cloud-sql-mysql', 'cloud-sql-postgres', 'cloud-sql-mssql', 'firestore', 'postgres', 'spanner', 'spanner-postgres'.")
```
averikitsch: Please remove firestore from this PR. 
I would like to propose `alloydb-postgres-admin` instead of 'alloydb-postgres-control-plane'

internal/server/config.go
```diff
@@ -34,13 +34,13 @@ type ServerConfig struct {
 	// Port is the port the server will listen on.
 	Port int
 	// SourceConfigs defines what sources of data are available for tools.
-	SourceConfigs SourceConfigs
+	SourceConfigs SourceConfigs `yaml:"sources"`
```
averikitsch: What is the purpose of adding these tags?

prernakakkar-google: It was not reading the tools unless we make the change with stdio mode

prernakakkar-google: removed

cmd/root.go
prernakakkar-google: Done

internal/prebuiltconfigs/tools/alloydb-postgres-admin.yaml
```diff
@@ -0,0 +1,123 @@
...
+    delay: 1s
+    maxDelay: 4m
+    multiplier: 2
+    maxRetries: 10
+  alloydb-create-instance:
+    kind: http
+    source: alloydb-api-source
+    method: POST
+    path: /v1/projects/{{.projectId}}/locations/{{.locationId}}/clusters/{{.clusterId}}/instances
+    description: "Creates a new AlloyDB instance (PRIMARY, READ_POOL, or SECONDARY) within a cluster. This is a long-running operation."
```
averikitsch: nit: does this return an operation id? If so it might be good to add that to the description so it knows how to connect it to the operations-get tool

internal/prebuiltconfigs/tools/alloydb-postgres-admin.yaml
```diff
@@ -0,0 +1,123 @@
+sources:
+  alloydb-api-source:
+    kind: http
+    baseUrl: https://alloydb.googleapis.com
+    headers:
+      Authorization: Bearer ${API_KEY}
```
averikitsch, line 6: Same as the wait for operations PR, I would expect my local credentials to work

internal/prebuiltconfigs/tools/alloydb-postgres-admin.yaml
prernakakkar-google: Yes it returns that, it uses the description of operation tools. But I can add here as well.

internal/prebuiltconfigs/tools/alloydb-postgres-admin.yaml
prernakakkar-google: The http tool here requires this as part of header since API needs it.

prernakakkar-google: The tokens are different and results in:
`
Details: Credential sent is invalid. Unknown token version 0 for token string: ${API_KEY} Trace ID: 0x5bb33598310d5b86\"\n      },\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.ErrorInfo\",\n        \"reason\": \"ACCESS_TOKEN_TYPE_UNSUPPORTED\",\n        \"metadata\": {\n          \"method\": \"google.cloud.alloydb.v1.AlloyDBAdmin.CreateCluster\",\n          \"service\": \"alloydb.googleapis.com\"\n        }\n      }\n    ]\n  }\n}\n"
        }
      ],
      "isError": true
    }
  }
}
`

PR #947, opened 2025-07-22
chore: fix dry run location

Genesis929:
Updated dry run in execute sql to also include a location, fix the potential issue in PR #925.


Yuan325 (approved):

Genesis929:
Turns out we don't need a default location even for restAPI. removed default setting.

Issue #922, opened 2025-07-17
FT.SEARCH command fails with "Syntax error" on a parameterized query for Redis DB [Docker image].


AnmolShukla2002:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

toolbox version 0.8.0

### Environment

1. OS type and version: Linux penguin 6.6.76-08024-gb30cb4a129c2
2. How are you running Toolbox: 
  - As a downloaded binary (curl -O https://storage.googleapis.com/genai-toolbox/v0.8.0/$OS/toolbox)

### Client

```
//hotel_agent.js
import "dotenv/config";
import { ChatVertexAI } from "@langchain/google-vertexai";
import { ToolboxClient } from "@toolbox-sdk/core";
import { tool } from "@langchain/core/tools";
import { HumanMessage, ToolMessage } from "@langchain/core/messages";

const prompt = `
You're a helpful hotel assistant. You handle hotel searching, booking, and
cancellations. When the user searches for a hotel, mention its name, id,
location and price tier. Always mention hotel ids while performing any
searches. This is very important for any operations. For any bookings or
cancellations, please provide the appropriate confirmation. Be sure to
update checkin or checkout dates if mentioned by the user.
Don't ask for confirmations from the user.
`;

const queries = [
  "Find hotels in Basel with Basel in its name.",
  "Can you book the Hilton Basel for me?",
  "Oh wait, this is too expensive. Please cancel it and book the Hyatt Regency instead.",
  "My check in dates would be from April 10, 2024 to April 19, 2024.",
];

async function runApplication() {
  console.log("Starting hotel agent...");

  const model = new ChatVertexAI({
    model: "gemini-2.0-flash-001",
    temperature: 0,
  });

  const client = new ToolboxClient("http://127.0.0.1:5000");
  const toolboxTools = await client.loadToolset("my-toolset");

  console.log(`Loaded ${toolboxTools.length} tools from Toolbox`);

  const tools = toolboxTools
    .map((t) => {
      return tool(t, {
        name: t.toolName,
        description: t.description,
        schema: t.params,
      });
    })
    .filter(Boolean);

  const modelWithTools = model.bindTools(tools);

  let messages = [new HumanMessage(prompt)];

  for (const query of queries) {
    console.log(`\nUser: ${query}`);

    messages.push(new HumanMessage(query));

    for (let step = 0; step < 5; step++) {
[...]

PR #954, opened 2025-07-22
fix: FT.SEARCH command fails with "Syntax error" on a parameterized query for Redis DB [Docker image].

AnmolShukla2002:
This PR fixes the issue [#922 ] that the toolbox's redis tool did not correctly substitute parameters when they appeared inside strings, such as within quotes (e.g., @name:'$name'). The original parameter substitution logic only replaced command parts that exactly matched a parameter placeholder, so any occurrence of $param embedded within a string was left unchanged, resulting in invalid queries and syntax errors from RediSearch. I fixed this by updating the substitution logic to replace all occurrences of $param within any command part, regardless of its position in the string.

### DIFF
--- a/internal/tools/redis/redis.go
+++ b/internal/tools/redis/redis.go
@@ -16,6 +16,7 @@ package redis
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	yaml "github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/sources"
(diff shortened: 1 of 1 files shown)


Yuan325:
Hi @AnmolShukla2002 I closed this PR for now since it is causing an issue to pull remotely.

AnmolShukla2002:
okay, no problem. sorry for the issue! @Yuan325

Yuan325:
No worries! :) @AnmolShukla2002

PR #958, opened 2025-07-23
fix(tools/redis): substitute parameters in query strings

AnmolShukla2002:
1) This PR fixes the issue #922. The Redis tool in Toolbox did not correctly substitute parameters when they appeared inside query strings, especially for RediSearch queries. This caused problems when i wrote commands like ["@name:'$name'"] in my YAML, as the $name parameter was not always replaced correctly, and array parameters could be expanded incorrectly. As a result, valid RediSearch queries would not work.

2) Updated the replaceCommandsParams function to:

- Replace all occurrences of $param inside any command string, not just when the whole part is a parameter placeholder.
- Ensure that quoted and embedded parameters are substituted properly, making the tool work for RediSearch and other Redis command patterns.

3) With this fix, the following types of queries are supported in YAML :

- Single field, quoted parameter:
      ["@name:'$name'"] or ["@name:\"$name\""]
      (for multi-word values)
- Single field, unquoted parameter:
      ["@name:$name"]
      (for single-word values)
- Multiple fields:
      ["@name:'$name' @location:'$location'"]
      (search by name and location)


duwenxin99:
Hi @AnmolShukla2002, thanks for raising the issue! I'll review the PR shortly and let you know if anything needs to be changed. Thank you!

AnmolShukla2002:
> Hi @AnmolShukla2002, thanks for raising the issue! I'll review the PR shortly and let you know if anything needs to be changed. Thank you!

Okay, sure!

internal/tools/redis/redis.go
```diff
@@ -189,21 +190,27 @@ func replaceCommandsParams(commands [][]string, params tools.Parameters, paramVa
...
 		for _, part := range cmd {
-			v, ok := paramMap[part]
-			if !ok {
-				// Command part is not a Parameter placeholder
-				newCmd = append(newCmd, part)
-				continue
+			replaced := part
+			isArray := false
+			var arrayItems []any
+			for _, p := range params {
```
duwenxin99: We can assume that all the parameters have their corresponding values in `paramMap` because we check for that during parameter parsing. Therefore we can remove the iteration here and directly jump to line 200.

internal/tools/redis/redis.go
```diff
@@ -189,21 +190,27 @@ func replaceCommandsParams(commands [][]string, params tools.Parameters, paramVa
...
+						isArray = true
+						arrayItems = v.([]any)
+					}
+				}
 			}
-			if typeMap[part] == "array" {
-				for _, item := range v.([]any) {
-					// Nested arrays will only be expanded once
-					// e.g., [A, [B, C]]  --> ["A", "[B C]"]
+			if isArray {
```
duwenxin99, line 214: Can we add a comment on array expansion behavior here? This would benefit maintenance in the long term and also make it easer for contributors to understand the substitution logic. Thanks!

duwenxin99 (changes requested):
Hi @AnmolShukla2002, left comments for a few small improvements. Everything else LGTM. Thank you!

AnmolShukla2002:
Okay, thanks for pointing out the improvements. I will push the changes soon. @duwenxin99

AnmolShukla2002:
Hi @duwenxin99, I have pushed the changes.  Thanks!

duwenxin99:
@AnmolShukla2002 Thanks for the update! The main concern I'm having (which is also why I originally designed the substitution logic to be exact match) is that what if the user tries to store some IDs/complicated strings into redis that contains special character `$name`, which will be substituted silently in our case. This will become a data integrity issue. For reference, SQL prepared statements does not substitute string literals to prevent this from happening.

AnmolShukla2002:
> @AnmolShukla2002 Thanks for the update! The main concern I'm having (which is also why I originally designed the substitution logic to be exact match) is that what if the user tries to store some IDs/complicated strings into redis that contains special character `$name`, which will be substituted silently in our case. This will become a data integrity issue. For reference, SQL prepared statements does not substitute string literals to prevent this from happening.

Yes that's a valid point....for that maybe we can allow $$ [like shell scripts] to represent literal $ and only substitute unescaped `$param` patterns [but then users must intentionally escape $$ for literal dollars] or we can use a different delimiter like use {{param}} or ${param} syntax instead of $param in order to reduce collision risk? I need to look around for possible solutions or if you come up with anything then let me know. Thanks!

duwenxin99:
> > @AnmolShukla2002 Thanks for the update! The main concern I'm having (which is also why I originally designed the substitution logic to be exact match) is that what if the user tries to store some IDs/complicated strings into redis that contains special character `$name`, which will be substituted silently in our case. This will become a data integrity issue. For reference, SQL prepared statements does not substitute string literals to prevent this from happening.
> 
> Yes that's a valid point....for that maybe we can allow $$ [like shell scripts] to represent literal $ and only substitute unescaped `$param` patterns [but then users must intentionally escape $$ for literal dollars] or we can use a different delimiter like use {{param}} or ${param} syntax instead of $param in order to reduce collision risk? I need to look around for possible solutions or if you come up with anything then let me know. Thanks!

Our other tools utilize [Template Parameters](https://googleapis.github.io/genai-toolbox/resources/tools/#template-parameters) as an "unsafe" option for the users if they really need complete flexibility over substitution. Maybe we could support it in Redis/Valkey as well. This does not support array expansions etc but it does work for other param types @AnmolShukla2002

AnmolShukla2002:
> > > @AnmolShukla2002 Thanks for the update! The main concern I'm having (which is also why I originally designed the substitution logic to be exact match) is that what if the user tries to store some IDs/complicated strings into redis that contains special character `$name`, which will be substituted silently in our case. This will become a data integrity issue. For reference, SQL prepared statements does not substitute string literals to prevent this from happening.
> > 
> > 
> > Yes that's a valid point....for that maybe we can allow $$ [like shell scripts] to represent literal $ and only substitute unescaped `$param` patterns [but then users must intentionally escape $$ for literal dollars] or we can use a different delimiter like use {{param}} or ${param} syntax instead of $param in order to reduce collision risk? I need to look around for possible solutions or if you come up with anything then let me know. Thanks!
> 
> Our other tools utilize [Template Parameters](https://googleapis.github.io/genai-toolbox/resources/tools/#template-parameters) as an "unsafe" option for the users if they really need complete flexibility over substitution. Maybe we could support it in Redis/Valkey as well. This does not support array expansions etc but it does work for other param types @AnmolShukla2002

Okay I will work on it and update here if it goes well or not.  Thanks!

AnmolShukla2002:
> This does not support array expansions etc but it does work for other param types

Hi @duwenxin99 ,
You mentioned that template parameters "does not support array expansions etc but it does work for other param types." However, the [docs](https://googleapis.github.io/genai-toolbox/resources/tools/#template-parameters) show SQL tools using {{array .columnNames}} syntax for array expansion. So, for the Redis tool implementation, should I:
Include array support using {{array .paramName}} syntax (like SQL tools do), or
Skip array support and only implement basic types (string, integer, float, boolean)?

Is there a Redis-specific reason to exclude arrays? 

Thanks!

duwenxin99:
> > This does not support array expansions etc but it does work for other param types
> 
> Hi @duwenxin99 , You mentioned that template parameters "does not support array expansions etc but it does work for other param types." However, the [docs](https://googleapis.github.io/genai-toolbox/resources/tools/#template-parameters) show SQL tools using {{array .columnNames}} syntax for array expansion. So, for the Redis tool implementation, should I: Include array support using {{array .paramName}} syntax (like SQL tools do), or Skip array support and only implement basic types (string, integer, float, boolean)?
> 
> Is there a Redis-specific reason to exclude arrays?
> 
> Thanks!

Hi @AnmolShukla2002, when I said that "template parameters does not support array expansions" I meant it doesn't separate arrays into multiple segments like we may require for the redis commands [here](https://github.com/googleapis/genai-toolbox/blob/513007dc5c9b2eae3b92f19ac5bdf1ba0eb8f72e/internal/tools/redis/redis.go#L212). Template parameters use Go template and it only outputs one single string after replacement. This isn't a problem for regular SQL queries because a SQL statement is a string while a Redis command is a list of string segments. It only works if the user wants to turn an array input into one single string.

AnmolShukla2002:
> > > This does not support array expansions etc but it does work for other param types
> > 
> > 
> > Hi @duwenxin99 , You mentioned that template parameters "does not support array expansions etc but it does work for other param types." However, the [docs](https://googleapis.github.io/genai-toolbox/resources/tools/#template-parameters) show SQL tools using {{array .columnNames}} syntax for array expansion. So, for the Redis tool implementation, should I: Include array support using {{array .paramName}} syntax (like SQL tools do), or Skip array support and only implement basic types (string, integer, float, boolean)?
> > Is there a Redis-specific reason to exclude arrays?
> > Thanks!
> 
> Hi @AnmolShukla2002, when I said that "template parameters does not support array expansions" I meant it doesn't separate arrays into multiple segments like we may require for the redis commands [here](https://github.com/googleapis/genai-toolbox/blob/513007dc5c9b2eae3b92f19ac5bdf1ba0eb8f72e/internal/tools/redis/redis.go#L212). Template parameters use Go template and it only outputs one single string after replacement. This isn't a problem for regular SQL queries because a SQL statement is a string while a Redis command is a list of string segments. It only works if the user wants to turn an array input into one single string.

Thanks for the clarification @duwenxin99! 
Which approach should I take for the Redis tool?
1) Template parameters only (no array expansion).
2) Keep existing $param syntax for array expansion and template parameters for single-values.

duwenxin99:
@AnmolShukla2002 
I think we should go with option2 so that we can accommodate both usages, and it also conforms to our other tools' design.

AnmolShukla2002:
> @AnmolShukla2002 I think we should go with option2 so that we can accommodate both usages, and it also conforms to our other tools' design.

Okk, thanks!

PR #959, opened 2025-07-23
fix: mark database field as required in the firestore prebuilt tools

trehanshakuntG:
Env variables cannot be used as optional parameters, hence marking database field as required.

### DIFF
--- a/docs/en/resources/sources/firestore.md
+++ b/docs/en/resources/sources/firestore.md
@@ -58,7 +58,7 @@ sources:
   my-firestore-source:
     kind: "firestore"
     project: "my-project-id"
-    database: "my-database"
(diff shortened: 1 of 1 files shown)


docs/en/resources/sources/firestore.md
```diff
@@ -58,7 +58,7 @@ sources:
   my-firestore-source:
     kind: "firestore"
     project: "my-project-id"
-    # database: "my-database"  # Optional, defaults to "(default)"
+    database: "my-database"
```
averikitsch: It may be required for prebuilt tools but that doesn't mean it's required for the source

trehanshakuntG: Reverted the change.

PR #973, opened 2025-07-23
docs: add Prebuilt documentation

averikitsch:


averikitsch:
I've added the Do not merge label because I am unsure if I am happy with this documentation.

PR #977, opened 2025-07-23
ci: Add MongoDB `aggregate` Tool and integration test

duwenxin99:


tests/mongodb/mongodb_integration_test.go
```diff
@@ -0,0 +1,759 @@
...
+	defer cancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	tests.RunToolGetTest(t)
+
+	select1Want, failInvocationWant, invokeParamWant, invokeIdNullWant, nullString, mcpInvokeParamWant := getMongoDBWants()
```
kurtisvg: why are we using a function instead of just declaring these?

duwenxin99: removed

PR #1014, opened 2025-07-28
docs: add link checker and fix broken links

averikitsch:


linkinator.config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "concurrency": 10,
+  "skip": [
+    "https://console.cloud.google.com",
+    "https://storage.googleapis.com",
+    "http://127.0.0.1",
+    "https://looker.example.com",
+    "https://xxx.cloud.dgraph.io/",
```
Yuan325, line 7: Did we have this in the docs?

linkinator.config.json
```diff
@@ -0,0 +1,14 @@
...
+    "mysql",
```
Yuan325: are we trying to skip any URL containing mysql (same with the "5000" and "8080")?

Yuan325 (approved):

linkinator.config.json
averikitsch: Yes, the looker and dgraph links are used as example sources.

linkinator.config.json
averikitsch: I have removed "mysql". There were some doc links that were coming up as a 403 when testing, but look fine now. There are issues with using "localhost" as a regex to ignore localhost:5000 for locally running toolbox. That's why I am using 5000 and 8080 to ignore these links.

PR #1015, opened 2025-07-28
feat(tools/looker-query-url): Add looker-query-url tool

drstrangelooker:


docs/en/resources/tools/looker/looker_query_url.md
```diff
@@ -0,0 +1,53 @@
...
+    query_url:
+        kind: looker-query-url
+        source: looker-source
+        description: |
+          Query URL Tool
+
+          This tool is used to generate the URL of a query in Looker.
+          The user can then explore the query further inside Looker.
+          The tool also returns the query_id and slug. The parameters
+          are the same as the query tool.
```
duwenxin99:
```suggestion
          are the same as the `looker-query` tool.
```

internal/prebuiltconfigs/tools/looker.yaml
```diff
@@ -114,6 +114,17 @@ tools:
 
           The result of the query sql tool is SQL text.
 
+    query_url:
```
duwenxin99, line 117: I think this was missed previously but we need to add Looker to the [prebuilt command description](https://github.com/googleapis/genai-toolbox/blob/dbf355d31ace10a2c1c12c34a325a538650c0f5d/cmd/root.go#L221)

internal/tools/looker/lookerqueryurl/lookerqueryurl_test.go
```diff
@@ -0,0 +1,116 @@
...
+				`,
+			want: server.ToolConfigs{
+				"example_tool": lkr.Config{
+					Name:         "example_tool",
+					Kind:         "looker-query-url",
+					Source:       "my-instance",
+					Description:  "some description",
+					AuthRequired: []string{},
+				},
+			},
```
duwenxin99, line 55: Can we add another config test `advanced example` with all the optional fields specified? Thanks!

duwenxin99 (changes requested):

internal/tools/looker/lookerqueryurl/lookerqueryurl_test.go
drstrangelooker: I didn't put support for templating parameters in my tools since they will likely just use the pre-built. I will add this in a different PR.

internal/prebuiltconfigs/tools/looker.yaml
drstrangelooker: Done now.

PR #1016, opened 2025-07-28
fix: minor rendering issue in markdown doc

drstrangelooker:
Tabpane wasn't rendered correctly.


drstrangelooker:
Nevermind. Another fix must have made it better.

PR #1019, opened 2025-07-28
ci: remove pull_request_target from link checker

averikitsch:


drstrangelooker:
There has been some flakeyness with the link tester. Sometimes good links are getting 429, 503, and 504 errors. I reran the check a few times until it passed.

PR #1020, opened 2025-07-29
fix(tools/bigquery-sql): ensure invoke always returns a non-null value

Genesis929:
This is to make bigquery-sql consistent with bigquery-execute-sql. May not be necessary to have.

- Added a dry run step to identify the query type (e.g., SELECT, DML), which allows the tool to correctly handle the query's output.
- The recommended high-level client, cloud.google.com/go/bigquery, does not expose the statement type from a dry run. To circumvent this limitation, the low-level BigQuery REST API client (google.golang.org/api/bigquery/v2) was added to gain access to these necessary details.


averikitsch (commented):
Overall, this lgtm. Tests conflicts need to be updated and we should confirm the use of dry run.

Genesis929:
> Overall, this lgtm. Tests conflicts need to be updated and we should confirm the use of dry run.

Updated, fixed the use of dry_run.

Issue #873, opened 2025-07-13
[bigquery] Add dataset access constraint for bigquery source


vksinha10:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

for bigquery, i have lots of datasets in my environment. i dont want to query all the datasets. i want only specific datasets. As of now we dont have that. 

### Suggested Solution(s)

can we add it something like this
sources:
  my-bigquery-source:
    kind: "bigquery"
    project: "my-project-id"
    dataset_id: "dataset_id"

Genesis929:
Hi, @vksinha10. Thank you for the feedback! It would be very helpful if you could provide a few sample chats showing how you expect your tools to interact with a default dataset.

This will be crucial for us to evaluate the feasibility and prioritize our work on this feature.

vksinha10:
Hi! Thanks for considering this feature request. Here are some sample chat interactions that demonstrate how I'd expect the tools to work with dataset-specific configuration:

Current behavior (querying across all datasets):
**User**: "Show me the schema for the users table"
**Assistant**: *searches across all datasets in the project, potentially finding multiple 'users' tables*

Expected behavior with dataset-specific configuration:
**User**: "Show me the schema for the users table"
**Assistant**: *queries only the specified dataset (e.g., 'production_analytics')* 
"Here's the schema for the users table in the production_analytics dataset..."

More specific examples:
Example 1: Table exploration
**User**: "What tables are available?"
**Assistant**: *lists only tables from the configured dataset instead of all datasets*
"Available tables in your configured dataset (production_analytics):
- users
- orders  
- products
- sessions"

**Example 2: Query execution**
**User**: "Get the count of active users from last week"
**Assistant**: *automatically scopes the query to the specified dataset*
"SELECT COUNT(*) FROM `my-project-id.production_analytics.users` 
WHERE status = 'active' AND created_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 7 DAY)"

**Example 3: Schema references**
**User**: "What columns does the orders table have?"
**Assistant**: *checks schema in the specific dataset only*
"The orders table in production_analytics has these columns: order_id, user_id, amount, created_at..."

Genesis929:
Hi @vksinha10,

Thanks so much for providing these detailed examples – they're incredibly helpful.

[...]

PR #1023, opened 2025-07-29
feat(sources/bigquery): support allowedDataset field in bigquery source

nikunj-badjatya:
Please check this PR for https://github.com/googleapis/genai-toolbox/issues/873

An additional argument to the toolbox is added - allowedDataset.
The value of which is via BIGQUERY_ALLOWED_DATASET env variable 
An empty "" value would mean queries have access to no datasets.
An "*" value would mean queries can run across all datasets
An explicit "datasetABC" would mean queries are allowed only on this dataset


nikunj-badjatya:
Hello @duwenxin99 
Please check this PR. Thanks!

internal/sources/bigquery/bigquery.go
```diff
@@ -68,11 +69,11 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
 	}
 
 	s := &Source{
-		Name:        r.Name,
-		Kind:        SourceKind,
-		Client:      client,
-		RestService: restService,
```
duwenxin99, line 74: The `RestService` param shouldn't be deleted

duwenxin99:
Hi @nikunj-badjatya, thanks for creating the PR! I want to clarify the use case here a little bit. I assume you want to limit the dataset access to the `bigquery-execute-sql` tool only, correct? In that case, I think it's better to add this field at a tool level to the `bigquery-execute-sql` tool specifically, since it won't be used by other BQ tool types.

nikunj-badjatya:
> Hi @nikunj-badjatya, thanks for creating the PR! I want to clarify the use case here a little bit. I assume you want to limit the dataset access to the `bigquery-execute-sql` tool only, correct? In that case, I think it's better to add this field at a tool level to the `bigquery-execute-sql` tool specifically, since it won't be used by other BQ tool types.

Ideally for all the tools for bigquery as the overall toolset evolves. The support allows overriding it as per the usecase and have "*" in the allowed dataset.
May be going forward we can have per tool level overriding capability. Thoughts ?

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -113,18 +117,94 @@ type Tool struct {
...
 	Parameters   tools.Parameters `yaml:"parameters"`
 	Client       *bigqueryapi.Client
+	Source       compatibleSource // Source for dataset validation
 	RestService  *bigqueryrestapi.Service
 	manifest     tools.Manifest
 	mcpManifest  tools.McpManifest
 }
 
-func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error) {
+func (t Tool) ValidateSQLDatasetAccess(sql string) error {
```
Genesis929, line 126: Thank you for your work! I can't directly run the test due to some issue, but it seems there are a few cases this may not work as expected:
1. select * from \`project\`.dataset.table limit 1
 This should pass, but it will fail
3. select * from \`project.dataset.table\` where s1 = "some text \`a.b.c\`" limit 1
 This should pass, but it will fail
5. select * from \`project.dataset.table\` join \`no\`.\`access\`.\`table\` on xxx limit 1
 This should fail, but it will pass

duwenxin99:
Hi @nikunj-badjatya, this is a security feature so we need to be extra careful that it works on all edge cases. @Genesis929 is actively working on this and we'll update to the issue in a week on the progress.

nikunj-badjatya:
> Hi @nikunj-badjatya, this is a security feature so we need to be extra careful that it works on all edge cases. @Genesis929 is actively working on this and we'll update to the issue in a week on the progress.

ok thanks. I will keep this on hold.

vksinha10:
Is there any timeline when this feature will be merged. We have usecase where we need this feature urgently. We have a similar feature in sqlserver also 
sources:
    my-mssql-source:
        kind: mssql
        host: 127.0.0.1
        port: 1433
        **database: my_db**
        user: ${USER_NAME}
        password: ${PASSWORD}

duwenxin99:
> Is there any timeline when this feature will be merged. We have usecase where we need this feature urgently. We have a similar feature in sqlserver also sources: my-mssql-source: kind: mssql host: 127.0.0.1 port: 1433 **database: my_db** user: ${USER_NAME} password: ${PASSWORD}

This will only be in the Bigquery source.

> Is there any timeline when this feature will be merged. We have usecase where we need this feature urgently. We have a similar feature in sqlserver also sources: my-mssql-source: kind: mssql host: 127.0.0.1 port: 1433 **database: my_db** user: ${USER_NAME} password: ${PASSWORD}

Hi @vksinha10, we have a [PR](https://github.com/googleapis/genai-toolbox/pull/1093/files) under review. Hopefully we can get it in the next release. Thanks

vksinha10:
@duwenxin99  We are still waiting for this feature to be merged. This is something which we require in our system to go as soon as possible.Thanks

duwenxin99:
> @duwenxin99 We are still waiting for this feature to be merged. This is something which we require in our system to go as soon as possible.Thanks

Hi @vksinha10, we already merged implementations of this feature for several BQ tools and are rolling out support for the rest. Please check out https://github.com/googleapis/genai-toolbox/issues/873 for updates.

PR #1047, opened 2025-07-31
feat(sources/postgres): add support for queryParams

### LINKED ISSUE #963: postgres does not support configuring sslmode
### Prerequisites

### Toolbox version

0.9.0

### Environment

   `Linux ip-192-168-4-30 5.4.0-216-generic #236-Ubuntu SMP Fri Apr 11 19:53:21 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

### Client

Haven't gotten this far as it's an issue related to the source.

### Expected Behavior

Most Postgres clients I've worked with typically allow specifying the `sslmode` as a configuration option so I was a bit surprised that the toolbox didn't allow that. I was digging through the code and saw it manually configured to `disable` in a few location, but was not specified as part of the postgres source.

### Current Behavior

I was attempting to use this tool to connect to a ClickHouse database using it's postgres port. When using the out of box postgres connector, it repeatedly failed to connect to the ClickHouse database during startup.

A mitigation for this is to use the MySQL connector as it will connect to the ClickHouse MySQL port without an issue. This will allow you to connect the tools easily, but will require that queries be written using MySQL statements.

### Steps to reproduce?

I haven't tried this with a postgres container, but I'm confident it can be reproduced consistently with a ClickHouse container. I have a copy of my configuration available in OpenSource for reference:

https://github.com/mjpitz/mjpitz/blob/main/infra/lab/metrics/docker-compose.yaml#L101-L124

If you change the `otel` source to use a `postgres` kind and the port to `9005`, then try to deploy the containers, you should be able to inspect the logs of the `lab-metrics-mcp-toolbox-1` to see it failing to connect to the clickhouse container.

### Additional Details

Was hard to tell if this was supposed to be a new feature or a bug... I would definitely classify it more as an enhancement to the existing implementation.

Yuan325:
Add support for `queryParams` for users that would like to connect with additional query parameters.

```
sources:
    my-pg-source:
        kind: postgres
        host: 127.0.0.1
        port: 5432
        database: my_db
        user: ${USER_NAME}
        password: ${PASSWORD}
        queryParams:
            sslmode: verify-full
            sslrootcert: /tmp/ca.crt
```

`queryParams` will be added as raw query of the database connection url.

Fixes #963


kurtisvg:
why `rawQuery`? 

Rather than a string, should we allow for a map of configurations?
```yaml
queryParams:
    sslmode: "verify-full"
    sslrootcert: "/tmp/ca.crt"
```

kurtisvg (approved):

PR #1053, opened 2025-08-01
chore: refactoring suggested by gemini

drstrangelooker:
Refactoring code in the Looker tools as suggested by Gemini


drstrangelooker:
@Yuan325 This is a minor update where I just refactored some of my code that I repeated into reusable chunks. For instance I have several tools that use the same parameters, so I factored out the parameter metadata and the processing of parameters into a helper function.

Yuan325 (approved):
Thank you!

PR #1062, opened 2025-08-04
feat(prebuiltconfigs/cloud-sql-postgres): Introduce additional parameter to limit context in list tables

prernakakkar-google:


internal/prebuiltconfigs/tools/cloud-sql-postgres.yaml
```diff
@@ -72,26 +72,36 @@ tools:
...
+                    ) 
+                  END AS object_details
             FROM table_info ti ORDER BY ti.schema_name, ti.table_name;
         parameters:
             - name: table_names
               type: string
               description: "Optional: A comma-separated list of table names. If empty, details for all tables in user-accessible schemas will be listed."
+            - name: output_format
+              type: string
+              description: "Use 'simple' for just names, or 'detailed' for the full schema."
```
averikitsch:
```suggestion
              description: "Optional: Use 'simple' to return table names only or  use 'detailed' to return the full information schema."
```
Thoughts on this updated description?

aarongeracci757-sys: > ```suggestion
>               description: "Optional: Use 'simple' to return table names only or  use 'detailed' to return the full information schema."
> ```
> Thoughts on this updated description?

Why not?

averikitsch (approved):

PR #1075, opened 2025-08-05
feat: Add bearer token support from code

prernakakkar-google:
image


Yuan325 (commented):
Can we update this in the docs as well? for the new field. Thank you! :)

internal/tools/http/http.go
```diff
@@ -309,6 +315,25 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
...
 	}
+
+	// If Bearer is true, generate an OAuth2 token and set the Authorization header.
+	// This will override any static Authorization header in the configuration.
+	if t.Bearer {
+		// This uses Application Default Credentials to generate an OAuth2 token.
+		// It assumes the tool is being run in an environment with appropriate credentials
+		// (e.g., a GCE VM, or with `gcloud auth application-default login`).
+		// The default scope is for Google Cloud Platform services.
+		tokenSource, err := google.DefaultTokenSource(ctx, "https://www.googleapis.com/auth/cloud-platform")
```
averikitsch, line 327: This seems fine for now, but at some point we may need this to be configurable https://developers.google.com/identity/protocols/oauth2/scopes

docs/en/resources/tools/http/http.md
```diff
@@ -139,6 +139,27 @@ my-http-tool:
...
+To enable this, set the `bearer` field to `true`:
+
+```yaml
+my-http-tool:
+    kind: http
+    source: my-http-source
+    method: GET
+    path: /some/google/cloud/api
+    description: A tool that calls a Google Cloud API
+    bearer: true
```
averikitsch: I'm not sure I want this on the tool. If I have the google api as my source and I need to authenticate all calls to that source, I may just want to set this once on the source. That does create the need to refresh the token though. 
We also need a better name to signal the Google credentials/ADC

internal/tools/http/http.go
prernakakkar-google: Had a chat with Andy and he mentioned we should restrict to just Database/Cloud Scope as a team. Let me know how to take it forward.

docs/en/resources/tools/http/http.md
prernakakkar-google: I can rename as you mentioned. To ensure it gets refreshed I chose tool. Let me know if we should have it in source.

prernakakkar-google: Changed it to googleCloudOauth2 to add clarity

internal/tools/http/http.go
averikitsch: No action needed for now

docs/en/resources/tools/http/http.md
averikitsch: Did you push these changes?

prernakakkar-google: Thanks for pointing out, I forgot to push on remote.

docs/en/resources/tools/http/http.md
```diff
@@ -139,6 +139,27 @@ my-http-tool:
         type: integer
 ```
 
+### Google Cloud OAuth2 Token
+
+Toolbox can automatically generate and add a Google Cloud OAuth2 token to the
+`Authorization` header using Google Application Default Credentials.
```
averikitsch, line 145: We should add more context and a link to ADC. This will only work if local gcloud credentials are set or if they are run Toolbox on Google Cloud

docs/en/resources/tools/http/http.md
averikitsch: > I can rename as you mentioned. To ensure it gets refreshed I chose tool. Let me know if we should have it in source.

I'm going to ask for feedback from the team if this is the experience we want or if we want to set it on the source. IMO, I don't want to add this field to every tool I want to add it on the source. We need to weigh this against the increased technical complexities

vksinha10:
just want to chcek will this support ADC(application default credentials) where i can use service account to authenticate my google api

averikitsch:
Due to security concerns, we will not be moving forward with this change.

PR #1076, opened 2025-08-05
ci: enable C++ compiler for duckDB

averikitsch:


averikitsch:
This will need more work. See https://github.com/marcboeker/go-duckdb/issues/94

PR #1086, opened 2025-08-05
feat: yaml builder with source templates

AlexTalreja:
Allow users to upload `.yaml` files, edit through the UI, and download the result. Also adds a `source` drop-down, allowing users to automatically populate the yaml with necessary fields based on the selected source.

### DIFF
--- a/internal/server/static/css/style.css
+++ b/internal/server/static/css/style.css
@@ -41,6 +41,7 @@ body {
   display: flex;
   flex-direction: column;
   padding: 15px;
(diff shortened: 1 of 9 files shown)


Yuan325:
This feature will add maintenance overhead (especially if we have to maintain the .json file that indicate config fields). Will hold this feature till we have a way to set config fields dynamically.

Yuan325:
Opened an issue for this feature -- #1618. Closing this PR since we won't be making updates or support this in the short term.

PR #1100, opened 2025-08-06
chore: add instructions and links to docs in UI

AlexTalreja:
Add instructions on basic usage + links to the public documentation on the main content area of the Homepage, Tools page, and Toolsets page.

### DIFF
--- a/internal/server/static/js/mainContent.js
+++ b/internal/server/static/js/mainContent.js
@@ -57,7 +57,7 @@ function getToolInstructions() {
         <p class="resource-description">
           Tools define actions an agent can take, such as running a SQL statement or interacting with a source. 
           You can define Tools as a map in the <code>tools</code> section of your <code>tools.yaml</code> file. <br><br>
-          Typically, tools also use <strong>parameters</strong>. Parameters for each Tool will define what inputs the agent will need to provide to invoke them. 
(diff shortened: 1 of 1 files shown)


internal/server/static/js/mainContent.js
```diff
@@ -30,11 +30,51 @@ function renderMainContent(containerId, idString) {
...
+function getToolInstructions() {
+    return `
+      <div class="resource-instructions">
+        <h1 class="resource-title">Tools</h1>
+        <p class="resource-intro">To inspect and test a tool, please click on one of your tools to the left.</p>
+        <h2 class="resource-subtitle">What are Tools?</h2>
+        <p class="resource-description">
+          Tools define actions an agent can take, such as running a SQL statement or interacting with a source. 
+          You can define Tools as a map in the <code>tools</code> section of your <code>tools.yaml</code> file. <br><br>
+          Typically, tools also use <strong>parameters</strong>. Parameters for each Tool will define what inputs the agent will need to provide to invoke them.
```
Yuan325:
```suggestion
          Some tools also uses <strong>parameters</strong>. Parameters for each Tool will define what inputs the agent will need to provide to invoke them. 
```

It's not uncommon for tools to not have parameters~ Some parameters are also pre-defined (such as `execute-sql` tools) so user don't have to configure it in tools.yaml.

Yuan325 (approved):
Thank you!

PR #1102, opened 2025-08-07
feat(source/http): Add User Agent to calls

prernakakkar-google:


internal/sources/http/http.go
```diff
@@ -93,6 +93,15 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
 	}
 
+	ua, err := util.UserAgentFromContext(ctx)
+	if err != nil {
+		fmt.Printf("Error in User Agent retrieval: %s", err)
+	}
+	if r.DefaultHeaders == nil {
+		r.DefaultHeaders = make(map[string]string)
+	}
+	r.DefaultHeaders["User-Agent"] = ua
```
averikitsch, line 106: We should append out user-agent if one already exists

averikitsch (changes requested):

internal/sources/http/http.go
prernakakkar-google: ADded similar to prebuilt  +"."

internal/sources/http/http.go
```diff
@@ -93,6 +93,19 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
+		fmt.Printf("Error in User Agent retrieval: %s", err)
+	}
+	if r.DefaultHeaders == nil {
+		r.DefaultHeaders = make(map[string]string)
+	}
+	if existingUA, ok := r.DefaultHeaders["User-Agent"]; ok && existingUA != "" {
+		r.DefaultHeaders["User-Agent"] += "." + ua
+	} else {
+		r.DefaultHeaders["User-Agent"] = ua
+	}
```
averikitsch: nit:
```suggestion
	if existingUA, ok := r.DefaultHeaders["User-Agent"]; ok && existingUA != "" {
		ua += existingUA + " " + ua
	}
 r.DefaultHeaders["User-Agent"] = ua
```

internal/sources/http/http.go
```diff
@@ -93,6 +93,19 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
 
+	ua, err := util.UserAgentFromContext(ctx)
+	if err != nil {
+		fmt.Printf("Error in User Agent retrieval: %s", err)
+	}
+	if r.DefaultHeaders == nil {
+		r.DefaultHeaders = make(map[string]string)
+	}
+	if existingUA, ok := r.DefaultHeaders["User-Agent"]; ok && existingUA != "" {
+		r.DefaultHeaders["User-Agent"] += "." + ua
```
averikitsch: User agents are separate by spaces

internal/sources/http/http.go
averikitsch: This is so we don't overwrite user set user agents. Prebuilt uses "+" to append metadata

internal/sources/http/http.go
```diff
@@ -93,6 +93,18 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
 
+	ua, err := util.UserAgentFromContext(ctx)
+	if err != nil {
+		fmt.Printf("Error in User Agent retrieval: %s", err)
+	}
+	if r.DefaultHeaders == nil {
+		r.DefaultHeaders = make(map[string]string)
+	}
+	if existingUA, ok := r.DefaultHeaders["User-Agent"]; ok && existingUA != "" {
+		ua += existingUA + " " + ua
```
averikitsch:
```suggestion
		ua = ua + " " + existingUA
```

kurtisvg (changes requested):
Can we include a description for the PR on what this is doing and why?

internal/sources/http/http.go
```diff
@@ -93,6 +93,18 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
+	ua, err := util.UserAgentFromContext(ctx)
+	if err != nil {
+		fmt.Printf("Error in User Agent retrieval: %s", err)
+	}
+	if r.DefaultHeaders == nil {
+		r.DefaultHeaders = make(map[string]string)
+	}
+	if existingUA, ok := r.DefaultHeaders["User-Agent"]; ok && existingUA != "" {
+		ua = ua + " " + existingUA
+	}
```
kurtisvg: The `existingUA != ""` check is redundant and not needed. 

```suggestion
	if existingUA, ok := r.DefaultHeaders["User-Agent"]; ok {
		ua = ua + " " + existingUA
	}
```

kurtisvg (approved):
LGTM, but let's make sure we have a good PR description before merge.

PR #1113, opened 2025-08-07
chore: add "waiting for response" label

averikitsch:


.github/labels.yaml
```diff
@@ -84,6 +84,10 @@
   color: 8befd7
   description: 'Status: waiting for feedback from community or issue author.'
 
+- name: 'waiting for response'
```
Yuan325:
```suggestion
- name: 'status: waiting for response'
```

nit: to be consistent with the other statuses

PR #1119, opened 2025-08-08
feat(source/bigquery): Add BigQuery client cache with end-user OAuth token

duwenxin99:


internal/sources/bigquery/bigquery.go
```diff
@@ -84,33 +84,46 @@ func (r Config) SourceConfigKind() string {
...
-
 func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.Source, error) {
 	if r.WriteMode == "" {
 		r.WriteMode = WriteModeAllowed
 	}
 
-	if r.WriteMode == WriteModeProtected && r.UseClientOAuth {
+    if r.WriteMode == WriteModeProtected && r.UseClientOAuth {
 		return nil, fmt.Errorf("writeMode 'protected' cannot be used with useClientOAuth 'true'")
+    }
```
kurtisvg: this is here and on L106-L107?

internal/sources/bigquery/bigquery.go
```diff
@@ -84,33 +84,46 @@ func (r Config) SourceConfigKind() string {
...
-	var restService *bigqueryrestapi.Service
-	var tokenSource oauth2.TokenSource
-	var clientCreator BigqueryClientCreator
-	var err error
-
 	if r.UseClientOAuth {
-		clientCreator, err = newBigQueryClientCreator(ctx, tracer, r.Project, r.Location, r.Name)
+		// use client OAUth
+		if r.WriteMode == WriteModeProtected {
+			return nil, fmt.Errorf("writeMode 'protected' cannot be used with useClientOAuth 'true'")
```
kurtisvg: nit: why is this?

internal/sources/bigquery/bigquery.go
```diff
@@ -170,6 +170,58 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
 
 }
 
+// setupClientCaching initializes caches and wraps the base client creator with caching logic.
+func (s *Source) setupClientCaching(baseCreator BigqueryClientCreator) {
```
kurtisvg: nit: should this be attached to Source or take in source as an input?

internal/sources/bigquery/bigquery.go
duwenxin99: It's supposed to exit early if both of these conditions are met. I could move it inside the if block on L106-107 but it seems to be less readable to me.

internal/sources/bigquery/bigquery.go
duwenxin99: I think it's because if `useOAuthClient` is true, a different session will be used each time during invocation since we create a different client each time. But `WriteModeProtected` checks for previous session queries. Details are in this [PR](https://github.com/googleapis/genai-toolbox/pull/1157). @Genesis929 Could you confirm and what's the planned future fix on this?

Genesis929: The main reason is that we cannot currently identify the user, so protected mode only has one session. This setup is not recommended for multiple users, and with useClientOAuth=True, it may not work at all.

A possible solution for useClientOAuth=True is to identify users based on their access token (to get some kind of global, long-term identifier) and then map each user to their own session. This is similar to the cache this PR is implementing, but given that sessions last quite long and access token only last one hour, the access token itself may not be the most appropriate key.

This is from one of my test, we can get a long term identifier "sub"(subject id) based on access token:
https://screenshot.googleplex.com/gWjXPn7yC3TJgNU

Though I'm not sure if this is considered PII, and can we use it.

internal/sources/bigquery/bigquery.go
duwenxin99: updated

internal/sources/bigquery/bigquery.go
kurtisvg: Wenxin and I chatted a bit about this today.

The problem is that with the current mechanism, the session is bound to the client and the client is bound to the token. We would need a way to update the token without changing the client. 

One way we could do this is to use [TokenSource](https://pkg.go.dev/golang.org/x/oauth2#TokenSource), and create a per-user client that uses a token source. We could update that token source when a different oauth2 token comes in. However there's some complication this will cause around race conditions -- if a user has two different sessions they will conflict. If we update the token for one, another tool call may be affected.

I'm OK with leaving this as is for now -- but we should document this somewhere (maybe in the code) why this behavior exists and the potential workaround (and complications for it).

PR #1120, opened 2025-08-09
feat(source/redis): add redis execute command support

### LINKED ISSUE #1024: Add support - redis executor
### Prerequisites

Currently, we need to define a series of preset commands to operate Redis, which is troublesome. I believe a more flexible usage would be more convenient, only requiring a single client for execution

### Suggested Solution(s)

like mysql-sql and mysql-execute-sql

### Alternatives Considered

_No response_

### Additional Details

_No response_

bpranava:
Fixes #1024 
This PR 
- Adds a execute command tool for redis
Proposed tool type is - 
```yaml
tools:
 execute_cmd_tool:
    kind: redis-execute-cmd
    source: my-redis-instance
    description: Use this tool to execute a redis command.
```
- Adds test cases to test for HGETALL, JSON.GET, RPUSH, LRANGE (array types)


internal/tools/redis/redisexecutecmd/redisexecutecmd.go
```diff
@@ -0,0 +1,186 @@
...
+func (t Tool) McpManifest() tools.McpManifest {
+	return t.mcpManifest
+}
+
+// ParseParams implements tools.Tool.
+func (t Tool) ParseParams(data map[string]any, claims map[string]map[string]any) (tools.ParamValues, error) {
+	return tools.ParseParams(t.Parameters, data, claims)
+}
+
+func toAnySlice(query string) []any {
```
duwenxin99: I think it doesn't work if a string with spaces is used as an input? For example, "HSET myhash field1 'Hello world'" will be parsed into ["HSET ", "myhash", "field1", "'Hello",  "world'"]

duwenxin99 (changes requested):
Hi @bpranava, thanks for the contribution!
It may be better to make the input a []any (a two dimensional slice) and we'll execute each slice contained as a command. This way we don't need deal with the parsing problem. Let me know what you think.

internal/tools/redis/redisexecutecmd/redisexecutecmd.go
bpranava: Thanks for pointing this out.

bpranava:
> Hi @bpranava, thanks for the contribution! It may be better to make the input a []any (a two dimensional slice) and we'll execute each slice contained as a command. This way we don't need deal with the parsing problem. Let me know what you think.

Thank you @duwenxin99, that makes sense. I'll test this out with LLMs and get back.

bpranava:
> > Hi @bpranava, thanks for the contribution! It may be better to make the input a []any (a two dimensional slice) and we'll execute each slice contained as a command. This way we don't need deal with the parsing problem. Let me know what you think.
> 
> Thank you @duwenxin99, that makes sense. I'll test this out with LLMs and get back.

@duwenxin99 , do you think it would be a good idea to introduce a new AnyParameter type among [parameters](https://github.com/googleapis/genai-toolbox/blob/main/internal/tools/parameters.go) to support this usecase?
Then I could replace the ["cmd" parameter ](https://github.com/bpranava/genai-toolbox/blob/86710adacbe9e72e300308a6afdf1fdc13be4118/internal/tools/redis/redisexecutecmd/redisexecutecmd.go#L81) with something like 

```go
tools.NewArrayParameter("cmd", "The command to execute.", tools.NewAnyParameter("entity", "Entity in a redis command"))
```

duwenxin99:
> > > Hi @bpranava, thanks for the contribution! It may be better to make the input a []any (a two dimensional slice) and we'll execute each slice contained as a command. This way we don't need deal with the parsing problem. Let me know what you think.
> > 
> > 
> > Thank you @duwenxin99, that makes sense. I'll test this out with LLMs and get back.
> 
> @duwenxin99 , do you think it would be a good idea to introduce a new AnyParameter type among [parameters](https://github.com/googleapis/genai-toolbox/blob/main/internal/tools/parameters.go) to support this usecase? Then I could replace the ["cmd" parameter ](https://github.com/bpranava/genai-toolbox/blob/86710adacbe9e72e300308a6afdf1fdc13be4118/internal/tools/redis/redisexecutecmd/redisexecutecmd.go#L81) with something like
> 
> ```go
> tools.NewArrayParameter("cmd", "The command to execute.", tools.NewAnyParameter("entity", "Entity in a redis command"))
> ```

Could you create an issue for supporting generic array parameters? We can certainly discuss it.
On the other hand, I think redis client libraries automatically convert any types to strings, meaning that an integer is converted to a string before being stored. So it should be the same to just initialize the input as a nested string array parameter. @bpranava

bpranava:
> > > > Hi @bpranava, thanks for the contribution! It may be better to make the input a []any (a two dimensional slice) and we'll execute each slice contained as a command. This way we don't need deal with the parsing problem. Let me know what you think.
> > > 
> > > 
> > > Thank you @duwenxin99, that makes sense. I'll test this out with LLMs and get back.
> > 
> > 
> > @duwenxin99 , do you think it would be a good idea to introduce a new AnyParameter type among [parameters](https://github.com/googleapis/genai-toolbox/blob/main/internal/tools/parameters.go) to support this usecase? Then I could replace the ["cmd" parameter ](https://github.com/bpranava/genai-toolbox/blob/86710adacbe9e72e300308a6afdf1fdc13be4118/internal/tools/redis/redisexecutecmd/redisexecutecmd.go#L81) with something like
> > ```go
> > tools.NewArrayParameter("cmd", "The command to execute.", tools.NewAnyParameter("entity", "Entity in a redis command"))
> > ```
> 
> Could you create an issue for supporting generic array parameters? We can certainly discuss it. On the other hand, I think redis client libraries automatically convert any types to strings, meaning that an integer is converted to a string before being stored. So it should be the same to just initialize the input as a nested string array parameter. @bpranava

Thanks @duwenxin99 , I have made the changes. I'll go ahead an create an issue to discuss about "AnyParameter"s. 
Just to confirm, the redis-execute-cmd now works only if the input is an array of strings and fails otherwise.

averikitsch:
Hi @bpranava , I apologize for the delay in the review. Do you have time to resolve the the conflicts?

PR #1123, opened 2025-08-11
docs: fix connectionstring in couchbase sources

Ganga4060:
Changed connection string localhost


docs/en/resources/sources/couchbase.md
```diff
@@ -22,7 +22,7 @@ allowing tools to execute SQL queries against it.
 sources:
     my-couchbase-instance:
         kind: couchbase
-        connectionString: couchbase://localhost:8091
+        connectionString: couchbase://localhost:127.0.0.1
```
twishabansal, line 25: localhost is already 127.0.0.1

PR #1150, opened 2025-08-13
chore(testing): Set up initial directory structure for sample testing

AnmolShukla2002:
This PR restructures the quickstart directories for Go and JavaScript to align with the new Automated Sample Testing Strategy.


twishabansal:
@AnmolShukla2002 thanks for raising this PR. But seems like there is a lot of content in this one. Would you mind breaking it down into multiple PRs to make it easy to review or revert back in case there are any issues?

PS: It would be nice if you can rebase to the include region tag PR and then include the new code files directly into the quickstart.

PR #1157, opened 2025-08-14
feat(bigquery): add optional write mode config

Genesis929:


docs/en/resources/sources/bigquery.md
```diff
@@ -96,3 +97,4 @@ sources:
 | kind      |  string  |     true     | Must be "bigquery".                                                           |
 | project   |  string  |     true     | Id of the GCP project that the cluster was created in (e.g. "my-project-id"). |
 | location  |  string  |    false     | Specifies the location (e.g., 'us', 'asia-northeast1') in which to run the query job. This location must match the location of any tables referenced in the query. The default behavior is for it to be executed in the US multi-region |
+| write_mode |  string  |    false     | Controls the write behavior for `bigquery-execute-sql`. `allowed` (default): All queries are permitted. `blocked`: Only `SELECT` statements are allowed. `protected`: Enables session-based execution. Only `SELECT` statements and writes to the session's temporary dataset are allowed. For `bigquery-sql`, the `writemode` restrictions do not apply; however, they will operate within the same session when the mode is `protected`. |
```
shobsi: suggest rewording to make it clear that SELECT works for all tables:

"Only `SELECT` statements and writes to the session's temporary dataset are allowed." -> "In addition to `SELECT` statements, writes to the session's temporary dataset are allowed."

docs/en/resources/tools/bigquery/bigquery-execute-sql.md
```diff
@@ -19,6 +19,16 @@ It's compatible with the following sources:
...
 
+## Write Mode Behavior
+
+The behavior of this tool is influenced by the `write_mode` setting on its `bigquery` source:
+
+- **`allowed` (default):** All SQL statements are permitted.
+- **`blocked`:** Only `SELECT` statements are allowed. Any other type of statement (e.g., `INSERT`, `UPDATE`, `CREATE`) will be rejected.
+- **`protected`:** This mode enables session-based execution. Only `SELECT` statements or write operations targeting the session's temporary dataset (e.g., `CREATE TEMP TABLE ...`) are allowed. This prevents modifications to permanent datasets while allowing stateful, multi-step operations within a secure session.
+
+> **Note:** This tool is intended for developer assistant workflows with human-in-the-loop and shouldn't be used for production agents.
```
shobsi, line 43: Is this note because of the default write mode being "allowed"? Can we not make it "blocked" by default (like we do in ADK)?

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -83,12 +84,24 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
...
 	}
 
-	sqlParameter := tools.NewStringParameter("sql", "The sql to execute.")
+	var sqlParamDescription string
+	switch s.BigQueryWriteMode() {
+	case bigqueryds.WriteModeBlocked:
+		sqlParamDescription = "The SQL to execute. In 'blocked' mode, only SELECT statements are allowed; " +
+			"other statement types will fail."
+	case bigqueryds.WriteModeProtected:
+		sqlParamDescription = fmt.Sprintf("The SQL to execute. In 'protected' mode, only SELECT statements and writes to "+
```
shobsi: LLM is going to have no clue how to interpret "In 'protected' mode", so maybe remove that part?

docs/en/resources/sources/bigquery.md
Genesis929: Updated.

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -137,11 +154,42 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error)
...
+			{Key: "session_id", Value: t.Session.ID},
+		}
+	}
+
+	dryRunJob, err := dryRunQuery(ctx, t.RestService, t.Client.Project(), query.Location, sql, query.ConnectionProperties)
 	if err != nil {
 		return nil, fmt.Errorf("query validation failed during dry run: %w", err)
 	}
 
+	if dryRunJob.Statistics == nil || dryRunJob.Statistics.Query == nil {
```
shobsi: Could we add a comment with an example query? Looks like this logic should be propagated to ask tools as well?

docs/en/resources/tools/bigquery/bigquery-execute-sql.md
Genesis929: No, this note is the same note for other tools in sources like mssql, mysql and more.

And if we make it blocked by default, it may be confusing, and will be a breaking change, and the tool will have a different default than other execute-sql tools.

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
Genesis929: I don't really have a example, this is just for safety, normally I won't expect the code can be ever called.

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
Genesis929: Removed.

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
Genesis929: Error message updated.

shobsi: as discussed offline, if we cannot explain "semantic errors" then its better to omit the term from the code and the error message

docs/en/resources/sources/bigquery.md
```diff
@@ -145,5 +147,6 @@ sources:
...
+| writeMode       |  string  |    false     | Controls the write behavior for `bigquery-execute-sql`. `allowed` (default): All queries are permitted. `blocked`: Only `SELECT` statements are allowed. `protected`: Enables session-based execution. `SELECT` statements can be used on all tables, while write operations are allowed only for the session's temporary dataset. For tools like `bigquery-sql`, `bigquery-forecast`, and `bigquery-analyze-contribution`, the `writeMode` restrictions do not apply; however, they will operate within the same session when the mode is `protected`. **Note:** The `protected` mode cannot be used with `useClientOAuth: true` and is not recommended for multi-user server environments as all users would share the same session. |
```
duwenxin99: Can we have a more clear definition of what a `session` is here (e.g every request to the same defined source will share the same session) ? I think some user may assume it to be different than it is. We should also detail the expiration time.

internal/tools/bigquery/bigquerysql/bigquerysql.go
```diff
@@ -187,7 +191,7 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken
 		if arrayParam, ok := p.(*tools.ArrayParameter); ok {
 			// Handle array types based on their defined item type.
 			lowLevelParam.ParameterType.Type = "ARRAY"
-			itemType, err := BQTypeStringFromToolType(arrayParam.GetItems().GetType())
+			itemType, err := bqutil.BQTypeStringFromToolType(arrayParam.GetItems().GetType())
```
duwenxin99, line 194: Does it handle nested arrays?

duwenxin99 (changes requested):

internal/tools/bigquery/bigquerysql/bigquerysql.go
Genesis929: Ah, I thought we don't support that. Also not sure how that is presented in yaml, is there an example?

docs/en/resources/sources/bigquery.md
Genesis929: Updated.

PR #1161, opened 2025-08-15
fix(tools/mysql): fix encoded text for mysql

### LINKED ISSUE #840: [tool:mysql-sql] Toolbox returns base64-encoded values for MySQL text columns (plain text in DB)
### Prerequisites

### Toolbox version

latest as of 7/9/25

### Environment

Toolbox Source: https://github.com/googleapis/genai-toolbox (commit 72a72827973b98186dced5589d71573411f6a3ab)
Built from source on 2025-07-09
Docker Engine: 24.0.7, macOS 14.5, M1 chip
Database: MySQL 8.0, containerized
Command: /toolbox toolbox --tools-file /app/tools.yaml --address 0.0.0.0 --port 8080

### Client

MCP Inspector (Community Tool)
Latest version as of 2025-07-09 (installed via npm)
URL: http://localhost:8888/mcp
Transport: streamable-http

### Expected Behavior

I expected the MCP Toolbox server to return string fields such as store, date, and total_amount as plain, human-readable text (e.g., "Trader Joe's", "2025-05-25", "148.23"), matching the output of direct SQL queries to the underlying MySQL database using JSON_UNQUOTE.

### Current Behavior

All string fields in the MCP Toolbox response are returned as base64-encoded values (e.g., "VHJhZGVyIEpvZSdz" instead of "Trader Joe's"). This occurs for all tools that query these fields, even though direct SQL queries to the database return standard string values. This unexpected encoding makes the output difficult to use with clients and downstream tools, and does not match the behavior described in the documentation.

```json
{
type:
"text"
text:
"{"date":"MjAyNS0wNS0yNg==","item_count":32,"receipt_id":20,"store":"VHJhZGVyIEpvZSdz","total_amount"..."
}
```

### Steps to reproduce?

1. Set up a minimal MySQL 8 container with a database and user.
2. Create a receipts table with a JSON column and insert test data:
```sql
CREATE TABLE receipts (
  receipt_id INT PRIMARY KEY AUTO_INCREMENT,
  storeName JSON,
  status VARCHAR(20)
);

INSERT INTO receipts (storeName, status)
VALUES ('{"verified": "Whole Foods"}', 'verified');
```
3. Create a minimal tools-test.yaml for MCP Toolbox:
```yaml
sources:
  test-mysql:
    kind: mysql
    host: mysql-test
    port: 3306
    database: testdb
    user: testuser
    password: testpass

tools:
  filter-test-receipts:
    kind: mysql-sql
    source: test-mysql
    description: Filter test receipts
    statement: |
      SELECT * FROM receipts
    parameters: []

toolsets:
  default:
    - filter-test-receipts
```
4. Start MCP Toolbox with config to connect to the MySQL and tools-test.yaml
5. Query the MCP server using MCP Inspector or HTTP POST
```json
{
  "tool": "filter-test-receipts",
  "params": {}
}
```
6. Observe the output - The value for storeName in the response will be base64-encoded, e.g.
```
"storeName": "eyJ2ZXJpZmllZCI6ICJXaG9sZSBGb29kcyJ9"
```
Decoding this gives the expected JSON: {"verified": "Whole Foods"}

### Additional Details

Values stored in DB are NOT base64 encoded

Yuan325:
Update the type conversion to according to the types that mysql driver returns: https://github.com/go-sql-driver/mysql/blob/v1.9.3/fields.go

All `scanTypeBytes` or `scanTypeString` or `scanTypeNullString` will now be converted into string. Remaining types (numeric types or `scanTypeUnknown`) are all returned as is.

Separately handle "JSON" to prevent double marshaling.

Fixes #840


kurtisvg (approved):
Do we need to make similar changes to oceanbase and mindsdb?

Yuan325:
@kurtisvg Do they have the same encoding issues?

kurtisvg:
@Yuan325 they both use mysql drivers under the hood -- so maybe? 

MindsDB is not merged yet. How hard is it to add a regression test here?

Yuan325:
I'll add Oceanbase's fix in a separate PR and have their engineer review it to make sure that they are aware. Same with MindsDB.

PR #1169, opened 2025-08-16
chore: add options for all tests

Yuan325:
Add options for `RunToolInvokeTest()`, `RunMCPToolCallMethod()`, `RunExecuteSqlToolInvokeTest()`.

When I was trying to update integration test for Firebird. They have a different syntax from the usual `SELECT 1` that the other DBs are using. If we want to update the `SELECT 1` statement in the test cases, we will have to add a new argument and update all the other integration tests.

With `Options`, we can just add customization without having the need to update all other integration tests. This allow the integration tests to be more flexible. With this, we can also remove the `GetNonSpannerWant()` since those can be set as default values for the `Options`.

### DIFF
--- a/tests/mongodb/mongodb_integration_test.go
+++ b/tests/mongodb/mongodb_integration_test.go
@@ -104,7 +104,7 @@ func TestMongoDBToolEndpoints(t *testing.T) {
 	myToolId3NameAliceWant := `[{"_id":5,"id":3,"name":"Alice"}]`
 	myToolById4Want := `[{"_id":4,"id":4,"name":null}]`
 	mcpMyFailToolWant := `invalid JSON input: missing colon after key `
-	mcpMyToolId3NameAliceWant := `{"jsonrpc":"2.0","id":"my-tool","result":{"content":[{"type":"text","text":"{\"_id\":5,\"id\":3,\"name\":\"Alice\"}"}]}}`
+	mcpMyToolId3NameAliceWant := `{"jsonrpc":"2.0","id":"my-simple-tool","result":{"content":[{"type":"text","text":"{\"_id\":5,\"id\":3,\"name\":\"Alice\"}"}]}}`
 
 	// Run tests
 	tests.RunToolGetTest(t)
@@ -131,9 +131,6 @@ func TestMongoDBToolEndpoints(t *testing.T) {
 	aggregate1Want := `[{"id":2}]`
 	aggregateManyWant := `[{"id":500},{"id":501}]`
 	RunToolAggregateInvokeTest(t, aggregate1Want, aggregateManyWant)
-
-	mcpInvokeParamWantFindOne := `{"jsonrpc":"2.0","id":"my-simple-tool","result":{"content":[{"type":"text","text":"{\"_id\":5,\"id\":3,\"name\":\"Alice\"}"}]}}`
-	tests.RunMCPToolCallMethod(t, mcpInvokeParamWantFindOne, failInvocationWant)
 }
 
 func RunToolDeleteInvokeTest(t *testing.T, delete1Want, deleteManyWant string) {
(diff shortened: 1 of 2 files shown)


tests/option.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Apply provided options
+	for _, option := range options {
+		option(invokeTestOption)
+	}
+
+	return invokeTestOption
+}
+
+func WithInvoketestSelect1Want(s string) InvokeTestOption {
```
duwenxin99: I think we should add a comment for each option to explain what it does. A brief example would be even better

tests/option.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+func WithDisableNullParam() InvokeTestOption {
+	return func(c *InvokeTestConfig) {
+		c.supportNullParam = false
+	}
+}
+
+func WithDisableArray() InvokeTestOption {
```
duwenxin99: Maybe `DisableArrayTest`?

tests/option.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+func WithNullString(s string) InvokeTestOption {
+	return func(c *InvokeTestConfig) {
+		c.nullString = s
+	}
+}
+
+func WithDisableNullParam() InvokeTestOption {
```
duwenxin99: Maybe `DisableNullParamTest`? Otherwise it's disabling null parameter which doesn't really make sense

tests/option.go
```diff
@@ -0,0 +1,268 @@
...
+	// Apply provided options
+	for _, option := range options {
+		option(templateParamOption)
+	}
+
+	return templateParamOption
+}
+
+// WithIgnoreDdl is the option function to configure ignoreDdl.
+func WithIgnoreDdl() TemplateParamOption {
```
duwenxin99: I think we need more explanatory comments than this for our contributors to understand.

tests/option.go
```diff
@@ -0,0 +1,268 @@
...
+	invokeParamWant  string
+	invokeIdNullWant string
+	nullString       string
+	supportNullParam bool
+	supportArray     bool
+}
+
+type InvokeTestOption func(*InvokeTestConfig)
+
+// NewInvokeTestConfig creates a new InvokeTestConfig instances with options.
```
duwenxin99: Can we include a brief example here of how this should be used?

duwenxin99: tbh I think using options for all the expected values is making it difficult to set up for external contributors. They will probably just use `InvokeTestConfig ` directly because it's much easier.

duwenxin99 (changes requested):
Can we separate the expected values input with the test config options?
For example, we have these values in a struct that are not set with options:
```
type InvokeTestExpectedValues struct {
	select1Want      string
	invokeParamWant  string
	invokeIdNullWant string
	nullString       string
}
```
We can create a `defaultExpectedValues` of this struct to be reused.
Then for test optional test cases, we can have test options like this to disable individual test case:
```
func DisableArrayTest(){}
```

The final usage would be
```
tests.RunToolInvokeTest(t, toolInvokeExpectedValues, test.DisableArrayTest())
```

I think it would make it easier for external contributors to understand our test structure and also more readable than setting expected values with options.

tests/option.go
Yuan325: @duwenxin99 the expected values are only needed if it differs from the default ones from `NewInvokeTestConfig()`.

Yuan325:
@duwenxin99 Golang only allow one functional Options per function. We can't separate it unless we define the `ExpectedValues` as a simple struct. But will that be more complicated for having the user to look into two different configuration options though (expected values & config specifics (config supports and create statements)). 

I also tried to reorganize the config to have all the wants at the top of the struct, and test specific configs at the bottom.


Added comments to all functions and added example at each `New...Config()` functions. Also updated options that disable specific support to `DisableXyzTest()` instead of `WithDisableXyz()`.

Yuan325:
For values that differs across all or most databases, those values will be kept as input arguments. Else, it is all moved to Options.

PR #1178, opened 2025-08-18
feat(server/mcp): Support `ping` mechanism

duwenxin99:
Send back an empty result with a `ping` request following the [MCP ping spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping).

Screenshot 2025-08-18 at 4 20 09 PM


internal/server/mcp/util/ping.go
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+package util
+
+import "github.com/googleapis/genai-toolbox/internal/server/mcp/jsonrpc"
+
+// PingResult defines the structure for the ping response payload.
+type PingResult struct {
+	Message string `json:"message"`
+}
```
Yuan325:
```suggestion
/* Ping */

// A ping, issued by either the server or the client, to check that the other party is still alive. The receiver must promptly respond, or else may be disconnected.
type PingResult struct {
	jsonrpc.Request
}
```

This will need to be included in all the 3 specification's `types.go`. 

The comments and documentation match official schema.ts (e.g. https://github.com/modelcontextprotocol/modelcontextprotocol/blob/3b3874b4fc010f34ef5b106f044a4867534a9499/schema/2025-06-18/schema.ts#L318C1-L318C11).

The reason we do this is incase there is a change in schema in the future, we can just update the version's specific types.go.

Also, from the official schema, it doesn't seem like they have a "Message" string.

internal/server/mcp/util/ping.go
```diff
@@ -0,0 +1,30 @@
...
+
+// PingHandler handles the "ping" method by returning a "pong" response.
```
Yuan325: Similarly, this will stay in each version's `method.go`.

internal/server/mcp_test.go
Yuan325, line 1: In addition to adding test in `TestMcpEndpointWithoutInitialized()`, probably also want to add the `ping` test in `TestMcpEndpoint()` for initialized server. That test will test all cases against all the different version.

Yuan325 (changes requested):

internal/server/mcp/util/ping.go
duwenxin99: This was supposed to be removed. Updated.

internal/server/mcp_test.go
duwenxin99: added tests

internal/server/mcp/v20241105/method.go
```diff
@@ -38,6 +40,15 @@ func ProcessMethod(ctx context.Context, id jsonrpc.RequestId, method string, too
 	}
 }
 
+// pingHandler handles the "ping" method by returning a "pong" response.
```
Yuan325:
```suggestion
// pingHandler handles the "ping" method by returning an empty response.
```

internal/server/mcp/v20250326/method.go
```diff
@@ -38,6 +40,15 @@ func ProcessMethod(ctx context.Context, id jsonrpc.RequestId, method string, too
 	}
 }
 
+// pingHandler handles the "ping" method by returning a "pong" response.
```
Yuan325:
```suggestion
// pingHandler handles the "ping" method by returning an empty response.
```

internal/server/mcp/v20250618/method.go
```diff
@@ -38,6 +40,15 @@ func ProcessMethod(ctx context.Context, id jsonrpc.RequestId, method string, too
 	}
 }
 
+// pingHandler handles the "ping" method by returning a "pong" response.
```
Yuan325:
```suggestion
// pingHandler handles the "ping" method by returning an empty response.
```

Yuan325 (approved):
Last little nit.

PR #1183, opened 2025-08-19
feat(prebuilt/alloydb)!: Add bearer token support for alloydb-wait-for-operation

prernakakkar-google:
## Description
---
The *alloydb-wait-for-operation* tool now automatically obtains a Google Cloud Platform OAuth2 token and uses it as a bearer token to authenticate requests to the AlloyDB Admin API. This ensures secure and proper communication with the API.
image
## PR Checklist
---
- [x] Make sure you reviewed
  [bug/issue](https://github.com/googleapis/langchain-google-alloydb-pg-python/issues/new/choose)

🛠️ Fixes https://github.com/googleapis/langchain-google-alloydb-pg-python/issues/456


internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
```diff
@@ -93,12 +91,11 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (tools.T
...
-	Name         string            `yaml:"name" validate:"required"`
-	Kind         string            `yaml:"kind" validate:"required"`
-	Source       string            `yaml:"source" validate:"required"`
-	Description  string            `yaml:"description" validate:"required"`
-	AuthRequired []string          `yaml:"authRequired"`
-	Headers      map[string]string `yaml:"headers"`
+	Name         string   `yaml:"name" validate:"required"`
+	Kind         string   `yaml:"kind" validate:"required"`
+	Description  string   `yaml:"description" validate:"required"`
+	AuthRequired []string `yaml:"authRequired"`
```
averikitsch, line 97: Should we also think about removing these fields as well?

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
```diff
@@ -142,6 +130,11 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 		InputSchema: inputSchema,
 	}
 
+	baseURL := cfg.BaseURL
```
averikitsch, line 133: I haven't seen customizable endpoints for AlloyDB. Do we want to expose this in the Config? If so we should also add it to the reference

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
```diff
@@ -251,9 +242,15 @@ func (t *Tool) Invoke(ctx context.Context, params tools.ParamValues) (any, error
 
 		req, _ := http.NewRequest(http.MethodGet, urlString, nil)
 
-		for k, v := range t.Headers {
-			req.Header.Set(k, v)
+		tokenSource, err := google.DefaultTokenSource(ctx, "https://www.googleapis.com/auth/cloud-platform")
```
averikitsch, line 249: We should note the specific permissions needed and how this is authenticated through environment credentials

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
```diff
@@ -371,5 +367,5 @@ func (t *Tool) McpManifest() tools.McpManifest {
 
 // Authorized checks if the tool is authorized.
 func (t *Tool) Authorized(verifiedAuthServices []string) bool {
```
averikitsch: We should move the authorized support if we are going to hard code this

averikitsch (commented):
Please also take a look at the pointer usage. We are using `&Tool{}` and the `(t *Tool)` were we can simplify to not use the pointers.

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
prernakakkar-google: This is just used for testing and we would not want it to be set any other way, hence not making part of reference.

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
prernakakkar-google: Required by tool config: "github.com/googleapis/genai-toolbox/internal/tools". Also description plays important role for agent to take decision.

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
prernakakkar-google: Since the tool implements tools.Tool this function is required.

internal/tools/utility/alloydbwaitforoperation/alloydbwaitforoperation.go
averikitsch: not-blocking: We are forcing the user to also copy or provide a description - why not hard code a good one?

docs/en/resources/tools/utility/alloydbwaitforoperation.md
averikitsch, line 1: Should we call out this tool has no source and that it uses the environment credentials to authenticate?

PR #1196, opened 2025-08-20
test: added tests for python quickstart

rapid-killer-9:
Added quickstart_test.py files for each Python sample, which compile and run the agent as a standalone application to validate its end-to-end functionality.
The test condition ensures the sample runs to completion and produces an output which confirms the agent is not breaking. Additionally, i introduced a secondary check for essential keywords from a golden.txt file, logging their presence without failing the test.

Running test file:
execute this cmd from terminal
```
ORCH_NAME=adk pytest
```


docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,76 @@
+import os
+import subprocess
+import sys
+import unittest
+from pathlib import Path
+
+
+class TestADKQuickstart(unittest.TestCase):
```
twishabansal: Let's use pytest in line with our existing python infrastructure. Please refer to the python sdk repo for more information.

docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,76 @@
...
+                    if line.strip():
+                        print(f"        {line}")
+            else:
+                print("        (No output)")
+            
+            # Check return code
+            if result.returncode != 0:
+                self.fail(f"Script execution failed with return code {result.returncode}")
+            else:
+                print("    quickstart_test.py:32: ✅ PASSED: Script ran successfully and produced output.")
```
twishabansal: nit: Can you fix this print statement?

docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,76 @@
...
+            golden_file_path = Path("../../golden.txt")
+            
+            if golden_file_path.exists():
+                try:
+                    with open(golden_file_path, 'r') as f:
+                        keywords = [line.strip() for line in f.readlines() if line.strip()]
+                    
+                    for keyword in keywords:
+                        if keyword in actual_output:
+                            print(f"    quickstart_test.py:41: ✅ INFO: Found keyword '{keyword}' in output.")
```
twishabansal: nit: Can you fix all the print statements

docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,76 @@
...
+                        else:
+                            print(f"    quickstart_test.py:43: ⚠️ INFO: Did not find keyword '{keyword}' in output.")
+                            
+                except Exception as e:
+                    print(f"    quickstart_test.py:45: ⚠️ WARNING: Could not read golden.txt: {e}")
```
twishabansal: We should use assert statements for test fails instead of print statements in tests

twishabansal:
Thanks for raising this. Let's add the cloud build trigger to this PR itself so that it shows things running e2e.

Let's make sure it includes

1. Lint check for the file
2. e2e test runs

You can refer to the existing test setup if you have any doubts

docs/en/getting-started/quickstart/python/core/quickstart_test.py
```diff
@@ -0,0 +1,129 @@
...
+        pytest.fail(f"Golden file not found: {golden_file_path}")
+    try:
+        with open(golden_file_path, 'r') as f:
+            return [line.strip() for line in f.readlines() if line.strip()]
+    except Exception as e:
+        pytest.fail(f"Could not read golden.txt: {e}")
+
+
+# --- Import Tests ---
+class TestCoreImports:
```
twishabansal: Why are we testing for imports here? Missing imports would anyway mean that the file fails at compile time

rapid-killer-9: We are checking for the imports are correct or not for the orch framework

We might have a case where
In folder of ADK  the code is using core imports instead of ADK Imports
so, for avoiding such error
Imports checks have been added

twishabansal: If I'm using ADK imports but using some other orch framework in my code, the code should not compile.

twishabansal: Whenever we're testing code, we usually test for the code behaviour. We don't want these tests because they're redundant as well as hard to maintain.

The code should perform checks like:

1. All code runs
2. No errors for responses

docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,133 @@
...
+                found_names = found_imports[expected_module]
+                for expected_name in expected_names:
+                    if expected_name not in found_names:
+                        missing_imports.append(f"'{expected_name}' not imported from '{expected_module}'")
+        
+        assert not missing_imports, f"Missing ADK imports: {missing_imports}"
+
+
+# --- Execution Tests ---
+class TestADKExecution:
```
twishabansal: Instead of script parsing, do you think running the tests for the function `run_application` might be a better idea?

You should be able to use pytest to directly test methods.
I think that would be a more maintainable approach. What do you think?

rapid-killer-9: Done

docs/en/getting-started/quickstart/python/core/quickstart_test.py
rapid-killer-9: Done

docs/en/getting-started/quickstart/python/langchain/quickstart_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+# --- Execution Tests ---
+class TestADKExecution:
+    """Test ADK framework execution and output validation."""
+
+    @pytest.fixture(autouse=True)
+    def check_prerequisites(self):
+        """Check for required environment variables."""
+        if not os.getenv("GOOGLE_API_KEY"):
```
twishabansal: Do we need tests for compile time errors?

rapid-killer-9: Yeah, 
It's not necessary 
I've removed this compile time test

docs/en/getting-started/quickstart/python/adk/quickstart.py
```diff
@@ -10,7 +10,7 @@
 
 # TODO(developer): replace this with your Google API key
 
-os.environ['GOOGLE_API_KEY'] = 'your-api-key'
```
twishabansal: This needs to be reverted?

docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_script_runs_without_errors(self, script_output):
+        """Test that the script runs and produces no stderr."""
+        assert script_output.err == "", f"Script produced stderr: {script_output.err}"
+
+    def test_script_produces_output(self, script_output):
+        """Test that the script produces some output to stdout."""
+        assert script_output.out.strip(), "Script produced no stdout"
+
+    def test_keywords_in_output(self, script_output, golden_keywords):
```
twishabansal: test_script_produces_output seems to be a subset of test_keywords_in_output

docs/en/getting-started/quickstart/python/adk/requirements.txt
```diff
@@ -0,0 +1,3 @@
+google-adk>=0.1.0
```
twishabansal: In requirements.txt files, we pin exact versions

docs/en/getting-started/quickstart/python/core/quickstart.py
```diff
@@ -29,7 +29,7 @@
     "My check in dates for my booking would be from April 10, 2024 to April 19, 2024.",
 ]
 
-async def run_application():
+async def main():
     async with ToolboxClient("http://127.0.0.1:5000") as toolbox_client:
```
twishabansal: Perhaps a rebase is required?

docs/en/getting-started/local_quickstart.md
```diff
@@ -26,6 +26,7 @@ This guide assumes you have already done the following:
 ## Step 2: Install and configure Toolbox
 {{< regionInclude "quickstart/shared/configure_toolbox.md" "configure_toolbox" >}}
 
+
```
twishabansal: Needs to be reverted?

docs/en/getting-started/quickstart/python/llamaindex/quickstart_test.py
```diff
@@ -0,0 +1,38 @@
...
+        pytest.fail(f"Golden file not found: {golden_file_path}")
+    try:
+        with open(golden_file_path, 'r') as f:
+            return [line.strip() for line in f.readlines() if line.strip()]
+    except Exception as e:
+        pytest.fail(f"Could not read golden.txt: {e}")
+
+
+# --- Execution Tests ---
+class TestADKExecution:
```
twishabansal: nit: Wrong name?

docs/en/getting-started/quickstart/python/langchain/quickstart_test.py
```diff
@@ -0,0 +1,38 @@
...
+        pytest.fail(f"Golden file not found: {golden_file_path}")
+    try:
+        with open(golden_file_path, 'r') as f:
+            return [line.strip() for line in f.readlines() if line.strip()]
+    except Exception as e:
+        pytest.fail(f"Could not read golden.txt: {e}")
+
+
+# --- Execution Tests ---
+class TestADKExecution:
```
twishabansal: Same here.

docs/en/getting-started/quickstart/python/core/quickstart_test.py
```diff
@@ -0,0 +1,38 @@
...
+        pytest.fail(f"Golden file not found: {golden_file_path}")
+    try:
+        with open(golden_file_path, 'r') as f:
+            return [line.strip() for line in f.readlines() if line.strip()]
+    except Exception as e:
+        pytest.fail(f"Could not read golden.txt: {e}")
+
+
+# --- Execution Tests ---
+class TestADKExecution:
```
twishabansal: nit: All these classes seem to have the same code. Is there a way to dedup this?

docs/en/getting-started/quickstart/python/adk/requirements.txt
```diff
@@ -0,0 +1,3 @@
+google-adk==0.1.0
```
twishabansal: Seems like a very old version. Can we make sure we use the latest possible versions of these frameworks?

rapid-killer-9: Done

docs/en/getting-started/quickstart/python/adk/quickstart_test.py
```diff
@@ -0,0 +1,38 @@
+import os
```
twishabansal: All the test files seem to have very similar code. Is there a way we can avoid that code duplication?

rapid-killer-9: Done

docs/en/getting-started/quickstart/python/quickstart_test.py
```diff
@@ -0,0 +1,48 @@
...
+            return [line.strip() for line in f.readlines() if line.strip()]
+    except Exception as e:
+        pytest.fail(f"Could not read golden.txt: {e}")
+
+
+# --- Execution Tests ---
+class TestExecution:
+    """Test framework execution and output validation."""
+
+
```
twishabansal:
```suggestion
```

docs/en/getting-started/quickstart/python/quickstart_test.py
```diff
@@ -0,0 +1,48 @@
+import os
+import pytest
+from pathlib import Path
+import asyncio
+import sys
+import importlib.util
+
+ORCH_NAME = os.environ.get("ORCH_NAME")
```
twishabansal, line 22: Please include in the PR description where this ORCH_NAME is coming from. Would be nice to show how the trigger/script sets the orch name.

twishabansal (changes requested):

docs/en/getting-started/quickstart/python/quickstart_test.py
```diff
@@ -0,0 +1,48 @@
...
+orch_dir = Path(__file__).parent / ORCH_NAME
```
twishabansal: Do you think it might be better to treat the samples as modules and then use module imports directly? Something like:
```python
ORCH_NAME = os.environ.get("ORCH_NAME")
module_path = f"your_package.{ORCH_NAME}.quickstart" 

try:
    quickstart = importlib.import_module(module_path)
except ModuleNotFoundError:
    ...
```

Our samples already have their own requirements files. Might make sense to add an `__init__` too.

docs/en/getting-started/quickstart/python/quickstart_test.py
rapid-killer-9: added in the description

docs/en/getting-started/quickstart/python/quickstart_test.py
rapid-killer-9: Done

docs/en/getting-started/quickstart/python/__init__.py
```diff
@@ -0,0 +1,4 @@
+# This file makes the 'adk' directory a Python package.
+
```
twishabansal, line 2: Is this file necessary for module type imports?

rapid-killer-9: yes, we needed that file for module type imports
And same pattern has been followed in python sdk  as well

docs/en/getting-started/quickstart/python/__init__.py
```diff
@@ -0,0 +1,4 @@
+# This file makes the 'adk' directory a Python package.
```
twishabansal:
```suggestion
# This file makes the 'quickstart' directory a Python package.
```

docs/en/getting-started/quickstart/python/__init__.py
twishabansal: In this PR, we are just importing other packages and not doing any exports. Can you explain why this is needed?

rapid-killer-9: The blank __init__.py file in the python folder is there to indicate that the directory is a Python package. While we are just importing other packages and not doing any exports, its presence ensures that Python recognizes the folder as a package. This is important for enabling imports and maintaining compatibility, even if the file itself is empty.

twishabansal: What imports does this file enable? Could you please elaborate?
And why do we need to see this folder as a python package?

docs/en/getting-started/quickstart/python/quickstart_test.py
```diff
@@ -0,0 +1,44 @@
+import os
```
twishabansal, line 15: Please add a license header

twishabansal (approved):

docs/en/getting-started/quickstart/python/__init__.py
rapid-killer-9: To run a single test file, we can use one of two methods:

1. Dynamic Path Modification
This was the previous approach. It involves programmatically altering Python's system path. so the test runner can locate the necessary modules at runtime.

2. Proper Packaging
This method treats the python directory as an installable package. It allows for clean, absolute imports in your tests without needing to manipulate sys.path. This is generally considered a more robust and modern solution.

PR #1208, opened 2025-08-22
docs(hugo): ignore quickstart folder from site indexing

rapid-killer-9:
Prevents Hugo from automatically indexing the quickstart folder in site navigation and search results.
This keeps documentation structure clean by excluding development/testing quickstart resources from public docs.

### DIFF
--- a/.hugo/hugo.toml
+++ b/.hugo/hugo.toml
@@ -8,6 +8,8 @@ defaultContentLanguageInSubdir = false
 enableGitInfo = true
 enableRobotsTXT = true
 
+ignoreFiles = ["quickstart/"]
+
 [languages]
   [languages.en]
     languageName ="English"


twishabansal:
Should we ignore quickstart/shared instead?

rapid-killer-9:
> Should we ignore quickstart/shared instead?
If we only ignore quick start/shared
Then same problem is occurring for quickstart/js, quickstart/go, quickstart/python

twishabansal:
> > Should we ignore quickstart/shared instead?
> > If we only ignore quick start/shared
> > Then same problem is occurring for quickstart/js, quickstart/go, quickstart/python

Maybe we can ignore those too? We might want to add a quickstart section in the docs and this would make it difficult to debug in the future

rapid-killer-9:
> > > Should we ignore quickstart/shared instead?
> > > If we only ignore quick start/shared
> > > Then same problem is occurring for quickstart/js, quickstart/go, quickstart/python
> 
> Maybe we can ignore those too? We might want to add a quickstart section in the docs and this would make it difficult to debug in the future

Done

PR #1229, opened 2025-08-25
fix(tools/firestore): add document/collection path validation

trehanshakuntG:


internal/tools/firestore/util/validator_test.go
```diff
@@ -0,0 +1,170 @@
...
+// limitations under the License.
+
+package util
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestValidateCollectionPath(t *testing.T) {
```
prernakakkar-google, line 22: Add test case for "users//user123", a trailing salsh "user/". Segments with only whitespace

prernakakkar-google:
Quick question: Did you try if just mentioning in description to use relative vs absolute helps.

internal/tools/firestore/util/validator_test.go
trehanshakuntG: Done.

trehanshakuntG:
> Quick question: Did you try if just mentioning in description to use relative vs absolute helps.

Do you mean in the description of the pre-built tool? Mentioning in the pre-built tool might help only for the firestore pre-built tool and not for someone who plans to use these tools individually.

prernakakkar-google (approved):

kurtisvg (approved):

PR #1235, opened 2025-08-25
feat(server): implement Tool call auth error propagation

duwenxin99:
For Toolbox protocol:
Before -  return 400 error for all tool invocation errors. 
After - Propagate auth-related errors (401 & 403) to the client if using client credentials. If using ADC, raise 500 error instead.

For MCP protocol:
Before -  return 200 with error message in the response body.
After - Propagate auth-related errors (401 & 403) to the client if using client credentials. If using ADC, raise 500 error instead.

### DIFF
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -152,25 +152,25 @@ steps:
           bigquery \
           bigquery
 
-  - id: "dataplex"
-    name: golang:1
-    waitFor: ["compile-test-binary"]
-    entrypoint: /bin/bash
-    env:
-      - "GOPATH=/gopath"
-      - "DATAPLEX_PROJECT=$PROJECT_ID"
-      - "SERVICE_ACCOUNT_EMAIL=$SERVICE_ACCOUNT_EMAIL"
-    secretEnv: ["CLIENT_ID"]
-    volumes:
-      - name: "go"
-        path: "/gopath"
-    args:
-      - -c
-      - |
-        .ci/test_with_coverage.sh \
-          "Dataplex" \
-          dataplex \
-          dataplex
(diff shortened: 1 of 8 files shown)


internal/server/api.go
```diff
@@ -222,7 +230,33 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+		if statusCode == http.StatusUnauthorized || statusCode == http.StatusForbidden {
+			if tool.RequiresClientAuthorization() {
+				// Propagate the original 401/403 error.
+				s.logger.DebugContext(ctx, fmt.Sprintf("error invoking tool. Client credentials lack authorization to the source: %v", err))
+				_ = render.Render(w, r, newErrResponse(err, statusCode))
+				return
+			}
+			// ADC lacking permission or credentials configuration error.
+			internalErr := fmt.Errorf("unexpected auth error occured during Tool invocation")
+			s.logger.ErrorContext(ctx, fmt.Sprintf("%s: %v", internalErr, err))
```
kurtisvg: If we are using %s, does internalErr need to be fmt.Errorf or fmt.Sprint?

(I think we should use fmt.Errorf with %w can wrap the causing error)

internal/server/mcp.go
```diff
@@ -405,16 +407,18 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	accessToken := tools.AccessToken(r.Header.Get("Authorization"))
 
 	v, res, err := processMcpMessage(ctx, body, s, protocolVersion, toolsetName, accessToken)
+
+	if err != nil {
+		s.logger.DebugContext(ctx, err.Error())
```
kurtisvg: nit: Does this logger need some more context around it?

PR #1222, opened 2025-08-22
feat(tools/looker): Control over whether hidden objects are surfaced

drstrangelooker:
The following config is added to the looker source:

* show_hidden_models
* show_hidden_explores
* show_hidden_fields

## Description
---
> Should include a concise description of the changes (bug or feature), it's
> impact, along with a summary of the solution

## PR Checklist
---
> Thank you for opening a Pull Request! Before submitting your PR, there are a
> few things you can do to make sure it goes smoothly:
- [ x] Make sure you reviewed
  [CONTRIBUTING.md](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md)
- [ x] Make sure to open an issue as a
  [bug/issue](https://github.com/googleapis/langchain-google-alloydb-pg-python/issues/new/choose)
  before writing your code!  That way we can discuss the change, evaluate
  designs, and agree on the general idea
- [ x] Ensure the tests and linter pass
- [ x] Code coverage does not decrease (if any source code was changed)
- [ x] Appropriate docs were updated (if necessary)
- [ x] Make sure to add `!` if this involve a breaking change

🛠️ Fixes #1214

PR #1248, opened 2025-08-26
feat: Add capability to set default for environment variable in config

### LINKED ISSUE #1001: Add a way for prebuilt configs and env variables in configs to have a default value
### Prerequisites

Today some of our prebuilt configs configure sources like this:
```
    mssql-source:
      kind: mssql
      host: ${MSSQL_HOST}
      port: ${MSSQL_PORT}
      database: ${MSSQL_DATABASE}
      user: ${MSSQL_USER}
      password: ${MSSQL_PASSWORD}
```

It would be nice if some of these optional parameters could be set with a reasonable default -- e.g. if port should default to 5432 if not set or if ipType = public. 

### Suggested Solution(s)

We can implement some kind of syntax for a fallback values -- e.g. ${MSSQL_PORT:5432}. 

### Alternatives Considered

_No response_

### Additional Details

_No response_

drstrangelooker:
Support `${ENV_VALUE:default_value}`

🛠️ Fixes #1001


drstrangelooker:
@Yuan325 This is the feature on its own.

drstrangelooker:
@Yuan325 I can't see why the dataplex integration test is failing here but not in https://github.com/googleapis/genai-toolbox/pull/1222 which is based on this pr.

drstrangelooker:
@Yuan325 this is ready to go now.

cmd/root.go
```diff
@@ -264,8 +264,9 @@ type ToolsFile struct {
 }
 
 // parseEnv replaces environment variables ${ENV_NAME} with their values.
+// also support ${ENV_NAME-default_value}.
```
```suggestion
// also support ${ENV_NAME:default_value}.
```

PR #1260, opened 2025-08-27
chore: update env var to allow empty string

Yuan325:
Update configuration file capability of reading from environment variable:
* allow default empty string
* fix if statement to check if `parts[2]` is empty instead since it will always have a length of 4.
* add unit tests.


drstrangelooker:
Thanks for this fix. In other environments the capture groups don't make an array element if they don't capture anything.

PR #1267, opened 2025-08-27
feat(tools/looker): Report field suggestions to agent

### LINKED ISSUE #1247: Add `suggestions` metadata to the `looker-get-dimensions` tool
### Prerequisites

Currently when giving prompts to the looker MCP, the query may fail to filter correctly as there is not enough context about what the available selectable options are for a filter. A query may produce no results if it filters incorrectly. The user would have to input the exact string text that would be required of the filter.

### Suggested Solution(s)

Enable the looker-get-dimensions tool to ingest the "suggestions" field from LookmlModelExploreField, so that we can add in the meta-data for what the selectable options could be when filtering.

### Alternatives Considered

Alternatively, when using the looker UI, users are able to click on a filter and it will look at the distinct values available for a field. It would be useful to be able to replicate it through the MCP.

### Additional Details

_No response_

drstrangelooker:
## Description
---
Report the LookML suggestions array or the suggst_explore and suggest_dimension to the caller so
that the caller can use those suggestions to improve filtering.

## PR Checklist
---
- [x] Make sure you reviewed
  [bug/issue](https://github.com/googleapis/langchain-google-alloydb-pg-python/issues/new/choose)

🛠️ Fixes #1247


internal/tools/looker/lookercommon/lookercommon.go
```diff
@@ -71,12 +71,21 @@ func ExtractLookerFieldProperties(ctx context.Context, fields *[]v4.LookmlModelE
...
 		}
-		if v.Tags != nil {
+		if v.Tags != nil && len(*v.Tags) > 0 {
 			vMap["tags"] = *v.Tags
 		}
-		if v.Synonyms != nil {
+		if v.Synonyms != nil && len(*v.Synonyms) > 0 {
 			vMap["synonyms"] = *v.Synonyms
 		}
+		if v.Suggestable != nil && *v.Suggestable {
```
Yuan325, line 127: can we add some tests to the `lookercommon_test.go` for these? thankyou!

Yuan325 (changes requested):

internal/tools/looker/lookercommon/lookercommon.go
drstrangelooker: done

drstrangelooker:
Can I get an LGTM?

Yuan325 (approved):
Thank you!

PR #1279, opened 2025-08-28
fix(bigquery): Move `useClientOAuth` config from tool to source

duwenxin99:


internal/sources/bigquery/bigquery.go
```diff
@@ -110,6 +126,10 @@ func (s *Source) BigQueryRestService() *bigqueryrestapi.Service {
 	return s.RestService
 }
 
+func (s *Source) BigQueryUseClientOAuth() bool {
```
kurtisvg: does this need to be "BigQuery..."?

duwenxin99: changed to `UseClientAuthorization` to match the tool method

duwenxin99:
> Is this a breaking change?

yes it is

PR #1283, opened 2025-08-29
fix(tools/oceanbase): fix encoded text with mysql driver

Yuan325:
Update the type conversion to according to the types that mysql driver returns: https://github.com/go-sql-driver/mysql/blob/v1.9.3/fields.go

All `scanTypeBytes` or `scanTypeString` or `scanTypeNullString` will now be converted into string. Remaining types (numeric types or `scanTypeUnknown`) are all returned as is.

Separately handle "JSON" to prevent double marshaling.

This update is needed for oceanbase since it uses mysql driver for database connection.

ref: #1161


Yuan325:
@xxsc0529 fyi on this change :)

duwenxin99 (approved):

PR #1285, opened 2025-08-29
fix: update mcp implementation comments

twishabansal:


internal/server/mcp.go
```diff
@@ -355,8 +355,10 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check if client have `Mcp-Session-Id` header
-	// if `Mcp-Session-Id` header is set, we are using v2025-03-26 since
-	// previous version doesn't use this header.
+	// if `Mcp-Session-Id` header is set, we are using v2025-03-26 or later
+	// since previous version doesn't use this header.
```
Yuan325: Only `v20250326` will be using this header since we only send this header to that specific version.

internal/server/mcp.go
```diff
@@ -355,8 +355,10 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check if client have `Mcp-Session-Id` header
-	// if `Mcp-Session-Id` header is set, we are using v2025-03-26 since
-	// previous version doesn't use this header.
+	// if `Mcp-Session-Id` header is set, we are using v2025-03-26 or later
+	// since previous version doesn't use this header. 
+	// Any versions later than v2025-03-26 would have the `MCP-Protocol-Version` 
+	// header. So the protocol version will be overridden in the subsequent code block.
```
Yuan325: We can add this comment in the part where it is looking for the `MCP-Protocol-Version` header. and mention something like "this header is only supported at v2025-06-18+" instead.

internal/server/mcp.go
twishabansal: Changed.

internal/server/mcp.go
twishabansal: Thanks for the review. Done!

Issue #1250, opened 2025-08-26
Dataplex search entry and search aspect integration tests failure


Yuan325:
### Prerequisites

- [x] I've searched the current open issues
- [ ] I've updated to the latest version of Toolbox

### Toolbox version

n/a

### Expected Behavior

Integration tests are passing up till this morning.

### Current Behavior

Tests started to fail for Search Entry and Search Aspect. Error: `"expected exactly one entry, but got 0"`

### Steps to reproduce?
Run integration tests for dataplex.

### Additional Details

Will be disabling dataplex integration tests for now.

PR #1289, opened 2025-08-29
ci: fix dataplex integration tests

estshy:
Restoring Dataplex integration after addressing test failure.

🛠️ Ref https://github.com/googleapis/genai-toolbox/issues/1250


Yuan325 (approved):
Thank you! Was there a recent update that changed the return format of Dataplex?

estshy:
> Thank you! Was there a recent update that changed the return format of Dataplex?

Yes, that would be my best guess. The previous code seems to be invalid for me, therefore if it worked for some time then it wasn't intentional.

kurtisvg (approved):

PR #1299, opened 2025-08-31
feat(source/mysql): support `queryParams` in MySQL source

### LINKED ISSUE #1286: mysql/postgresql ssl-mode
### Prerequisites

Hello, could you advise how I can pass the ssl_mode through this configuration? If not through this one, then how?
https://googleapis.github.io/genai-toolbox/how-to/connect-ide/mysql_mcp/

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

govegasbudapest:
Fixes #1286

### Motivation
* Allow secure connections to mysql without custom code.

#### Sources
* `mysql`: `Config.QueryParams` added; DSN building rewritten via `url.Values`.


internal/prebuiltconfigs/tools/mysql.yaml
Yuan325, line 1: Can we separate the changes in prebuiltconfig to a separate PRs? thank you!

internal/sources/postgres/postgres.go
Yuan325, line 1: Existing users can easily use `queryParam` to add `ssl_mode`. The reason we did this is so that user can have more control over what parameter that they want to add (e.g. user can add just the `ssl_mode` or adding `ssl_mode` + `sslkey` etc.). This will also help keep things simpler to prevent adding more shortcuts in the future.

If the usage is not clear, we can look into improving the documentation to add examples for `queryParams`. What do you think about this?

internal/sources/mysql/mysql.go
Yuan325, line 1: Please add a new test case in `mysql_test.go` with query params with this new addition.

Yuan325 (changes requested):
Hi @govegasbudapest thankyou for contributing! If you can update the docs that will be great too! ref: https://github.com/googleapis/genai-toolbox/pull/1047

internal/sources/postgres/postgres.go
govegasbudapest: Thank you for the clarification. You are absolutely right.

I agree completely with this approach. I will remove the changes I made, as I overlooked the existing queryParam parameter.

internal/sources/mysql/mysql.go
govegasbudapest: I've added the tests, thanks for mentioning it.

internal/sources/mysql/mysql.go
```diff
@@ -54,14 +55,18 @@ type Config struct {
 	Password     string `yaml:"password" validate:"required"`
 	Database     string `yaml:"database" validate:"required"`
 	QueryTimeout string `yaml:"queryTimeout"`
+	// QueryParams passes arbitrary key/value pairs to the driver DSN
+	// (e.g. tls=preferred, tls=skip-verify), enabling TLS configuration
+	// without changing the code.
```
Yuan325:
```suggestion
```
Since this was added in docs, I think we can remove it from here.

internal/sources/mysql/mysql_test.go
```diff
@@ -80,9 +85,68 @@ func TestParseFromYamlCloudSQLMySQL(t *testing.T) {
...
+					Kind:        mysql.SourceKind,
+					Host:        "0.0.0.0",
+					Port:        "my-port",
+					Database:    "my_db",
+					User:        "my_user",
+					Password:    "my_pass",
+					QueryParams: map[string]string{},
+				},
+			},
+		},
```
Yuan325:
```suggestion
```

This is not necessary since it will be tested in the other test cases that does not have `queryParams` defined. :)

internal/sources/mysql/mysql_test.go
```diff
@@ -80,9 +85,68 @@ func TestParseFromYamlCloudSQLMySQL(t *testing.T) {
...
 	}
 	for _, tc := range tcs {
+		tc := tc
```
Yuan325:
```suggestion
```

internal/sources/mysql/mysql_test.go
```diff
@@ -101,9 +165,10 @@ func TestParseFromYamlCloudSQLMySQL(t *testing.T) {
 
 func TestFailParseFromYaml(t *testing.T) {
 	tcs := []struct {
-		desc string
-		in   string
-		err  string
+		desc     string
+		in       string
+		err      string
+		contains bool
```
Yuan325: Do we need this? It should be working without the `contains` field.

internal/sources/mysql/mysql_test.go
```diff
@@ -131,11 +197,30 @@ func TestFailParseFromYaml(t *testing.T) {
...
+				my-mysql-instance:
+					kind: mysql
+					host: 0.0.0.0
+					port: 3306
+					database: my_db
+					user: my_user
+					password: my_pass
+					queryParams: not-a-map
+			`,
+			err:      "cannot decode",
```
Yuan325: This test case should trigger the unmarshal error in golang, we can copy the whole error string here too. Thank you for adding this :)

internal/sources/mysql/mysql_test.go
```diff
@@ -131,11 +197,30 @@ func TestFailParseFromYaml(t *testing.T) {
...
+			contains: true,
 		},
 	}
 	for _, tc := range tcs {
+		tc := tc
```
Yuan325:
```suggestion
```

internal/sources/mysql/mysql_test.go
```diff
@@ -145,9 +230,38 @@ func TestFailParseFromYaml(t *testing.T) {
...
+				if errStr != tc.err {
+					t.Fatalf("unexpected error: got %q, want %q", errStr, tc.err)
+				}
 			}
 		})
 	}
 }
+
+// TestInitializeInvalidQueryTimeout verifies that an invalid QueryTimeout string is rejected during initialization without attempting a DB connection.
+func TestInitializeInvalidQueryTimeout(t *testing.T) {
```
Yuan325:
```suggestion
// TestFailInitialization test error during initialization without attempting a DB connection.
func TestFailInitialization(t *testing.T) {
```

Let's provide a more generic name so this could be extended in the future.

internal/sources/mysql/mysql_test.go
```diff
@@ -145,9 +230,38 @@ func TestFailParseFromYaml(t *testing.T) {
...
+	t.Parallel()
+
+	cfg := mysql.Config{
+		Name:         "instance",
+		Kind:         mysql.SourceKind,
```
Yuan325:
```suggestion
		Kind:         "mysql",
```

internal/prebuiltconfigs/tools/mysql.yaml
Yuan325: reminder: please move prebuiltconfig updates to a separate PR, thank you! 
Also instead of having `tls: ${MYSQL_TLS}`, we might just want to check if we can do `queryParams: {$QUERYPARAM}` and pass the map via env var instead. If `MYSQL_TLS` is not provided, we would want `queryParams` field to remain empty/nil.

Yuan325 (changes requested):
Thank you for your prompt response! Left a couple more comments on the changes.

internal/prebuiltconfigs/tools/mysql.yaml
govegasbudapest: Corrected it. I had it in my head for some reason that this was about a situation where Postgres is present.
https://github.com/googleapis/genai-toolbox/pull/1318

internal/sources/mysql/mysql_test.go
govegasbudapest: Fixed

internal/sources/mysql/mysql_test.go
govegasbudapest: Fixed

PR #1312, opened 2025-09-02
fix(tools/bigquerysql)!: remove `useClientOAuth` from tools config

Yuan325:
## Description
---
The `useClientOAuth` config was moved from the tool into `bigquery` source in #1279, however the config was not removed from the source. The value is now retrieved from the source instead of setting it directly.

### DIFF
--- a/internal/tools/bigquery/bigquerysql/bigquerysql.go
+++ b/internal/tools/bigquery/bigquerysql/bigquerysql.go
@@ -127,17 +127,17 @@ type Tool struct {
 	Name               string           `yaml:"name"`
 	Kind               string           `yaml:"kind"`
 	AuthRequired       []string         `yaml:"authRequired"`
+	UseClientOAuth     bool             `yaml:"useClientOAuth"`
 	Parameters         tools.Parameters `yaml:"parameters"`
 	TemplateParameters tools.Parameters `yaml:"templateParameters"`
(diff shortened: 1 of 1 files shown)


internal/tools/bigquery/bigquerysql/bigquerysql.go
```diff
@@ -128,17 +127,17 @@ type Tool struct {
 	Name               string           `yaml:"name"`
 	Kind               string           `yaml:"kind"`
 	AuthRequired       []string         `yaml:"authRequired"`
-	UseClientOAuth     bool             `yaml:"useClientOAuth"`
```
duwenxin99: We want to preserve this in the Tool struct as it inherits from the source

duwenxin99 (dismissed):

internal/tools/bigquery/bigquerysql/bigquerysql.go
Yuan325: it's not removed~ It's moved to the bottom of the struct where those values are set from the source. and removed the `yaml:"useClientOAuth"` since its not defined in the tools anymore.

Yuan325: Re-added the `yaml` since this is required for the manifest.

PR #1326, opened 2025-09-03
ci: Auto-remove the docs preview deploy label

duwenxin99:
Automatically removes the label once the workflow triggered.


averikitsch:
Unfortunately, this goes against current recommendations for pull_request_target. We could use it with pull_request.

PR #1354, opened 2025-09-06
chore: update unmarshal function for ToolsFile

Yuan325:
Update `parseToolsFile()` to use the file conversion function. Update unmarshaling function.

This PR will not pass since the updates for sources/auth/tools will be in subsequent PRs. Breaking it down to keep review simpler.


Yuan325:
Closing due to priority shift. Will resume in the future.

PR #1363, opened 2025-09-08
chore(testing): testcases for JS

AnmolShukla2002:


docs/en/getting-started/quickstart/js/quickstart.test.js
```diff
@@ -0,0 +1,70 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { describe, test, before, after } from "node:test";
+import assert from "node:assert/strict";
+import fs from "fs";
+import path from "path";
+import { fileURLToPath } from "url";
+
+const ORCH_NAME = process.env.ORCH_NAME;
```
twishabansal, line 21: Same as https://github.com/googleapis/genai-toolbox/pull/1196/files#r2335861165

docs/en/getting-started/quickstart/js/quickstart.test.js
```diff
@@ -0,0 +1,70 @@
...
+    console.log = (msg) => {
+      capturedOutput.push(msg);
+    };
+  });
+
+  after(() => {
+    console.log = originalLog;
+  });
+
+  test("Assert", async () => {
```
twishabansal: nit: please rename test

AnmolShukla2002: done!

docs/en/getting-started/quickstart/js/quickstart.test.js
AnmolShukla2002: But Here the sample code file is being treated as a module only and the test file imports its main function directly. In package.json we have set type: module so it helps to import .JS file as a module that can be imported directly by its path.

```
const { main: runAgent } = await import(quickstartPath);
```

twishabansal: Thanks!

twishabansal: If we're installing them as modules. We shouldn't need the directory paths I suppose?

PR #1364, opened 2025-09-08
chore: modified quickstart.py files for support of quickstart sample testing

rapid-killer-9:


twishabansal:
Is this PR needed now?

docs/en/getting-started/quickstart/python/adk/quickstart.py
```diff
@@ -35,7 +35,7 @@ async def main():
 
       session_service = InMemorySessionService()
       artifacts_service = InMemoryArtifactService()
-      session = await session_service.create_session(
+      session = session_service.create_session(
```
twishabansal: This would not work for the new adk versions

docs/en/getting-started/quickstart/python/adk/quickstart.py
```diff
@@ -8,9 +8,9 @@
 import asyncio
 import os
 
-# TODO(developer): replace this with your Google API key
+# TODO(developer): export your google api key in terminal
```
twishabansal: Perhaps we can ask them to add their api key in the file

docs/en/getting-started/quickstart/python/adk/quickstart.py
```diff
@@ -8,9 +8,9 @@
...
 
-os.environ['GOOGLE_API_KEY'] = 'your-api-key'
+os.environ['GOOGLE_API_KEY']
```
twishabansal: Does this work?

docs/en/getting-started/quickstart/python/core/quickstart.py
```diff
@@ -39,7 +40,7 @@ async def main():
         # provided wrapper packages, which handle framework-specific boilerplate.
         toolbox_tools = await toolbox_client.load_toolset("my-toolset")
         genai_client = genai.Client(
-            vertexai=True, project="project-id", location="us-central1"
+            vertexai=True, project=os.environ.get("GCP_PROJECT"), location="us-central1"
```
twishabansal: Let's make it a global var like the api key?

docs/en/getting-started/quickstart/python/core/quickstart.py
```diff
@@ -68,41 +69,48 @@ async def main():
...
-                    function_result = await toolbox_tools[1](**function_call.args)
-                else:
-                    raise ValueError("Function name not present.")
-                function_response = {"result": function_result}
-                function_response_part = Part.from_function_response(
-                    name=function_call.name,
-                    response=function_response,
-                )
-                function_response_parts.append(function_response_part)
+
```
twishabansal, line 74: Do we need this change?

docs/en/getting-started/quickstart/python/core/quickstart.py
```diff
@@ -68,41 +69,48 @@ async def main():
...
-                model="gemini-2.0-flash-001",
-                contents=history,
-                config=GenerateContentConfig(
-                    tools=genai_tools,
-                ),
-            )
-            final_model_response_content = response2.candidates[0].content
-            history.append(final_model_response_content)
-            print(response2.text)
+                response2 = genai_client.models.generate_content(
```
twishabansal, line 103: Do we need this change?

docs/en/getting-started/quickstart/python/core/quickstart.py
```diff
@@ -68,41 +69,48 @@ async def main():
...
+                    ),
+                )
+                final_model_response_content = response2.candidates[0].content
+                history.append(final_model_response_content)
+                print(response2.text)
+            else:
+                print(response.text)
 
-asyncio.run(main())
+if __name__ == "__main__":
```
twishabansal: Do we need this change?

docs/en/getting-started/quickstart/python/llamaindex/quickstart.py
```diff
@@ -34,7 +34,7 @@ async def main():
     # TODO(developer): replace this with another model if needed
     llm = GoogleGenAI(
         model="gemini-2.0-flash-001",
-        vertexai_config={"project": "project-id", "location": "us-central1"},
+        vertexai_config={"project": os.environ.get("GCP_PROJECT", "project-id"), "location": "us-central1"},
```
twishabansal: Same as https://github.com/googleapis/genai-toolbox/pull/1364/files#r2336025415

docs/en/getting-started/quickstart/python/core/quickstart.py
rapid-killer-9: https://github.com/googleapis/genai-toolbox/pull/1385/files#r2335984283

docs/en/getting-started/quickstart/python/core/quickstart.py
rapid-killer-9: https://github.com/googleapis/genai-toolbox/pull/1385/files#r2335984283

docs/en/getting-started/quickstart/python/adk/quickstart.py
rapid-killer-9: yes

docs/en/getting-started/quickstart/python/llamaindex/quickstart.py
```diff
@@ -13,6 +13,8 @@
 
 from toolbox_llamaindex import ToolboxClient
 
+project = os.environ.get("GCP_PROJECT", "project-id")
```
twishabansal, line 16: Please see how the js quickstart does it. We should allow users to set their project in the file.

docs/en/getting-started/quickstart/python/adk/quickstart.py
twishabansal: Let's allow the users to set the api key in the file (See [JS quickstart](https://googleapis.github.io/genai-toolbox/getting-started/local_quickstart_js/))

docs/en/getting-started/quickstart/python/core/quickstart.py
twishabansal: Was the original code not working?

rapid-killer-9: no, during test its was failing

docs/en/getting-started/quickstart/python/adk/quickstart.py
```diff
@@ -10,7 +10,7 @@
 
 # TODO(developer): replace this with your Google API key
 
-os.environ['GOOGLE_API_KEY'] = 'your-api-key'
+api_key = os.environ.get("GOOGLE_API_KEY") or "your-api-key"
```
twishabansal:
```suggestion
api_key = os.environ.get("GOOGLE_API_KEY") or "your-api-key" # Set your API key here
```

docs/en/getting-started/quickstart/python/core/quickstart.py
```diff
@@ -68,41 +71,46 @@ async def main():
...
-                else:
-                    raise ValueError("Function name not present.")
-                function_response = {"result": function_result}
-                function_response_part = Part.from_function_response(
-                    name=function_call.name,
-                    response=function_response,
-                )
-                function_response_parts.append(function_response_part)
+
+            if response.function_calls:
```
twishabansal, line 75: Any reason for this if check?

rapid-killer-9: Actually,
With the original code. when we try to run the quickstart sample we are getting the output.
but during the testcase we try its fails for the output produce. coz, sometimes llm directly response without calling the tools.
and only tools response have been considered by the original code. so, need to modify with this check to allow llm and tools both response

twishabansal: Do you mean that the llm is giving the response without actually calling the tools? That should mean test failure since it wouldn't be able to look into the database and search for the hotels.

rapid-killer-9: our original code are failing that why we need this change

PR #1366, opened 2025-09-08
tests: added Go quickstart tests

AnmolShukla2002:


docs/en/getting-started/quickstart/go/genkit/quickstart_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestAgentOutputAndKeywords(t *testing.T) {
```
twishabansal: All quickstart tests seem quite similar. Can we avoid this code duplication?

docs/en/getting-started/quickstart/go/genkit/quickstart_test.go
```diff
@@ -0,0 +1,60 @@
...
+	if os.Getenv("GOOGLE_API_KEY") == "" {
+		t.Skip("Skipping integration test: GOOGLE_API_KEY environment variable is not set.")
+	}
+
+	buildCmd := exec.Command("go", "build", "-o", "quickstart_test_binary", ".")
```
twishabansal: Is direct testing without compilation not possible in go?

docs/en/getting-started/quickstart/go/genkit/quickstart_test.go
AnmolShukla2002: Yes, sure I will push the changes.

docs/en/getting-started/quickstart/go/genkit/quickstart_test.go
AnmolShukla2002: No....since go is a compiled language so all the tests must be compiled to create a kind of executable file.

docs/en/getting-started/quickstart/go/quickstart_test.go
```diff
@@ -0,0 +1,81 @@
...
+			t.Skipf("Skipping test for %s: GOOGLE_API_KEY environment variable is not set.", framework)
+		}
+	}
+
+	sampleDir := filepath.Join(".", framework)
+	if _, err := os.Stat(sampleDir); os.IsNotExist(err) {
+		t.Fatalf("Test setup failed: directory for framework '%s' not found.", framework)
+	}
+
+	cmd := exec.Command("go", "run", ".")
```
dishaprakash, line 46: Is there a reason why we are not using the `go test ./...` command?

AnmolShukla2002: When I tried replacing go run . with go test ./... inside the command, the test failed...... that’s because go test generates the output of the test runner (like "PASS" or "FAIL"), not the actual output of my application, so the keyword check was failing which we do not want. We want to take the user's output to stdout and then verify against out goldent.txt file.

dishaprakash: Got it! Sounds good!

dishaprakash (approved):

PR #1381, opened 2025-09-09
Fix(tools/bigquery-conversational-analytics): fix authentication scope error in Cloud Run

### LINKED ISSUE #1378: GitHub Issue: BigQuery Conversational Analytics Authentication Scope Error
### Prerequisites

### Toolbox version

0.14.0

## Environment

- **✅ Working**: Local development → MCP Toolbox
- **❌ Failing**: Cloud Run → MCP Toolbox
- **✅ Working**: All other BigQuery tools in both environments
- **❌ Failing**: Only `bigquery-conversational-analytics` tool

Linux johanesa45 6.12.32-1rodete1-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.32-1rodete1 (2025-06-26) x86_64 GNU/Linux

## Client

Via [ADK](https://github.com/johanesalxd/bq-agent-app/blob/main/bq_multi_agent_app/tools.py) deployed on Cloud Run.

## Expected Behavioud

- Working on all env.

## Problem

The `bigquery-conversational-analytics` tool fails with `ACCESS_TOKEN_SCOPE_INSUFFICIENT` error in Cloud Run environments, while all other BigQuery tools work fine.

## Error Details

```json
{
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "status": "PERMISSION_DENIED",
    "details": [{
      "@type": "type.googleapis.com/google.rpc.ErrorInfo",
      "reason": "ACCESS_TOKEN_SCOPE_INSUFFICIENT",
      "domain": "googleapis.com",
      "metadata": {
        "method": "google.cloud.geminidataanalytics.v1alpha.DataChatService.Chat",
        "service": "geminidataanalytics.googleapis.com"
      }
    }]
  }
}
```

## Root Cause

The conversational analytics tool uses `TokenSource` with limited BigQuery scope (`bigqueryapi.Scope`), but the Gemini Data Analytics API requires broader `cloud-platform` scope.

**Key difference from working tools:**
- Other BigQuery tools use `ClientCreator` pattern for flexible scope handling
- Conversational analytics tool makes direct HTTP calls to `geminidataanalytics.googleapis.com` with insufficient scopes

## Reproduction

Use this test script to reproduce the error:

```python
#!/usr/bin/env python3
import os
import json
import requests
import google.auth
from google.auth.transport.requests import Request

def test_conversational_analytics():
    # Get cloud-platform scoped token
    credentials, _ = google.auth.default(scopes=["https://www.googleapis.com/auth/cloud-platform"])
    request = Request()
    credentials.refresh(request)

    headers = {
        "Content-Type": "application/json",
        "Authorization": f"Bearer {credentials.token}"
    }

    payload = {
        "jsonrpc": "2.0",
        "id": 1,
        "method": "tools/call",
        "params": {
            "name": "bigquery-conversational-analytics",
            "arguments": {
                "user_query_with_context": "What is the structure of this table?",
                "table_references": json.dumps([{
                    "projectId": "your-project",
                    "datasetId": "samples",
                    "tableId": "shakespeare"
                }])
            }
        }
    }

    response = requests.post("YOUR_TOOLBOX_URL/mcp", headers=headers, json=payload)
    print(f"Status: {response.status_code}")
    print(response.text)

if __name__ == "__main__":
    test_conversational_analytics()
```

## Files Affected

- `internal/tools/bigquery/bigqueryconversationalanalytics/bigqueryconversationalanalytics.go`

## Impact

This issue prevents the conversational analytics tool from working in production Cloud Run environments where other BigQuery tools work perfectly. The tool is the only BigQuery tool that exhibits this authentication scope limitation.

johanesalxd:
## Problem
Fixes #1378 - BigQuery Conversational Analytics tool fails with `ACCESS_TOKEN_SCOPE_INSUFFICIENT` error in Cloud Run environments while other BigQuery tools work fine.

## Root Cause
The conversational analytics tool was using `BigQueryTokenSource()` with limited `bigqueryapi.Scope`, but the Gemini Data Analytics API (`geminidataanalytics.googleapis.com`) requires broader `cloud-platform` scope.

## Solution
- Replace ADC token acquisition to use `google.DefaultTokenSource` with `cloud-platform` scope
- Remove dependency on limited BigQuery scope from source
- Maintain compatibility with existing OAuth flow

## Testing
- ✅ **Local testing confirmed**: Tool now works perfectly
- ✅ **Test script**: https://github.com/johanesalxd/bq-agent-app/blob/main/setup/mcp_toolbox_ca_issue/test_detailed_error.py
- ✅ **Successful response**: Returns proper schema information and conversational answers
- ✅ **All BigQuery tool tests pass**: No regression in other tools
- ✅ **Build successful**: No compilation errors

## Impact
- **Fixes**: Cloud Run deployment authentication errors
- **Maintains**: Local development functionality  
- **Preserves**: All existing BigQuery tool functionality


duwenxin99:
Hi @Genesis929, could you take a look at this PR? Thanks!

Genesis929 (approved):

duwenxin99:
Since we are removing the `TokenSource` from the BQ source, let's also remove it from the init function here: https://github.com/googleapis/genai-toolbox/blob/aa3972470fd0f6f5901c5d85dd05f1e2ae973e7b/internal/sources/bigquery/bigquery.go#L226

internal/tools/bigquery/bigqueryconversationalanalytics/bigqueryconversationalanalytics.go
```diff
@@ -196,11 +193,12 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken
 			return nil, fmt.Errorf("error parsing access token: %w", err)
 		}
 	} else {
-		// Use ADC
-		if t.TokenSource == nil {
-			return nil, fmt.Errorf("ADC is missing a valid token source")
+		// Use ADC with cloud-platform scope for Gemini Data Analytics API
+		tokenSource, err := google.DefaultTokenSource(ctx, "https://www.googleapis.com/auth/cloud-platform")
```
Genesis929: Maybe set the scope/set a different tokensource in https://github.com/googleapis/genai-toolbox/blob/main/internal/sources/bigquery/bigquery.go#L216 instead of adding a new tokensource here.

johanesalxd: hi @Genesis929 , wdyt if i add a new method instead? i think we can keep the original one with more narrow scope (bigquery) and use the new method if we need wider one: https://github.com/googleapis/genai-toolbox/pull/1381/files#diff-ebabb25527e9628a15f43d8d77a1d9bf2a3cfa4c93481e8023275fa4d224242fR193

Genesis929: Hi @johanesalxd. Thanks for replying, yes this is a great approach! 

I have one minor point: Instead of passing the entire `BigQuerySource` object, we should pass only the specific attribute or function from within `BigQuerySource`. You can follow the pattern of `BigQueryClientCreator` instead of passing the `BigQuerySource`. Example links: https://github.com/googleapis/genai-toolbox/blob/main/internal/sources/bigquery/bigquery.go#L294, https://github.com/googleapis/genai-toolbox/blob/main/internal/tools/bigquery/bigquerygettableinfo/bigquerygettableinfo.go#L105

Thanks again for driving this improvement!

Genesis929: An alternative approach could be to simply obtain this specific TokenSource directly within the Initialize function instead of Invoke function，as we don't need to create a new token source every tool call.

johanesalxd: > Hi @johanesalxd. Thanks for replying, yes this is a great approach!
> 
> I have one minor point: Instead of passing the entire `BigQuerySource` object, we should pass only the specific attribute or function from within `BigQuerySource`. You can follow the pattern of `BigQueryClientCreator` instead of passing the `BigQuerySource`. Example links: https://github.com/googleapis/genai-toolbox/blob/main/internal/sources/bigquery/bigquery.go#L294, https://github.com/googleapis/genai-toolbox/blob/main/internal/tools/bigquery/bigquerygettableinfo/bigquerygettableinfo.go#L105
> 
> Thanks again for driving this improvement!

Noted on this, let me take a look!

> An alternative approach could be to simply obtain this specific TokenSource directly within the Initialize function instead of Invoke function.

Yes, I'm debating this myself as well. I previously thought to just replace the _bigquery scope_ with _cloud-platform scope_ as per your suggestion, but i wonder if its a good practice or not. I don't think it will break anything though, since wider scope doesn't do anything if the service account itself doesn't have the permission (afaik). But i think maybe having a separation between CA tool which need the wider scope (and only get called during _invoke_) and the rest of tools which rely on _bigquery scope_ is a better approach (hence need to put it on _invoke_ and not _init_ func)? wdyt @Genesis929 ?

(I'm thinking the current error that I got is a good one since it only failed on the CA tool and not the rest due to more narrow _bigquery scope_, where the CA tool itself need the wider one, probably due to API call)

Genesis929: Sorry for the confusion, I also think it's better to separate the scope.  By "obtain this specific TokenSource directly within the Initialize " I mean in the Initialize function of the tool.  So the scope still tool-specific, but you don't need to create a token source every time, and you don't need to make change in Invoke and Tool config.

Something like:

```
t := Tool{
		Name:               cfg.Name,
		Kind:               kind,
		Project:            s.BigQueryProject(),
		Location:           s.BigQueryLocation(),
		Parameters:         parameters,
		AuthRequired:       cfg.AuthRequired,
		Client:             s.BigQueryClient(),
		UseClientOAuth:     s.UseClientAuthorization(),

		TokenSource:        s.BigQueryTokenSourceWithScope(xxx), # Or just s.BigQueryCloudPlatformTokenSource() since this is the only tool that uses access token.

		manifest:           tools.Manifest{Description: cfg.Description, Parameters: parameters.Manifest(), AuthRequired: cfg.AuthRequired},
		mcpManifest:        mcpManifest,
		MaxQueryResultRows: s.GetMaxQueryResultRows(),
	}
```

johanesalxd: ah i see noted. i've updated the code and yes its much cleaner now. and also its getting the token during init and not for every invocation anymore. thanks @Genesis929

johanesalxd:
thanks @Genesis929 and @duwenxin99 . just want to check if anything else needed from my side?

Genesis929:
> thanks @Genesis929 and @duwenxin99 . just want to check if anything else needed from my side?

Thanks for your great work! We will merge it soon.

PR #1386, opened 2025-09-09
fix(bigquery)!: Add `Bearer` parsing to auth token

duwenxin99:
Previously we propagate tokens directly to the BQ API. But MCP inspector adds a "Bearer" prefix to all authorization header. We will need to parse the token accordingly to make it work.

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 
 	yaml "github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/sources"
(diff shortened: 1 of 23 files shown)


internal/tools/tools.go
```diff
@@ -66,6 +67,14 @@ type ToolConfig interface {
 
 type AccessToken string
 
+func (token AccessToken) ParseBearerToken() (string, error) {
+	headerParts := strings.Split(string(token), " ")
+	if len(headerParts) != 2 || strings.ToLower(headerParts[0]) != "bearer" {
+		return "", fmt.Errorf("authorization header must be in the format 'Bearer <token>: %w", ErrUnauthorized)
```
kurtisvg:
```suggestion
		return "", fmt.Errorf("authorization header must be in the format 'Bearer <token>': %w", ErrUnauthorized)
```

kurtisvg (approved):

PR #1390, opened 2025-09-09
chore: update yaml tag in source and auth services

Yuan325:
Update yaml tag in source and auth services from `kind` to `type`.

We'll need all yaml tag updates in order for the tests to pass.


Yuan325:
Closing due to priority shift. Will resume in the future.

PR #1395, opened 2025-09-09
feat: add metadata in MCP Manifest for Toolbox auth

Yuan325:
Add `_meta` for `tools/list` method in MCP Toolbox.

If there are authorized invocation, the following will be return in `_meta`:
```
{
    "name":"my-tool-name",
    "description":"my tool description",
     "inputSchema":{
        "type":"object",
         "properties":{
             "user_id":{"type":"string","description":"user's name from google login"}
         },
         "required":["user_id"]
     },
     "_meta":{
         "toolbox/authParam":{"user_id":["my_auth"]}
     }
}
```

If there are authenticated parameter, the following will be return in `_meta`:
```
{
    "name":"my-tool-name",
    "description":"my tool description",
    "inputSchema":{
        "type":"object",
        "properties":{
            "sql":{"type":"string","description":"The sql to execute."}
        },
        "required":["sql"]
    },
    "_meta":{
        "toolbox/authInvoke":["my_auth"]
    }
}
```

If there are no authorized invocation or authenticated prameter, the `_meta` field will be omitted.

With this feature, the following were updated in the source code: 
* In each `func(p CommonParameter) McpManifest()`, we will return a `[]string` for the list of authenticated parameters. This is similar to how Manifest() return the list of authNames in non-MCP Toolbox's manifest.
* The `func(ps Parameters) McpManifest()` will return a `map[string][]string` that with key as param's name, and value as the param's auth.
* Added a new function `GetMcpManifest()` in `tools.go`. This function will consctruct the McpManifest, and add the `Metadata` field.
* Associated tests were added or updated.

### DIFF
--- a/internal/server/common_test.go
+++ b/internal/server/common_test.go
@@ -75,7 +75,7 @@ func (t MockTool) RequiresClientAuthorization() bool {
 func (t MockTool) McpManifest() tools.McpManifest {
 	properties := make(map[string]tools.ParameterMcpManifest)
 	required := make([]string, 0)
-	authParam := make(map[string][]string)
+	authParams := make(map[string][]string)
 
 	for _, p := range t.Params {
 		name := p.GetName()
@@ -84,7 +84,7 @@ func (t MockTool) McpManifest() tools.McpManifest {
 		required = append(required, name)
 
 		if len(authParamList) > 0 {
-			authParam[name] = authParamList
+			authParams[name] = authParamList
 		}
 	}
 
@@ -100,9 +100,9 @@ func (t MockTool) McpManifest() tools.McpManifest {
 		InputSchema: toolsSchema,
 	}
 
-	if len(authParam) > 0 {
+	if len(authParams) > 0 {
 		mcpManifest.Metadata = map[string]any{
-			"toolbox/authParam": authParam,
+			"toolbox/authParams": authParams,
 		}
 	}
 


internal/tools/parameters.go
```diff
@@ -437,11 +441,15 @@ func (p *CommonParameter) GetRequired() bool {
 }
 
 // McpManifest returns the MCP manifest for the Parameter.
-func (p *CommonParameter) McpManifest() ParameterMcpManifest {
+func (p *CommonParameter) McpManifest() (ParameterMcpManifest, []string) {
+	authNames := make([]string, len(p.AuthServices))
```
duwenxin99: I think `authNames` is a little vague. Can we use `authServiceNames` instead?

duwenxin99 (changes requested):

internal/tools/parameters.go
Yuan325: Updated. Also updated the `authNames` used in toolbox manifest (non-mcp). Created a common function to dedup.

internal/server/common_test.go
```diff
@@ -75,11 +75,17 @@ func (t MockTool) RequiresClientAuthorization() bool {
 func (t MockTool) McpManifest() tools.McpManifest {
 	properties := make(map[string]tools.ParameterMcpManifest)
 	required := make([]string, 0)
+	authParam := make(map[string][]string)
```
duwenxin99:
```suggestion
	authParams := make(map[string][]string)
```

Yuan325: will keep `authParam` to be consistent with the metadata key name.

PR #1463, opened 2025-09-16
ci: add code signing signatures to binary releases

duwenxin99:
https://cloud.google.com/build/docs/building/build-containers#use-cosign

### DIFF
--- a/.ci/versioned.release.cloudbuild.yaml
+++ b/.ci/versioned.release.cloudbuild.yaml
@@ -17,6 +17,7 @@ steps:
     waitFor: ['-']
     script: |
         #!/usr/bin/env bash
+        set -e
         export VERSION=$(cat ./cmd/version.txt)
         docker buildx create --name container-builder --driver docker-container --bootstrap --use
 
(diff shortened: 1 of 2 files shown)


.ci/versioned.release.cloudbuild.yaml
```diff
@@ -26,6 +27,41 @@ steps:
...
+  - id: "get-docker-digest"
+    name: "gcr.io/cloud-builders/gcloud"
+    waitFor:
+      - "build-docker"
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      export VERSION=$(cat ./cmd/version.txt)
+      IMAGE_DIGEST=$(\
+        gcloud container images describe ${_DOCKER_URI}:$VERSION \
```
Yuan325, line 47: we push docker with 2 tags `$VERSION` and the `latest` if its a latest tag. will this work for both (same image with different tag)?

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -52,14 +88,31 @@ steps:
...
     script: |
         #!/usr/bin/env bash
+        set -e
         export VERSION=v$(cat ./cmd/version.txt)
-        gcloud storage cp toolbox.linux.amd64 gs://$_BUCKET_NAME/$VERSION/linux/amd64/toolbox
+        
+        gcloud storage cp toolbox.linux.amd64
+        gs://$_BUCKET_NAME/$VERSION/linux/amd64/toolbox
+
+        gcloud storage cp toolbox.linux.amd64.sig gs://$_BUCKET_NAME/$VERSION/linux/amd64/toolbox.sig
```
Yuan325: does it need to have the `.sig` suffix? And do we need to store both? can we just store the signed ones without the `.sig` suffix? Same with the rest.

duwenxin99: The .sig file does not contain the original file itself, but contains the signature and certificate that will be used to verify the original file.

.ci/versioned.release.cloudbuild.yaml
duwenxin99: I think the 'latest' tag shouldn't matter here? We are getting the image with this specific version tag so whether there are other tags on it shouldn't make a difference.

averikitsch:
Closing for now.

PR #1472, opened 2025-09-16
feat: add support for user agent in cloud monitoring source

dolphin1999:
## Description

---
> Add support for user agent in cloud monitoring source
> Make changes to the toolset names to be in sync with the overall toolbox

## PR Checklist

---

- [x] Make sure you reviewed
- [] Make sure to add `!` if this involve a breaking change

### DIFF
--- a/internal/sources/cloudmonitoring/cloud_monitoring.go
+++ b/internal/sources/cloudmonitoring/cloud_monitoring.go
@@ -44,7 +44,7 @@ func (rt *userAgentRoundTripper) RoundTrip(req *http.Request) (*http.Response, e
 	if ua == "" {
 		newReq.Header.Set("User-Agent", rt.userAgent)
 	} else {
-		newReq.Header.Set("User-Agent", rt.userAgent+" "+ua)
+		newReq.Header.Set("User-Agent", ua+" "+rt.userAgent)
 	}
 	return rt.next.RoundTrip(&newReq)
 }


prernakakkar-google (approved):

internal/sources/cloudmonitoring/cloud_monitoring.go
```diff
@@ -29,6 +29,26 @@ import (
...
+	newReq := *req
+	newReq.Header = make(http.Header)
+	for k, v := range req.Header {
+		newReq.Header[k] = v
+	}
+	ua := newReq.Header.Get("User-Agent")
+	if ua == "" {
+		newReq.Header.Set("User-Agent", rt.userAgent)
+	} else {
+		newReq.Header.Set("User-Agent", rt.userAgent+" "+ua)
```
Yuan325: Correct me if I'm wrong but seems like best practice when appending UA is to append to the back instead of the front? 

```suggestion
		newReq.Header.Set("User-Agent", ua+" "+rt.userAgent)
```

internal/sources/cloudmonitoring/cloud_monitoring.go
```diff
@@ -65,14 +85,24 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
 		// Use Application Default Credentials
 		creds, err := google.FindDefaultCredentials(ctx, monitoring.MonitoringScope)
 		if err != nil {
 			return nil, fmt.Errorf("failed to find default credentials: %w", err)
 		}
-		client = oauth2.NewClient(ctx, creds.TokenSource)
+		baseClient := oauth2.NewClient(ctx, creds.TokenSource)
+		baseClient.Transport = &userAgentRoundTripper{
+			userAgent: ua,
+			next:      baseClient.Transport,
```
Yuan325, line 103: Just curious, will this be empty? or just the default Transport value when creating a new client? What is the difference from `http.DefaultTransport`?

Yuan325 (approved):
Thank you! Left a comment and a nit.

internal/sources/cloudmonitoring/cloud_monitoring.go
dolphin1999: https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.31.0:transport.go;l=20-28 

I think using baseClient.Transport ensures that custom User-Agent header is added before the request is passed to the OAuth layer that handles authentication.

The Execution flow with http.DefaultTransport :
*userAgentRoundTripper → http.DefaultTransport → Network

The Execution flow with oauth2.Transport :
*userAgentRoundTripper → *oauth2.Transport → http.DefaultTransport → Network

internal/sources/cloudmonitoring/cloud_monitoring.go
dolphin1999: Following the same format as done in csqladmin source. 

https://github.com/googleapis/genai-toolbox/blob/main/internal/sources/cloudsqladmin/cloud_sql_admin.go#L38C1-L51C2

dolphin1999: Made the change as suggested after discussing with @prernakakkar-google

PR #1496, opened 2025-09-17
fix(sources/looker) Allow Looker to be configured without setting a Client Id or Secret

drstrangelooker:
Fixed a minor oversight from previous work on oauth.


drstrangelooker:
@Yuan325 feel free to approve at any time, but there is no urgency. This can wait until after the gemini-cli extensions ship.

Yuan325 (approved):

PR #1497, opened 2025-09-17
fix(tools/looker): Refactor run-inline-query logic to helper function

drstrangelooker:
Inline queries are directed to an undocumented endpoint so that they can be tracked in Looker System Activity separately. The logic is to try the undocumented endpoint first, and if there is any error, fall back to the documented endpoint.

This was done in multiple places. This PR combines that logic into one helper function, reducing the duplication of code.


drstrangelooker:
@Yuan325 feel free to approve at any time, but there is no urgency. This can wait until after the gemini-cli extensions ship.

internal/tools/looker/lookerquery/lookerquery.go
```diff
@@ -127,27 +127,11 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken
...
-		Query: *wq,
-		RenderOpts: lookercommon.RenderOptions{
-			Format: "json",
-		},
-		QueryApiClientCtx: lookercommon.QueryApiClientContext{
-			Name: "MCP Toolbox",
-		},
-	}
-	resp, err := lookercommon.RunInlineQuery2(sdk, req2, t.ApiSettings)
+	resp, err := lookercommon.RunInlineQuery(ctx, sdk, wq, "json", t.ApiSettings)
```
drstrangelooker, line 130: @LukaFontanilla I'm combining all this logic into one spot.

Yuan325 (approved):

PR #1503, opened 2025-09-18
ci: add binaries for extensions

Yuan325:


.ci/versioned.release.cloudbuild.yaml
```diff
@@ -61,6 +61,30 @@ steps:
...
+    env: 
+      - 'GOPATH=/gopath'
+    volumes:
+      - name: 'go'
+        path: '/gopath'
+    script: |
+        #!/usr/bin/env bash
+        export VERSION=$(cat ./cmd/version.txt)
+        CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
+          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.buildType=extensions -X github.com/googleapis/genai-toolbox/cmd.commitSha=$(git rev-parse HEAD)" -o toolbox.extensions.linux.amd64
```
averikitsch: nit: We may want to use `cmd.buildType=extensions.binary` if we think there's going to be other install methods in the future. I am also fine with this approach

PR #1528, opened 2025-09-22
chore(tests): Add database cleanup functions

### LINKED ISSUE #1304: Integration tests do not clean up database tables from previous runs
### Prerequisites

### Toolbox version

toolbox version 0.13.0

### Environment

_No response_

### Expected Behavior

Each integration test run must start with a clean and predictable database state. There should be no data or tables present from the previous test runs.

### Current Behavior

Integration tests (for postgres, cloudsql postgres, etc) do not clean up tables from previous test runs (probably from aborted test runs). This causes accumulation of orphan tables (like `param_table_uuid`, `auth_table_uuid`) in the test databases. This can make subsequent tests unreliable. 

### Steps to reproduce?

1. Set up the environment variables for a database (ex. postgres) to run integration tests.
2. Use `go test -timeout` or anything else to abort the test run.
3. Inspect the test database to find orphan tables from this test run.

### Additional Details

_No response_

Myst9:
## Description

---
This change adds helper functions to cleanup databases (PostgreSQL, MySQL, and MSSQL). They can be called at the beginning of each test run.

## PR Checklist

---

- [ ] Make sure you reviewed

🛠️ Fixes #1304


tests/common.go
```diff
@@ -789,3 +790,98 @@ func TestCloudSQLMySQL_IPTypeParsingFromYAML(t *testing.T) {
...
+			continue
+		}
+		tablesToDrop = append(tablesToDrop, fmt.Sprintf("`%s`", tableName))
+	}
+
+	if len(tablesToDrop) == 0 {
+		return
+	}
+
+	// Disable foreign key checks, drop all tables and re-enable
```
duwenxin99, line 868: Why do we need this?

Myst9: If we don't disable foreign key checks, the `DROP TABLE` query can fail if there are any dependent tables (based on the order in which the tables are dropped). The current test tables don't use foreign keys but I added this just in case future tests do.

duwenxin99: Sounds good.

PR #1564, opened 2025-09-25
feat: add new enum parameter type

### LINKED ISSUE #779: Mitigations for potential security risks in template parameter
### Prerequisites

The addition of `templateParameters` is valuable for Toolbox users. However, since template parameters can directly replace identifiers, column names, and table names, they are prone to SQL injections. 

Do note that basic parameters are always preferred for performance and safety reasons.

### Suggested Solution(s)
We are proposing two solutions to mitigate this risk.

(1) Adding an `escape` field to template parameter. If user uses `templateParameter` with this field, we will add double quotes (or backticks/square brackets depending on the db) to escape identifiers such as column names or table names. 
```
tools:
 select_columns_from_table:
    kind: postgres-sql
    source: my-pg-instance
    statement: SELECT * FROM {{.tableName}};
    description: Use this tool to list all information from a specific table.
    templateParameters:
      - name: tableName
        type: string
        description: Table to select from
        escape: true
```
In the example above, the statement will resolve to `SELECT * FROM "flight_table";`. This method does reduces some risks but still prone to SQL injections.

(2) Adding a new `allowedValues` field in `templateParameters` that allows users to provide an allow-list. In the example below, `tableName` could be either `flights_table` or `tickets_table`. Any other values provided will cause an error and the query will NOT be performed.
```
tools:
 select_columns_from_table:
    kind: postgres-sql
    source: my-pg-instance
    statement: SELECT * FROM {{.tableName}}
    description: Use this tool to list all information from a specific table.
    templateParameters:
      - name: tableName
        type: string
        description: Table to select from
        allowedValues:
          - flights_table
          - tickets_table
```

### Alternatives Considered

(2a) adding a new `enum` parameter type. Similar to `allowedValues`, we will instead add a new parameter type `enum`: 
```
    templateParameters:
      - name: tableName
        type: enum
        description: Table to select from.
        enumType: string
        values:
          - flights_table
          - tickets_table
```
With this type, user can provide a list of values. There's also potential to support retrieving an allow-list from the description.

(3) Adding a flag `--allow-unsafe-tools` if user wants to use the `templateParameters`. Without the flag, users WILL NOT be able to use tools that have `templateParameters`.

(4) Adding a field for user to indicate `runSingleStatement`. The field will default to `false`. When specified, Toolbox will only grab the first statement (the statement before the first `;`).
```
tools:
 select_columns_from_table:
    kind: postgres-sql
    source: my-pg-instance
    statement: SELECT * FROM {{.tableName}}
    runSingleStatement: true
    description: Use this tool to list all information from a specific table.
```

### Additional Details

_No response_

Yuan325:
## Description

---
Add new `enum` parameter type. `enum` parameter type can be used as follows (can be used in the `parameter` field or `templateParameter` field:

```
parameter:
    - name: tableName
       type: enum
       description: table name.
       enumType: string # this will be one of { string / integer / float / boolean / array}
       allowedValues:
            - flights_table
            - tickets_table
```

The `enumType` field could be any of the valid parameter type.

Here, we introduce 3 new fields as compared to the regular parameters:
| parameter name | type | required | description |
|------------------|-----|---------|-------------|
| enumType | string | true | This indicates the type of the value. Accepted values are any of the valid parameter type . |
| escape | boolean | false | This indicates if the value will be escaped. For escaping, we will add double quotes (or backticks/square brackets depending on the db) to escape identifiers when used in **template parameter**. This is defaulted to `false`. |
| allowedValues | []string or string | true | We will check input value against this. User can either provide a list of allowed values. | 

Future PR changes: 
* expand `allowedValues` field to support regex. Currently it supports a list of allowedValues

## PR Checklist

---

- [ ] Make sure you reviewed

🛠️ Fixes #779


internal/tools/parameters.go
```diff
@@ -179,6 +181,11 @@ func GetParams(params Parameters, paramValuesMap map[string]any) (ParamValues, e
 		if !ok {
 			return nil, fmt.Errorf("missing parameter %s", k)
 		}
+		if p.GetType() == typeEnum {
+			if p.(*EnumParameter).GetEscape() {
```
averikitsch: Does this need to be set based on database?

internal/tools/parameters.go
```diff
@@ -1230,3 +1254,144 @@ func (p *MapParameter) McpManifest() ParameterMcpManifest {
 		AdditionalProperties: additionalProperties,
 	}
 }
+
+// NewEnumParameter is a convenience function for initializing a EnumParameter.
+func NewEnumParameter(param Parameter, escape bool, allowedValues []any) *EnumParameter {
+	d := param.GetDefault()
```
averikitsch: Do we need a NewEnumParameterWithDefault method? This can probably wait until requested.

internal/tools/parameters.go
```diff
@@ -1230,3 +1254,144 @@ func (p *MapParameter) McpManifest() ParameterMcpManifest {
...
+	AllowedValues   []any  `yaml:"allowedValues"`
+	EnumItem        Parameter
+}
+
+// Ensure EnumParameter implements the Parameter interface.
+var _ Parameter = &EnumParameter{}
+
+// UnmarshalYAML handles parsing the EnumParameter from YAML input.
+func (p *EnumParameter) UnmarshalYAML(ctx context.Context, unmarshal func(interface{}) error) error {
+	var rawItem map[string]any
```
averikitsch: nit: is it possible to use a struct here?

internal/tools/parameters.go
```diff
@@ -1230,3 +1254,144 @@ func (p *MapParameter) McpManifest() ParameterMcpManifest {
...
+	delete(rawItem, "allowedValues")
+
+	// create a util.DelayedUnmarshaler from the remaining fields
+	m, err := yaml.Marshal(rawItem)
+	if err != nil {
+		return fmt.Errorf("error marshaling remaining fields from enum parameter")
+	}
+	var delayedUnmarshaler util.DelayedUnmarshaler
+	if err = yaml.UnmarshalContext(ctx, m, &delayedUnmarshaler); err != nil {
+		return fmt.Errorf("error unmarhaling into DelayedUnmarshaler")
```
averikitsch: nit: typo

internal/tools/parameters.go
```diff
@@ -1230,3 +1254,144 @@ func (p *MapParameter) McpManifest() ParameterMcpManifest {
...
+	if err := unmarshal(&rawItem); err != nil {
+		return fmt.Errorf("error parsing enum parameter: %w", err)
+	}
+
+	// extract enum parameter known fields
+	enumType, ok := rawItem["enumType"].(string)
+	if !ok {
+		return fmt.Errorf("error parsing 'enumType' field")
+	}
+	escape := false
```
averikitsch: Does escape get an updated value?

Yuan325:
@averikitsch refactored PR to the latest design.

internal/tools/parameters.go
```diff
@@ -436,6 +440,40 @@ func (p *CommonParameter) GetRequired() bool {
...
 }
 
+// GetAllowedValues returns the allowed values for the Parameter.
+func (p *CommonParameter) GetAllowedValues() []any {
+	return p.AllowedValues
+}
+
+// IsAllowedValues checks if the value is allowed.
+func (p *CommonParameter) IsAllowedValues(v any) bool {
+	a := p.AllowedValues
```
duwenxin99: nit: I think it's better to remove this assignment. "a" is quite meaningless and reduces readability.

PR #1568, opened 2025-09-25
feat: support regex in enum type

Yuan325:
Support regex in enum type.


Yuan325:
Regex is added in another PR. Closing due to obsolete.

Issue #1491, opened 2025-09-17
MongoDB "find" tool not using the limit parameter


abdul0807:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

toolbox version 0.14.0

### Expected Behavior

Based on the below configuration:

```
tools:
  find_documents_in_collection:
    kind: mongodb-find
    source: mongodb-source
    description: >-
      Finds and lists multiple documents from the given collection.
      Use this for general searches or when you expect a list of results.
    database: xxxxxx
    collection: xxxxxx
    limit: 5
```

Only 5 records should be returned because the limit is 5.

### Current Behavior

The limit parameter is not being applied to the MongoDB query.

### Steps to reproduce?

1. create a tools.yaml file to connect to MongoDB
2. Use the "find tool". Sample configuration is shared below for reference:
```
tools:
  find_documents_in_collection:
    kind: mongodb-find
    source: mongodb-source
    description: >-
      Finds and lists multiple documents from the given collection.
      Use this for general searches or when you expect a list of results.
    database: xxxxxx
    collection: xxxxxx
    limit: 5
```
3. The toolbox returns all the documents and does not respect the limit value

duwenxin99:
Hi @abdul0807, thanks for opening the issue. Let me take a look into this.

PR #1570, opened 2025-09-25
fix(tool/mongodb-find): fix find tool `limit` field

duwenxin99:
The projection checking block in `getOptions` exists early previously, resulting in the limit not being set.

🛠️ Fixes https://github.com/googleapis/genai-toolbox/issues/1491

### DIFF
--- a/internal/tools/mongodb/mongodbfind/mongodbfind.go
+++ b/internal/tools/mongodb/mongodbfind/mongodbfind.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/goccy/go-yaml"
 	mongosrc "github.com/googleapis/genai-toolbox/internal/sources/mongodb"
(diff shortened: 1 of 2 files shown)


internal/tools/mongodb/mongodbfind/mongodbfind.go
```diff
@@ -54,7 +55,7 @@ type Config struct {
 	Database       string           `yaml:"database" validate:"required"`
 	Collection     string           `yaml:"collection" validate:"required"`
 	FilterPayload  string           `yaml:"filterPayload" validate:"required"`
-	FilterParams   tools.Parameters `yaml:"filterParams" validate:"required"`
+	FilterParams   tools.Parameters `yaml:"filterParams"`
```
Yuan325, line 58: Might wanna update docs too if we change this.

PR #1588, opened 2025-09-29
feat: Use project from env for alloydb and cloud sql control plane tools

Myst9:
## Description

---
This change introduces the `DefaultProject` field for the `alloydb-admin` and `cloud-sql-admin` sources. This field allows the alloydb and cloud sql control plane tools to use the project value from the environment variables (Ex: `ALLOYDB_POSTGRES_PROJECT`) if it is already set instead of asking the user.

## PR Checklist

---

- [x] Make sure you reviewed

🛠️ Fixes: https://github.com/gemini-cli-extensions/alloydb/issues/47


prernakakkar-google (approved):

docs/en/resources/sources/alloydb-admin.md
```diff
@@ -1,12 +1,10 @@
 ---
-title: "AlloyDB Admin"
-linkTitle: "AlloyDB Admin"
+title: AlloyDB Admin
+linkTitle: AlloyDB Admin
 type: docs
 weight: 2
```
averikitsch: nit: this may fix the alphabetical order
```suggestion
weight: 1
```

internal/sources/alloydbadmin/alloydbadmin.go
```diff
@@ -110,11 +111,14 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
 		return nil, fmt.Errorf("error creating new alloydb service: %w", err)
 	}
 
+	defaultProject := util.ExpandEnv(r.DefaultProject)
```
averikitsch: Env vars in the config files get automatically parsed when the tools file is loaded via https://github.com/googleapis/genai-toolbox/blob/5aed4e136d0091731d2ded10ec076ee789e1987c/cmd/root.go#L343. Is there a reason why we need to parse it again?

internal/tools/alloydb/alloydblistusers/alloydblistusers.go
```diff
@@ -70,8 +70,16 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 		return nil, fmt.Errorf("invalid source for %q tool: source kind must be `%s`", kind, alloydbadmin.SourceKind)
 	}
 
+	project := s.DefaultProject
+	var projectParam tools.Parameter
+	if project != "" {
```
averikitsch, line 75: nit: do we want to add "This is pre-configured; you can verify the default project with the user"

docs/en/resources/sources/alloydb-admin.md
```diff
@@ -36,7 +35,9 @@ sources:
...
 
 ## Reference
+
 | **field**      | **type** | **required** | **description**                                                                                                                                |
-|----------------|:--------:|:------------:|------------------------------------------------------------------------------------------------------------------------------------------------|
+| -------------- | :------: | :----------: | ---------------------------------------------------------------------------------------------------------------------------------------------- |
 | kind           |  string  |     true     | Must be "alloydb-admin".                                                                                                                       |
-| useClientOAuth | boolean  |    false     | If true, the source will use client-side OAuth for authorization. Otherwise, it will use Application Default Credentials. Defaults to `false`. |
\ No newline at end of file
+| defaultProject |  string  |     false    | The environment variable containing the Google Cloud project ID to use for all AlloyDB operations.                                             |
```
averikitsch: The prebuilt tools read the env var but in general this could be hard coded for custom tools

averikitsch (changes requested):

internal/sources/alloydbadmin/alloydbadmin.go
Myst9: You're right, it is not needed. Modified the regex in the parseEnv function to allow hyphens in the default value and updated this code.

internal/tools/alloydb/alloydblistusers/alloydblistusers.go
Myst9: Yes, without this the agent still asks the user for the project id even when it is set

docs/en/resources/sources/cloud-sql-admin.md
```diff
@@ -37,6 +36,7 @@ sources:
 ## Reference
 
 | **field**      | **type** | **required** | **description**                                                                                                                                |
-|----------------|:--------:|:------------:|------------------------------------------------------------------------------------------------------------------------------------------------|
+| -------------- | :------: | :----------: | ---------------------------------------------------------------------------------------------------------------------------------------------- |
 | kind           |  string  |     true     | Must be "cloud-sql-admin".                                                                                                                     |
-| useClientOAuth | boolean  |    false     | If true, the source will use client-side OAuth for authorization. Otherwise, it will use Application Default Credentials. Defaults to `false`. |
+| defaultProject |  string  |     false    | For prebuilt tools, this is the environment variable containing the Google Cloud project ID.                                                   |
```
averikitsch: Please add any prebuilt info to the reference page: https://googleapis.github.io/genai-toolbox/reference/prebuilt-tools/. These doc pages should just reference the field ie "The Google Cloud project ID to use for Cloud SQL infrastructure tools."

Myst9: Hi @averikitsch , I updated the doc pages for the sources. Can you please clarify what I should add here: https://googleapis.github.io/genai-toolbox/reference/prebuilt-tools? As the `defaultProject` field is defined in the source and not the tools.

averikitsch (approved):

docs/en/resources/sources/cloud-sql-admin.md
averikitsch: We shouldn't need to update the ref page. I was just outlining the different scopes of the pages.

averikitsch:
@Myst9 please see lint error

PR #1711, opened 2025-10-14
fix: githubaction broken links

rahulpinto19:
The GitHub Action is configured to check all broken links found in .MD (Markdown) files, including both external URLs and local file paths, while specifically ignoring any localhost links.

### DIFF
--- a/.github/workflows/action.yaml
+++ b/.github/workflows/action.yaml
@@ -1,0 +1,37 @@
+name: Link Check
+
+on:
+  push:
+
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v5
+
+      - name: Restore lychee cache
+        uses: actions/cache@v4
+        with:
+          path: .lycheecache
+          key: cache-lychee-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


.lycheeignore
```diff
@@ -0,0 +1,15 @@
+
+
+^https?:\\/\\/([a-zA-Z0-9-]+\\.)?example\\.com(:\\d+)?(\\/.*)?$
```
twishabansal, line 3: Why are we using regexes in the ignore file?

.github/workflows/action.yaml
```diff
@@ -0,0 +1,37 @@
+name: Link Check
```
twishabansal, line 1: Let's add a license header here

.lycheeignore
```diff
@@ -0,0 +1,15 @@
+
```
twishabansal, line 1: Can we add a license header to this file?

.lycheeignore
rahulpinto19: This we can remove , if there are any example.com domain then we need to ignore

.lycheeignore
rahulpinto19: it's not required it only checks the .MD files

Issue #1737, opened 2025-10-19
HTTP tools send optional nil query parameters as empty strings instead of omitting them


Px8Studio:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

latest (main branch)

### Environment

1. OS type and version: Any (affects all platforms)
2. How are you running Toolbox: Local Docker image

### Client

1. Client: Any HTTP tool (see example below)
2. Version: N/A
3. Example: Minimal YAML config and curl example below

### Expected Behavior

When invoking an HTTP tool with optional query parameters not specified (or set to `null`/`nil`), the generated HTTP request URL should **omit** those parameters. For example, if both `page` and `pageSize` are optional and not provided, the URL should be:
```
https://api.example.com/endpoint
```
This matches REST API standards, avoids sending empty query values, and prevents API errors.

### Current Behavior

The HTTP tool implementation in `internal/tools/http/http.go` currently converts `nil` values for optional query parameters to empty strings and **always includes them** in the request URL (e.g., `?page=&pageSize=`). Strict APIs reject these requests with HTTP 400 Bad Request errors, even though the intention is to omit the optional parameters entirely.

#### Minimal Example
```yaml
tools:
  my-http-tool:
    kind: http
    source: my-http-source
    method: GET
    path: /data
    queryParams:
      - name: page
        type: integer
        required: false
      - name: pageSize
        type: integer
        required: false
```
Invoke with:
```bash
curl -X POST http://localhost:5000/api/tool/my-http-tool/invoke \
  -H "Content-Type: application/json" \
  -d '{}'
```
Resulting URL (buggy):
```
https://api.example.com/data?page=&pageSize=
```
API returns: `HTTP 400 Bad Request`

### Steps to reproduce?

1. Create an HTTP tool with optional query parameters as above
2. Invoke the tool with `{}` (no parameters)
3. Observe the generated request URL and HTTP 400 response from the upstream API

### Additional Details

- **Root Cause:**
  In `internal/tools/http/http.go`, the `getURL()` function (lines 210-229) adds ALL query parameters, converting `nil` to `""` (empty string) and always adding them to the query string.
  ```go
  for _, p := range queryParams {
      v := paramsMap[p.GetName()]
      if v == nil {
          v = ""
      }
      query.Add(p.GetName(), fmt.Sprintf("%v", v))
  }
  ```
- **Impact:**
[...]

PR #1762, opened 2025-10-21
fix(tools/http): omit optional nil query parameters

duwenxin99:
🛠️ Fixes https://github.com/googleapis/genai-toolbox/issues/1737

### DIFF
--- a/docs/en/resources/tools/_index.md
+++ b/docs/en/resources/tools/_index.md
@@ -77,13 +77,17 @@ the parameter.
         description: Airline unique 2 letter identifier
 ```
 
-| **field**   | **type**        | **required** | **description**                                                             |
-|-------------|:---------------:|:------------:|-----------------------------------------------------------------------------|
-| name        |  string         |     true     | Name of the parameter.                                                      |
-| type        |  string         |     true     | Must be one of "string", "integer", "float", "boolean" "array"              |
-| description |  string         |     true     | Natural language description of the parameter to describe it to the agent.  |
-| default     |  parameter type |     false    | Default value of the parameter. If provided, `required` will be `false`.    |
-| required    |  bool           |     false    | Indicate if the parameter is required. Default to `true`.                   |
+| **field**     |    **type**    | **required** | **description**                                                                                                                                                                                                                        |
+|---------------|:--------------:|:------------:|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| name          |     string     |     true     | Name of the parameter.                                                                                                                                                                                                                 |
+| type          |     string     |     true     | Must be one of "string", "integer", "float", "boolean" "array"                                                                                                                                                                         |
+| description   |     string     |     true     | Natural language description of the parameter to describe it to the agent.                                                                                                                                                             |
+| default       | parameter type |    false     | Default value of the parameter. If provided, `required` will be `false`.                                                                                                                                                               |
+| required      |      bool      |    false     | Indicate if the parameter is required. Default to `true`.                                                                                                                                                                              |
+| allowedValues |    []string    |    false     | Input value will be checked against this field. Regex is also supported.                                                                                                                                                               |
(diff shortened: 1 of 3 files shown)


tests/http/http_integration_test.go
```diff
@@ -305,8 +328,91 @@ func TestHttpToolEndpoints(t *testing.T) {
...
+		requestBody io.Reader
+		want        string
+		isErr       bool
+	}{
+		{
+			name:        "invoke query-param-tool (optional omitted)",
+			api:         "http://127.0.0.1:5000/api/tool/my-query-param-tool/invoke",
+			requestBody: bytes.NewBuffer([]byte(`{"reqId": "test1"}`)),
+			want:        `"reqId=test1"`,
+			isErr:       false,
```
Yuan325: if isErr is always false, we probably don't need it here.

tests/http/http_integration_test.go
```diff
@@ -305,8 +328,91 @@ func TestHttpToolEndpoints(t *testing.T) {
...
+			api:         "http://127.0.0.1:5000/api/tool/my-query-param-tool/invoke",
+			requestBody: bytes.NewBuffer([]byte(`{"reqId": "test2", "page": "5"}`)),
+			want:        `"page=5\u0026reqId=test2"`, // 'filter' omitted
+			isErr:       false,
+		},
+		{
+			name:        "invoke query-param-tool (required param nil)",
+			api:         "http://127.0.0.1:5000/api/tool/my-query-param-tool/invoke",
+			requestBody: bytes.NewBuffer([]byte(`{"reqId": null, "page": "1"}`)),
+			want:        `"page=1\u0026reqId="`, // reqId becomes "",
```
Yuan325, line 365: I thought we should omit reqId overall? so the want would be just "page=1"?

tests/http/http_integration_test.go
```diff
@@ -305,8 +328,91 @@ func TestHttpToolEndpoints(t *testing.T) {
...
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatalf("unable to send request: %s", err)
+			}
+			defer resp.Body.Close()
+
+			if resp.StatusCode != http.StatusOK {
+				if tc.isErr == true {
+					return
+				}
```
Yuan325: Probably don't need this since isErr is always false.

tests/http/http_integration_test.go
```diff
@@ -470,7 +588,7 @@ func getHTTPToolsConfig(sourceConfig map[string]any, toolKind string) map[string
 				"method":      "get",
 				"path":        "/{{.path}}?id=2",
 				"description": "some description",
-				"headers": map[string]string{
+		"headers": map[string]string{
```
Yuan325: please fix indentation

Yuan325 (changes requested):

tests/http/http_integration_test.go
duwenxin99: We preserver `reqId=` if it is a required parameter, otherwise we omit it. It will depend on the users whether they want to send a null query parameter.

tests/http/http_integration_test.go
```diff
@@ -305,8 +329,84 @@ func TestHttpToolEndpoints(t *testing.T) {
...
+			}
+
+			if got != tc.want {
+				t.Fatalf("unexpected value: got %q, want %q", got, tc.want)
+			}
+		})
+	}
+}
+
+// END: New test runner function
```
Yuan325:
```suggestion
}
```

tests/http/http_integration_test.go
```diff
@@ -68,11 +69,35 @@ func multiTool(w http.ResponseWriter, r *http.Request) {
...
+
+    err := enc.Encode(r.URL.RawQuery)
+    if err != nil {
+        http.Error(w, "Failed to write response", http.StatusInternalServerError)
+        return
+    }
+}
+
+
 // handler function for the test server
```
Yuan325, line 100:
```suggestion
}

// handler function for the test server
```

Yuan325 (approved):
last nits~

PR #1763, opened 2025-10-21
feat: support escape in string parameter

Yuan325:
supporting `backticks`, `double-quotes`, `single-quotes`, `square-brackets` as escaping delimiters.

Example to apply escaping delimiters:
```
parameters:
      - name: airline
        type: string
        description: Airline unique 2 letter identifier
        escape: double-quotes
```

escaping delimiters can be used for identifiers or string literals. If `allowedValues` were used, Toolbox will check for allowed values before applying delimiters.


docs/en/resources/tools/_index.md
```diff
@@ -233,6 +236,7 @@ tools:
           name: column
           type: string
           description: Name of a column to select
+          escape: double-quotes
```
duwenxin99: nit: can we add an output example of how the statement will look like after escaping?

duwenxin99 (approved):

Issue #1689, opened 2025-10-10
Add tools for Google Serverless Spark


dborowitz:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

I'm from the Google Cloud Dataproc team, and I'd like to add support for one of our runtime engines, Serverless Spark (full name [Google Cloud Serverless for Apache Spark](https://cloud.google.com/dataproc-serverless/docs)).

Mostly, the exposed tools should be thin wrappers around our batch and session [CRUD APIs](https://cloud.google.com/dataproc-serverless/docs/reference/rest), as well as some undocumented APIs for fetching debugging information.

Sample tools:
- batch_list: list all batches in a project/location
- batch_get: get info on a single batch
- batch_create: create a batch (Spark workload)
- operation_cancel: cancel a long-running operation, such as a batch
- batch_metrics_get: get runtime metrics for a batch that may be useful for debugging

Example CUJs include:
- Integrating with coding agents to build and launch batches from project source code
- Checking the status of running batches.
- Debugging performance or other runtime issues with current or historical batches.

Yuan325:
Closing this issue due to resolve. Will open new issue for future tools.

PR #1783, opened 2025-10-24
feat(serverless-spark): Add get_batch tool

dborowitz:
## Description

Add a new tool `get_batch` taking a batch name and returning the full batch info JSON.

## PR Checklist

- [ ] Make sure you reviewed

🛠️ Part of #1689

### DIFF
--- a/internal/tools/serverlessspark/serverlesssparkgetbatch/serverlesssparkgetbatch.go
+++ b/internal/tools/serverlessspark/serverlesssparkgetbatch/serverlesssparkgetbatch.go
@@ -1,0 +1,171 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package serverlesssparkgetbatch
+
+import (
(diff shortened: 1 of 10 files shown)


dborowitz:
The integration test failure seems to be from the server returning a different error messages for a particular failure case than what I'm seeing locally. I will try to figure out the cause, or failing that just weaken the test assertion.

internal/tools/serverlessspark/serverlesssparkgetbatch/serverlesssparkgetbatch.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, fmt.Errorf("invalid source for %q tool: source kind must be `%s`", kind, serverlessspark.SourceKind)
+	}
+
+	desc := cfg.Description
+	if desc == "" {
+		desc = "Gets a Serverless Spark (aka Dataproc Serverless) batch"
+	}
+
+	allParameters := tools.Parameters{
+		tools.NewStringParameter("name", "The short name of the batch, e.g. for \"projects/my-project/locations/us-central1/batches/my-batch\", pass \"my-batch\" (the project and location are pre-configured in the tool)"),
```
duwenxin99: Should it be "the project and location are inherited from the source" instead? The tools doesn't have these fields so it may confuse the user.

duwenxin99 (approved):

internal/tools/serverlessspark/serverlesssparkgetbatch/serverlesssparkgetbatch.go
dborowitz: Good catch, I think I meant "tools.yaml" but "source" is clearer. Done.

dborowitz:
> The integration test failure seems to be from the server returning a different error messages for a particular failure case than what I'm seeing locally. I will try to figure out the cause, or failing that just weaken the test assertion.

Done. Couldn't figure out the nondeterminism, ended up adding a short-circuit check in invoke.

PR #1804, opened 2025-10-28
feat(tools/looker): tools to retrieve the connections, schemas, databases, and column metadata from a looker system.

drstrangelooker:
## Description

Add new tools to get metadata from databases through Looker

* get_connections
* get_connection_schemas
* get_connection_databases
* get_connection_tables
* get_connection_table_columns


internal/tools/looker/lookergetconnectiondatabases/lookergetconnectiondatabases.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"context"
+	"fmt"
+
+	yaml "github.com/goccy/go-yaml"
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	lookersrc "github.com/googleapis/genai-toolbox/internal/sources/looker"
+	"github.com/googleapis/genai-toolbox/internal/tools"
+	"github.com/googleapis/genai-toolbox/internal/tools/looker/lookercommon"
+	//"github.com/googleapis/genai-toolbox/internal/util"
```
duwenxin99: remove this import?

internal/tools/looker/lookergetconnectiondatabases/lookergetconnectiondatabases.go
```diff
@@ -0,0 +1,155 @@
...
+	Client             *v4.LookerSDK
+	ApiSettings        *rtl.ApiSettings
+	AuthRequired       []string         `yaml:"authRequired"`
+	Parameters         tools.Parameters `yaml:"parameters"`
+	manifest           tools.Manifest
+	mcpManifest        tools.McpManifest
+}
+
+func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken tools.AccessToken) (any, error) {
+	//logger, err := util.LoggerFromContext(ctx)
```
duwenxin99: Do we want to uncomment the logger?

internal/tools/looker/lookergetconnectionschemas/lookergetconnectionschemas.go
```diff
@@ -0,0 +1,164 @@
...
+	Client             *v4.LookerSDK
+	ApiSettings        *rtl.ApiSettings
+	AuthRequired       []string         `yaml:"authRequired"`
+	Parameters         tools.Parameters `yaml:"parameters"`
+	manifest           tools.Manifest
+	mcpManifest        tools.McpManifest
+}
+
+func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken tools.AccessToken) (any, error) {
+	//logger, err := util.LoggerFromContext(ctx)
```
duwenxin99: Uncomment the logger? And we could add debugging logs below

internal/tools/looker/lookergetconnectionschemas/lookergetconnectionschemas.go
```diff
@@ -0,0 +1,164 @@
...
+
+	sdk, err := lookercommon.GetLookerSDK(t.UseClientOAuth, t.ApiSettings, t.Client, accessToken)
+	if err != nil {
+		return nil, fmt.Errorf("error getting sdk: %w", err)
+	}
+	req := v4.RequestConnectionSchemas{
+		ConnectionName: conn,
+	}
+	if db != "" {
+		req.Database = &db
```
duwenxin99, line 129: Why do we need this line here? What happens if `db` is empty string?

drstrangelooker: Then we don't set the element at all. But IIRC it will be a "zero" so an empty string anyway. I'll take this out.

internal/tools/looker/lookergetconnectionschemas/lookergetconnectionschemas.go
drstrangelooker: The stuff I would want to log is actually pointers, so I get a bunch of numbers. not useful

internal/tools/looker/lookergetconnectiondatabases/lookergetconnectiondatabases.go
drstrangelooker: Taking out for now

internal/tools/looker/lookergetconnectiondatabases/lookergetconnectiondatabases.go
drstrangelooker: will do

internal/tools/looker/lookergetconnectionschemas/lookergetconnectionschemas.go
drstrangelooker: Made a mistake here. The Database element is a string pointer. So it needs to be `nil` if db is empty.

drstrangelooker:
@duwenxin99 This should be ready for review. I'll incorporate main after you give a LGTM so we don't have to run all the tests over again.

PR #1818, opened 2025-10-29
feat: Added ExcludeValues + Fixed issue with regex matching type.

### LINKED ISSUE #1792: Introduce excludeValues as a complement to allowedValues
### Prerequisites

The current library provides an allowedValues list, which is perfect for "allow-listing" defining a small, finite set of values that are acceptable for a parameter.

However, there is no simple way to support "deny-listing" defining a small set of values that are unacceptable, while allowing all other values of that type.

Use Case: I have two common scenarios that are currently difficult to implement:

- I want to define a parameter that accepts any integer except 12345.

- I want to define a user parameter that accepts any username except 'admin' or 'root'.

Current allowedValues (Impossible): For the integer case, I would have to list every other integer [...12344, 12346, ...], which is infinite and not possible.

Current regex (Difficult/Fragile): This is the only workaround but this requires a negative lookahead, which is not supported (e.g., `^(?!12345$)\d+$)` or `^(?!(admin|root)$).*$`).

This limitation makes it difficult to implement common validation rules.

### Suggested Solution(s)

I propose introducing a new excludeValues parameter. This parameter would be a list of items that are explicitly forbidden.

To make this feature powerful and flexible, an item in the excludeValues list could be either a literal value OR a regex string.

### Alternatives Considered

_No response_

### Additional Details

_No response_

philBatardiere:
### This PR introduces a new excludedValues field for tool parameters, enhancing validation capabilities.

This change introduces a new excludedValues field for tool parameters. This field allows developers to specify a list of values that are not allowed for a parameter. 
The excludedValues field supports both exact value matching and regular expression matching.

The changes include:
- Updating the tool parameter documentation to include the excludedValues field.
- Adding the excludedValues field to the CommonParameter struct.
- Implementing the logic to check for excluded values in the Parse method of each parameter type.
- Updating the MatchStringOrRegex function to support non-string inputs by converting them to strings before regex matching. This makes the allowedValues and excludedValues checks more robust.
- Adding unit tests for allowedValues to verify the MatchStringOrRegex change on parameters.
- Adding unit tests to verify the excludedValues functionality.

## PR Checklist

- [x] Make sure you reviewed
CONTRIBUTING.md (httpshttps://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md)
bug/issue (https://github.com/googleapis/genai-toolbox/issues/new/choose)
- [ ] Make sure to add ! if this involve a breaking change

🛠️ Fixes #1792

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/jackc/pgx/v5 v5.7.6
 	github.com/json-iterator/go v1.1.12
-	github.com/looker-open-source/sdk-codegen/go v0.25.16
+	github.com/looker-open-source/sdk-codegen/go v0.25.18
 	github.com/microsoft/go-mssqldb v1.9.3
 	github.com/nakagami/firebirdsql v0.9.15
 	github.com/neo4j/neo4j-go-driver/v5 v5.28.4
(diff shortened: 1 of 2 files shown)


philBatardiere:
Hello, I just signed the CLA. Thanks

averikitsch (approved):

averikitsch:
Thank you @philBatardiere for the contribution!

PR #1822, opened 2025-10-30
fix(source/cloud-sql-mssql): remove ip address field

### LINKED ISSUE #1549: Remove IP address from Cloud SQL SQL Server source
### Prerequisites

### Toolbox version

0.15.0

### Environment

  - As a downloaded binary (e.g. from `curl -O https://storage.googleapis.com/genai-toolbox/v$VERSION/linux/amd64/toolbox`)
  - Compiled from source (include the command used to build)

### Client

Gemini CLI

### Expected Behavior

IP address is not an input for Cloud SQL SQL Server source.

### Current Behavior

IP Address is label as a required input for the Cloud SQL SQL server source. However, any ip address works. The connector has uses localhost as an example, so this input most likely is not needed.

### Steps to reproduce?

1. ?
2. ?
3. ?
...

### Additional Details

_No response_

Yuan325:
## Description

Removing the `ipAddress` field since it is not an input for Cloud SQL SQL Server source. 

Kept the variable in Source's config but removed this variable from everywhere else in the code. This will PREVENT a breaking change since the validator won't flag it as an "extra field".

**Will have to update the following as well:**
(1) Cloud docs https://cloud.google.com/sql/docs/sqlserver/pre-built-tools-with-mcp-toolbox
(2) gemini-cli-extensions https://github.com/gemini-cli-extensions/cloud-sql-sqlserver

## PR Checklist

- [x] Make sure you reviewed

🛠️ Fixes #1549

### DIFF
--- a/docs/en/resources/tools/bigquery/bigquery-analyze-contribution.md
+++ b/docs/en/resources/tools/bigquery/bigquery-analyze-contribution.md
@@ -46,6 +46,13 @@ The behavior of this tool is influenced by the `writeMode` setting on its `bigqu
   tools using the same source. This allows the `input_data` parameter to be a query that references temporary resources (e.g., 
   `TEMP` tables) created within that session.
 
+The tool's behavior is also influenced by the `allowedDatasets` restriction on the `bigquery` source:
+
+- **Without `allowedDatasets` restriction:** The tool can use any table or query for the `input_data` parameter.
+- **With `allowedDatasets` restriction:** The tool verifies that the `input_data` parameter only accesses tables within the allowed datasets.
+  - If `input_data` is a table ID, the tool checks if the table's dataset is in the allowed list.
+  - If `input_data` is a query, the tool performs a dry run to analyze the query and rejects it if it accesses any table outside the allowed list.
+
 
 ## Example
 
--- a/docs/en/resources/tools/bigquery/bigquery-get-dataset-info.md
+++ b/docs/en/resources/tools/bigquery/bigquery-get-dataset-info.md
@@ -15,10 +15,19 @@ It's compatible with the following sources:
 
 - [bigquery](../../sources/bigquery.md)
 
-`bigquery-get-dataset-info` takes a `dataset` parameter to specify the dataset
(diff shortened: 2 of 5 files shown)


internal/sources/cloudsqlmssql/cloud_sql_mssql.go
```diff
@@ -49,16 +49,15 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
 
 type Config struct {
 	// Cloud SQL MSSQL configs
-	Name      string         `yaml:"name" validate:"required"`
-	Kind      string         `yaml:"kind" validate:"required"`
-	Project   string         `yaml:"project" validate:"required"`
-	Region    string         `yaml:"region" validate:"required"`
-	Instance  string         `yaml:"instance" validate:"required"`
-	IPAddress string         `yaml:"ipAddress" validate:"required"`
```
averikitsch, line 57: Should we keep the field listed here but deprecate it (similar to AuthSources) so it's not a breaking change?

internal/sources/cloudsqlmssql/cloud_sql_mssql.go
```diff
@@ -125,7 +124,6 @@ func initCloudSQLMssqlConnection(ctx context.Context, tracer trace.Tracer, name,
 	url := &url.URL{
 		Scheme:   "sqlserver",
 		User:     url.UserPassword(user, pass),
-		Host:     ipAddress,
```
averikitsch, line 128: I assume we don't need to set a default to localhost, right?

Yuan325: Doesn't seem like it. I tested this and it worked without a host

internal/sources/cloudsqlmssql/cloud_sql_mssql.go
Yuan325: Removed the required flag but kept this field so validator won't throw an error. This variable will not be used in the code.

averikitsch (approved):

PR #1827, opened 2025-10-30
feat(serverless-spark): add cancel-batch tool

dborowitz:
## Description

Another tool for the new serverless-spark source, this time to cancel a running batch. Straightforward implementation, although a little trickier (and slower) integration testing since we need to a) create a batch before we cancel it, b) block until the async cancel operation completes.

## PR Checklist

- [ ] Make sure you reviewed

🛠️ Part of #1689

### DIFF
--- a/internal/tools/serverlessspark/serverlesssparkcancelbatch/serverlesssparkcancelbatch.go
+++ b/internal/tools/serverlessspark/serverlesssparkcancelbatch/serverlesssparkcancelbatch.go
@@ -1,0 +1,162 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


internal/tools/serverlessspark/serverlesssparkcancelbatch/serverlesssparkcancelbatch.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, fmt.Errorf("failed to get operations client: %w", err)
+	}
+
+	paramMap := params.AsMap()
+	operation, ok := paramMap["operation"].(string)
+	if !ok {
+		return nil, fmt.Errorf("missing required parameter: operation")
+	}
+
+	if strings.Contains(operation, "/") {
```
duwenxin99, line 127: Why do we specifically check for '/'?

dborowitz: This is the same issue as in #1783, where I just now realized you maybe didn't review the final snapshot.

Basically I want this behavior:
1. user passes `"operation": "projects/my-project/operations/abcd"` -> fail with a deterministic error message
2. user passes `"operation": "abcd"` -> succeed

The issue I had with the tests in #1783 was that, for some reason I gave up on figuring out, the service API may return two different error messages for (1). Eagerly checking the input for `/` was a simple way to make the error message deterministic.

I'm open to other suggestions, though.

duwenxin99 (approved):

PR #1860, opened 2025-11-05
feat(tools/looker-run-look): modify run_look to show query origin

drstrangelooker:
… as coming from MCP Toolbox

### DIFF
--- a/docs/en/getting-started/quickstart/python/adk/requirements.txt
+++ b/docs/en/getting-started/quickstart/python/adk/requirements.txt
@@ -1,3 +1,3 @@
-google-adk==1.15.1
+google-adk==1.18.0
 toolbox-core==0.5.2
 pytest==8.4.2


internal/tools/looker/lookerrunlook/lookerrunlook.go
```diff
@@ -126,17 +126,30 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken
 
 	look_id := paramsMap["look_id"].(string)
 	limit := int64(paramsMap["limit"].(int))
+	limitStr := fmt.Sprintf("%d", limit)
```
averikitsch, line 129: nit: you may be able to simplify to  `limitStr := fmt.Sprintf("%d", paramsMap["limit"])`

averikitsch (approved):

PR #1863, opened 2025-11-05
fix: remove custom promptset support

twishabansal:
## Description

## PR Checklist

- [x] Make sure you reviewed

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -101,7 +101,6 @@ import (
 	_ "github.com/googleapis/genai-toolbox/internal/tools/firestore/firestoreupdatedocument"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/firestore/firestorevalidaterules"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/http"
-	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookercreateprojectfile"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookeradddashboardelement"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookerconversationalanalytics"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookercreateprojectfile"


cmd/root.go
```diff
@@ -530,14 +519,14 @@ func handleDynamicReload(ctx context.Context, toolsFile ToolsFile, s *server.Ser
 		panic(err)
 	}
 
-	sourcesMap, authServicesMap, toolsMap, toolsetsMap, promptsMap, promptsetsMap, err := validateReloadEdits(ctx, toolsFile)
+	sourcesMap, authServicesMap, toolsMap, toolsetsMap, promptsMap, _, err := validateReloadEdits(ctx, toolsFile)
```
Yuan325, line 525: Correct me if I'm wrong~ I believe that we need to keep this. Since there will be a default promptset with name "" that will be used for retrieving prompts.

cmd/root.go
```diff
@@ -530,14 +519,14 @@ func handleDynamicReload(ctx context.Context, toolsFile ToolsFile, s *server.Ser
...
 	if err != nil {
 		errMsg := fmt.Errorf("unable to validate reloaded edits: %w", err)
 		logger.WarnContext(ctx, errMsg.Error())
 		return err
 	}
 
-	s.ResourceMgr.SetResources(sourcesMap, authServicesMap, toolsMap, toolsetsMap, promptsMap, promptsetsMap)
+	s.ResourceMgr.SetResources(sourcesMap, authServicesMap, toolsMap, toolsetsMap, promptsMap, nil)
```
Yuan325, line 532: we'll probaly need the promptsets map

cmd/root.go
Yuan325: Please check and make sure that the default promptset is still generated both during the start of a NewServer, and during dynamic reloading.

twishabansal: Please see https://github.com/googleapis/genai-toolbox/pull/1863#discussion_r2509378338

cmd/root.go
twishabansal: Currently, the way toolsets and promptsets are designed, in this line, only toolsets or promptsets defined explicitly in the yaml file are passed here. 

The default toolset is created in the [server.go](https://github.com/googleapis/genai-toolbox/blob/3f0a57be58e87fc726373ab901f96c9ec6ab31a1/internal/server/server.go#L260) file which would contain all the tools, and the same goes for prompts.

Therefore, using nil here means that the user defined promptset is nil. Dynamic reloading would anyway occur at the prompt level (in root.go) and then the default promptset will be constructed in server.go (Just as for tools).

I have tested this out for both cases and it works. I am also adding comments in the code to clarify this behaviour. Let me know if it makes sense or if you would like a change in behaviour.

Yuan325: But here, in if we send promptset as `nil` to `s.ResourceMgr.SetResources()`, we will be setting `r.promptsets = nil`.

And when we use `s.ResourceMgr.GetPromptset(promptsetName)` (https://github.com/googleapis/genai-toolbox/blob/9c839c7530e2b82c176eff9d312b0647796ef566/internal/server/mcp.go#L518), it will throw error of promptset does not exist.

just tested it, and I did receive the promptset does not exist error.

The log will still say "initialized 1 promptset", but because we ignore that value (line 525), it will not get set to the resource manager.

twishabansal: Seems like I might have forgotten to restart the server on code change. Thanks for pointing out. I have fixed this.

PR #1864, opened 2025-11-05
chore: embed Config into Source

Yuan325:
To keep a persistent backend storage for configuration, we will have to keep a single source of truth. This involves supporting bi-directional conversion between Config and Source.

This PR make the following changes:
* Embed Config in Source
* Add `ToConfig()` to extract Config from Source.

### DIFF
--- a/internal/sources/cloudhealthcare/cloud_healthcare.go
+++ b/internal/sources/cloudhealthcare/cloud_healthcare.go
@@ -119,17 +119,12 @@ func (c Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
 		allowedDICOMStores[store] = struct{}{}
 	}
 	s := &Source{
-		name:               c.Name,
-		kind:               SourceKind,
-		project:            c.Project,
-		region:             c.Region,
-		dataset:            c.Dataset,
+		Config:             c,
 		service:            service,
 		serviceCreator:     serviceCreator,
 		tokenSource:        tokenSource,
 		allowedFHIRStores:  allowedFHIRStores,
 		allowedDICOMStores: allowedDICOMStores,
-		useClientOAuth:     c.UseClientOAuth,
 	}
 	return s, nil
 }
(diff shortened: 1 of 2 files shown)


internal/sources/cloudhealthcare/cloud_healthcare.go
```diff
@@ -259,5 +253,5 @@ func (s *Source) IsDICOMStoreAllowed(storeID string) bool {
 }
 
 func (s *Source) UseClientAuthorization() bool {
-	return s.useClientOAuth
+	return s.UseClientOAuth
```
averikitsch, line 256: Does this need to be from s.Config.UseClientOauth?

internal/sources/cloudhealthcare/cloud_healthcare.go
```diff
@@ -187,33 +182,32 @@ func initHealthcareConnection(ctx context.Context, tracer trace.Tracer, name str
...
 }
 
 func (s *Source) Region() string {
-	return s.region
+	return s.Config.Region
 }
 
 func (s *Source) DatasetID() string {
-	return s.dataset
+	return s.Dataset
```
averikitsch, line 210: Does this need to be s.Config.Dataset?

internal/sources/spanner/spanner.go
```diff
@@ -64,33 +64,33 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
+	return s.Config
+}
+
 func (s *Source) SpannerClient() *spanner.Client {
 	return s.Client
 }
 
 func (s *Source) DatabaseDialect() string {
-	return s.Dialect
+	return s.Dialect.String()
```
averikitsch, line 93: Should this be in config now?

internal/sources/cloudhealthcare/cloud_healthcare.go
Yuan325: Nope, since Config is embeded

internal/sources/cloudhealthcare/cloud_healthcare.go
Yuan325: Nope~

internal/sources/spanner/spanner.go
Yuan325: I think this can still be here, since this function is mainly for the `Source` struct instead of `Config` struct.

PR #1865, opened 2025-11-05
chore: embed Config into Auth

Yuan325:
To keep a persistent backend storage for configuration, we will have to keep a single source of truth. This involves supporting bi-directional conversion between Config and AuthService.

This PR make the following changes:
* Embed Config in AuthService
* Add `ToConfig()` to extract Config from AuthService.

### DIFF
--- a/internal/auth/google/google.go
+++ b/internal/auth/google/google.go
@@ -43,8 +43,7 @@ func (cfg Config) AuthServiceConfigKind() string {
 // Initialize a Google auth service
 func (cfg Config) Initialize() (auth.AuthService, error) {
 	a := &AuthService{
-		Config:   cfg,
-		ClientID: cfg.ClientID,
+		Config: cfg,
 	}
 	return a, nil
 }
@@ -54,7 +53,6 @@ var _ auth.AuthService = AuthService{}
 // struct used to store auth service info
 type AuthService struct {
 	Config
-	ClientID string `yaml:"clientId"`
 }
 
 // Returns the auth service kind


internal/auth/google/google.go
```diff
@@ -64,6 +62,10 @@ func (a AuthService) AuthServiceKind() string {
 	return AuthServiceKind
 }
 
+func (a AuthService) ToConfig() auth.AuthServiceConfig {
+	return a.Config
+}
+
 // Returns the name of the auth service
```
duwenxin99, line 67: This wouldn't work anymore right?

internal/auth/auth.go
```diff
@@ -30,4 +30,5 @@ type AuthService interface {
 	AuthServiceKind() string
 	GetName() string
```
duwenxin99, line 31: Do we need to remove the `GetName` method?

duwenxin99 (changes requested):

internal/auth/auth.go
Yuan325: Nope, GetName still work. Since Config is embed, all fields under `Config` will be accessible under `AuthService` as well.

internal/auth/google/google.go
Yuan325: This will still work~

duwenxin99 (approved):

internal/auth/google/google.go
```diff
@@ -54,8 +53,7 @@ var _ auth.AuthService = AuthService{}
 
 // struct used to store auth service info
 type AuthService struct {
-	Name     string `yaml:"name"`
-	Kind     string `yaml:"kind"`
+	Config
 	ClientID string `yaml:"clientId"`
```
averikitsch: Is ClientID not in config as well?

PR #1852, opened 2025-11-04
chore: introduce `orderedmap` to preserve column order in SQL results during marshal

banaras-be-coding:
This commit introduces a new `orderedmap` package to preserve the column order of SQL query results when they are marshaled to JSON.

The default Go `json.Marshal` function sorts map keys, which was causing the column order to be lost in the output of the database tools.

This commit updates the following tools to use the new `orderedmap` package:
- `mysqlexecutesql`
- `mssqlexecutesql`
- `postgresexecutesql`
- `spannerexecutesql`
- `sqliteexecutesql`
- `bigqueryexecutesql`

A new test has been added to the `mysqlexecutesql` tool to verify that the column order is preserved.

## Description

> Should include a concise description of the changes (bug or feature), it's
> impact, along with a summary of the solution

## PR Checklist

> Thank you for opening a Pull Request! Before submitting your PR, there are a
> few things you can do to make sure it goes smoothly:

- [ ] Make sure you reviewed
  [CONTRIBUTING.md](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md)
- [ ] Make sure to open an issue as a
  [bug/issue](https://github.com/googleapis/genai-toolbox/issues/new/choose)
  before writing your code! That way we can discuss the change, evaluate
  designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)
- [ ] Make sure to add `!` if this involve a breaking change

🛠️ Fixes #1492

PR #1884, opened 2025-11-06
fix: bigquery result column order

Yuan325:
Assigning job iterator value to an array instead of map to preserve column order.

When assigning incoming values to map, sometimes the result is not in order of statement. E.g. `SELECT id, name from ...` might turn into `{"name": "name_value", "id":1}` rather than `{"id":1, "name": "name_value"}`. Previously, during json marshaling, it will ALWAYS order the map in alphabetical order. so that wasn't an issue.

With the implementation of `orderedmap` (#1852), the bigquery execute sql tool will now preserves the column order during the marshaling process. Due to this, bigquery's integration test is flaky and failed when the map is reordered. This update assign incoming value as array instead, preserving the actual order.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -127,6 +127,7 @@ import (
 	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookerquery"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookerquerysql"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookerqueryurl"
+	_ "github.com/googleapis/genai-toolbox/internal/tools/looker/lookerrundashboard"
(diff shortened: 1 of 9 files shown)


Yuan325:
@averikitsch updated the PR description

Genesis929 (approved):

duwenxin99:
The other BQ tools seem to have the same issue: https://github.com/googleapis/genai-toolbox/blob/27663f0a1224706165d4d0052c1cee2231564a51/internal/tools/bigquery/bigquerysql/bigquerysql.go#L273 Can we fix them as well? Thanks!

Yuan325:
@duwenxin99 orderedmap is currently only implemented in execute sql, so other tools will not need this fix yet

PR #1897, opened 2025-11-07
Add DDL support for Spanner

banaras-be-coding:
## Description

## PR Checklist

- [ ] Make sure you reviewed

🛠️ Fixes #52


internal/tools/spanner/spannerexecutesql/spannerexecutesql_test.go
Yuan325, line 1: We should add the integration test in `/tests/` folder instead. This test will be running in our unit test environment, so there won't be a local mysql server running (which means this added test will always be skipped).

Yuan325 (changes requested):

internal/tools/spanner/spannerexecutesql/spannerexecutesql.go
```diff
@@ -155,6 +173,28 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken
 	}
 	logger.DebugContext(ctx, "executing `%s` tool query: %s", kind, sql)
 
+	isDDL := strings.HasPrefix(strings.ToUpper(sql), "CREATE") ||
```
Yuan325, line 176: Is it possible for user to use DDL batching?

For example: 
```
START BATCH DDL;
CREATE TABLE NewTable1 (Id INT64 NOT NULL) PRIMARY KEY (Id);
CREATE TABLE NewTable2 (Name STRING(MAX) NOT NULL) PRIMARY KEY (Name);
RUN BATCH;
```

internal/tools/spanner/spannerexecutesql/spannerexecutesql_test.go
```diff
@@ -12,18 +12,40 @@
...
+	"github.com/googleapis/genai-toolbox/internal/tools"
+	"github.com/googleapis/gax-go/v2"
+	databasepb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
 )
 
+type mockDatabaseAdminClient struct {
+	spannerDatabaseAdminClient
+	updateDatabaseDdl func(context.Context, *databasepb.UpdateDatabaseDdlRequest, ...gax.CallOption) (*database.UpdateDatabaseDdlOperation, error)
+	close func() error
+}
```
Yuan325, line 36:
```suggestion
var _ spannerDatabaseAdminClient = mockDatabaseAdminClient{}

type mockDatabaseAdminClient struct {
	UpdateDatabaseDdl func(context.Context, *databasepb.UpdateDatabaseDdlRequest, ...gax.CallOption) (*database.UpdateDatabaseDdlOperation, error)
	Close func() error
}
```

In order to implement the interface with mockDatabaseAdminClient, we can use a validation. The mock have to implement the exact function name (`Update...` instead of `update...`).

internal/tools/spanner/spannerexecutesql/spannerexecutesql_test.go
```diff
@@ -12,18 +12,40 @@
...
+
+func (m *mockDatabaseAdminClient) UpdateDatabaseDdl(ctx context.Context, req *databasepb.UpdateDatabaseDdlRequest, opts ...gax.CallOption) (*database.UpdateDatabaseDdlOperation, error) {
+	return m.updateDatabaseDdl(ctx, req, opts...)
```
Yuan325, line 39:
```suggestion
	return nil, errors.New("test error")
```

We can just return nil if this is the mock behavior we want.

internal/tools/spanner/spannerexecutesql/spannerexecutesql_test.go
```diff
@@ -12,18 +12,40 @@
...
+}
+
+func (m *mockDatabaseAdminClient) Close() error {
+	if m.close != nil {
+		return m.close()
+	}
```
Yuan325, line 45:
```suggestion
```

Same with this as well.

internal/tools/spanner/spannerexecutesql/spannerexecutesql_test.go
```diff
@@ -91,5 +113,64 @@ func TestParseFromYamlExecuteSql(t *testing.T) {
...
+		{
+			desc:          "ddl execution in read-only mode",
+			params:        tools.ParamValues{{Name: "sql", Value: "ALTER TABLE Singers ADD COLUMN FirstName STRING(1024)"}},
+			readOnly:      true,
+			expectedError: "cannot execute DDL statements in read-only mode",
+		},
+		{
+			desc: "ddl execution error",
+			adminClient: &mockDatabaseAdminClient{
+				updateDatabaseDdl: func(ctx context.Context, req *databasepb.UpdateDatabaseDdlRequest, opts ...gax.CallOption) (*database.UpdateDatabaseDdlOperation, error) {
```
Yuan325, line 144: will not need these, since we directly `return nil, errors.New("test error")` on the mock

internal/tools/spanner/spannerexecutesql/spannerexecutesql_test.go
```diff
@@ -91,5 +113,64 @@ func TestParseFromYamlExecuteSql(t *testing.T) {
...
+					return nil, errors.New("test error")
+				},
+			},
+			params:        tools.ParamValues{{Name: "sql", Value: "ALTER TABLE Singers ADD COLUMN FirstName STRING(1024)"}},
+			expectedError: "error executing DDL statement: test error",
+		},
```
Yuan325, line 150: can we also add a successful test for DDL statement? (case where it returns `"DDL statement executed successfully"`)

Yuan325 (changes requested):

Yuan325:
@TanmayVartak This is to provide a more secure `executeSQL` tool in cases where we want to limit the tool towards non DDL statements. Do you have recommendations to define DDL vs non-DDL statement other than string matching?

ref: #435

averikitsch:
Closing for now due to no current updates.

PR #1903, opened 2025-11-07
ci: Add CGO cross compilation support to release pipeline

duwenxin99:
- Support CGO cross compilation for multiple architectures using Zig.
- Download and link MacOSX SDK as needed by the MacOSX cross compilation. There is no official release for MacOSC SDK so I had to download from a third party repo.
- Update dockerfile from using `gcr.io/distroless/static:nonroot` to `gcr.io/distroless/cc-debian12:nonroot` for C libraries that is needed for dynamic linking.

### DIFF
--- a/.ci/continuous.release.cloudbuild.yaml
+++ b/.ci/continuous.release.cloudbuild.yaml
@@ -303,6 +303,6 @@ options:
 substitutions:
   _REGION: us-central1
   _AR_HOSTNAME: ${_REGION}-docker.pkg.dev
-  _AR_REPO_NAME: test
-  _BUCKET_NAME: test-cross-comp
+  _AR_REPO_NAME: toolbox-dev
+  _BUCKET_NAME: genai-toolbox-dev
   _DOCKER_URI: ${_AR_HOSTNAME}/${PROJECT_ID}/${_AR_REPO_NAME}/toolbox
--- a/.ci/versioned.release.cloudbuild.yaml
+++ b/.ci/versioned.release.cloudbuild.yaml
@@ -350,7 +350,7 @@ options:
 substitutions:
   _REGION: us-central1
   _AR_HOSTNAME: ${_REGION}-docker.pkg.dev
-  _AR_REPO_NAME: test
-  _BUCKET_NAME: test-cross-comp
+  _AR_REPO_NAME: toolbox
+  _BUCKET_NAME: genai-toolbox
   _DOCKER_URI: ${_AR_HOSTNAME}/${PROJECT_ID}/${_AR_REPO_NAME}/toolbox
-  _PUSH_LATEST: "false"
+  _PUSH_LATEST: "false" # Substituted in trigger


.ci/continuous.release.cloudbuild.yaml
```diff
@@ -95,32 +147,57 @@ steps:
...
-          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.commitSha=$(git rev-parse --short HEAD)" -o toolbox.darwin.amd64
+        SDK_PATH="/macos-sdk/MacOSX14.5.sdk"
+        export MACOS_MIN_VER=10.14
+
+        CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 \
+        CGO_LDFLAGS="-mmacosx-version-min=${MACOS_MIN_VER} --sysroot ${SDK_PATH} -F${SDK_PATH}/System/Library/Frameworks -L/usr/lib" \
+        COMMON_FLAGS="-mmacosx-version-min=${MACOS_MIN_VER} -target x86_64-macos.11.0.0-none -isysroot ${SDK_PATH} -iwithsysroot /usr/include -iframeworkwithsysroot /System/Library/Frameworks" \
+        CC="/zig-tools/zig cc ${COMMON_FLAGS}" \
+        CXX="/zig-tools/zig c++ ${COMMON_FLAGS}" \
+         go build -trimpath -buildmode=pie -ldflags "-s -w -X github.com/googleapis/genai-toolbox/cmd.commitSha=$(git rev-parse --short HEAD)" -o toolbox.darwin.amd64
```
kurtisvg: nit: should this be a script or something? We are calling it in 2 places.

.ci/continuous.release.cloudbuild.yaml
```diff
@@ -132,31 +209,48 @@ steps:
...
+        path: '/zig-tools'
+      - name: 'macos-sdk'
+        path: '/macos-sdk'
     script: |
         #!/usr/bin/env bash
-        export VERSION=$(cat ./cmd/version.txt)
-        CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 \
-          go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.commitSha=$(git rev-parse --short HEAD)" -o toolbox.geminicli.darwin.amd64
+        SDK_PATH="/macos-sdk/MacOSX14.5.sdk"
+        export MACOS_MIN_VER=10.14
```
kurtisvg: nit: do we need to consider bumping this? Or can it stay the same forever?

.ci/continuous.release.cloudbuild.yaml
```diff
@@ -132,31 +209,48 @@ steps:
...
       - name: 'go'
         path: '/gopath'
+      - name: 'zig'
+        path: '/zig-tools'
     script: |
         #!/usr/bin/env bash
-        CGO_ENABLED=0 GOOS=windows GOARCH=amd64 \
+        CGO_ENABLED=1 GOOS=windows GOARCH=amd64 \
+        CC="/zig-tools/zig cc -target x86_64-windows-gnu" \
+        CXX="/zig-tools/zig c++ -target x86_64-windows-gnu" \
```
kurtisvg: nit: should these env vars be under `env`?

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -240,7 +338,7 @@ options:
 substitutions:
   _REGION: us-central1
   _AR_HOSTNAME: ${_REGION}-docker.pkg.dev
-  _AR_REPO_NAME: toolbox
-  _BUCKET_NAME: genai-toolbox
+  _AR_REPO_NAME: test
+  _BUCKET_NAME: test-cross-comp
   _DOCKER_URI: ${_AR_HOSTNAME}/${PROJECT_ID}/${_AR_REPO_NAME}/toolbox
-  _PUSH_LATEST: "true"
+  _PUSH_LATEST: "false"
```
kurtisvg: nit: look into if this should be default false (so no one pushes by accident) and override in the trigger.

kurtisvg (approved):

.ci/continuous.release.cloudbuild.yaml
duwenxin99: We only need to bump it (together with the macos SDK used to compile the binary) if we ever introduce a dependency that relies on the newer sdk version. Otherwise we can keep using this stable sdk version forever.

.ci/versioned.release.cloudbuild.yaml
duwenxin99: updated the default to false

.ci/continuous.release.cloudbuild.yaml
duwenxin99: moved to env

Issue #1768, opened 2025-10-22
Add a new source - Maria DB


georgie76:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

The existing *mysql* source kind is not comaptible with Maria Db.
We need to add the new source kind: **"mariadb"**.

duwenxin99:
Hi @georgie76, we are welcoming contributions for new data sources. Please refer to this guide for contribution [instructions](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md#adding-a-new-database-source-or-tool).

pkalsi97:
@duwenxin99 I have already started working on this because I also need this feature, can I directly raise the PR or I'll have to submit a proposal here ? If I have to just let me know. Thanks.

duwenxin99:
> [@duwenxin99[<img alt="" width="15" height="15" src="chrome-extension://dlebflppeeemcdpidccbiblndppbmjmh/ospo-chrome-ext-logo.png">](https://teams.googleplex.com/duwenxin@google.com)](https://github.com/duwenxin99) I have already started working on this because I also need this feature, can I directly raise the PR or I'll have to submit a proposal here ? If I have to just let me know. Thanks.

Hi @pkalsi97, is Maria DB compatible with MySQL? Have you tried using the current MySQL source and tools to connect to a Maria DB?

pkalsi97:
@duwenxin99 

**Few points** 
1. Both use the same protocol so the implementation of source is very similar. Clients, drivers and database/sql connections treat them the same. 
2. MySQL and MariaDB are compatible for simple SQL queries, they differ once you query metadata such as information_schema, column metadata, table stats.
3. Analysis of existing tooling 
- `mysql-execute-sql` -> Will work for MariaDB
- `mysql-sql` -> Maybe if SQL template is same 
For other 4 tools, they depend on MySQL-specific metadata structures and system tables that do not work with MariaDB.


I am working on it, I have set up some tools for MariaDB and I am figuring out integration test for them but i plan to implement these tools 
1. Execute arbitrary SQL
2. Template-based SQL execution
3. List Active Queries (Avoids information_schema.innodb_trx which MariaDB does not implement)
4. List Table Fragmentation
5. List Tables 
6. List Missing Unique Indexes

If this Looks good to you I'll continue working and most probably finish this by weekend and raise PR.

[...]

PR #1908, opened 2025-11-08
feat(sources/mariadb): add MariaDB integration docs and tests

### LINKED ISSUE #1712: [source] Maria DB source and tools
### Prerequisites

Hello,

Currently this tooldbox doesn't seem to support Maria DB. Could you let me know if you can add that? I tried used mysql but it didn't work. It could not create a connection pool. 

Let me know if you can add Maria DB support. 

Thank you very much!!

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

pkalsi97:
## Description
This PR 
1. Adds **MariaDB** as a Source  - Implementation is similar to **MySQL** source
2. Utilises pre implemented **MySQL** Tools
- `mysql-execute-sql`
- `mysql-list-active-queries`
- `mysql-list-table-fragmentation`
- `mysql-list-tables`
- `mysql-list-tables-missing-unique-indexes`
- `mysql-sql`
**Note:** After discussion with @duwenxin99 in issue #1768, I initially assumed MariaDB required new tools due to different metadata structures and system tables. That is true for older MariaDB versions, but current MySQL tooling already works with MariaDB (verified), so a separate tool set was not needed. 

3. Adds a source doc for **MariaDB** in docs 
4. Adds MariaDB integration tests using the existing MySQL test flow.
Note: The test file is based on the MySQL integration test, but `GetMariaDBWants()` and
`RunMariDBListTablesTest()` are implemented because MariaDB returns different metadata
and list-tables output, so the assertions must be MariaDB-specific.
5. Updates CI 

Lastly
I considered adding a MariaDB-exclusive Galera cluster monitoring tool, but skipped it because it requires a multi-node Galera setup for integration testing and would significantly increase CI complexity with unclear usage demand.

## PR Checklist
- [x] Make sure you reviewed

🛠️ Fixes #1712 #1768

### DIFF
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -903,7 +903,8 @@ steps:
     args:
       - -c
       - |
-        go test ./tests/mariadb //skip coverage check as it re-uses current MySQL implementation
+        # skip coverage check as it re-uses current MySQL implementation
+        go test ./tests/mariadb
 
 
 availableSecrets:


duwenxin99:
Hi @pkalsi97, the PR seems good to me. Could you provide a link to a free version (preferably docker image) of the MariaDB that we can use in the CI/CD pipeline? I'll set up the test instance and come back to trigger the integration test on this PR. Meanwhile, just asking in case of potential GTMs: are you from the MariaDB team?

pkalsi97:
@duwenxin99 Thanks for the review!

You can run docker pull on this:
docker.io/library/mariadb:latest

No I am not from mariaDB team.

pkalsi97:
@duwenxin99 Lint and unit tests were failing due to the upstream change adding `ToConfig`, which wasn’t yet implemented in the MariaDB source. That part is fixed now.

I’m not sure why the integration tests are failing. I suspect it was triggered by the same upstream changes.

duwenxin99:
@pkalsi97 There seems to be a connection issue to the test instance. I'll debug it and get back to you.

pkalsi97:
@duwenxin99 Thanks, Let me know if you need anything from my side.

tests/tool.go
```diff
@@ -3189,3 +3189,194 @@ func RunRequest(t *testing.T, method, url string, body io.Reader, headers map[st
 	defer resp.Body.Close()
 	return resp, respBody
 }
+
```
duwenxin99, line 3458: Can we move this to the `mariadb_integration_test.go` since it's not used by other sources? Thanks.

tests/common.go
```diff
@@ -989,3 +989,12 @@ func CleanupMSSQLTables(t *testing.T, ctx context.Context, pool *sql.DB) {
 	}
 
 }
+
+// GetMariaDBWants return the expected wants for mariaDB
```
duwenxin99: Same here. Move it to the mariadb test file.

duwenxin99 (commented):
Hi @pkalsi97, the integration tests are now passing. I saw that there are different several parameters used for connecting to MariaDB vs. MySQL but they use the same driver. What differentiate the MariaDB source from the MySQL source? As the latest MariaDB fully supports the MySQL toolsets, I assume there won't be new tools added specifically for the MariaDB source? I'm concerned about the long term maintenance of the MariaDB source. If it is completely the same as the MySQL source, then users could choose to use the MySQL source instead.

pkalsi97:
@duwenxin99, Thanks for making the Integration test work!

I agree, sources share a wire protocol, so the MySQL source works with MariaDB. We can definitely remove this source and update the MariaDB documentation to state explicitly that it is supported through the MySQL data source.

Regarding tools, The goal was to provide reasonable overlap since **MySQ**L and **MariaDB** differ substantially in metadata and engine support. 

I had plan to support (metadata based tools)
1. Sequence objects tool
2. Galera Cluster tool

Since actual demand around metadata specific tooling is unclear right now, we can skip it for now. 
If future requests for these tools arise, you can ping me in the issue and I’ll help or implement based on availability.

#### ToDos:
1. Update MariaDB docs - state explicitly that it is supported through the MySQL data source.
2. Remove the MariaDB source 
3. Update integration test to initialise using `initMySQLConnectionPool` and remove `initMariaDB`

duwenxin99:
Hi @pkalsi97, can we update the MariaDB doc to make it clear to the users that they should be using the MySQL source and include link to redirect to the MySQL source page? Thanks! For the integration test, I think we can skip it for now if there are no new tools added since it would just be exactly the same as the MySQL test. If we add in MariaDB-specific tools in the future, we can then include the integration test in this PR.
Maybe we should open a new PR instead of modifying the current one so that we can save the code change in case we want to add in MariaDB-specific tools later. How does that sound to you?

pkalsi97:
@duwenxin99 sure
Sure, I’ll update the docs and remove the source file along with its usage in the MySQL tools.

I prefer keeping the integration test since it diverges slightly from the MySQL version, even with the current toolset. I leave it to your better judgement. Please let me know. I'll do accordingly.

I kept this PR medium-sized to keep the review manageable, but I definitely plan to add `Sequence objects tool`.

duwenxin99:
> @duwenxin99 sure Sure, I’ll update the docs and remove the source file along with its usage in the MySQL tools.
> 
> I prefer keeping the integration test since it diverges slightly from the MySQL version, even with the current toolset. I leave it to your better judgement. Please let me know. I'll do accordingly.
> 
> I kept this PR medium-sized to keep the review manageable, but I definitely plan to add `Sequence objects tool`.

@pkalsi97, that sounds good to me. Let's keep the integration tests so that it also makes things easier when you add in a sequence objects tool later. Please tag me again when you updated this PR so I can help you merge it in. Thanks again!

pkalsi97:
@duwenxin99 I have updated the docs, removed the source and updated integration test.

Just curious, does the CI internally grants `PROCESS` privilege required by `RunMySQLListActiveQueriesTest` to `MySQL` and `MariaDB` container internally ?

pkalsi97:
@duwenxin99 PTAL

duwenxin99:
> @duwenxin99 I have updated the docs, removed the source and updated integration test.
> 
> Just curious, does the CI internally grants `PROCESS` privilege required by `RunMySQLListActiveQueriesTest` to `MySQL` and `MariaDB` container internally ?

Hi @pkalsi97, sorry for the late reply as I just came back from Thanksgiving vacation. I believe the privilege should be granted unless it requires special setup for the root user.

pkalsi97:
@duwenxin99 can we run the ci again ?

pkalsi97:
@duwenxin99 Thanks! 😄

Issue #1705, opened 2025-10-13
Migrate the ADK portion of the Python Quickstart to asynchronous toolbox client.


vladkol:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

Instructions on [Introduction](https://googleapis.github.io/genai-toolbox/getting-started/introduction/) still use `ToolboxSyncClient`, and so does [example in ADK docs](https://google.github.io/adk-docs/tools/google-cloud-tools/#toolbox-tools-for-databases).
Code snippet on [Deploy to Cloud Run](https://googleapis.github.io/genai-toolbox/how-to/deploy_toolbox/) page abruptly ends after creating the client.

At the same time, there is a number closed issues (in [this repo](https://github.com/google/adk-python/issues/28), and in [ADK repo](https://github.com/google/adk-python/issues/28)) that highlight the same problem.

Here is the code snippet with an example:

```python
from toolbox_core import ToolboxClient
from google.adk.agents import Agent

async with ToolboxClient(
    URL
) as toolbox:
    tools = await toolbox.load_toolset()
        root_agent = Agent(
            model='gemini-2.0-flash',
            name='assistant_agent',
            description='A helpful AI assistant.',
            instruction="You are a helpful assistant that answers user's questions.",
            tools=[*tools] # <-- this! 
        )
```

anubhav756:
Thanks for raising this issue and highlighting the inconsistency in our documentation. Following recent updates to the underlying libraries, I want to provide the reasoning behind it.

**Our official recommendation is to now standardize on using `ToolboxSyncClient` for all ADK-related examples.**

### Detailed Reasoning

Previously, we faced significant deployment issues (https://github.com/googleapis/mcp-toolbox-sdk-python/issues/341) to the agent engine with both clients, especially when using the common ADK pattern of a separate script that prepares the agent for the `adk run` or `adk web` commands.

*   With **`ToolboxSyncClient`**, users would encounter a `"cannot pickle '_contextvars.Context' object"` error.
*   With **`ToolboxClient`** (the async version), users would face `"no running event loop"` or `"'await' outside function"` errors. A similar issue was also filed in the `adk-python` repo.

The key distinction is that these errors primarily occurred when using the ADK framework's script-based method. When ADK was simply imported as a library into a custom web application (the method used in the quickstart), both clients generally worked without these deployment issues.

[...]

PR #1937, opened 2025-11-12
docs: Update the ADK quickstart sample

anubhav756:
Addresses https://github.com/googleapis/genai-toolbox/issues/1705

## Description

The current ADK quickstart example used the `ToolboxSyncClient` while most of the examples use the async `ToolboxClient`. This PR updates the quickstart to make it use the `ToolboxClient` and making it consistent with the other samples.

### DIFF
--- a/docs/en/getting-started/quickstart/python/adk/quickstart.py
+++ b/docs/en/getting-started/quickstart/python/adk/quickstart.py
@@ -3,7 +3,7 @@ from google.adk.runners import Runner
 from google.adk.sessions import InMemorySessionService
 from google.adk.artifacts.in_memory_artifact_service import InMemoryArtifactService
 from google.genai import types
-from toolbox_core import ToolboxSyncClient
+from toolbox_core import ToolboxClient
 
 import asyncio
 import os
@@ -14,7 +14,7 @@ api_key = os.environ.get("GOOGLE_API_KEY") or "your-api-key" # Set your API key 
 os.environ["GOOGLE_API_KEY"] = api_key
 
 async def main():
-  with ToolboxSyncClient("http://127.0.0.1:5000") as toolbox_client:
+  async with ToolboxClient("http://127.0.0.1:5000") as toolbox_client:
 
       prompt = """
         You're a helpful hotel assistant. You handle hotel searching, booking and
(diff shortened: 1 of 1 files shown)


averikitsch (approved):

anubhav756:
Closing this PR.

Earlier, we encountered deployment issues to the agent engine with both `ToolboxSyncClient` (used in the current example) and `ToolboxClient` (which this PR introduces). With `ToolboxSyncClient`, we saw a `"cannot pickle '_contextvars.Context' object"` error (https://github.com/googleapis/mcp-toolbox-sdk-python/issues/341), while `ToolboxClient` resulted in `"no running event loop"` (https://github.com/googleapis/mcp-toolbox-sdk-python/issues/341#issuecomment-3485005631) or `"'await' outside function"` errors (https://github.com/googleapis/mcp-toolbox-sdk-python/issues/341#issuecomment-3325770346). These issues were reproducible, and a bug was filed for the official `ToolboxToolset` example in the `adk-python` repo which had the same `"no event loop running"` error (https://github.com/google/adk-python/issues/3492).

Note that the ADK offers two main ways to create an agent: by importing ADK as a library into an existing application, or by creating a separate script that the ADK framework uses to prepare and run the agent. The quickstart sample in this PR uses the former method. Interestingly, the deployment issues with both clients only appeared with the latter method; the former method worked fine with both `ToolboxSyncClient` and `ToolboxClient`.

Given that, and a request to make our ADK samples async-first for consistency with other quickstarts (https://github.com/googleapis/genai-toolbox/issues/1705), I initially created this PR to switch to `ToolboxClient`.

However, the situation has changed with the latest ADK release (`google-adk` `v1.18.0`). The new release has fixed the deployment issue for `ToolboxSyncClient` when using the second, script-based agent creation method (https://github.com/googleapis/mcp-toolbox-sdk-python/issues/341#issuecomment-3530327715). Now, `ToolboxSyncClient` works with _both_ agent creation methods, while `ToolboxClient` only works with the first.

This makes `ToolboxSyncClient` the more robust and versatile choice. Users can take the current example (using the first method) and adapt it to the second method without running into deployment issues. This PR, if merged, would inadvertently lead users to face the same deployment problems we initially encountered.

Therefore, I'm closing this PR to keep the quickstart guide using `ToolboxSyncClient`. I have also update the original request for consistency, explaining why `ToolboxSyncClient` is now the preferred client for ADK (https://github.com/googleapis/genai-toolbox/issues/1705#issuecomment-3538791291).

anubhav756:
Quick update: We've updated the Python quickstart according to the latest and recommended way to get started with ADK and integrate Toolbox: https://github.com/googleapis/genai-toolbox/pull/1962

> [!NOTE]
> We use the `ToolboxSyncClient` because the async alternatives, `ToolboxClient` and `ToolboxToolset` from ADK, fail when deployed on Agent Engine, despite working locally. (https://github.com/google/adk-python/issues/3492, https://github.com/google/adk-python/issues/3532)

Issue #1779, opened 2025-10-23
Add Oracle OCI Driver Support for TNS Aliases and Wallet Authentication


cmedianu:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

## Description

Add a new Oracle source that uses the official Oracle OCI driver (godror) to enable connections to Oracle databases that require TNS aliases and wallet-based mutual TLS authentication.

## Motivation

The existing `oracle` source kind uses the go-ora driver (pure Go implementation), which is excellent for simple direct connections but lacks support for:
- TNS aliases from tnsnames.ora
- Oracle Wallet for mutual TLS authentication
- Mutual TLS (mTLS) connections required by Oracle Cloud databases

## Use Cases

This feature would enable users to connect to:
- **Oracle Autonomous Database** - Requires wallet-based mutual TLS authentication
- **Oracle Cloud Database services** - Often require mutual TLS for secure connections
- **TNS-configured deployments** - Enterprises using tnsnames.ora for connection management
- **Wallet-secured environments** - Any Oracle deployment using wallet authentication

### Suggested Solution(s)

Add a new `oracle-oci` source kind that:
- Uses the godror driver (official Oracle driver with full OCI support)
- Supports TNS aliases from tnsnames.ora
- Supports Oracle Wallet authentication for mutual TLS
- Supports all standard connection methods (TNS alias, connection string, host+port+serviceName)
- Requires Oracle Instant Client at runtime
- Uses build tags (`-tags oracleoci`) to remain optional and non-breaking

duwenxin99:
Hi @cmedianu, we actually originally had the Oracle sources implemented using godror but [switched to go-ora](https://github.com/googleapis/genai-toolbox/pull/1685) because our current release pipeline cannot handle cross-compilation. We do plan on switching back to godror for the Oracle source. However, we need to first tackle the cross-compilation challenge in our release pipeline before doing that.

cmedianu:
Makes sense, looking forward to the fix. I'll use my patch in the meantime... Thanks!

PR #1945, opened 2025-11-13
feat(sources/oracle): Add Oracle OCI and Wallet support

duwenxin99:
Previously we used go-ora (a pure Go Oracle driver) because our release pipeline did not support cross-compilation with CGO. Now that it's fixed, we want to add support for Oracle OCI driver for advanced features including digital wallet etc. 

Users will be able to configure a source to use OCI by specifying a `UseOCI: true` field. The source defaults to use the pure Go driver otherwise. 

Oracle Wallet:
- OCI users should use the `tnsAdmin` to set the wallet location
- Non-OCI users can should use the `walletLocation` field.

fix: https://github.com/googleapis/genai-toolbox/issues/1779

### DIFF
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -212,6 +212,26 @@ steps:
           bigquery \
           bigquery
 
+  - id: "cloud-gda"
+    name: golang:1
+    waitFor: ["compile-test-binary"]
+    entrypoint: /bin/bash
+    env:
+      - "GOPATH=/gopath"
+      - "CLOUD_GDA_PROJECT=$PROJECT_ID"
+      - "SERVICE_ACCOUNT_EMAIL=$SERVICE_ACCOUNT_EMAIL"
+    secretEnv: ["CLIENT_ID"]
+    volumes:
+      - name: "go"
+        path: "/gopath"
+    args:
+      - -c
+      - |
+        .ci/test_with_coverage.sh \
+          "Cloud Gemini Data Analytics" \
+          cloudgda \
+          cloudgda
+
   - id: "dataplex"
     name: golang:1
     waitFor: ["compile-test-binary"]
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -73,6 +73,7 @@ import (
 	_ "github.com/googleapis/genai-toolbox/internal/tools/clickhouse/clickhouselistdatabases"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/clickhouse/clickhouselisttables"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/clickhouse/clickhousesql"
+	_ "github.com/googleapis/genai-toolbox/internal/tools/cloudgda"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/cloudhealthcare/cloudhealthcarefhirfetchpage"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/cloudhealthcare/cloudhealthcarefhirpatienteverything"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/cloudhealthcare/cloudhealthcarefhirpatientsearch"
--- a/docs/en/resources/sources/cloud-gda.md
+++ b/docs/en/resources/sources/cloud-gda.md
@@ -1,0 +1,40 @@
+---
+title: "Gemini Data Analytics"
+type: docs
+weight: 1
+description: >
+  A "cloud-gemini-data-analytics" source provides a client for the Gemini Data Analytics API.
+aliases:
+  - /resources/sources/cloud-gemini-data-analytics
+---
+
+## About
+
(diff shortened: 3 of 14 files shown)


internal/sources/oracle/oracle.go
```diff
@@ -1,4 +1,3 @@
-// Copyright © 2025, Oracle and/or its affiliates.
```
averikitsch, line 1: Keep this copyright

averikitsch (commented):
Is this just reverting the changes required for the first release?

duwenxin99:
> Is this just reverting the changes required for the first release?

@averikitsch Not really. This adds a config field `UseOCI` that allows users to freely choose whether or not to use the OCI driver. The initial PR only used the OCI driver.

.ci/integration.cloudbuild.yaml
```diff
@@ -806,8 +806,8 @@ steps:
           cassandra
 
   - id: "oracle"
-    name: golang:1
-    waitFor: ["compile-test-binary"]
+    name: ghcr.io/oracle/oraclelinux8-instantclient:21
```
averikitsch: Can we continue to test with coverage? We could possiblyinstall these dependencies in a different step than the test runner

docs/en/resources/sources/oracle.md
```diff
@@ -15,23 +15,45 @@ description: >
...
+   installation, but it **lacks support for advanced features** like Oracle
+   Wallets or Kerberos authentication.
+
+2. **OCI-Based Driver (`useOCI: true`):** Uses the `godror` library, which
+   provides access to **advanced Oracle features** like Digital Wallet support.
+
+If you set `useOCI: true`, you **must** install the **Oracle Instant Client**
+libraries on the machine where this tool runs.
+
+You can download the Instant Client from the official Oracle website: [Oracle
```
averikitsch, line 51: We should ensure that any error message is clear for needing these packages

internal/sources/oracle/oracle.go
```diff
@@ -145,28 +156,39 @@ func initOracleConnection(ctx context.Context, tracer trace.Tracer, config Confi
...
-		config.User, config.Password, serverString)
+	var driverName string
+	var finalConnStr string
+
+	if config.UseOCI {
+		// Use godror driver (requires OCI/instant client)
+		driverName = "godror"
+		finalConnStr = fmt.Sprintf(`user="%s" password="%s" connectString="%s"`,
+			config.User, config.Password, connectStringBase)
+		logger.DebugContext(ctx, fmt.Sprintf("Using godror driver (OCI-based) with connectString: %s\n", connectStringBase))
```
averikitsch, line 185: nit: should we have password in the debug logs?

internal/tools/oracle/oracleexecutesql/oracleexecutesql.go
```diff
@@ -113,7 +113,7 @@ func (t Tool) Invoke(ctx context.Context, params tools.ParamValues, accessToken
 	if err != nil {
 		return nil, fmt.Errorf("error getting logger: %s", err)
 	}
-	logger.DebugContext(ctx, fmt.Sprintf("executing `%s` tool query: %s", kind, sqlParam))
+	logger.DebugContext(ctx, "executing `%s` tool query: %s", kind, sqlParam)
```
averikitsch, line 113: I don't think debugcontext handles the rendering of %s

.ci/integration.cloudbuild.yaml
duwenxin99: I was having problem getting Toolbox to connect to the downloaded OCI in the test step :(

internal/sources/oracle/oracle.go
duwenxin99: It's only logging the `connectStringBase` which doesn't include the username or password

internal/tools/oracle/oracleexecutesql/oracleexecutesql.go
duwenxin99: I just discovered this as well but the DebugContext method signature is 
`DebugContext(ctx context.Context, format string, args ...interface{})`
so it should be able to format strings

.ci/integration.cloudbuild.yaml
duwenxin99: I will spend some time trying it again

averikitsch (approved):

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -350,7 +350,7 @@ options:
...
   _REGION: us-central1
   _AR_HOSTNAME: ${_REGION}-docker.pkg.dev
-  _AR_REPO_NAME: toolbox
-  _BUCKET_NAME: genai-toolbox
+  _AR_REPO_NAME: test
+  _BUCKET_NAME: test-cross-comp
   _DOCKER_URI: ${_AR_HOSTNAME}/${PROJECT_ID}/${_AR_REPO_NAME}/toolbox
-  _PUSH_LATEST: "false" # Substituted in trigger
\ No newline at end of file
+  _PUSH_LATEST: "false" # Substituted in trigger
```
averikitsch: These should be reverted, right?

Issue #1691, opened 2025-10-10
Create custom tool kinds for Postgres prebuilt tools


averikitsch:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

Tools need to be converted from using postgres-sql to custom tool kinds for all postgres prebuilt toolsets.
* list_memory_configurations
* list_top_bloated_tables
* list_replication_slots
* list_invalid_indexes
* get_query_plan

averikitsch:
We need to assign this to @isaurabhuttam once added to the repo.

srividyareddy786:
We are planning to add more pre-built tools for PostgreSQL. Here are a few tools:
- list_databases
- list_schemas
- list_indexes
- list_sequences
- list_views
- list_roles_and_privileges
- list_roles_and_privileges

averikitsch:
Hi @srividyareddy786 , I am reopening this issue to ensure we convert tools like https://github.com/googleapis/genai-toolbox/blob/5abad5d56c6cc5ba86adc5253b948bf8230fa830/internal/prebuiltconfigs/tools/alloydb-postgres.yaml#L75 to prebuilt tools

PR #1976, opened 2025-11-18
feat: add list-query-stats and get-column-cardinality functions

ravisiddhu:
Adds the following tools for Postgre
(1) get_column_cardinality - to fetch the estimates of the distinct column counts in the table for particular column or all columns
(2) list-query-stats - to obtain the query level statistics in a database. This tool requires pg_stat_statements extension as a prerequisite.

image

image

PR Checklist
[Y] Make sure you reviewed
[Y] Make sure to open an issue as a
[Y] Ensure the tests and linter pass
[Y] Code coverage does not decrease (if any source code was changed)
[Y] Appropriate docs were updated (if necessary)
[Y] Make sure to add ! if this involve a breaking change

🛠️ Fixes https://github.com/googleapis/genai-toolbox/issues/1691

### DIFF
--- a/internal/tools/postgres/postgreslistquerystats/postgreslistquerystats.go
+++ b/internal/tools/postgres/postgreslistquerystats/postgreslistquerystats.go
@@ -113,7 +113,7 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 		cfg.Description = "Lists performance statistics for executed queries ordered by total time, filtering by database name pattern if provided. This tool requires the pg_stat_statements extension to be installed. The tool returns the database name, query text, execution count, timing metrics (total, min, max, mean), rows affected, and buffer cache I/O statistics (hits and reads)."
 	}
 
-	mcpManifest := tools.GetMcpManifest(cfg.Name, cfg.Description, cfg.AuthRequired, allParameters)
+	mcpManifest := tools.GetMcpManifest(cfg.Name, cfg.Description, cfg.AuthRequired, allParameters, nil)
 
 	// finish tool setup
 	return Tool{


internal/tools/postgres/postgresgetcolumncardinality/postgresgetcolumncardinality.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// verify the source is compatible
+	s, ok := rawS.(compatibleSource)
+	if !ok {
+		return nil, fmt.Errorf("invalid source for %q tool: source kind must be one of %q", kind, compatibleSources)
+	}
+
+	allParameters := parameters.Parameters{
+		parameters.NewStringParameterWithRequired("schema_name", "Required: The schema name in which the table is present.", true),
```
averikitsch: nit: do we want to set a default to public?

internal/tools/postgres/postgresgetcolumncardinality/postgresgetcolumncardinality.go
```diff
@@ -0,0 +1,203 @@
...
+		parameters.NewStringParameterWithRequired("table_name", "Required: The table name in which the column is present.", true),
+		parameters.NewStringParameterWithRequired("column_name", "Optional: The column name for which the cardinality is to be found. If not provided, cardinality for all columns will be returned.", false),
```
averikitsch, line 108: Should this just use WithDefault instead of WithRequired?

internal/tools/postgres/postgresgetcolumncardinality/postgresgetcolumncardinality.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse row: %w", err)
+		}
+		rowMap := make(map[string]any)
+		for i, field := range fields {
+			rowMap[string(field.Name)] = values[i]
+		}
+		out = append(out, rowMap)
+	}
+
```
averikitsch, line 181: I recommend adding https://github.com/ravisiddhu/genai-toolbox/blob/768bf01966229f0bdfc5a0d761dd04f7e3045c84/internal/tools/postgres/postgresexecutesql/postgresexecutesql.go#L150 to catch additional errors

internal/tools/postgres/postgresgetcolumncardinality/postgresgetcolumncardinality.go
ravisiddhu: Done.

internal/tools/postgres/postgresgetcolumncardinality/postgresgetcolumncardinality.go
ravisiddhu: I have set column name as required=false, as I do not want to give a default of empty string but rather to have a null value here if unset.

internal/tools/postgres/postgresgetcolumncardinality/postgresgetcolumncardinality.go
ravisiddhu: Done

Issue #1633, opened 2025-10-06
Add OpenTelemetry tracing for STDIO, MCP methods, and tool execution


ShubhamPalriwala:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

Critical gaps in OpenTelemetry instrumentation prevent production observability:

**1. Debug STDIO client connections (Claude Desktop, MCP Inspector)**
- STDIO transport has 0% instrumentation coverage
- Cannot trace connection lifecycle or message flow
- Cannot measure performance or see errors
- Primary desktop client transport is a complete blind spot

**2. Measure MCP protocol performance**
- Cannot distinguish between MCP methods (`initialize`, `tools/list`, `tools/call`, `ping`)
- MCP tool invocations **don't record metrics** while Native API does (`api.go:154`)
- Majority of production usage is invisible in monitoring dashboards

**3. Analyze tool execution performance**
- No per-tool tracing for all the third-party tools
- Cannot measure individual tool execution time
- Cannot identify slow performance blocker 
- Cannot trace external API calls from tools

**4. Track authentication failures**
- No tracing for token validation, claims extraction, or authorization checks
- Cannot measure auth overhead or debug auth-related failures

**Current Coverage**:
| Transport | Traces | Metrics | Coverage |
|-----------|--------|---------|----------|
| Native | ✅ | ✅ | 100% |
| HTTP/SSE | ⚠️ Partial | ⚠️ Partial | ~60% |
| STDIO | ❌ None | ❌ None | 0% |

**Impact**: Cannot effectively monitor, debug, or optimize production deployments. Server-side metrics are essential for SLOs/SLIs but currently missing for MCP flows.

### Suggested Solution(s)

Implement comprehensive OpenTelemetry instrumentation across Toolbox:

**STDIO Transport Instrumentation**
- Add connection lifecycle spans in `ServeStdio()`
- Add message read/write tracing in `readInputStream()`
- Add STDIO-specific metrics (connections, messages, size)

**MCP Method-Level Tracing**
- Pass instrumentation object to MCP method handlers
- Add spans for each method: `initialize`, `tools/list`, `tools/call`, `ping`
- **Critical fix**: Add `instrumentation.ToolInvoke.Add()` in `toolsCallHandler()` (currently missing)
- Apply consistently to all 3 MCP protocol versions (v20241105, v20250326, v20250618)

[...]

PR #1987, opened 2025-11-19
feat!: Update/add detailed telemetry for mcp endpoint compliant with OTEL semantic convention

AjmeraParth132:
## Description

This PR adds consistent and actionable telemetry for MCP sessions across HTTP and STDIO transports, enabling quick visibility into toolset discovery and tool invocation activity with minimal setup.

## PR Checklist

- [x] Make sure you reviewed  [CONTRIBUTING.md](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md)
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/genai-toolbox/issues/new/choose)
  before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

### DIFF
--- a/internal/sources/cockroachdb/cockroachdb.go
+++ b/internal/sources/cockroachdb/cockroachdb.go
@@ -34,14 +34,13 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-const SourceKind string = "cockroachdb"
 const SourceType string = "cockroachdb"
 
 var _ sources.SourceConfig = Config{}
 
 func init() {
-	if !sources.Register(SourceKind, newConfig) {
-		panic(fmt.Sprintf("source kind %q already registered", SourceKind))
+	if !sources.Register(SourceType, newConfig) {
+		panic(fmt.Sprintf("source type %q already registered", SourceType))
 	}
 }
 
@@ -94,10 +93,6 @@ type Config struct {
 	ClusterID        string `yaml:"clusterID"`        // Optional cluster identifier for telemetry
 }
 
-func (r Config) SourceConfigKind() string {
-	return SourceKind
-}
-
 func (r Config) SourceConfigType() string {
 	return SourceType
 }
@@ -125,10 +120,6 @@ var _ sources.Source = &Source{}
 type Source struct {
 	Config
 	Pool *pgxpool.Pool
-}
-
-func (s *Source) SourceKind() string {
-	return SourceKind
 }
 
 func (s *Source) SourceType() string {
@@ -379,7 +370,7 @@ func (s *Source) EmitTelemetry(ctx context.Context, event TelemetryEvent) {
 
 func initCockroachDBConnectionPoolWithRetry(ctx context.Context, tracer trace.Tracer, name, host, port, user, pass, dbname string, queryParams map[string]string, maxRetries int, baseDelay time.Duration) (*pgxpool.Pool, error) {
 	//nolint:all
-	ctx, span := sources.InitConnectionSpan(ctx, tracer, SourceKind, name)
+	ctx, span := sources.InitConnectionSpan(ctx, tracer, SourceType, name)
 	defer span.End()
 
 	userAgent, err := util.UserAgentFromContext(ctx)
--- a/internal/tools/cockroachdb/cockroachdbexecutesql/cockroachdbexecutesql.go
+++ b/internal/tools/cockroachdb/cockroachdbexecutesql/cockroachdbexecutesql.go
@@ -30,11 +30,11 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
-const kind string = "cockroachdb-execute-sql"
+const resourceType string = "cockroachdb-execute-sql"
 
 func init() {
-	if !tools.Register(kind, newConfig) {
-		panic(fmt.Sprintf("tool kind %q already registered", kind))
+	if !tools.Register(resourceType, newConfig) {
+		panic(fmt.Sprintf("tool type %q already registered", resourceType))
 	}
 }
 
@@ -50,7 +50,7 @@ type compatibleSource interface {
 	Query(ctx context.Context, sql string, args ...interface{}) (pgx.Rows, error)
 }
 
-var compatibleSources = [...]string{cockroachdb.SourceKind}
+var compatibleSources = [...]string{cockroachdb.SourceType}
 
 type Config struct {
 	Name         string   `yaml:"name" validate:"required"`
@@ -62,12 +62,8 @@ type Config struct {
 
 var _ tools.ToolConfig = Config{}
 
-func (cfg Config) ToolConfigKind() string {
-	return kind
-}
-
 func (cfg Config) ToolConfigType() string {
-	return kind
+	return resourceType
 }
 
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
@@ -78,7 +74,7 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 
 	_, ok = rawS.(compatibleSource)
 	if !ok {
-		return nil, fmt.Errorf("invalid source for %q tool: source kind must be one of %q", kind, compatibleSources)
+		return nil, fmt.Errorf("invalid source for %q tool: source type must be one of %q", resourceType, compatibleSources)
 	}
 
 	sqlParameter := parameters.NewStringParameter("sql", "The sql to execute.")
--- a/internal/tools/cockroachdb/cockroachdblistschemas/cockroachdblistschemas.go
+++ b/internal/tools/cockroachdb/cockroachdblistschemas/cockroachdblistschemas.go
@@ -29,7 +29,7 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
-const kind string = "cockroachdb-list-schemas"
+const resourceType string = "cockroachdb-list-schemas"
 
 const listSchemasStatement = `
 	SELECT
@@ -44,8 +44,8 @@ const listSchemasStatement = `
 `
 
 func init() {
-	if !tools.Register(kind, newConfig) {
-		panic(fmt.Sprintf("tool kind %q already registered", kind))
+	if !tools.Register(resourceType, newConfig) {
+		panic(fmt.Sprintf("tool type %q already registered", resourceType))
 	}
 }
 
@@ -63,7 +63,7 @@ type compatibleSource interface {
 
 var _ compatibleSource = &cockroachdb.Source{}
 
-var compatibleSources = [...]string{cockroachdb.SourceKind}
+var compatibleSources = [...]string{cockroachdb.SourceType}
 
 type Config struct {
 	Name         string   `yaml:"name" validate:"required"`
@@ -75,12 +75,8 @@ type Config struct {
 
 var _ tools.ToolConfig = Config{}
 
-func (cfg Config) ToolConfigKind() string {
-	return kind
-}
-
 func (cfg Config) ToolConfigType() string {
-	return kind
+	return resourceType
 }
 
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
@@ -91,7 +87,7 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 
 	_, ok = rawS.(compatibleSource)
 	if !ok {
-		return nil, fmt.Errorf("invalid source for %q tool: source kind must be one of %q", kind, compatibleSources)
+		return nil, fmt.Errorf("invalid source for %q tool: source type must be one of %q", resourceType, compatibleSources)
 	}
 
 	allParameters := parameters.Parameters{}
--- a/internal/tools/cockroachdb/cockroachdblisttables/cockroachdblisttables.go
+++ b/internal/tools/cockroachdb/cockroachdblisttables/cockroachdblisttables.go
@@ -29,7 +29,7 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
-const kind string = "cockroachdb-list-tables"
+const resourceType string = "cockroachdb-list-tables"
 
 const listTablesStatement = `
 	WITH desired_relkinds AS (
@@ -104,8 +104,8 @@ const listTablesStatement = `
 `
 
 func init() {
-	if !tools.Register(kind, newConfig) {
-		panic(fmt.Sprintf("tool kind %q already registered", kind))
+	if !tools.Register(resourceType, newConfig) {
+		panic(fmt.Sprintf("tool type %q already registered", resourceType))
 	}
 }
 
@@ -123,7 +123,7 @@ type compatibleSource interface {
 
 var _ compatibleSource = &cockroachdb.Source{}
 
-var compatibleSources = [...]string{cockroachdb.SourceKind}
+var compatibleSources = [...]string{cockroachdb.SourceType}
 
 type Config struct {
 	Name         string   `yaml:"name" validate:"required"`
@@ -135,12 +135,8 @@ type Config struct {
 
 var _ tools.ToolConfig = Config{}
 
-func (cfg Config) ToolConfigKind() string {
-	return kind
-}
-
 func (cfg Config) ToolConfigType() string {
-	return kind
+	return resourceType
 }
 
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
@@ -151,7 +147,7 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 
 	_, ok = rawS.(compatibleSource)
 	if !ok {
-		return nil, fmt.Errorf("invalid source for %q tool: source kind must be one of %q", kind, compatibleSources)
+		return nil, fmt.Errorf("invalid source for %q tool: source type must be one of %q", resourceType, compatibleSources)
 	}
 
 	allParameters := parameters.Parameters{
--- a/internal/tools/cockroachdb/cockroachdbsql/cockroachdbsql.go
+++ b/internal/tools/cockroachdb/cockroachdbsql/cockroachdbsql.go
@@ -30,11 +30,11 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
-const kind string = "cockroachdb-sql"
+const resourceType string = "cockroachdb-sql"
 
 func init() {
-	if !tools.Register(kind, newConfig) {
-		panic(fmt.Sprintf("tool kind %q already registered", kind))
+	if !tools.Register(resourceType, newConfig) {
+		panic(fmt.Sprintf("tool type %q already registered", resourceType))
 	}
 }
 
@@ -52,7 +52,7 @@ type compatibleSource interface {
 
 var _ compatibleSource = &cockroachdb.Source{}
 
-var compatibleSources = [...]string{cockroachdb.SourceKind}
+var compatibleSources = [...]string{cockroachdb.SourceType}
 
 type Config struct {
 	Name               string                `yaml:"name" validate:"required"`
@@ -67,12 +67,8 @@ type Config struct {
 
 var _ tools.ToolConfig = Config{}
 
-func (cfg Config) ToolConfigKind() string {
-	return kind
-}
-
 func (cfg Config) ToolConfigType() string {
-	return kind
+	return resourceType
 }
 
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
@@ -83,7 +79,7 @@ func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error)
 
 	_, ok = rawS.(compatibleSource)
 	if !ok {
-		return nil, fmt.Errorf("invalid source for %q tool: source kind must be one of %q", kind, compatibleSources)
+		return nil, fmt.Errorf("invalid source for %q tool: source type must be one of %q", resourceType, compatibleSources)
 	}
 
 	allParameters, paramManifest, err := parameters.ProcessParameters(cfg.TemplateParameters, cfg.Parameters)
--- a/tests/cockroachdb/cockroachdb_integration_test.go
+++ b/tests/cockroachdb/cockroachdb_integration_test.go
@@ -31,8 +31,8 @@ import (
 )
 
 var (
-	CockroachDBSourceKind = "cockroachdb"
-	CockroachDBToolKind   = "cockroachdb-sql"
+	CockroachDBSourceType = "cockroachdb"
+	CockroachDBToolType   = "cockroachdb-sql"
 	CockroachDBDatabase   = getEnvOrDefault("COCKROACHDB_DATABASE", "defaultdb")
 	CockroachDBHost       = getEnvOrDefault("COCKROACHDB_HOST", "localhost")
 	CockroachDBPort       = getEnvOrDefault("COCKROACHDB_PORT", "26257")
@@ -53,7 +53,7 @@ func getCockroachDBVars(t *testing.T) map[string]any {
 	}
 
 	return map[string]any{
-		"type":           CockroachDBSourceKind,
+		"type":           CockroachDBSourceType,
 		"host":           CockroachDBHost,
 		"port":           CockroachDBPort,
 		"database":       CockroachDBDatabase,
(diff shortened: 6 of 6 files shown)


Yuan325:
Hi @AjmeraParth132 , thank you for your contribution! I'll make sure to review this tomorrow. :)

In the meantime, please reopen this issue #1633. There's also some outstanding comments regarding the naming of the traces. Ideally we would want to come into an agreement on the names before making any official changes since updating trace name is considered breaking change~

AjmeraParth132:
Hi @Yuan325, thanks for the update. Sure, we'll just internally discuss the latest scope of work and reopen the updated issue quickly. Will inform you here once done.

internal/server/api.go
```diff
@@ -58,6 +58,7 @@ func toolsetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	toolsetName := chi.URLParam(r, "toolsetName")
 	s.logger.DebugContext(ctx, fmt.Sprintf("toolset name: %s", toolsetName))
 	span.SetAttributes(attribute.String("toolset_name", toolsetName))
+	span.SetAttributes(attribute.String("transport", "native"))
```
Yuan325:
```suggestion
	span.SetAttributes(attribute.String("endpoint", "native"))
```

What do you think about `endpoint` instead? if using transport, i might think of the transport protocol in mcp. Also, since the trace name for /api vs /mcp endpoint will be different, do we need this attribute?

Please include this update in the proposed design in #1633 as well

internal/server/mcp.go
```diff
@@ -143,12 +143,35 @@ func (s *stdioSession) readInputStream(ctx context.Context) error {
...
-		v, res, err := processMcpMessage(ctx, []byte(line), s.server, s.protocol, "", "", nil)
-		if err != nil {
+
+		// Parse the message to determine the trace name
+		var baseMessage jsonrpc.BaseMessage
+		traceName := "toolbox/server/mcp"
+		if err := json.Unmarshal([]byte(line), &baseMessage); err == nil && baseMessage.Method != "" {
+			switch baseMessage.Method {
+			case "tools/list":
+				traceName = "toolbox/server/toolset/get"
```
Yuan325: please refer to the comment (https://github.com/googleapis/genai-toolbox/issues/1633#issuecomment-3439457943) Will it be better to add it in the processMcpMessage() function? so that all transport does get the same child span? and since there's where the method function is processed (ideally we want to prevent checking for baseMessage.Method here).

AjmeraParth132:
@Yuan325 thanks for your review suggestions. 

- On the transport/endpoint convention, we started with connection-type, but Kurtis suggested transport in the review doc. Happy to rename it as endpoint if that feels more intuitive.
- Setting traceName, toolName and toolsetName attributes to processMcpMessage() seems a clear way - just to confirm, we might need to pass the span context for trace telemetry through the function's arguments further down, I hope that's fine?
- Meanwhile, I have opened a new issue - [#2222](https://github.com/googleapis/genai-toolbox/issues/2222) for tracking with latest set of actionables as discussed. 

Will rebase the PR and update here :)

Yuan325:
Hi @AjmeraParth132 

Some questions: (please advise on best practice within the field if there's any) -- 
1. Second thought --do we need the transport/endpoint attribute? since it can be determine through the trace name itself right?
2. Would like to clarify what you mean by span context? We can definitely add `ctx context.Context` to the functions if that's what you meant.

Please add the full visualization of the traces/attributes/metrics that you will be adding in that new issue -- if you're planning to just continue to work from there? (similar to https://github.com/googleapis/genai-toolbox/issues/1633#issuecomment-3375999151) It will be easier for future contributors to reference, and also a way to keep track of the changes that will be applied :) Thank you very much!! I'll try to keep a closer eye on your issue/PR so that we can progress faster. Apologies for the slow response

AjmeraParth132:
Hi @Yuan325 ,

Apologies for the delayed response, we’ve been AFK this week due to some family commitments and New Year festivities. We’ll be back to work after this week and will pick this up with the highest priority to help move things faster. Responding to your points:

1. The plan is to unify trace names across connection types and represent transports as attributes. Could you please share your email ID? I’ll send over the document we previously shared with Kurtis that outlines the proposed changes.
2. Yes!

We’ll also add a complete visualization of the new traces, attributes, and metrics here shortly, so we can align on the overall approach before proceeding further with development. Thanks.

Yuan325:
@AjmeraParth132 Sounds good, sent an email to you :)

Yuan325:
@AjmeraParth132 
The design LGTM, just a little comment: for transport attribute, can we separate it to `mcp-shttp` (streamable http), `mcp-stdio`, `mcp-httpsse` (for sse).

Please lmk once this is rebased and updated (include moving span to `processMCPMessage()`). Thank you again!! :)

AjmeraParth132:
@Yuan325 perfect, updating the same. though, one more clarification: moving/initializing trace in processMCPMessage() or processMethod() will eventually narrow down the span scope - ie, span time and capturing error will now be limited to the new scope. what we can do here is:

1. either we start with a temporary trace name, and update it later down in processMethod() or processMCPMessage()
2. or we decode the request body in httpHandler start itself to find out the calling method 
3. or work with limited span scope starting from the lower down functions 

which one do you prefer?

Yuan325:
@AjmeraParth132 What do you think about this (kind of combination between 1 and 3):
1. Keep the existing `toolbox/server/mcp` span that starts in `httpHandler()`. There's also a span -- `toolbox/server/mcp/sse` in `sseHandler()`. If we want to make it consistent, we can add a new span in stdio `toolbox/server/mcp/stdio`.
2. Add the general `toolbox/server/toolset/get` and `toolbox/server/tool/invoke` in `processMCPMessage()`. Since we will have the method decoded there (and error handling when decode is not successful), we can prevent decoding it twice. Or if there's any changes on decoding in the future, we won't have to update it at all the different places.

AjmeraParth132:
@Yuan325 That makes sense. But rather than spamming with 2 traces for the same case, what I feel a better approach is -

To start with `toolbox/server/mcp/sse`, `toolbox/server/mcp/stdio` and `toolbox/server/mcp/shttp` respectively, and rename the trace later in `processMCPMessage()`. In that case, we will have full span duration and error attribute reporting with the main trace itself, and if everything runs right, will be renamed to the unified naming anyways.

You can check the PR - it's updated and rebased as well. Lmk if we prefer any other way though.

Yuan325:
@AjmeraParth132  Isn't it generally bad practice to rename telemetry traces since it will generate inconsistent traces, and potentially make it harder to track?

AjmeraParth132:
Agreed @Yuan325 , but the downsides felt more concerning!

For starting `tool/invoke` and `toolset/get` in `processMCPMessage()` - 

1. We’ll loose the errors captured in handlers for mcp calls in the unified traces
2. For someone observing the telemetry, the native ones will have timespan from handler start, while the mcp ones will have ‘em from `processMCPMessage()`, making latency an unreliable attribute. 
3. And obviously, unnecessary 2 trace spamming for the collector 

I feel, if we can let decoding the method in handler itself, that could be a solution OR the renaming way itself as the likelihood of inconsistent traces is almost null (as the traces will always be renamed unless we catch an error in handler parsing prior to `processMCPMessage()` call which is very unlikely, and that too will be captured under `/mcp/<transport>` trace anyway). 

Would love to hear your thoughts!

Yuan325:
@AjmeraParth132 Got it, thanks for the insight. Let me discuss this with the team before finalizing this. Will keep this as priority on my end and circle back to you asap :)

Yuan325:
@AjmeraParth132 I discovered that OpenTelemetry actually has established semantic conventions for MCP ([here](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/)). I wonder if you/your team had a chance to review those yet?

I also wanted to flag that we are planning to deprecate the native endpoint transport in the future. Given that, it may not be worth the effort to consolidate spans from both endpoints right now.

If you’d prefer to sync up to resolve this quickly, feel free to send over a meeting invite.

AjmeraParth132:
Hey @Yuan325! Sorry for the delay. Saw this today - that just didn’t pop in my email.

Yes we have been following the OTel convention developments closely, and I think that’s a great way to instrument for end to end telemetry use cases, across multiple services. Given that, while the full fruit of reorganizing our telemetry will come up after we instrument the complete genAI stack, it is still always the best bet to align with the standard conventions.

Also, deprecating the native endpoint makes it easier. We can now get the MCP spans in `processMCPMessage` directly where we have it decoded anyway. Note that it implies missing error captures in the handlers, which I guess is fine (or we can keep the existing custom `toolbox/server/mcp` as well for that purpose). 

On telemetry, let us reorganize `tools/list` and `tools/call` spans and their attributes as per the convention, and continue with the custom `toolset/get` as `mcp.method.name`, given there’s none defined for that purpose in [[server convention](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#server)] method name list.

Do let me know if you have any thoughts or need any changes in here. Will update the PR with this plan if it looks good? And once done, we can then align metrics as well with the conventions?

Yuan325:
@AjmeraParth132 Awesome, I do agree that deprecating the native endpoint will make the traces much simpler (that way we don't have to keep consistency between the 2 endpoints).

1. Keep the existing `toolbox/server/mcp/sse` for the endpoint that it's used to establish SSE connection.
2. Rename the span `toolbox/server/mcp` to `toolbox/server/mcp/http` and add a new span `toolbox/server/stdio` for stdio. This way we can track if there's any error happening during the handlers? WDYT about this?
3. Implement the sem conv for spans within `processMcpMessage()`. Seems like we'll also have to retrieve properties inside `params._meta` property bag for parent trace's informations.
4. For the telemetry name, agreed to use the names per convention -- `{mcp.method.name} {target}` (e.g. `tools/call get-weather`). Let's also try to add the recommended attributes -- `network.transport`, `network.protocol.*`, `mcp.protocol.version`,... especially since it will be used to differentiate transport protocol etc.
5. We currently don't have a toolset/get method within our mcp endpoint, it uses `tools/list` under the hood, so the `toolset/get` will not be necessary? WDYT about keeping with `tools/list`, and we can add our custom attributes `toolset.name`? (or do you think it makes more sense to add the toolset name as part of the `{target}` attribute. Example: 
   a. `tools/list` (attribute `toolset.name=toolset_a`) and `tools/call get-weather` (attribute `toolset.name=toolset_a`)
   or
   b. `tools/list toolset_a` and `tools/call toolset_a/get-weather` for invoking a specific toolset; `tools/list` and `tools/call get-weather` for default toolset (default toolset name is `""`)
   
Example or spans: 
STDIO initialize:
```
initialize (CLIENT, trace=t1, span=s1) # FROM MCP Client
|
--- toolbox/server/mcp/stdio (SERVER, trace=t1, span=s2, parent=s1) # IN TOOLBOX
     |
     --- initialize (SERVER, trace=t1, span=s3, parent=s2) # IN TOOLBOX
```
STDIO tool call:
```
tools/call get-weather (CLIENT, trace=t1, span=s1) # FROM MCP Client
|
--- toolbox/server/mcp/stdio (SERVER, trace=t1, span=s2, parent=s1) # IN TOOLBOX
     |
     --- tools/call get-weather (SERVER, trace=t1, span=s3, parent=s2) # IN TOOLBOX
```

SSE connection:
```
connection (CLIENT, trace=t1, span=s2, parent=s1) # FROM MCP Client
|
--- toolbox/server/mcp/sse (SERVER, trace=t1, span=s3, parent=s2) # IN TOOLBOX
```

HTTP initialize:
```
initialize (CLIENT, trace=t1, span=s1) # FROM MCP Client
|
--- toolbox/server/mcp/http (SERVER, trace=t1, span=s2, parent=s1) # IN TOOLBOX
     |
     --- initialize (SERVER, trace=t1, span=s3, parent=s2) # IN TOOLBOX
```
HTTP tool call:
```
tools/call get-weather (CLIENT, trace=t1, span=s1) # FROM MCP Client
|
--- toolbox/server/mcp/http (SERVER, trace=t1, span=s2, parent=s1) # IN TOOLBOX
     |
     --- tools/call get-weather (SERVER, trace=t1, span=s3, parent=s2) # IN TOOLBOX
```

If this looks good to you, feel free to update this PR, thank you so much once again!
And yes, once this is done, we can then align the metrics with the conventions :)

AjmeraParth132:
That's great @Yuan325! 

1,2,4 - perfectly aligned
3 - yes, will check with few mcp clients as well if they are consistently passing parent trace information in the meta object and will add them in our server telemetry
5 - correct. I think it is better to put that in attribute(a) as there's no written convention around toolsets, but either way is fine.

Yes, the flows are exactly what they should be! Will start working on the PR 🚀

Yuan325:
@AjmeraParth132 awesome, thank you! Will take a look when Let's add the toolset name in attribute then 👍 

I'll make sure to prioritize for review once this is updated! Thank you so much again :)

AjmeraParth132:
Hey @Yuan325, updated the PR with the required changes, lmk your thoughts.

internal/server/mcp.go
```diff
@@ -143,12 +223,23 @@ func (s *stdioSession) readInputStream(ctx context.Context) error {
...
+
+		v, res, err := processMcpMessage(msgCtx, []byte(line), s.server, s.protocol, "", "", nil, "stdio", "")
 		if err != nil {
 			// errors during the processing of message will generate a valid MCP Error response.
 			// server can continue to run.
-			s.server.logger.ErrorContext(ctx, err.Error())
+			s.server.logger.ErrorContext(msgCtx, err.Error())
+			span.SetStatus(codes.Error, err.Error())
 		}
+		span.End()
```
Yuan325: might want to `defer` this?

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+	case "prompts/get":
+		var req struct {
+			Params struct {
+				Name string `json:"name"`
+			} `json:"params"`
+		}
+		if json.Unmarshal(body, &req) == nil && req.Params.Name != "" {
+			span.SetAttributes(attribute.String("gen_ai.prompt.name", req.Params.Name))
+		}
+	}
```
Yuan325: Similar to `getMethodSpanName()`, we can add the attribute within the method handlers itself when we have access to those information.

internal/server/mcp.go
```diff
@@ -117,6 +119,84 @@ type stdioSession struct {
...
+		}
+		return otel.GetTextMapPropagator().Extract(ctx, carrier)
+	}
+
+	return ctx
+}
+
+// getMethodSpanName returns the span name for a given MCP method following semantic conventions
+// Format: {mcp.method.name} {target} (e.g., "tools/call get-weather")
+func getMethodSpanName(method string, body []byte) string {
```
Yuan325: In cases where this method fail to retrieve the target, it will return `{mcp.method.name}`. In this case, I'm leaning towards removing this function, and naming the span with the `{mcp.method.name}` in `processMcpMessage()` when we start the span.

Then, within each method handlers (e.g. https://github.com/googleapis/genai-toolbox/blob/a15a12873f936b0102aeb9500cc3bcd71bb38c34/internal/server/mcp/v20241105/method.go#L87), we can use `trace.SpanFromContext(ctx)` to retrieve tha ctive span -> update name (e.g. `span.SetName(fmt.Springf("%s %s", TOOLS_CALL, toolsName))`. We'll have to update this in all mcp supported mcp version here (https://github.com/googleapis/genai-toolbox/tree/main/internal/server/mcp)

p/s: in the example above, `TOOLS_CALL` is a const defined in types.go (e.g. https://github.com/googleapis/genai-toolbox/blob/a15a12873f936b0102aeb9500cc3bcd71bb38c34/internal/server/mcp/v20241105/types.go#L33)`

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+
+	// Create method-specific span with semantic conventions
+	// Note: Trace context is already extracted and set in ctx by the caller
+	spanName := getMethodSpanName(baseMessage.Method, body)
+	ctx, span := s.instrumentation.Tracer.Start(ctx, spanName,
+		trace.WithSpanKind(trace.SpanKindServer),
+	)
+	defer span.End()
+
+	// Set required semantic attributes
```
Yuan325:
```suggestion
	// Set required semantic attributes for span according to OTEL MCP semcov
	// ref: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#server
```

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+		networkProtocolName = "http"
+	}
+
+	// Create method-specific span with semantic conventions
+	// Note: Trace context is already extracted and set in ctx by the caller
+	spanName := getMethodSpanName(baseMessage.Method, body)
+	ctx, span := s.instrumentation.Tracer.Start(ctx, spanName,
+		trace.WithSpanKind(trace.SpanKindServer),
+	)
+	defer span.End()
```
Yuan325: Might want to move this a little higher  (before notifications).

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+				switch rpcErr.Error.Code {
+				case jsonrpc.METHOD_NOT_FOUND:
+					span.SetAttributes(attribute.String("error.type", "method_not_found"))
+				case jsonrpc.INVALID_PARAMS:
+					span.SetAttributes(attribute.String("error.type", "invalid_params"))
+				case jsonrpc.INTERNAL_ERROR:
+					span.SetAttributes(attribute.String("error.type", "internal_error"))
+				default:
+					span.SetAttributes(attribute.String("error.type", "jsonrpc_error"))
+				}
```
Yuan325: can we move this switch-case logic into `internal/server/mcp/jsonrpc/jsonrpc.go`? Perhaps something like 
```
func (e Error) String() string {
	switch e.Code {
	case METHOD_NOT_FOUND:
		return "method_not_found"
	...
}
```

In here, we can use `rpcErr.Error.String()`

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+			return "", result, err
 		}
-		return v, res, err
+		return version, result, err
 	default:
 		toolset, ok := s.ResourceMgr.GetToolset(toolsetName)
 		if !ok {
 			err = fmt.Errorf("toolset does not exist")
+			span.SetStatus(codes.Error, err.Error())
+			span.SetAttributes(attribute.String("error.type", "invalid_toolset"))
```
Yuan325: For these, should we keep to the rpc code?
And update the code as follows (checkout the other comment down below regarding the `rpcErr.Error.String()` function): 
```
rpcErr := jsonrpc.NewError(baseMessage.Id, jsonrpc.INVALID_REQUEST, err.Error(), nil)
span.SetStatus(codes.Error, err.Error())
span.SetAttributes(attribute.String("error.type", rpcErr.Error.String()))
return "", rpcErr, err
```

Appreciate if you could update the others as well, thank you!

internal/server/mcp.go
```diff
@@ -240,7 +331,9 @@ func mcpRouter(s *Server) (chi.Router, error) {
 
 // sseHandler handles sse initialization and message.
 func sseHandler(s *Server, w http.ResponseWriter, r *http.Request) {
-	ctx, span := s.instrumentation.Tracer.Start(r.Context(), "toolbox/server/mcp/sse")
+	ctx, span := s.instrumentation.Tracer.Start(r.Context(), "toolbox/server/mcp/sse",
+		trace.WithSpanKind(trace.SpanKindServer),
+	)
```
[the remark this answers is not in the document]
Yuan325: irrelevant to this PR

internal/server/mcp.go
```diff
@@ -336,9 +429,27 @@ func methodNotAllowed(s *Server, w http.ResponseWriter, r *http.Request) {
 func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
-	ctx, span := s.instrumentation.Tracer.Start(r.Context(), "toolbox/server/mcp")
+	ctx := r.Context()
+	ctx = util.WithLogger(ctx, s.logger)
+
+	// Read body first so we can extract trace context
+	body, err := io.ReadAll(r.Body)
```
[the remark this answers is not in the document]
Yuan325: irrelevant to this PR.

internal/server/mcp.go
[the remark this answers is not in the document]
Yuan325: leaning towards removing the function. Please refer to my comments~

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+		}
+		if json.Unmarshal(body, &req) == nil && req.Params.Name != "" {
+			span.SetAttributes(attribute.String("gen_ai.prompt.name", req.Params.Name))
+		}
+	}
+
+	// Process the method
+	var result any
+	var version string
+
```
Yuan325: Do we need to define these here?

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+			span.SetAttributes(attribute.String("jsonrpc.request.id", idStr))
+		} else if idNum, ok := baseMessage.Id.(float64); ok {
+			span.SetAttributes(attribute.String("jsonrpc.request.id", fmt.Sprintf("%v", idNum)))
+		}
+	}
+
+	// Set toolset name if available
+	if toolsetName != "" {
+		span.SetAttributes(attribute.String("toolset.name", toolsetName))
+	}
```
Yuan325:
```suggestion
	span.SetAttributes(attribute.String("toolset.name", toolsetName))
```
Let's set the toolset name regardless~ the default toolset name is an empty string --`""`

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+		trace.WithSpanKind(trace.SpanKindServer),
+	)
+	defer span.End()
+
+	// Set required semantic attributes
+	span.SetAttributes(
+		attribute.String("mcp.method.name", baseMessage.Method),
+		attribute.String("network.transport", networkTransport),
+		attribute.String("network.protocol.name", networkProtocolName),
+		attribute.String("mcp.transport.type", transportType),
```
Yuan325: Doesn't seem like they have a `mcp.transport.type` attribute in the semcov? Might not be necessary. The transport could be deduced from the `network.transport` and `network.protocol.version` attributes (https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#recording-mcp-transport).

We can probably remove the `transportType` input parameter.

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+	}
+
+	// Process the method
+	var result any
+	var version string
+
 	switch baseMessage.Method {
 	case mcputil.INITIALIZE:
-		res, v, err := mcp.InitializeResponse(ctx, baseMessage.Id, body, s.version)
+		result, version, err = mcp.InitializeResponse(ctx, baseMessage.Id, body, s.version)
```
Yuan325: we might want to do a `span.SetAttributes` for the attribute `"mcp.protocol.version"` here.

internal/server/mcp.go
```diff
@@ -504,25 +606,125 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+		span.SetAttributes(attribute.String("mcp.protocol.version", protocolVersion))
+	}
+
+	// Set request ID
+	if baseMessage.Id != nil {
+		if idStr, ok := baseMessage.Id.(string); ok {
+			span.SetAttributes(attribute.String("jsonrpc.request.id", idStr))
+		} else if idNum, ok := baseMessage.Id.(float64); ok {
+			span.SetAttributes(attribute.String("jsonrpc.request.id", fmt.Sprintf("%v", idNum)))
+		}
```
Yuan325:
```suggestion
		span.SetAttributes(attribute.String("jsonrpc.request.id", fmt.Sprintf("%v", baseMessage.Id)))
```

using `%v` directly should be okay. It should automaticallly work for both types.

AjmeraParth132:
Thanks @Yuan325 for the detailed feedback, rightly agreed. Incorporated the above requested changes to the PR.

internal/server/mcp.go
Yuan325:
```suggestion
			if err = s.write(msgCtx, res); err != nil {
```

internal/server/mcp.go
```diff
@@ -498,31 +571,92 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
...
+	ctx, span := s.instrumentation.Tracer.Start(ctx, baseMessage.Method,
+		trace.WithSpanKind(trace.SpanKindServer),
+	)
+	defer span.End()
+
 	// Check if message is a notification
 	if baseMessage.Id == nil {
 		err := mcp.NotificationHandler(ctx, body)
 		return "", nil, err
 	}
```
Yuan325: Might want to move this to after setting the span attributes since span are added before notification handling? Also will appreciate if you can setStatus as error if there's error.
```
		if err != nil {
			span.SetStatus(codes.Error, err.Error())
		}
        return "", nil, err
```

AjmeraParth132:
Perfect! Pushed the changes.

PR #2007, opened 2025-11-20
feat: support for annotations

### LINKED ISSUE #927: Add optional annotations for tagging tools as readOnly and destructive.
### Prerequisites

Nothing currently. This is just a nice thing to add.

### Suggested Solution(s)

https://modelcontextprotocol.io/docs/concepts/tools#tool-definition-structure

According to the latest spec, tools can have the following annotations:
```
    readOnlyHint?: boolean;    // If true, the tool does not modify its environment
    destructiveHint?: boolean; // If true, the tool may perform destructive updates
    idempotentHint?: boolean;  // If true, repeated calls with same args have no additional effect
    openWorldHint?: boolean;   // If true, tool interacts with external entities
```

Agents and users can use this to decide whether to run a tool.

In MCP Toolbox a tool that runs a select or fetches metadata should have the readOnly hint. A tool that runs INSERT/UPDATE/DELETE should have a destructive hint.

### Alternatives Considered

_No response_

### Additional Details

_No response_

drstrangelooker:
## Description

The MCP spec supports tool annotations like the below structure in the 2025-06-18 version of the spec.
https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations

```
{
  destructiveHint?: boolean;
  idempotentHint?: boolean;
  openWorldHint?: boolean;
  readOnlyHint?: boolean;
}
```

Added a ToolAnnotations structure, an Annotations member to the McpManifest structure, and a nil initializer for the Annotations member to all calls to GetMcpManifest.

The ToolAnnotations structure and the member annotations are all defined as pointers so that they are omited when not set. There are times when the zero value is meaningful so this was the only way to make sure that we distinguish between not setting the annotation and setting it with a zero value.

## PR Checklist

- [x] Make sure you reviewed

🛠️ Fixes #927

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -68,10 +68,10 @@ type ToolConfig interface {
 
 // https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations
 type ToolAnnotations struct {
-	DestructiveHint *bool `json:"destructiveHint,omitempty"`
-	IdempotentHint  *bool `json:"idempotentHint,omitempty"`
-	OpenWorldHint   *bool `json:"openWorldHint,omitempty"`
-	ReadOnlyHint    *bool `json:"readOnlyHint,omitempty"`
+	DestructiveHint *bool `json:"destructiveHint,omitempty" yaml:"destructiveHint,omitempty"`
+	IdempotentHint  *bool `json:"idempotentHint,omitempty" yaml:"idempotentHint,omitempty"`
+	OpenWorldHint   *bool `json:"openWorldHint,omitempty" yaml:"openWorldHint,omitempty"`
+	ReadOnlyHint    *bool `json:"readOnlyHint,omitempty" yaml:"readOnlyHint,omitempty"`
 }
 
 type AccessToken string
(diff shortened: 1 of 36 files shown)


internal/tools/tools_test.go
```diff
@@ -84,7 +84,7 @@ func TestGetMcpManifestMetadata(t *testing.T) {
 	}
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			got := tools.GetMcpManifest(tc.name, tc.description, tc.authInvoke, tc.params)
+			got := tools.GetMcpManifest(tc.name, tc.description, tc.authInvoke, tc.params, nil)
```
duwenxin99: Can we add a test case to make sure that the annotations are printed correctly in the manifest? Thanks!

internal/tools/tools.go
averikitsch, line 1: @kurtisvg can you review the use of pointers?

drstrangelooker: I wish there was a better way, but we want the hint to not be there at all if it isn't set true or false.

drstrangelooker:
> What do we think about expose this in the custom tools.yaml?

I think that is a great idea.

drstrangelooker:
> What do we think about expose this in the custom tools.yaml?

If I understand how yaml serialization is done it should be pretty trivial.

internal/tools/tools.go
drstrangelooker: A lot of the pointer usage goes away now that I moved the config into the yaml file.

drstrangelooker:
> What do we think about expose this in the custom tools.yaml?

@averikitsch I moved it into the tools config. Take a look at the prebuilt looker.yaml for examples.

PR #2032, opened 2025-11-25
ci: add gemini bot for issue triage

Yuan325:
## Description

Added issue triage workflows. Consolidated the [automated](https://github.com/google-gemini/gemini-cli/blob/main/.github/workflows/gemini-automated-issue-triage.yml) and [scheduled](https://github.com/google-gemini/gemini-cli/blob/main/.github/workflows/gemini-scheduled-issue-triage.yml) issue triage workflow from gemini-cli repo.

The workflow will be triggered with the following:
* Daily at 9am - will catch any issues that are missed, or issues with the `need-triage` label.
* Whenever there's a new issue opened (reopen does not trigger this workflow)
* manually dispatch

This workflow will do the following:
 * find any issues that needs to be triaged (including issues without any labels, or issues with the `need-triage` label).
 * for each issue, the bot will determine the `product:` label, `type:` label, and `priority:` label that best matches the issue.
 * if issue doesn't have sufficient information, it will leave a comment to request for relevant information.
 * If any issues is identify as `priority: p0` label, it will add a `status: manual-triage` label with a note so that maintainer can manually check those issues.
 * If the bot are not able to identify the `priority:` label, it will also apply the `status: manual-triage` label.
 * after adding labels, it will remove the `status: need-triage` label.
 * blunderbuss will assign the PR based on `product:` label (if set by the bot).

With this, this PR added a couple of labels:
* status: need-triage -- maintainer can add this label to get the issue triaged by the automation.
* status: manual triage -- triage automation will add this label if the issue needs to get triaged manually.

This PR also updated docs to `CONTRIBUTING.md` for new sources to add their labels.


Yuan325:
@averikitsch what do you think about adding both the automated and schedule? Or should we start with the automated for now? (removing the scheduled workflow) to keep it simpler.

.github/workflows/gemini_automated_issue_triage.yaml
```diff
@@ -0,0 +1,461 @@
...
+            echo "Issue #${ISSUE_NUMBER_INPUT} already has 'product:' or 'priority:' labels. Stopping workflow."
+            exit 1
+          fi
+
+          echo "Manual triage checks passed."
+
+      - name: 'Checkout'
+        uses: 'actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8' # ratchet:actions/checkout@v5
+
+      - name: 'Generate GitHub App Token'
```
averikitsch: Why do we need this?

.github/workflows/gemini_automated_issue_triage.yaml
```diff
@@ -0,0 +1,461 @@
...
+        id: 'get_labels'
+        uses: 'actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea'
+        with:
+          github-token: '${{ steps.generate_token.outputs.token }}'
+          script: |-
+            const { data: labels } = await github.rest.issues.listLabelsForRepo({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            });
+            const allowedLabels = [
```
averikitsch: Let's not do it if its too complicated but can we pull these in from the labels.yaml?

.github/workflows/gemini_scheduled_issue_triage.yml
```diff
@@ -0,0 +1,350 @@
...
+          ${{ steps.find_issues.outputs.issues_to_triage != '[]' }}
+        uses: 'google-github-actions/run-gemini-cli@a3bf79042542528e91937b3a3a6fbc4967ee3c31' # ratchet:google-github-actions/run-gemini-cli@v0
+        id: 'gemini_issue_analysis'
+        env:
+          GITHUB_TOKEN: '' # Do not pass any auth token here since this runs on untrusted inputs
+          ISSUES_TO_TRIAGE: '${{ steps.find_issues.outputs.issues_to_triage }}'
+          REPOSITORY: '${{ github.repository }}'
+          AVAILABLE_LABELS: '${{ steps.get_labels.outputs.available_labels }}'
+        with:
+          gcp_workload_identity_provider: '${{ vars.GCP_WIF_PROVIDER }}'
```
averikitsch: Are we setting this up?

.github/workflows/gemini_scheduled_issue_triage.yml
```diff
@@ -0,0 +1,350 @@
...
+                   "labels_to_remove": ["status: need-triage"],
+                   "explanation": "This issue is a bug within the alloydb tool that needs to be addressed with medium priority."
+                 }
+               ]
+               ```
+              If an issue cannot be classified, do not include it in the output array.
+            9. If you see that the issue doesn't look like it has sufficient information, leave a comment politely requesting the relevant information, eg.. if repro steps are missing request for repro steps. if version information is missing request for version information into the explanation section below.
+              - After identifying appropriate labels to an issue, add "status: need-triage" label to labels_to_remove in the output.
+            10. If you think an issue might be a Priority: P0 do not apply the priority: p0 label. Instead apply a status: manual-triage label and include a note in your explanation.
+            11. If you are uncertain and have not been able to apply one each of product:, type: and priority: , apply the status: manual-triage label.
```
averikitsch: would quotes around the labels help?

.github/workflows/gemini_automated_issue_triage.yaml
```diff
@@ -0,0 +1,461 @@
...
+name: '🏷️ Gemini Automated Issue Triage'
+
+on:
+  issues:
+    types:
+      - 'opened'
+      - 'reopened'
+  issue_comment:
+    types:
+      - 'created'
```
averikitsch: I would assume open an issue is a good trigger - not sure if we need to do comments or reopen. The scheduled might be a better workflow - every morning we can check the triage is right and it will pick up anything missed. However, I do like the triage on opening and it should be triaged automatically. We should only need one or another

Yuan325: Agree, I'll reduce it to running every morning at 9am, and also automatically triaged during opening of an issue.

.github/workflows/gemini_automated_issue_triage.yaml
Yuan325: Removed and use the default `${{ secrets.GITHUB_TOKEN }}` instead. Seems to be sufficient for what we are doing.

Yuan325:
currently setting up vars and secrets for this.

.github/workflows/gemini_issue_triage.yaml
```diff
@@ -0,0 +1,396 @@
...
+        uses: 'actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea'
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const rawOutput = process.env.LABELS_OUTPUT;
+            core.info(`Raw output from model: ${rawOutput}`);
+            let parsedLabels;
+
+            try {
+              const jsonMatch = rawLabels.match(/```json\s*([\s\S]*?)\s*```/);
```
dishaprakash, line 314: Should this be `rawOutput.match()` instead?

.github/workflows/gemini_issue_triage.yaml
```diff
@@ -0,0 +1,396 @@
...
+            SINGLE_DATA="$(gh issue view "${ISSUE_NUMBER}" \
+              --repo "${GITHUB_REPOSITORY}" \
+              --json number,title,body)"
+            
+            ISSUES="[${SINGLE_DATA}]"
+          else
+            echo "📅 Bulk Mode: Running full triage scan..."
+            echo '🔍 Finding issues without labels...'
+            NO_LABEL_ISSUES="$(gh issue list --repo "${GITHUB_REPOSITORY}" \
+              --search 'is:open is:issue no:label' --json number,title,body)"
```
dishaprakash, line 134: Should we make this search more specific to find issues which do not have a priority / triage label?
Edge case - Would it fail in the scenario that a new issue is opened and the author adds a `bug` / `question` / etc. tag?

Yuan325:
Closing until further updates on gemini-cli github workflow.

PR #2050, opened 2025-11-26
feat(source/cloudsqlmysql): add support for IAM authentication in Cloud SQL MySQL source

dishaprakash:
## Description

This PR adds the support for IAM authentication in the Cloud SQL MySQL source

## PR Checklist

- [x] Make sure you reviewed


dishaprakash:
The disabled tests (dataplex & healthcare) are currently failing and will be enabled before merging

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -53,9 +53,9 @@ type Config struct {
 	Project  string         `yaml:"project" validate:"required"`
 	Region   string         `yaml:"region" validate:"required"`
 	Instance string         `yaml:"instance" validate:"required"`
-	IPType   sources.IPType `yaml:"ipType"`
-	User     string         `yaml:"user" validate:"required"`
-	Password string         `yaml:"password" validate:"required"`
+	IPType   sources.IPType `yaml:"ipType" validate:"required"`
```
averikitsch: I believe we hard code this to public if not set. We shouldn't update this because then it's a breaking change to require it in the yaml.

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -100,31 +100,78 @@ func (s *Source) MySQLPool() *sql.DB {
...
+		return user, pass, useIAM, nil
+	}
+
+	// If username is empty, fetch email from ADC
+	// otherwise, use username as IAM email
+	if user == "" {
+		if pass != "" {
+			return "", "", useIAM, fmt.Errorf("password is provided without a username. Please provide both a username and password, or leave both fields empty")
+		}
+		email, err := sources.GetIAMPrincipalEmailFromADC(ctx)
```
averikitsch: For MySQL: For an IAM user account, this is the user's email address, without the @ or domain name. For example, for test-user@gmail.com, set the user field to test-user. For a service account, this is the service account's email address without the @project-id.iam.gserviceaccount.com suffix.

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -100,31 +100,78 @@ func (s *Source) MySQLPool() *sql.DB {
...
 	if err != nil {
 		return nil, err
 	}
 
-	if !slices.Contains(sql.Drivers(), "cloudsql-mysql") {
-		_, err = mysql.RegisterDriver("cloudsql-mysql", opts...)
+	// Register a unique driver name for this specific configuration
+	// This prevents conflicts between database sources using
+	// IAM authentication and Password authentication
+	driverName := fmt.Sprintf("cloudsql-mysql-%s", name)
```
averikitsch, line 151: I don't think it's necessary to have a custom driver name

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -100,31 +100,78 @@ func (s *Source) MySQLPool() *sql.DB {
...
+
+	if !slices.Contains(sql.Drivers(), driverName) {
+		_, err = mysql.RegisterDriver(driverName, opts...)
 		if err != nil {
 			return nil, fmt.Errorf("unable to register driver: %w", err)
 		}
 	}
 	// Tell the driver to use the Cloud SQL Go Connector to create connections
-	dsn := fmt.Sprintf("%s:%s@cloudsql-mysql(%s:%s:%s)/%s?connectionAttributes=program_name:%s", user, pass, project, region, instance, dbname, url.QueryEscape(userAgent))
+	dsn := fmt.Sprintf("%s:%s@%s(%s:%s:%s)/%s?connectionAttributes=program_name:%s",
```
averikitsch: We need 2 dsn 1 for both user and password and one for "user@driver"

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
dishaprakash: Got it. Reverted it back

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
dishaprakash: Since this is string formatting and not variable assignment (like postgres), we are able to use 1 dsn instead, 
the tests are also passing. Although I can use 2 for cleaner code

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
dishaprakash: I've added custom formatting for postgres vs mysql now. We only retain the username before `@` for mysql, since it only occurs once in the email (`@` is a forbidden character in all service account / user emails).

Please take a look and let me know what you think

internal/sources/util.go
```diff
@@ -83,9 +83,25 @@ func GetIAMPrincipalEmailFromADC(ctx context.Context) (string, error) {
...
+
+	fullEmail, ok := emailValue.(string)
+	if !ok {
+		return "", fmt.Errorf("email field is not a string")
+	}
+
+	// Format the username based on Database Type
+	switch strings.ToLower(dbType) {
+	case "mysql":
+		username, _, _ := strings.Cut(fullEmail, "@")
```
averikitsch: Should we make sure username is not empty?

averikitsch (approved):

internal/sources/util.go
dishaprakash: Added a final validation check before returning the value for empty string

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
dishaprakash: As mentioned, I have created 2 dsn. For some reason the test was not catching the issue before.

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -121,8 +153,36 @@ func initCloudSQLMySQLConnectionPool(ctx context.Context, tracer trace.Tracer, n
...
-	dsn := fmt.Sprintf("%s:%s@cloudsql-mysql(%s:%s:%s)/%s?connectionAttributes=program_name:%s", user, pass, project, region, instance, dbname, url.QueryEscape(userAgent))
+	if useIAM {
+		token, err := sources.GetIAMAccessToken(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get IAM access token: %w", err)
+		}
+		// Set allowCleartextPasswords to true for IAM
+		dsn = fmt.Sprintf("%s:%s@cloudsql-mysql(%s:%s:%s)/%s?connectionAttributes=program_name:%s&allowCleartextPasswords=true",
+			user,
+			token,
```
averikitsch: We shouldn't need a token for auth. We should be able to use dsn `%s@cloudsql-mysql(%s:%s:%s)/%s`

dishaprakash: Removed the use of token for auth.

PR #2061, opened 2025-11-27
chore(tools/bigtable,tools/cassandra,tools/clickhouse): decouple config from tool

Yuan325:
## Description

## PR Checklist

- [ ] Make sure you reviewed

### DIFF
--- a/internal/tools/clickhouse/clickhouseexecutesql/clickhouseexecutesql.go
+++ b/internal/tools/clickhouse/clickhouseexecutesql/clickhouseexecutesql.go
@@ -28,12 +28,12 @@ import (
 const executeSQLKind string = "clickhouse-execute-sql"
 
 func init() {
-	if !tools.Register(executeSQLKind, newConfig) {
+	if !tools.Register(executeSQLKind, newExecuteSQLConfig) {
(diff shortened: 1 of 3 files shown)


internal/tools/bigtable/bigtable.go
```diff
@@ -156,6 +135,15 @@ func getMapParamsType(tparams parameters.Parameters, params parameters.ParamValu
 }
 
 func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, params parameters.ParamValues, accessToken tools.AccessToken) (any, error) {
+	s, ok := resourceMgr.GetSource(t.Source)
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve source %s in tool %s", t.Source, t.Name)
+	}
+	source, ok := s.(compatibleSource)
```
averikitsch, line 142: Do we only have a compatible source check on invoke? When Toolbox starts up it should ensure compatibility.

Yuan325: @averikitsch Yes. In cases where user adds the tool before adding the source, we don't want the server to fail. (that way the initialization of tool does not depend on the source).

PR #2062, opened 2025-11-27
chore(tools/cloudhealthcare,tools/cloudmonitoring): decouple source config from tools

Yuan325:


internal/sources/cloudmonitoring/cloud_monitoring.go
```diff
@@ -118,9 +118,9 @@ var _ sources.Source = &Source{}
 
 type Source struct {
 	Config
-	BaseURL   string `yaml:"baseUrl"`
-	Client    *http.Client
-	UserAgent string
+	baseURL   string `yaml:"baseUrl"`
```
averikitsch: Is the yaml:BaseUrl required here if it's not in the YAML?

Yuan325: removed

PR #2070, opened 2025-11-27
chore(tools/sqlite,tools/tidb,tools/trino,tools/utility,tools/valkey,tools/yugabytedbsql): decouple source config from tools

Yuan325:


internal/tools/sqlite/sqliteexecutesql/sqliteexecutesql_test.go
averikitsch, line 1: Will tests be updated to init the db or was this extra?

Yuan325: This was tested as part of the integration tests~

PR #2099, opened 2025-12-03
fix(tools/mongodb): remove `required` tag from the `canonical` field

duwenxin99:
The `required` tag raises validation failure error when a boolean field is defined as `false`:
```
ERROR "unable to parse tool file at "mongodb_tools.yaml": unable to parse tool "insert-one-device" as kind "mongodb-insert-one": [2:12] Key: 'Config.Canonical' Error:Field validation for 'Canonical' failed on the 'required' tag\n   1 | authRequired: []\n>  2 | canonical: false\n                  ^\n   3 | collection: Device\n   4 | database: xiar\n   5 | description: Inserts a single new document into the Device collection. The 'data' parameter must be a string containing the JSON object to insert.\n   6 | "
```
All the `required` tags are removed from the boolean `canonical` field of the MongoDB tools. Unit tests are added.

### DIFF
--- a/.ci/quickstart_test/js.integration.cloudbuild.yaml
+++ b/.ci/quickstart_test/js.integration.cloudbuild.yaml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 steps:
-  - name: 'node:20'
+  - name: 'node:22'
     id: 'js-quickstart-test'
     entrypoint: 'bash'
     args:
--- a/README.md
+++ b/README.md
@@ -515,6 +515,36 @@ For more detailed instructions on using the Toolbox Core SDK, see the
     ```
 
   </details>
+  <details>
+    <summary>ADK</summary>
+
+1. Install [Toolbox ADK SDK][toolbox-adk-js]:
+
+    ```bash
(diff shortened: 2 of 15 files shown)


internal/tools/mongodb/mongodbinsertmany/mongodbinsertmany.go
```diff
@@ -54,7 +54,7 @@ type Config struct {
 	Description  string   `yaml:"description" validate:"required"`
 	Database     string   `yaml:"database" validate:"required"`
 	Collection   string   `yaml:"collection" validate:"required"`
-	Canonical    bool     `yaml:"canonical" validate:"required"` //i want to force the user to choose
+	Canonical    bool     `yaml:"canonical"`
```
averikitsch, line 57: We could also have this as required and use *bool

duwenxin99: I think it's ok to just keep them non-required. It only determines the output format so the users can configure it if they want to make the format canonical.

PR #2107, opened 2025-12-04
fix: list tables tools null fix

### LINKED ISSUE #2027: `list_tables` Returns Null Instead of Empty List for Empty Responses
### Prerequisites

Currently, the `list_tables` command, particularly within the Postgres tools, returns `null` when there are no tables in the database. This behavior is unexpected; it should return an empty list (`[]`) instead. This inconsistency can lead to issues in how empty responses are handled by other tools or clients.

### Suggested Solution(s)

Declare an empty slice in the `postgreslisttables.go` file (e.g., `internal/tools/postgres/postgreslisttables/postgreslisttables.go#L231`) to ensure an empty list is returned when no tables are found. This would align the behavior with expectations for empty responses.

### Alternatives Considered

Return a message if no results are found.

### Additional Details

_No response_

gRedHeadphone:
## Description

Return empty list instead of null in list tables tools when no tables found

## PR Checklist

- [x] Make sure you reviewed

🛠️ Fixes #2027

### DIFF
--- a/tests/mariadb/mariadb_integration_test.go
+++ b/tests/mariadb/mariadb_integration_test.go
@@ -282,7 +282,7 @@ func RunMariDBListTablesTest(t *testing.T, databaseName, tableNameParam, tableNa
 				if err := json.Unmarshal([]byte(resultString), &tables); err != nil {
 					t.Fatalf("failed to unmarshal outer JSON array into []tableInfo: %v", err)
 				}
-				var details []map[string]any
+				details := []map[string]any{}
 				for _, table := range tables {
 					var d map[string]any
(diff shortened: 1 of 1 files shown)


averikitsch (approved):

averikitsch:
Can you look to see if MariaDB list tables needs to be updated?

gRedHeadphone:
> Can you look to see if MariaDB list tables needs to be updated?

@averikitsch It seems MariaDB uses MySQL source and tools, and I have updated MySQL list tables https://github.com/googleapis/genai-toolbox/pull/2107/files#diff-3a1dc16d3164a7f1344ee3ea73ab4505f57f15b9baf49ba095f4b3be6f08f84d so it should be fine

PR #2121, opened 2025-12-04
feat: Add `embeddingModel` support

duwenxin99:
First part of the implementation to support semantic search in tools.
Second part: https://github.com/googleapis/genai-toolbox/pull/2151


docs/en/resources/embeddingModels/_index.md
```diff
@@ -0,0 +1,84 @@
...
+      FROM documents
+      ORDER BY distance LIMIT 1
+    parameters:
+      - name: semantic_search_string
+        type: string
+        description: The search query that will be converted to a vector.
+        embeddedBy: gemini-model # refers to the name of a defined embedding model
+```
+
+## Kinds of Embedding Models
```
averikitsch, line 84: What is expected to go in this section?

docs/en/resources/embeddingModels/gemini.md
```diff
@@ -0,0 +1,60 @@
...
+type: docs
+weight: 1
+description: >
+  Use Google's Gemini models to generate high-performance text embeddings for vector databases.
+---
+
+## Getting Started
+
+Google Gemini provides state-of-the-art embedding models that convert text into
+high-dimensional vectors. To use Gemini embeddings, you must first obtain an API
```
averikitsch: Could we also use ADC instead of an API key?

docs/en/resources/embeddingModels/gemini.md
```diff
@@ -0,0 +1,60 @@
...
+### Automatic Vectorization
+
+When a tool parameter is configured with `embeddedBy: <your-gemini-model-name>`,
+the Toolbox intercepts the raw text input from the client and sends it to the
+Gemini API. The resulting numerical array is then formatted before being passed
+to your database source.
+
+### Dimension Matching
+
+The `dimension` field must match the expected size of your database column
```
averikitsch, line 41: nit: add link to a model page for more info

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,118 @@
...
+func (cfg Config) EmbeddingModelConfigKind() string {
+	return EmbeddingModelKind
+}
+
+// Initialize a Gemini embedding model
+func (cfg Config) Initialize() (embeddingmodels.EmbeddingModel, error) {
+	ctx := context.Background()
+
+	// Create new Gemini API client
+	client, err := genai.NewClient(ctx, &genai.ClientConfig{APIKey: cfg.ApiKey})
```
averikitsch: Would a vertex library support both api key and ADC credentials?

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		log.Fatal(err)
```
averikitsch: we should return the error isn't of using log.fatal

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Returns the embedding model kind
+func (cfg Config) EmbeddingModelConfigKind() string {
+	return EmbeddingModelKind
+}
+
+// Initialize a Gemini embedding model
+func (cfg Config) Initialize() (embeddingmodels.EmbeddingModel, error) {
+	ctx := context.Background()
```
averikitsch: I would expect ctx to be pass into this initialize method. @Yuan325 can you confirm?

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (m EmbeddingModel) ToConfig() embeddingmodels.EmbeddingModelConfig {
+	return m.Config
+}
+
+func (m EmbeddingModel) EmbedParameters(ctx context.Context, parameters []string) ([][]float32, error) {
+	logger, err := util.LoggerFromContext(ctx)
+	if err != nil {
+		panic(err)
```
averikitsch: Can we return the error instead of panic? Also can we provide a better error message?

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	contents := convertStringsToContents(parameters)
+
+	embedConfig := &genai.EmbedContentConfig{
+		TaskType: "SEMANTIC_SIMILARITY",
```
averikitsch, line 90: nit: in the future we may want this as a config option

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	if m.Dimension > 0 {
+		embedConfig.OutputDimensionality = genai.Ptr(m.Dimension)
+	}
+
+	result, err := m.Client.Models.EmbedContent(ctx, m.Model, contents, embedConfig)
```
averikitsch, line 97: Does this need to be a batch embed method? Does this return multiple vectors?

duwenxin99: Yes. If a tool has multiple parameters using the same embedding model, we want to send them in a batch request instead of one by one.

docs/en/resources/embeddingModels/_index.md
duwenxin99: This will show the list of embedding model docs like this: https://googleapis.github.io/genai-toolbox/previews/PR-2121/resources/embeddingmodels/#kinds-of-embedding-models

docs/en/resources/embeddingModels/gemini.md
duwenxin99: Yes. I updated the API key field to be optional so that the users can choose to only use the ADC. updated doc as well

docs/en/resources/embeddingModels/gemini.md
duwenxin99: added

internal/embeddingmodels/gemini/gemini.go
duwenxin99: yes it does.

internal/embeddingmodels/gemini/gemini.go
duwenxin99: It seems that we don't have ctx passed in to initialize() for other interfaces.

Yuan325: we do pass in context for sources ([example](https://github.com/googleapis/genai-toolbox/blob/18885f64333196a986bb3a670f40107a671fe9ca/internal/sources/sources.go#L59)). If embeddingmodels require ctx, we should pass it in during initialize()~

duwenxin99: updated

cmd/root.go
```diff
@@ -479,6 +482,15 @@ func mergeToolsFiles(files ...ToolsFile) (ToolsFile, error) {
 			}
 		}
 
+		// Check for conflicts and merge embeddingModels
+		for name, cfg := range file.EmbeddingModels {
```
```suggestion
		for name, em := range file.EmbeddingModels {
```

cmd/root.go
```diff
@@ -479,6 +482,15 @@ func mergeToolsFiles(files ...ToolsFile) (ToolsFile, error) {
...
+			if _, exists := merged.EmbeddingModels[name]; exists {
+				conflicts = append(conflicts, fmt.Sprintf("embedding model '%s' (file #%d)", name, fileIndex+1))
+			} else {
+				merged.EmbeddingModels[name] = cfg
```
```suggestion
				merged.EmbeddingModels[name] = em
```

docs/en/resources/embeddingModels/_index.md
```diff
@@ -0,0 +1,84 @@
...
+embeddingModels:
+  gemini-model: # Name of the embedding model
+    kind: gemini
+    model: gemini-embedding-001
+    apiKey: ${GOOGLE_API_KEY}
+    dimension: 768
+
+```
+
+### Step 1 - Embed Tool Parameters
```
Yuan325:
```suggestion
### Step 2 - Embed Tool Parameters
```

docs/en/resources/embeddingModels/gemini.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "Gemini Embedding"
+type: docs
+weight: 1
+description: >
+  Use Google's Gemini models to generate high-performance text embeddings for vector databases.
```
Yuan325, line 6:
```suggestion
  Use Google's Gemini models to generate high-performance text embeddings for vector databases.
aliases: [/resources/embeddingModels/gemini]
```

docs/en/resources/embeddingModels/gemini.md
```diff
@@ -0,0 +1,73 @@
...
+---
+
+## Getting Started
```
Yuan325:
```suggestion
## About
```

internal/embeddingmodels/gemini/gemini.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/googleapis/genai-toolbox/internal/embeddingmodels"
+	"github.com/googleapis/genai-toolbox/internal/util"
+	"google.golang.org/genai"
+)
+
+const EmbeddingModelKind = "gemini"
```
Yuan325:
```suggestion
const EmbeddingModelKind string = "gemini"
```

internal/embeddingmodels/gemini/gemini.go
Yuan325, line 1: do we need a gemini_test.go? to check for unmarshaling?

internal/server/config.go
```diff
@@ -205,6 +209,50 @@ func (c *AuthServiceConfigs) UnmarshalYAML(ctx context.Context, unmarshal func(i
...
+		kind, ok := v["kind"]
+		if !ok {
+			return fmt.Errorf("missing 'kind' field for embedding model %q", name)
+		}
+
+		dec, err := util.NewStrictDecoder(v)
+		if err != nil {
+			return fmt.Errorf("error creating decoder: %w", err)
+		}
+		switch kind {
```
Yuan325, line 242: Just a heads up, at some point we might want to do the registration method similar to source and tool? Feel free to resolve this.

docs/en/resources/embeddingModels/gemini.md
duwenxin99: I don't think we need aliases for it here because we are not referring to this path anywhere else. The older docs needed them because we migrated them to a different folder.

internal/embeddingmodels/gemini/gemini.go
duwenxin99: added

PR #2151, opened 2025-12-10
feat(tools/postgressql): Add Parameter `embeddedBy` config support

duwenxin99:
Add parameter `embeddedBy` field to support vector embedding & semantic search.
Major change in `internal/util/parameters/parameters.go`

This PR only adds vector formatter for the postgressql tool. Other tools requiring vector formatting may not work with embeddedBy.

Second part of the Semantic Search support. First part: https://github.com/googleapis/genai-toolbox/pull/2121

### DIFF
--- a/internal/tools/postgres/postgresliststoredprocedure/postgresliststoredprocedure.go
+++ b/internal/tools/postgres/postgresliststoredprocedure/postgresliststoredprocedure.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	yaml "github.com/goccy/go-yaml"
+	"github.com/googleapis/genai-toolbox/internal/embeddingmodels"
 	"github.com/googleapis/genai-toolbox/internal/sources"
 	"github.com/googleapis/genai-toolbox/internal/sources/alloydbpg"
 	"github.com/googleapis/genai-toolbox/internal/sources/cloudsqlpg"
@@ -186,6 +187,10 @@ func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, para
 
 func (t Tool) ParseParams(data map[string]any, claims map[string]map[string]any) (parameters.ParamValues, error) {
 	return parameters.ParseParams(t.allParams, data, claims)
+}
+
+func (t Tool) EmbedParams(ctx context.Context, paramValues parameters.ParamValues, embeddingModelsMap map[string]embeddingmodels.EmbeddingModel) (parameters.ParamValues, error) {
+	return parameters.EmbedParams(ctx, t.allParams, paramValues, embeddingModelsMap, nil)
 }
 
 func (t Tool) Manifest() tools.Manifest {


internal/tools/bigquery/bigquerysql/bigquerysql.go
```diff
@@ -299,6 +300,10 @@ func (t Tool) ParseParams(data map[string]any, claims map[string]map[string]any)
 	return parameters.ParseParams(t.AllParams, data, claims)
 }
 
+func (t Tool) EmbedParams(ctx context.Context, paramValues parameters.ParamValues, embeddingModelsMap map[string]embeddingmodels.EmbeddingModel) (parameters.ParamValues, error) {
```
shobsi, line 220: Would appreciate a bit of context, like how this change is going to help BigQuery tools?

duwenxin99: This parameter configuration allows users configure a string param that must be vectorized before being inserted into a semantic search query. Specifically, Toolbox will send a request to an text embedding model API when a parameter with `embeddedBy` is used in a tool call. After receiving the vector result, the vector is sent as part of the query instead of the original string. 
This `EmbedParams` function allows each source to customize how it handles the returned vector. Different sources require different vector formats to be used in a query. A vector formatting function can be passed into `parameters.EmbedParams()`

internal/util/parameters/parameters.go
```diff
@@ -162,6 +163,82 @@ func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[st
...
+			var finalValue any = rawVector
+
+			if formatter == nil {
+				paramValues[item.Index].Value = finalValue
+				continue
+			}
+			if formatter != nil {
+				formattedVector := formatter(rawVector)
+
+				if err != nil {
```
averikitsch: where is the err coming from?

internal/embeddingmodels/embeddingmodels.go
```diff
@@ -27,3 +31,17 @@ type EmbeddingModel interface {
 	ToConfig() EmbeddingModelConfig
 	EmbedParameters(context.Context, []string) ([][]float32, error)
 }
+
+type VectorFormatter func(vectorFloats []float32) any
+
+// formatVectorForPgvector converts a slice of floats into a PostgreSQL vector literal string: '[x, y, z]'
+func FormatVectorForPgvector(vectorFloats []float32) any {
+	str := make([]string, len(vectorFloats))
```
averikitsch: Allocating a slice of strings and then joining them might be slow for very large vectors (e.g., 1536 dims). A strings.Builder or bytes.Buffer would be more performant.

```
if len(vectorFloats) == 0 {
        return "[]"
    }

    var b strings.Builder
    // Heuristic: Pre-allocate memory to avoid re-sizing. 
    // Assuming roughly 10 characters per float + commas.
    b.Grow(len(vectorFloats) * 10) 

    b.WriteByte('[')
    
    for i, f := range vectorFloats {
        if i > 0 {
            b.WriteByte(',')
        }
        // strconv is much faster than fmt.Sprintf for simple types
        b.WriteString(strconv.FormatFloat(float64(f), 'g', -1, 32))
    }
    
    b.WriteByte(']')
    
    return b.String()
```
Why this is better
Reduced Allocations (strings.Builder):

Old way: You were allocating a new string for every single float in the vector (1536 allocations for a standard OpenAI embedding), plus a slice to hold them, plus the final joined string.

New way: strings.Builder manages a single internal buffer. By using b.Grow, we attempt to allocate the necessary memory just once.

Faster Conversion (strconv vs fmt):

Old way: fmt.Sprintf("%g", f) uses reflection to determine the type of f at runtime, which is computationally expensive inside a hot loop.

New way: strconv.FormatFloat is the direct, low-level function for converting floats to strings. It is significantly more performant.

Zero-Copy Return:

b.String() in the new implementation converts the underlying bytes to a string without copying the data (in modern Go versions), whereas strings.Join has to iterate and copy data into a new buffer.

internal/tools/bigquery/bigquerygetdatasetinfo/bigquerygetdatasetinfo.go
```diff
@@ -181,6 +182,10 @@ func (t Tool) ParseParams(data map[string]any, claims map[string]map[string]any)
 	return parameters.ParseParams(t.Parameters, data, claims)
 }
 
+func (t Tool) EmbedParams(ctx context.Context, paramValues parameters.ParamValues, embeddingModelsMap map[string]embeddingmodels.EmbeddingModel) (parameters.ParamValues, error) {
```
averikitsch, line 164: Not all tools have parameters (example https://googleapis.github.io/genai-toolbox/resources/tools/bigquery/bigquery-analyze-contribution/#reference), do we need to support this everywhere?

internal/util/parameters/parameters.go
averikitsch, line 405: are we missing the error for type float?

internal/util/parameters/parameters.go
duwenxin99: removed this block

internal/tools/bigquery/bigquerygetdatasetinfo/bigquerygetdatasetinfo.go
duwenxin99: For Tools that don't have parameters, the `EmbedParams` method would not be useful just like `ParseParams()`. But we still need to require it for the Tool interface so that we can use it during parameter processing.

internal/util/parameters/parameters.go
duwenxin99: updated

internal/embeddingmodels/embeddingmodels.go
duwenxin99: Updated

PR #2155, opened 2025-12-11
feat(sources/trino): add ssl verification options and fix docs example

### LINKED ISSUE #1910: Support for disabling SSL verification for Trino
### Prerequisites

Similar to what was done for http, can there be a flag added to disable ssl verification for Trino?

Or is there a way to include trust store certificate.

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

gRedHeadphone:
## Description

Adds options such as disableSslVerification, sslCert and sslCertPath to trino source. Also fixes trino-sql docs on params

## PR Checklist

- [x] Make sure you reviewed

🛠️ Fixes #1910


internal/sources/trino/trino.go
```diff
@@ -102,17 +107,28 @@ func (s *Source) TrinoDB() *sql.DB {
...
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client := &http.Client{Transport: tr}
+		clientName := fmt.Sprintf("insecure_trino_client_%s", name)
+		if err := trinogo.RegisterCustomClient(clientName, client); err != nil {
+			return nil, fmt.Errorf("failed to register custom client: %w", err)
+		}
+		dsn = fmt.Sprintf("%s&custom_client=%s", dsn, clientName)
+	}
```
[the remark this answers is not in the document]
gRedHeadphone: added warning

internal/sources/trino/trino.go
```diff
@@ -102,17 +107,28 @@ func (s *Source) TrinoDB() *sql.DB {
 	return s.Pool
 }
 
-func initTrinoConnectionPool(ctx context.Context, tracer trace.Tracer, name, host, port, user, password, catalog, schema, queryTimeout, accessToken string, kerberosEnabled, sslEnabled bool) (*sql.DB, error) {
-	//nolint:all // Reassigned ctx
-	ctx, span := sources.InitConnectionSpan(ctx, tracer, SourceKind, name)
+func initTrinoConnectionPool(ctx context.Context, tracer trace.Tracer, name, host, port, user, password, catalog, schema, queryTimeout, accessToken string, kerberosEnabled, sslEnabled bool, sslCertPath, sslCert string, disableSslVerification bool) (*sql.DB, error) {
+	_, span := sources.InitConnectionSpan(ctx, tracer, SourceKind, name)
```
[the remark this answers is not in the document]
gRedHeadphone: made the change

internal/sources/trino/trino.go
```diff
@@ -102,17 +107,30 @@ func (s *Source) TrinoDB() *sql.DB {
...
 
 	// Build Trino DSN
-	dsn, err := buildTrinoDSN(host, port, user, password, catalog, schema, queryTimeout, accessToken, kerberosEnabled, sslEnabled)
+	dsn, err := buildTrinoDSN(host, port, user, password, catalog, schema, queryTimeout, accessToken, kerberosEnabled, sslEnabled, sslCertPath, sslCert)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build DSN: %w", err)
 	}
 
+	if disableSslVerification {
+		fmt.Printf("WARNING: SSL verification is disabled for trino source %q. This is an insecure setting and should not be used in production.", name)
```
Yuan325: Instead of using `fmt.Printf`, let's grab the logger from ctx and use the `WarnContext()` to log a warn. (example: https://github.com/googleapis/genai-toolbox/blob/0f27f956c7d85e6f8a3f4c4589cf8020bc1d3d3b/internal/sources/http/http.go#L72)

internal/sources/trino/trino.go
```diff
@@ -102,17 +107,30 @@ func (s *Source) TrinoDB() *sql.DB {
...
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client := &http.Client{Transport: tr}
+		clientName := fmt.Sprintf("insecure_trino_client_%s", name)
+		if err := trinogo.RegisterCustomClient(clientName, client); err != nil {
```
Yuan325, line 184: Why do we need to create a custom client?

internal/sources/trino/trino.go
```diff
@@ -47,26 +49,29 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
...
+	User                   string `yaml:"user"`
+	Password               string `yaml:"password"`
+	Catalog                string `yaml:"catalog" validate:"required"`
+	Schema                 string `yaml:"schema" validate:"required"`
+	QueryTimeout           string `yaml:"queryTimeout"`
+	AccessToken            string `yaml:"accessToken"`
+	KerberosEnabled        bool   `yaml:"kerberosEnabled"`
+	SSLEnabled             bool   `yaml:"sslEnabled"`
+	SSLCertPath            string `yaml:"sslCertPath"`
+	SSLCert                string `yaml:"sslCert"`
```
Yuan325, line 66: What are the `SSLCertPath` and `SSLCert` used for?

internal/sources/trino/trino.go
```diff
@@ -47,26 +49,29 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
...
+	DisableSslVerification bool   `yaml:"disableSslVerification"`
```
Yuan325, line 67: Instead of `disableSslVerification`, can we use `sslVerify` instead (with default to `true` <- if that's the existing default value; to prevent breaking change)?

Also to clarify, is this value supposed to be a boolean? In the issue, it seems like the example uses `NONE`.

internal/sources/trino/trino_test.go
```diff
@@ -215,6 +230,41 @@ func TestParseFromYamlTrino(t *testing.T) {
...
+					port: "8443"
+					user: testuser
+					catalog: hive
+					schema: default
+					sslEnabled: true
+					sslCertPath: /path/to/cert.pem
+					sslCert: |-
+						-----BEGIN CERTIFICATE-----
+						...
+						-----END CERTIFICATE-----
```
Yuan325, line 249: Was this tested manually? I wonder if this string will parse correctly with the dsn param query.

Yuan325:
@gRedHeadphone Thank you for opening this PR. :) Please let me know if you need more clarifications, thank you!

internal/sources/trino/trino.go
gRedHeadphone: done

internal/sources/trino/trino.go
gRedHeadphone: to make the trino driver use the http client with custom transport config which has ssl verification disabled

internal/sources/trino/trino.go
gRedHeadphone: They are configs which are used in query params when building the DSN

https://pkg.go.dev/github.com/trinodb/trino-go-client/trino#:~:text=config%20path%20(optional)-,SSLCertPath,-string%20%20%20%20%20%20%20%20%20%20%20%20//%20The%20SSL

internal/sources/trino/trino.go
gRedHeadphone: I chose `disableSslVerification` so it is consistent with http config options https://github.com/googleapis/genai-toolbox/blob/0f27f956c7d85e6f8a3f4c4589cf8020bc1d3d3b/internal/sources/http/http.go#L56 .

It should be bool because we are not passing it to uri query params, the implementation of disabling ssl verification in a go client is to register a custom http client.

internal/sources/trino/trino_test.go
gRedHeadphone: The unit test passes, as for manually testing if connection was working, I was getting some error like target IP not matching but that I feel was probably because of the setup I had for testing (some docker config with SSL)

Issue #1220, opened 2025-08-22
Add ability to use both prebuilt config and tool file at once


drstrangelooker:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

Add the ability to run a prebuilt config and specify additional tools in a tool file.

PR #2188, opened 2025-12-16
feat: Support combining prebuilt and custom tool configurations

dishaprakash:
## Description

This PR updates the CLI to allow the --prebuilt flag to be used simultaneously with custom tool flags (--tools-file, --tools-files, or --tools-folder). This enables users to extend a standard prebuilt environment with their own custom tools and configurations.

### Key changes

- Sequential Loading: Load prebuilt configurations first, then accumulate any specified custom configurations before merging.

- Smart Defaults: Updated logic to only default to tools.yaml if no configuration flags are provided. 

- Legacy Auth Compatibility: Implemented an additive merge strategy for authentication. Legacy authSources from custom files are merged into the modern authServices map used by prebuilt tools.

- Strict Validation: To prevent ambiguity, the server will throw an explicit error if a legacy authSource name conflicts with an existing authService name (e.g., from a prebuilt config).

## PR Checklist

- [x] Make sure you reviewed

🛠️ Fixes https://github.com/googleapis/genai-toolbox/issues/1220

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -169,6 +169,7 @@ import (
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mssql/mssqllisttables"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mssql/mssqlsql"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqlexecutesql"
+	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqlgetqueryplan"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqllistactivequeries"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqllisttablefragmentation"
 	_ "github.com/googleapis/genai-toolbox/internal/tools/mysql/mysqllisttables"
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -338,7 +338,7 @@ steps:
         .ci/test_with_coverage.sh \
           "Spanner" \
           spanner \
-          spanner
+          spanner || echo "Integration tests failed." # ignore test failures
 
   - id: "neo4j"
     name: golang:1
--- a/docs/en/resources/sources/cloud-sql-mysql.md
+++ b/docs/en/resources/sources/cloud-sql-mysql.md
@@ -31,6 +31,9 @@ to a database by following these instructions][csql-mysql-quickstart].
 - [`mysql-list-active-queries`](../tools/mysql/mysql-list-active-queries.md)
   List active queries in Cloud SQL for MySQL.
 
+- [`mysql-get-query-plan`](../tools/mysql/mysql-get-query-plan.md)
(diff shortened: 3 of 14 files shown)


cmd/root.go
```diff
@@ -460,6 +460,9 @@ func mergeToolsFiles(files ...ToolsFile) (ToolsFile, error) {
 			if _, exists := merged.AuthSources[name]; exists {
 				conflicts = append(conflicts, fmt.Sprintf("authSource '%s' (file #%d)", name, fileIndex+1))
 			} else {
+				if merged.AuthSources == nil {
```
averikitsch, line 465: Is this needed? I don't mind the extra check but seems unnecessary.

docs/en/reference/cli.md
averikitsch, line 1: Can we add an example to this page?

docs/en/reference/prebuilt-tools.md
```diff
@@ -13,6 +13,8 @@ allowing developers to interact with and take action on databases.
 See guides, [Connect from your IDE](../how-to/connect-ide/_index.md), for
 details on how to connect your AI tools (IDEs) to databases via Toolbox and MCP.
 
+You can now leverage both prebuilt and custom tools simultaneously for broader versatility.
```
averikitsch: Once we add a example to the other doc page, can we link it here?

cmd/root.go
```diff
@@ -856,72 +853,99 @@ func run(cmd *Command) error {
...
+	cmd.cfg.PromptConfigs = finalToolsFile.Prompts
 
-	authSourceConfigs := toolsFile.AuthSources
+	authSourceConfigs := finalToolsFile.AuthSources
 	if authSourceConfigs != nil {
 		cmd.logger.WarnContext(ctx, "`authSources` is deprecated, use `authServices` instead")
-		cmd.cfg.AuthServiceConfigs = authSourceConfigs
+		if cmd.cfg.AuthServiceConfigs == nil {
+			cmd.cfg.AuthServiceConfigs = make(server.AuthServiceConfigs)
+		}
```
averikitsch: This extra check should not be required

cmd/root_test.go
averikitsch, line 1: Can we add a test case for a naming collisions?

cmd/root.go
dishaprakash: This is needed to avoid a panic. Because AuthSources is not initialized like the rest of the parts (Sources, AuthServices, etc.) above this.

This is because I've added a test which uses AuthSources (since even though it is deprecated, it is still supported currently)

cmd/root.go
dishaprakash: Removed!

cmd/root_test.go
dishaprakash: We do have a test for tool name collision, but I've increased the scope to toolset and source name collisions as well!

docs/en/reference/prebuilt-tools.md
dishaprakash: Done!

docs/en/reference/cli.md
dishaprakash: Done!

cmd/root.go
averikitsch: Thank you! Sources and Services look very similar

PR #2193, opened 2025-12-17
chore: update dataplex aspecttypes integration tests

### LINKED ISSUE #2057: Ensure that dataplex testing removes aspect types
### Prerequisites

### Toolbox version

0.21.0

### Environment

  - As a downloaded binary (e.g. from `curl -O https://storage.googleapis.com/genai-toolbox/v$VERSION/linux/amd64/toolbox`)
  - Compiled from source (include the command used to build)

### Client

1. Client: <name and link to the client are you using>
2. Version: <what exact version of the client are you using> 
3. Example: If possible, please include your code of configuration:

```python
# Code goes here! 
```

### Expected Behavior

Integration test to clean up resources after every run.

### Current Behavior

Seems like the aspect types is not removed. We hit 250 quota limit.

### Steps to reproduce?

After awhile of running integration test, it will hit the quota.

### Additional Details

_No response_

sahaajaaa:
## Description

Addresses an issue where Dataplex AspectTypes created during integration tests were not consistently deleted. This accumulation led to exceeding the AspectType quota.
To prevent this, the test setup now includes a step to list and delete all existing AspectTypes within the test project and location *before* attempting to create any new ones. This ensures a clean state for each test run and avoids hitting the quota.

## PR Checklist

- [ ] Make sure you reviewed

🛠️ Fixes #2057

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ To run Toolbox from binary:
 To run the server after pulling the [container image](#installing-the-server):
 
 ```sh
-export VERSION=0.11.0 # Use the version you pulled
+export VERSION=0.24.0 # Use the version you pulled
 docker run -p 5000:5000 \
 -v $(pwd)/tools.yaml:/app/tools.yaml \
 us-central1-docker.pkg.dev/database-toolbox/toolbox/toolbox:$VERSION \
--- a/docs/en/resources/tools/alloydbainl/alloydb-ai-nl.md
+++ b/docs/en/resources/tools/alloydbainl/alloydb-ai-nl.md
@@ -91,8 +91,8 @@ visible to the LLM.
     https://cloud.google.com/alloydb/docs/parameterized-secure-views-overview
 
 {{< notice tip >}} Make sure to enable the `parameterized_views` extension
-before running this tool. You can do so by running this command in the AlloyDB
-studio:
+to utilize PSV feature (`nlConfigParameters`) with this tool. You can do so by
+running this command in the AlloyDB studio:
 
 ```sql
 CREATE EXTENSION IF NOT EXISTS parameterized_views;
--- a/internal/sources/looker/looker.go
+++ b/internal/sources/looker/looker.go
@@ -15,7 +15,9 @@ package looker
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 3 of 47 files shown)


tests/dataplex/dataplex_integration_test.go
```diff
@@ -102,6 +138,9 @@ func TestDataplexToolEndpoints(t *testing.T) {
 		t.Fatalf("unable to create Dataplex connection: %s", err)
 	}
 
+	// Cleanup ALL existing AspectTypes in the project
+        cleanupExistingAspectTypes(t, ctx, dataplexClient)
```
divitwork: Can we assert the aspect type count is 0 after this cleanup? 

Can we also check the performance (time taken) after introduction of these steps.

sahaajaaa: Added a verify step in the helper method.

checked the performance by testing before and after modifcation locally and the time taken is ~128sec for both.

sahaajaaa:
> I don't think this solution works if there are multiple tests running at the same time, but it does cover if the tests end early.

You're right, thanks for spotting the concurrency issue. I've updated the code to only delete AspectTypes older than 2 hours at the start, making it safe for parallel runs while still cleaning up leaks from tests that ended early.

tests/dataplex/dataplex_integration_test.go
```diff
@@ -85,13 +85,58 @@ func initDataplexConnection(ctx context.Context) (*dataplex.CatalogClient, error
...
+		Parent:   parent,
+		Filter:   filter,
+		PageSize: 8, // Fetch up to 8 items to delete
+	}
+
+	const maxDeletes = 8 // Explicitly limit the number of deletions
+	it := client.ListAspectTypes(ctx, listReq)
+	var aspectTypesToDelete []string
+	for {
+		if len(aspectTypesToDelete) >= maxDeletes {
```
divitwork: We already have a PageSize filter in the listReq, do we need this check?

tests/dataplex/dataplex_integration_test.go
```diff
@@ -85,13 +85,58 @@ func initDataplexConnection(ctx context.Context) (*dataplex.CatalogClient, error
...
+	}
+
+	if len(aspectTypesToDelete) == 0 {
+		t.Logf("cleanupOldAspectTypes: No aspect types found older than %s to delete.", oldThreshold.String())
+		return
+	}
+
+	for _, aspectTypeName := range aspectTypesToDelete {
+		deleteReq := &dataplexpb.DeleteAspectTypeRequest{Name: aspectTypeName}
+		if _, err := client.DeleteAspectType(ctx, deleteReq); err != nil {
```
divitwork: This returns an LRO, we should wait for it to complete using:
op.Wait(ctx);

tests/dataplex/dataplex_integration_test.go
sahaajaaa: PageSize in the ListAspectTypes API acts as a suggestion for how many items the server should return per page. The iterator pattern used by client.ListAspectTypes might still fetch more than maxDeletes items in total if there are multiple pages of results. The explicit if len(aspectTypesToDelete) >= maxDeletes check ensures we strictly limit the number of items we process and add to the delete list to 8, regardless of the server's pagination behavior.

tests/dataplex/dataplex_integration_test.go
sahaajaaa: Done

tests/dataplex/dataplex_integration_test.go
```diff
@@ -85,13 +85,61 @@ func initDataplexConnection(ctx context.Context) (*dataplex.CatalogClient, error
...
+// cleanupOldAspectTypes Deletes AspectTypes older than the specified duration.
+func cleanupOldAspectTypes(t *testing.T, ctx context.Context, client *dataplex.CatalogClient, oldThreshold time.Duration) {
+	parent := fmt.Sprintf("projects/%s/locations/us", DataplexProject)
+	olderThanTime := time.Now().Add(-oldThreshold)
+	filterTimeStr := olderThanTime.UTC().Format(time.RFC3339)
+	filter := fmt.Sprintf("create_time < \"%s\"", filterTimeStr)
+
+	listReq := &dataplexpb.ListAspectTypesRequest{
+		Parent:   parent,
+		Filter:   filter,
```
estshy: According to public docs endpoint doesn't support filter by create_time field, therefore we might relay here on undocumented behavior that might break at some point.

https://docs.cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations.aspectTypes/list#query-parameters

I would consider to retrieve ~100 aspect types and filter/sort in memory instead.

sahaajaaa: Done

PR #2210, opened 2025-12-19
feat(bigquery): Make credentials scope configurable

### LINKED ISSUE #1942: Adding drive.readonly scope in BigQuery ADC initialization for external Google Sheets
### Prerequisites

### Question

Hi!
This is a follow-up question related to my previous issue #1601 about accessing external tables backed by Google Sheets from BigQuery via Toolbox.

In that issue, the error was:

> Access Denied: BigQuery ... Permission denied while getting Drive credentials

On my **local environment** and in **docker-compose**, I was able to fix this by using ADC with explicit scopes, e.g.:

```bash
gcloud auth application-default login \
  --scopes=https://www.googleapis.com/auth/drive.readonly,https://www.googleapis.com/auth/bigquery,https://www.googleapis.com/auth/cloud-platform
```
After doing that, Toolbox could successfully query the external table backed by Google Sheets, and everything worked as expected. 🙌

Kubernetes environment: cannot easily customize ADC scopes
Now I’m trying to run Toolbox in a Kubernetes environment.

According to the ADC docs (especially for Kubernetes workloads), there is no interactive gcloud auth application-default login --scopes=... flow, and ADC is obtained from the environment (Workload Identity / metadata server, or a mounted service account key).

In this setup, Toolbox’s BigQuery source eventually calls:

```go
// internal/sources/bigquery/bigquery.go
cred, err := google.FindDefaultCredentials(ctx, bigqueryapi.Scope)
```
(ref:
https://github.com/googleapis/genai-toolbox/blob/dc7c62c951590c97d8b67d786a7358fdc8036089/internal/sources/bigquery/bigquery.go#L519)

This means ADC is created with only bigqueryapi.Scope (i.e. https://www.googleapis.com/auth/bigquery), but not with a Drive scope such as https://www.googleapis.com/auth/drive.readonly.

My understanding is:

When querying an external table backed by Google Sheets,
BigQuery needs a token that also has the Google Drive scope,
Which is why it works locally when I login with both BigQuery + Drive scopes,
But fails in Kubernetes where the Toolbox server itself never asks for the Drive scope.

Question: Can we add drive.readonly scope here?
Would it be acceptable (or recommended) to change this line:

```go
cred, err := google.FindDefaultCredentials(ctx, bigqueryapi.Scope)
to something like:
```

```go
cred, err := google.FindDefaultCredentials(
    ctx,
    bigqueryapi.Scope,
    "https://www.googleapis.com/auth/drive.readonly",
)
```

Right now:
Local / docker-compose: works fine when I explicitly set ADC scopes (including Drive) via gcloud auth application-default login.
Kubernetes: I’m relying on ADC from the environment, and I don’t see a practical way to “add” the Drive scope there unless Toolbox itself requests it in FindDefaultCredentials.
If I’m misunderstanding how BigQuery external Google Sheets authentication is supposed to work here, I’d really appreciate some guidance on the intended setup for Toolbox in Kubernetes.

Thanks again for all your help and for maintaining this project! 🙏

### Code

_No response_

### Additional Details

_No response_

shobsi:
## Description

This change addresses the ask where the user may want to use custom scopes. For instance, the default scope (bigquery) falls short from running sql that utilizes integration with other google products, such as Drive, Vertex AI, Cloud Run etc. With this change the user would be able to configure custom scopes depending on their use case.

The custom scopes can be configured in the tools.yaml file, e.g.:

```yaml
sources:
  bigquery-source:
    kind: "bigquery"
    project: ${BIGQUERY_PROJECT}
    location: ${BIGQUERY_LOCATION:}
    useClientOAuth: ${BIGQUERY_USE_CLIENT_OAUTH:false}
    scopes:
      - "https://www.googleapis.com/auth/bigquery"
      - "https://www.googleapis.com/auth/drive"
```

and if the [bigquery prebuilt config](https://github.com/googleapis/genai-toolbox/blob/main/internal/prebuiltconfigs/tools/bigquery.yaml) is being used, then it can be set in the environment variable as well:

```shell
$ export BIGQUERY_SCOPES="https://www.googleapis.com/auth/bigquery,https://www.googleapis.com/auth/drive"
```

## PR Checklist

- [ ] Make sure you reviewed

🛠️ Fixes #1942

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -484,11 +484,11 @@ func mergeToolsFiles(files ...ToolsFile) (ToolsFile, error) {
 		}
 
 		// Check for conflicts and merge embeddingModels
-		for name, model := range file.EmbeddingModels {
+		for name, em := range file.EmbeddingModels {
 			if _, exists := merged.EmbeddingModels[name]; exists {
 				conflicts = append(conflicts, fmt.Sprintf("embedding model '%s' (file #%d)", name, fileIndex+1))
 			} else {
-				merged.EmbeddingModels[name] = model
+				merged.EmbeddingModels[name] = em
 			}
 		}
 
--- a/internal/embeddingmodels/embeddingmodels.go
+++ b/internal/embeddingmodels/embeddingmodels.go
@@ -14,7 +14,11 @@
 
 package embeddingmodels
 
-import "context"
+import (
+	"context"
+	"strconv"
+	"strings"
+)
 
 // EmbeddingModelConfig is the interface for configuring embedding models.
 type EmbeddingModelConfig interface {
@@ -27,3 +31,29 @@ type EmbeddingModel interface {
 	ToConfig() EmbeddingModelConfig
 	EmbedParameters(context.Context, []string) ([][]float32, error)
 }
+
+type VectorFormatter func(vectorFloats []float32) any
+
+// FormatVectorForPgvector converts a slice of floats into a PostgreSQL vector literal string: '[x, y, z]'
+func FormatVectorForPgvector(vectorFloats []float32) any {
+	if len(vectorFloats) == 0 {
+		return "[]"
+	}
+
+	// Pre-allocate the builder.
+	var b strings.Builder
+	b.Grow(len(vectorFloats) * 10)
+
+	b.WriteByte('[')
+	for i, f := range vectorFloats {
+		if i > 0 {
+			b.WriteString(", ")
+		}
+		b.Write(strconv.AppendFloat(nil, float64(f), 'g', -1, 32))
+	}
+	b.WriteByte(']')
+
+	return b.String()
+}
+
+var _ VectorFormatter = FormatVectorForPgvector
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -246,6 +246,14 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	}
 	s.logger.DebugContext(ctx, fmt.Sprintf("invocation params: %s", params))
 
+	params, err = tool.EmbedParams(ctx, params, s.ResourceMgr.GetEmbeddingModelMap())
+	if err != nil {
+		err = fmt.Errorf("error embedding parameters: %w", err)
+		s.logger.DebugContext(ctx, err.Error())
+		_ = render.Render(w, r, newErrResponse(err, http.StatusBadRequest))
+		return
+	}
+
 	res, err := tool.Invoke(ctx, s.ResourceMgr, params, accessToken)
 
 	// Determine what error to return to the users.
--- a/internal/server/common_test.go
+++ b/internal/server/common_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/go-chi/chi/v5"
+	"github.com/googleapis/genai-toolbox/internal/embeddingmodels"
 	"github.com/googleapis/genai-toolbox/internal/log"
 	"github.com/googleapis/genai-toolbox/internal/prompts"
 	"github.com/googleapis/genai-toolbox/internal/server/resources"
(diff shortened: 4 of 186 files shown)


internal/sources/bigquery/bigquery.go
```diff
@@ -502,12 +504,27 @@ func initBigQueryConnection(
...
+		} else {
+			credScopes = []string{"https://www.googleapis.com/auth/cloud-platform"}
+		}
+	} else {
+		if len(scopes) > 0 {
+			credScopes = scopes
+		} else {
+			credScopes = []string{bigqueryapi.Scope}
+		}
+	}
```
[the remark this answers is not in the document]
shobsi: Done!

internal/sources/bigquery/bigquery.go
```diff
@@ -630,11 +648,18 @@ func initDataplexConnection(
 	} else {
 		var opts []option.ClientOption
 
+		var credScopes []string
+		if len(scopes) > 0 {
+			credScopes = scopes
+		} else {
+			credScopes = []string{"https://www.googleapis.com/auth/cloud-platform"}
+		}
```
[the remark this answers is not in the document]
shobsi: Done!

docs/BIGQUERY_README.md
```diff
@@ -68,6 +68,7 @@ The BigQuery MCP server is configured using environment variables.
 export BIGQUERY_PROJECT="<your-gcp-project-id>"
 export BIGQUERY_LOCATION="<your-dataset-location>"  # Optional
 export BIGQUERY_USE_CLIENT_OAUTH="true"  # Optional
+export BIGQUERY_SCOPES="<comma-separated-scopes>"  # Optional
```
duwenxin99, line 71: Could you add more detailed instruction on how to set the scope? We should also add this new field and its explanation to docs/en/resources/sources/bigquery.md.

internal/sources/bigquery/bigquery.go
```diff
@@ -72,14 +75,42 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
...
+	UseClientOAuth            bool                `yaml:"useClientOAuth"`
+	ImpersonateServiceAccount string              `yaml:"impersonateServiceAccount"`
+	Scopes                    StringOrStringSlice `yaml:"scopes"`
+}
+
+// StringOrStringSlice is a custom type that can unmarshal both a single string
+// (which it splits by comma) and a sequence of strings into a string slice.
+type StringOrStringSlice []string
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
```
duwenxin99, line 98: Why do we need this instead of keep using []string?

duwenxin99 (changes requested):

internal/sources/bigquery/bigquery.go
shobsi: That's what I had in the beginning, but then realized in the local debugging that we needed custom handling to convert the comma-separated list of scopes set via environment variable BIGQUERY_SCOPES (mentioned in the description) to an array of string.

docs/BIGQUERY_README.md
shobsi: Done, thanks for the call!

internal/sources/bigquery/bigquery.go
```diff
@@ -77,14 +80,42 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
...
+func (s *StringOrStringSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
duwenxin99: We use `any` instead of `interface{}` for this repo because it is more idiomatic in modern Go. Could you replace all the `interface{}` with `any` in this function? Thanks!

shobsi: Done, thanks!

PR #2211, opened 2025-12-19
docs: migrate to `toolbox-adk` and simplified `ToolboxToolset`

anubhav756:
Updates all quickstart guides and samples to use the new `toolbox-adk` package instead of the legacy `toolbox-core`. Also updates `ToolboxToolset` usage to rely on the simplified constructor (implicit authentication) and ensures correct dependency installation.

> [!NOTE]
> The integration tests are failing because the `google-adk` package is not released yet with the newer changes from `toolbox-adk`. This is expected behavior until the [package update](cl/853799009) is released.

### DIFF
--- a/docs/en/how-to/deploy_adk_agent.md
+++ b/docs/en/how-to/deploy_adk_agent.md
@@ -46,10 +46,10 @@ with the necessary configuration for deployment to Vertex AI Agent Engine.
     process will generate deployment configuration files (like a `Makefile` and
     `Dockerfile`) in your project directory.
 
-4.  Add `toolbox-adk` as a dependency to the new project:
+4.  Add `google-adk[toolbox]` as a dependency to the new project:
 
     ```bash
-    uv add toolbox-adk
+    uv add google-adk[toolbox]
     ```
 
 ## Step 3: Configure Google Cloud Authentication


docs/en/getting-started/quickstart/python/adk/quickstart.py
```diff
@@ -1,15 +1,18 @@
 from google.adk import Agent
 from google.adk.apps import App
-from toolbox_core import ToolboxSyncClient
+from google.adk.tools.toolbox_toolset import ToolboxToolset
 
 # TODO(developer): update the TOOLBOX_URL to your toolbox endpoint
-client = ToolboxSyncClient("http://127.0.0.1:5000")
+toolset = ToolboxToolset(
+    server_url="http://127.0.0.1:5000",
+    toolset_name="my-toolset"
```
averikitsch: is toolset required?

averikitsch: I want us to focus on the beginner experience of no toolset just all tools if possible.

anubhav756: Good idea and makes sense. Simplified it. Thanks for the suggestion!

docs/en/how-to/deploy_adk_agent.md
```diff
@@ -46,10 +46,10 @@ with the necessary configuration for deployment to Vertex AI Agent Engine.
     process will generate deployment configuration files (like a `Makefile` and
     `Dockerfile`) in your project directory.
 
-4.  Add `toolbox-core` as a dependency to the new project:
+4.  Add `toolbox-adk` as a dependency to the new project:
 
     ```bash
-    uv add toolbox-core
+    uv add toolbox-adk
```
averikitsch: Why not use `google-adk[toolbox]`?

anubhav756: Missed it. Thanks for pointing that out 🙂

PR #2238, opened 2025-12-24
refactor: move source implementation in Invoke() function to Source

Yuan325:
Move source-related queries from `Invoke()` function into Source.

The following sources were updated in this PR:
* mssql
* cloudsql-mssql
* mysql
* cloudsql-mysql

This is an effort to generalizing tools to work with any Source that implements a specific interface. This will provide a better segregation of the roles for Tools vs Source.

Tool's role will be limited to the following:
* Resolve any pre-implementation steps or parameters (e.g. template parameters)
* Retrieving Source
* Calling the source's implementation


internal/tools/mysql/mysqllisttablesmissinguniqueindexes/mysqllisttablesmissinguniqueindexes.go
```diff
@@ -135,57 +135,7 @@ func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, para
...
-		}
-		out = append(out, vMap)
-	}
-
-	if err := results.Err(); err != nil {
-		return nil, fmt.Errorf("errors encountered during row iteration: %w", err)
-	}
-
-	return out, nil
+	return source.RunSQL(ctx, listTablesMissingUniqueIndexesStatement, []any{table_schema, table_schema, limit})
```
shuzhou-gc, line 138: How does this change affect some future tools we plan to add, which involves more complicated logic and more than one query...

Should those future tools just call multiple RunSQL() when needed and parse returns from RunSQL() ?

Yuan325: @shuzhou-gc The tool can handle it's own logic to grab all the information needed --> parse the statement + params to `RunSQL()`. If there are need to process the results further, the tool can handle that as well

shuzhou-gc: Got it, thanks for the clarification.

internal/tools/mysql/mysqlgetqueryplan/mysqlgetqueryplan.go
```diff
@@ -108,30 +110,27 @@ func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, para
...
-	} else {
+	// extract and return only the query plan object
+	resSlice, ok := result.([]any)
+	if !ok || len(resSlice) == 0 {
 		return nil, fmt.Errorf("no query plan returned")
 	}
-
-	if err := results.Err(); err != nil {
-		return nil, fmt.Errorf("errors encountered during row iteration: %w", err)
+	row, ok := resSlice[0].(orderedmap.Row)
```
shuzhou-gc, line 122: Just want to confirm, this maps to the whole `{}` json object in the result, right?

Something like 
```
{
 "query_block": {
    ...
  }
}
```

Yuan325: @shuzhou-gc it maps to a []Column of the json object. But yes, it will map to the whole `{}` json object similar to what was returned previously. There's an integration test that checks for this (https://github.com/googleapis/genai-toolbox/blob/c088d4ed4224d3c6702471399c8ab4ebc6d8719f/tests/tool.go#L3390). Let me know if you have any concerns about this~

PR #2254, opened 2026-01-04
feat: add allowed-hosts flag

Yuan325:
## Description

Previously added `allowed-origins` (for CORs) is not sufficient for preventing DNS rebinding attacks. We'll have to check host headers.

To test, run Toolbox with the following:
```
go run . --allowed-hosts=127.0.0.1:5000
```

Test with the following:
```
// curl successfully
curl -H "Host: 127.0.0.1:5000" http://127.0.0.1:5000

// will show Invalid Host Header error
curl -H "Host: attacker:5000" http://127.0.0.1:5000
```

## PR Checklist

- [ ] Make sure you reviewed

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -949,6 +949,7 @@ func run(cmd *Command) error {
 
 	cmd.cfg.SourceConfigs = finalToolsFile.Sources
 	cmd.cfg.AuthServiceConfigs = finalToolsFile.AuthServices
+	cmd.cfg.EmbeddingModelConfigs = finalToolsFile.EmbeddingModels
 	cmd.cfg.ToolConfigs = finalToolsFile.Tools
 	cmd.cfg.ToolsetConfigs = finalToolsFile.Toolsets
 	cmd.cfg.PromptConfigs = finalToolsFile.Prompts


docs/en/reference/cli.md
```diff
@@ -8,25 +8,26 @@ description: >
...
-|              | `--allowed-origins`        | Specifies a list of origins permitted to access this server.                                                                                                                                  | `*`         |
-| `-v`         | `--version`                | version for toolbox                                                                                                                                                                           |             |
+|              | `--ui`                     | Launches the Toolbox UI web server.                                                                                                                                              |             |
+|              | `--allowed-origins`        | Specifies a list of origins permitted to access this server.                                                                                                                     | `*`         |
```
duwenxin99: We may need to explain the difference between these two flags better.

Yuan325: The explanation is included in the docs (deploy to gke/docker/cloud etc).

Yuan325: Also added more description to the CLI. (CORs vs DNS rebinding attacks)

PR #2288, opened 2026-01-10
refactor(tools/cloudgda)!: update description and parameter name for cloudgda tool

wangauone:
- Refactors the 'cloud-gemini-data-analytics-query' tool to update its default description with detailed tool guidance and usage guidance. 
- Append the default description to the tools.yaml description no matter whether the tools.yaml description exists since this guidance will always be useful to the agent on how to use the tool. 
- Renames the input parameter from 'prompt' to 'query' for better consistency.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,13 @@ You can manually trigger the bot by commenting on your Pull Request:
 *   `/gemini summary`: Posts a summary of the changes in the pull request.
 *   `/gemini help`: Overview of the available commands
 
+## Guidelines for Pull Requests
+
+1. Please keep your PR small for more thorough review and easier updates. In case of regression, it also allows us to roll back a single feature instead of multiple ones.
+1. For non-trivial changes, consider opening an issue and discussing it with the code owners first.
+1. Provide a good PR description as a record of what change is being made and why it was made. Link to a GitHub issue if it exists.
+1. Make sure your code is thoroughly tested with unit tests and integration tests. Remember to clean up the test instances properly in your code to avoid memory leaks.
+
 ## Adding a New Database Source or Tool
 
 Please create an
@@ -109,6 +116,8 @@ implementation](https://github.com/googleapis/genai-toolbox/blob/main/internal/s
 
 We recommend looking at an [example tool
 implementation](https://github.com/googleapis/genai-toolbox/tree/main/internal/tools/postgres/postgressql).
+
+Remember to keep your PRs small. For example, if you are contributing a new Source, only include one or two core Tools within the same PR, the rest of the Tools can come in subsequent PRs. 
 
 * **Create a new directory** under `internal/tools` for your tool type (e.g., `internal/tools/newdb/newdbtool`).
 * **Define a configuration struct** for your tool in a file named `newdbtool.go`.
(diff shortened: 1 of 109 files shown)


internal/tools/cloudgda/cloudgda.go
```diff
@@ -68,11 +68,31 @@ func (cfg Config) ToolConfigKind() string {
```
g-lynnzee, line 83: Non-blocking for this change, but wanted to point this out:

With the CA work, we will want to also expose Chat() method. Will the structure of this tool/file support adding a second method (and potentially more methods) from GDA API?

internal/tools/cloudgda/cloudgda.go
```diff
@@ -68,11 +68,31 @@ func (cfg Config) ToolConfigKind() string {
...
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
 	// Define the parameters for the Gemini Data Analytics Query API
-	// The prompt is the only input parameter.
+	// The query is the only input parameter.
 	allParameters := parameters.Parameters{
-		parameters.NewStringParameterWithRequired("prompt", "The natural language question to ask.", true),
+		parameters.NewStringParameterWithRequired("query", "A natural language formulation of a database query.", true),
+	}
+	// The input and outputs are for tool guidance, usage guidance is for multi-turn interaction.
+	guidance := "Tool guidance:\n" +
```
g-lynnzee: Have you tested removing this part of the prompt from the ADK app and having it rely on the tool prompt?

internal/tools/cloudgda/cloudgda.go
```diff
@@ -68,11 +68,31 @@ func (cfg Config) ToolConfigKind() string {
...
+		"    1. disambiguation_question: Clarification questions or comments where the tool needs the users' input.\n" +
+		"    2. generated_query: The generated query for the user query.\n" +
+		"    3. intent_explanation: An explanation for why the tool produced `generated_query`.\n" +
+		"    4. query_result: The result of executing `generated_query`.\n" +
+		"    5. natural_language_answer: The natural language answer that summarizes the `query` and `query_result`.\n" +
+		"\n" +
+		"Usage guidance:\n" +
+		"  1. If `disambiguation_question` is produced, then solicit the needed inputs from the user and try the tool with a new `query` that has the needed clarification.\n" +
+		"  2. If `natural_language_answer` is produced, use `intent_explanation` and `generated_query` to see if you need to clarify any assumptions for the user.\n" +
+		"  3. If the tool output indicates failure or empty results, explain that clearly using the provided reasoning."
```
g-lynnzee: I think we should leave #3 out, since there is no interpretation here and the behavior is more application specific.

g-lynnzee (approved):

internal/tools/cloudgda/cloudgda.go
wangauone: Yes, we just call different methods from the client. I am also working on updating to use our Go SDK client, it will be much easier to support a new method.

internal/tools/cloudgda/cloudgda.go
wangauone: Yes, was tested by removing all the tool related prompts. Verified the prompt shows up in the tool side. Also, the E2E example worked as before.

internal/tools/cloudgda/cloudgda.go
wangauone: SG. Removed.

Yuan325 (approved):
Thank you! Marked this as breaking change since it changes the parameter name

PR #2308, opened 2026-01-14
feat(tools/serverlessspark): Adding get_session_template tool to serverlessspark 

vinodhinic:
## Description

Adding get_session_template tool to Serverless spark

🛠️ Part of https://github.com/googleapis/genai-toolbox/issues/2405

### DIFF
--- a/.github/workflows/docs_preview_build.yaml
+++ b/.github/workflows/docs_preview_build.yaml
@@ -1,0 +1,93 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Build Docs Preview"
+
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs**'
+      - '.hugo/**'
+
+# Read-only token for code execution
+permissions: read-all
+
+jobs:
+  build-preview:
+    runs-on: ubuntu-24.04
+    env:
+      PR_NUMBER: ${{ github.event.number }}
+      HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+    concurrency:
+      group: "preview-${{ github.event.number }}"
+      cancel-in-progress: true
+    defaults:
+      run:
+        working-directory: .hugo
+    steps:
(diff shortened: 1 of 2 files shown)


vinodhinic:
@duwenxin99 Can you please help the integration tests using /gcbrun?

internal/sources/serverlessspark/serverlessspark.go
```diff
@@ -292,3 +305,36 @@ func (s *Source) GetBatch(ctx context.Context, name string) (map[string]any, err
...
+		return nil, fmt.Errorf("failed to unmarshal session template JSON: %w", err)
+	}
+
+	consoleUrl, err := SessionTemplateConsoleURLFromProto(sessionTemplatePb)
+	if err != nil {
+		return nil, fmt.Errorf("error generating console url: %v", err)
+	}
+
+	wrappedResult := map[string]any{
+		"consoleUrl":		consoleUrl,
```
dborowitz: How necessary is this?

I added URLs for batches because there is a lot of info about batch performance, logs, etc. which is not directly captured in the JSON output, and there aren't other MCP servers for, but which a human can usefully access via the Cloud Console. Plus the URLs themselves (particularly Cloud Logging URLs) are hard for an LLM to generate correctly.

My impression for session templates is that all the necessary info is in the JSON response, and even if not, the Console URLs are easier for LLMs to generate.

tests/serverlessspark/serverless_spark_integration_test.go
```diff
@@ -161,6 +161,15 @@ func TestServerlessSparkToolEndpoints(t *testing.T) {
 				"source":       "my-spark",
 				"authRequired": []string{"my-google-auth"},
 			},
+			"get-session-template": map[string]any{
```
duwenxin99, line 182: We should add test cases to call these new tools

internal/sources/serverlessspark/serverlessspark.go
vinodhinic: Got it, I will remove this then,

vinodhinic: This is done.

tests/serverlessspark/serverless_spark_integration_test.go
vinodhinic: Added integration tests.

internal/sources/serverlessspark/serverlessspark.go
```diff
@@ -298,6 +308,58 @@ func (s *Source) GetBatch(ctx context.Context, name string) (map[string]any, err
...
+		sessionTemplate := SessionTemplate{
+			Name:        sessionTemplatePb.Name,
+			Description: sessionTemplatePb.Description,
+			Creator:     sessionTemplatePb.Creator,
+			CreateTime:  sessionTemplatePb.CreateTime.AsTime().Format(time.RFC3339),
+		}
+		sessionTemplates = append(sessionTemplates, sessionTemplate)
+	}
+	return sessionTemplates, nil
+}
```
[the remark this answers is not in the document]
vinodhinic: Ack. this helper function was also added for list-session-templates tool. It will be used within sources/ in the follow up PR.

PR #2310, opened 2026-01-14
fix(embeddingModel): add embedding model to MCP handler

duwenxin99:
- Add embedding model to mcp handlers
- Add integration tests


internal/server/mcp/v20241105/method.go
Yuan325, line 1: Are we only adding it to v20241105?

duwenxin99: WIll add it to all the other versions as well.

internal/server/mcp/v20250618/method.go
Yuan325, line 1: `internal/server/map/v20251125` just got merged~ please add that as well~

Yuan325 (approved):

PR #2314, opened 2026-01-15
chore: add preprocessing function to convert tools file

Yuan325:
Add preprocessing function during parsing of tools file to convert v2 tools file.

Related #817

Future updates will include: 
* Update unmarshal function for ToolsFile to convert config file (test will fail since the yaml tag is not yet updated).
* Update yaml tag (test will pass).

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -449,7 +449,7 @@ func convertToolsFile(raw []byte) ([]byte, error) {
 			}
 			// check if the key is config file v1's key
 			if slices.Contains(v1keys, key) {
-				// check if value convertion to yaml.MapSlice successfully
+				// check if value conversion to yaml.MapSlice successfully
 				// fields such as "tools" in toolsets might pass the first check but
 				// fail to convert to MapSlice
 				if slice, ok := item.Value.(yaml.MapSlice); ok {


cmd/test.db
averikitsch, line 1: Is this file required?

cmd/root.go
```diff
@@ -425,6 +426,106 @@ func parseEnv(input string) (string, error) {
...
+	resourceOrder := []string{}
+	lookup := make(map[string]yaml.MapSlice)
+	for _, item := range input {
+		key, ok := item.Key.(string)
+		if !ok {
+			return nil, fmt.Errorf("unexpected non-string key in input: %v", item.Key)
+		}
+		if slice, ok := item.Value.(yaml.MapSlice); ok {
+			// convert authSources to authServices
+			if key == "authSources" {
```
averikitsch: Do we need to continue to convert authSources? Should we throw a deprecation error?

cmd/root.go
```diff
@@ -425,6 +426,106 @@ func parseEnv(input string) (string, error) {
...
+			// works even if lookup[key] is nil
+			lookup[key] = append(lookup[key], slice...)
+			// preserving the resource's order of original toolsFile
+			if !slices.Contains(resourceOrder, key) {
+				resourceOrder = append(resourceOrder, key)
+			}
+		} else {
+			// toolsfile is already v2
+			if key == "kind" {
+				return raw, nil
```
averikitsch: Is there any chance we would return early but need to check more keys?

cmd/test.db
Yuan325: removed.

cmd/root.go
Yuan325: @averikitsch do we want to stop support backwards compatible for this?

cmd/root.go
Yuan325: Only if user combine both v1 and v2 formatting in the same config file. Do we want to support that? (I feel like it will get messy) If user have one file in the v1 format, and another in v2, this will still work.

cmd/root.go
Yuan325: keeping for backwards compatibility.

cmd/root.go
averikitsch: I see this as a potential situation. I think we should ensure that this edge case is handled.

Yuan325: Updated, and added test case for that.

cmd/root.go
```diff
@@ -425,6 +426,129 @@ func parseEnv(input string) (string, error) {
...
+			return nil, err
+		}
+		for _, item := range input {
+			key, ok := item.Key.(string)
+			if !ok {
+				return nil, fmt.Errorf("unexpected non-string key in input: %v", item.Key)
+			}
+			// check if the key is config file v1's key
+			if slices.Contains(v1keys, key) {
+				// check if value convertion to yaml.MapSlice successfully
```
averikitsch:
```suggestion
				// check if value conversion to yaml.MapSlice successfully
```

PR #2323, opened 2026-01-16
fix(oracle): enable DML operations and resolve incorrect array type error

### LINKED ISSUE #2026: Oracle tool invocation with DML SQL raises error
### Prerequisites

### Toolbox version

toolbox version 0.21.0

### Environment

Users encountered an “incorrect array type” error when running DML statements (insert, delete etc.). DQL queries worked fine (e.g select). 
Log shows that the parameter is passed in as a string variable correctly, so the issue may happen when the driver is parsing a response from the Oracle server. 

### Client

_No response_

### Expected Behavior

DML queries should be supported

### Current Behavior

Image
When using this configuration, the following error is raised:
Image

### Steps to reproduce?

1. Set up tools.yaml with a oracle-sql or oracle-execute-sql tool that uses a DML statement.
2. Invoke the tool

### Additional Details

_No response_

ishatilwani1301:
This pull request resolves an issue where executing DML statements (UPDATE, INSERT, DELETE) caused "incorrect array type" errors. Previously, the Oracle source forced all operations to use QueryContext, expecting rows to be returned. This caused failures when running write operations that do not return rows.

**Changes Implemented:**
The modification updates the Oracle source and tool definitions to distinguish between Read-Only (SELECT) and Action (DML) operations.

- **internal/sources/oracle/oracle.go:**

1. Updated RunSQL signature to accept a readOnly boolean.
2. Added conditional logic: If readOnly is false, it now uses ExecContext and returns rows_affected instead of attempting to scan non-existent rows.

- **internal/tools/oracle/oraclesql/oraclesql.go:**

1. Added a readonly field to the tool configuration (YAML).
2. Updated the Invoke method to pass this flag to the source. Defaults to true (Read-Only) for backward compatibility.

- **internal/tools/oracle/oraclesql/oraclesql_test.go:**

1. Added unit tests to verify that readonly: false is correctly parsed from the YAML configuration.

🛠️ Fixes #2026


internal/tools/oracle/oraclesql/oraclesql.go
```diff
@@ -41,6 +41,7 @@ type Config struct {
 	Source             string                `yaml:"source" validate:"required"`
 	Description        string                `yaml:"description" validate:"required"`
 	Statement          string                `yaml:"statement" validate:"required"`
+	ReadOnly           *bool                 `yaml:"readonly"`
```
duwenxin99:
```suggestion
	ReadOnly           *bool                 `yaml:"readOnly"`
```

internal/tools/oracle/oraclesql/oraclesql_test.go
```diff
@@ -65,6 +69,29 @@ func TestParseFromYamlOracleSql(t *testing.T) {
...
 		},
+		{
+			desc: "example with readonly flag set to false (DML)",
+			in: `
+            tools:
+                update_user:
+                    kind: oracle-sql
+                    source: db-prod
+                    description: Updates user email.
+                    readonly: false   # <--- We are testing this specific line
```
duwenxin99:
```suggestion
                    readOnly: false   # <--- We are testing this specific line
```

internal/tools/oracle/oraclesql/oraclesql.go
duwenxin99: If the default is false, we should have test cases:
1. does not specify readOnly and check if it is false
2. specify readOnly to true

duwenxin99 (changes requested):
We should also update the documentation to reflect this change. If the coverage is failing, we should look into how to raise the coverage. 20% is not a very high threshold.

internal/tools/oracle/oraclesql/oraclesql_test.go
ishatilwani1301: Fixed

internal/tools/oracle/oraclesql/oraclesql.go
ishatilwani1301: The default value is true. I have added three test  cases:

    1. does not specify readOnly and checking if it is working
    2. specified readOnly to true
    3. specified readOnly to false

Still coverage is same 19.6% . Working on it to create more test cases.

internal/tools/oracle/oraclesql/oraclesql_test.go
```diff
@@ -11,6 +11,10 @@ import (
 	"github.com/googleapis/genai-toolbox/internal/tools/oracle/oraclesql"
 )
 
+func boolPtr(b bool) *bool {
```
duwenxin99: If this is just used in one place, lets remove the helper and just do it in place

ishatilwani1301: Done

internal/tools/oracle/oraclesql/oraclesql_test.go
```diff
@@ -60,6 +65,53 @@ func TestParseFromYamlOracleSql(t *testing.T) {
...
+			statement: "SELECT * FROM orders"
+			`,
+			want: server.ToolConfigs{
+				"safe_query": oraclesql.Config{
+					Name:         "safe_query",
+					Type:         "oracle-sql",
+					Source:       "db-prod",
+					Description:  "Safe read operation.",
+					Statement:    "SELECT * FROM orders",
+					ReadOnly:     &valTrue,
```
duwenxin99, line 91: Can we just specify "true" and "false" directly?

tests/oracle/oracle_integration_test.go
```diff
@@ -242,3 +285,47 @@ func dropAllUserTables(t *testing.T, ctx context.Context, db *sql.DB) {
 		}
 	}
 }
+
+// runInvokeToolTest sends a JSON-RPC request to the running test server
+// and asserts that the response body contains the expected substring.
+func runInvokeToolTest(t *testing.T, toolName, paramsJSON, wantResponseSubStr string) {
```
duwenxin99: Can we be more specific by renaming this test to something like `testDmlQueries`

tests/oracle/oracle_integration_test.go
```diff
@@ -103,6 +106,39 @@ func TestOracleSimpleToolEndpoints(t *testing.T) {
 	tmplSelectCombined, tmplSelectFilterCombined := tests.GetMySQLTmplToolStatement()
 	toolsFile = tests.AddTemplateParamConfig(t, toolsFile, OracleToolType, tmplSelectCombined, tmplSelectFilterCombined, "")
 
+	// Configure a DML tool (Update) to verify the 'readonly: false' logic.
```
duwenxin99: Can we clean up the comments a little? Only save those that are necessary

internal/sources/oracle/oracle_test.go
```diff
@@ -191,3 +192,36 @@ func TestFailParseFromYaml(t *testing.T) {
 		})
 	}
 }
+
+// TestRunSQLExecutesDML verifies that RunSQL correctly routes operations to
+// ExecContext instead of QueryContext when the readOnly flag is set to false.
+func TestRunSQLExecutesDML(t *testing.T) {
```
duwenxin99, line 198: Not sure if this will work... Have you run the unit test locally?

internal/tools/oracle/oraclesql/oraclesql_test.go
ishatilwani1301: We cannot directly specify "true" and "false" because ReadOnly is a pointer(*bool). We can again add a small helper function that we removed boolPtr(b bool) *bool so we can write ReadOnly: boolPtr(true) or ReadOnly: boolPtr(false).

tests/oracle/oracle_integration_test.go
ishatilwani1301: Yes, we can. I have renamed it.

tests/oracle/oracle_integration_test.go
ishatilwani1301: Done

internal/sources/oracle/oracle_test.go
ishatilwani1301: Yes I have run this locally, it is passing.

internal/tools/oracle/oraclesql/oraclesql_test.go
duwenxin99: Why does this field need to be a pointer though?

ishatilwani1301: This field is a pointer because if used as a regular bool, an omitted field would automatically default to false. Since false grants write access, that would make the tool dangerous by default. Using *bool means an omitted field is nil, allowing us to safely default to true (read-only) instead.

PR #2331, opened 2026-01-20
chore(ui): prevent script execution in Toolbox UI rendering

threatpointer:
# Defensive Security Hardening: Prevent Script Execution in Toolbox UI Rendering

> **Note:** This issue was identified during security research and reviewed previously.  
> While typical deployments operate within a trusted configuration model, addressing this behavior was recommended as a defense-in-depth improvement. This PR describes the implemented fix.

## Overview

This change improves the safety of the GenAI Toolbox UI by preventing unintended JavaScript execution when rendering values derived from tool configuration files.

Previously, certain fields from tool definitions were rendered directly into HTML contexts without escaping. As a result, tool definitions containing embedded HTML or script payloads could trigger JavaScript execution when viewed in the dashboard. While this occurs within the same trust boundary as the configuration owner, escaping these values by default avoids unexpected execution and improves robustness.

### 1. New Utility
- Added `sanitize.js` which exports a strict `escapeHtml()` function.
- Escapes dangerous characters: `&`, `<`, `>`, `"`, `'`, `/`, `` ` ``.
- Performs strict type checking, rendering `null` and `undefined` values as empty strings.

### 2. Input Handling
- Updated `internal/server/static/js/toolDisplay.js` to wrap `tool.name` and `tool.description` with `escapeHtml()` prior to rendering them into the DOM.

### 3. Error Handling
- Updated `internal/server/static/js/loadTools.js` to sanitize error messages that may reflect user-controlled or derived input before rendering.

## Validation

- Verified behavior using tool definition files containing common script execution vectors.
- Confirmed that embedded HTML and script payloads are rendered as literal text.
- Verified that standard and existing tool definitions continue to render correctly without functional regression.

## Notes

This change is a defense-in-depth hardening measure.  
It does not modify the existing trust model or intended usage patterns, but ensures safer default rendering behavior and avoids unintended script execution in the UI.

## Attribution

**Contributor:** Mohammed Tanveer (threatpointer)

### DIFF
--- a/internal/server/api.go
+++ b/internal/server/api.go
@@ -26,6 +26,7 @@ import (
 	"github.com/go-chi/render"
 	"github.com/googleapis/genai-toolbox/internal/tools"
 	"github.com/googleapis/genai-toolbox/internal/util"
+	"github.com/googleapis/genai-toolbox/internal/util/parameters"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
@@ -231,7 +232,7 @@ func toolInvokeHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	params, err := tool.ParseParams(data, claimsFromAuth)
+	params, err := parameters.ParseParams(tool.GetParameters(), data, claimsFromAuth)
 	if err != nil {
 		// If auth error, return 401
 		if errors.Is(err, util.ErrUnauthorized) {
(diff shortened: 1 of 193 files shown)


threatpointer:
Based on code review comments, updated the PR to also sanitize `authProfileName` in `createGoogleAuthMethodItem()` within `auth.js`.

The `escapeHtml()` utility is now applied consistently across all user-facing string interpolations.

Yuan325 (commented):
@threatpointer Thank you for your contribution!  :)

Yuan325 (approved):

PR #2340, opened 2026-01-21
feat(prebuiltconfigs/alloydb-omni): Implement Alloydb omni dataplane tools

ankel:
## Description

Implement Dataplane tools for alloydb omni using Postgres source and postgres tools. I also add more omni-specific tool (columnar engine).

### DIFF
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -923,7 +923,7 @@ steps:
         tar -C /usr/local -xzf go.tar.gz
         export PATH="/usr/local/go/bin:$$PATH"
 
-        go test -v ./internal/sources/oracle/... \
+        go test -v ./tests/oracle/... \
           -coverprofile=oracle_coverage.out \
           -coverpkg=./internal/sources/oracle/...,./internal/tools/oracle/...
 
@@ -931,8 +931,8 @@ steps:
         total_coverage=$(go tool cover -func=oracle_coverage.out | grep "total:" | awk '{print $3}')
         echo "Oracle total coverage: $total_coverage"
         coverage_numeric=$(echo "$total_coverage" | sed 's/%//')
-        if awk -v cov="$coverage_numeric" 'BEGIN {exit !(cov < 20)}'; then
-            echo "Coverage failure: $total_coverage is below 20%."
+        if awk -v cov="$coverage_numeric" 'BEGIN {exit !(cov < 60)}'; then
+            echo "Coverage failure: $total_coverage is below 60%."
             exit 1
         fi
 


tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/googleapis/genai-toolbox/internal/testutils"
+	"github.com/googleapis/genai-toolbox/tests"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+var (
+	AlloyDBSourceKind = "alloydb-omni-source"
```
Yuan325:
```suggestion
	AlloyDBSourceType = "alloydb-omni-source"
```

We had recently refactored our configuration (ref: https://github.com/googleapis/genai-toolbox/pull/2369)

note: there won;t be need to update the prebuilt tools config file yet, we will run the migration eventually.

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	AlloyDBToolKind   = "postgres-sql"
```
Yuan325:
```suggestion
	AlloyDBToolType   = "postgres-sql"
```

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	AlloyDBUser       = "postgres"
+	AlloyDBPass       = "mysecretpassword"
```
Yuan325: Do we need this? seems like password is optional

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	AlloyDBDatabase   = "postgres"
+)
+
+// Copied over from postgres.go
+func initAlloyDBConnectionPool(host, port, user, pass, dbname string) (*pgxpool.Pool, error) {
```
Yuan325: nit: since this is technically using the postgres source~

```suggestion
func initPostgresConnectionPool(host, port, user, pass, dbname string) (*pgxpool.Pool, error) {
```

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	mappedPort, err := container.MappedPort(ctx, "5432")
+	if err != nil {
+		cleanup()
+		t.Fatalf("failed to get container mapped port: %s", err)
+	}
+
+	return host, mappedPort.Port(), cleanup
+}
+
+func TestAlloyDB(t *testing.T) {
```
Yuan325:
```suggestion
func TestAlloyDBOmni(t *testing.T) {
```

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	args := []string{"--prebuilt", "alloydb-omni"}
+
+	pool, err := initAlloyDBConnectionPool(AlloyDBHost, AlloyDBPort, AlloyDBUser, AlloyDBPass, AlloyDBDatabase)
+	if err != nil {
+		t.Fatalf("unable to create alloydb connection pool: %s", err)
+	}
+
+	// cleanup test environment
+	tests.CleanupPostgresTables(t, ctx, pool)
```
Yuan325:
```suggestion
	}
```

This is for tests that are running in a static instance. Since we are using testcontainers, I don't think we need to run this.

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	waitCtx, waitCancel := context.WithTimeout(ctx, 30*time.Second)
+	defer waitCancel()
+
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	// Run Postgres prebuilt tool tests
```
Yuan325, line 136: Are there any reasons we are leaving out certain tests (RunPostgresListTablesTest, etc.)?

tests/alloydbomni/alloydb_omni_integration_test.go
Yuan325, line 1: Thank you for adding tests! :)

Yuan325 (changes requested):
Hi, thank you for your contributions!

We're missing the following:
1. unit test for prebuilt tools in `cmd/root_test.go` (ref: https://github.com/googleapis/genai-toolbox/blob/cf477b529a6158cbb19994350c5c2318ff2b81c9/cmd/root_test.go#L1355)

2. documentation for prebuilt tools (https://googleapis.github.io/genai-toolbox/reference/prebuilt-tools/): Code for the docs is at `docs/en/reference/prebuilt-tools.md`

Thank you!

tests/alloydbomni/alloydb_omni_integration_test.go
ankel: Two things that are left out:

* The RunPostgresListTable depends on the tables that are created before running the test. I left them out because I thought they are not super relevant to the current test. I can add another test to cover that case if needed
* The RunPostgresListQueryStatsTest relies on pg_stat_statements extension which is enabled by default in cloud but disabled by default in Omni, so that's why it couldn't run.

tests/alloydbomni/alloydb_omni_integration_test.go
ankel: Password is indeed optional, but in this case, it's passed in as an env var to the container (line 63) and is also used in the postgres url (line 42). 

Happy to remove it if you think we should trim it out.

tests/alloydbomni/alloydb_omni_integration_test.go
ankel: My pleasure!

tests/alloydbomni/alloydb_omni_integration_test.go
```diff
@@ -0,0 +1,157 @@
...
+		Host:   fmt.Sprintf("%s:%s", host, port),
+		Path:   dbname,
+	}
+	pool, err := pgxpool.New(context.Background(), url.String())
+	if err != nil {
+		return nil, fmt.Errorf("Unable to create connection pool: %w", err)
+	}
+
+	return pool, nil
+}
```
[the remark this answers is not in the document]
Yuan325: It's okay for now to keep this.

PR #2344, opened 2026-01-21
chore: update unmarshal function for ToolsFile

Yuan325:
Update `parseToolsFile()` to use the file conversion function. Update unmarshalling function since we will be unmarshalling per-doc (separated by `---`).

`AuthSources` will be converted to `AuthServices` during the file conversion stage. Hence, removing it from `ToolsFile` struct. ToolsFile v2 will not support `AuthSources`. Double checked that all docs do not reference to `AuthSources` anymore; new users is expected to be using `AuthServices` instead.

This PR will not pass the unit test since the updates for resources's yaml tag will be in subsequent PRs. Breaking it down to keep review simpler. 

Related #817

Upcoming PR:
* Update yaml tag

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -170,7 +170,7 @@ func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, Au
 		var kind, name string
 		var ok bool
 		if kind, ok = resource["kind"].(string); !ok {
-			return nil, nil, nil, nil, nil, nil, fmt.Errorf("missing 'kind' field or it is not a string")
+			return nil, nil, nil, nil, nil, nil, fmt.Errorf("missing 'kind' field or it is not a string: %v", resource)
 		}
 		if name, ok = resource["name"].(string); !ok {
 			return nil, nil, nil, nil, nil, nil, fmt.Errorf("missing 'name' field or it is not a string")
@@ -302,7 +302,11 @@ func UnmarshalYAMLToolConfig(ctx context.Context, name string, r map[string]any)
 
 func UnmarshalYAMLToolsetConfig(ctx context.Context, name string, r map[string]any) (tools.ToolsetConfig, error) {
 	var toolsetConfig tools.ToolsetConfig
-	justTools := map[string]any{"tools": r["tools"]}
+	toolList, ok := r["tools"].([]string)
+	if !ok {
+		return toolsetConfig, fmt.Errorf("tools is missing or not a list of strings: %v", r)
+	}
+	justTools := map[string]any{"tools": toolList}
 	dec, err := util.NewStrictDecoder(justTools)
 	if err != nil {
 		return toolsetConfig, fmt.Errorf("error creating decoder: %s", err)


internal/server/config.go
```diff
@@ -127,269 +128,211 @@ func (s *StringLevel) Type() string {
...
+	// promptset configs is not yet supported
+
+	decoder := yaml.NewDecoder(bytes.NewReader(raw))
+	// for loop to unmarshal documents with the `---` separator
+	for {
+		var resource map[string]any
+		if err := decoder.DecodeContext(ctx, &resource); err != nil {
+			if err.Error() == "EOF" {
+				break
 			}
```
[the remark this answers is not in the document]
Yuan325: resolve to `if err == io.EOF`

cmd/root.go
```diff
@@ -395,7 +395,6 @@ func NewCommand(opts ...Option) *Command {
 
 type ToolsFile struct {
 	Sources         server.SourceConfigs         `yaml:"sources"`
-	AuthSources     server.AuthServiceConfigs    `yaml:"authSources"` // Deprecated: Kept for compatibility.
```
averikitsch, line 398: We should track this deprecation/breaking change

internal/server/config.go
```diff
@@ -127,269 +129,211 @@ func (s *StringLevel) Type() string {
...
-		}
-		switch kind {
-		case google.AuthServiceType:
-			actual := google.Config{Name: name}
-			if err := dec.DecodeContext(ctx, &actual); err != nil {
-				return fmt.Errorf("unable to parse as %q: %w", kind, err)
+// PromptConfigs is a type used to allow unmarshal of the prompt configs
+type PromptsetConfigs map[string]prompts.PromptsetConfig
+
+func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, AuthServiceConfigs, EmbeddingModelConfigs, ToolConfigs, ToolsetConfigs, PromptConfigs, error) {
```
averikitsch, line 150: nit: this in the future it may be better to refactor this to return a struct of configs instead of managing all outputs

internal/server/config.go
```diff
@@ -127,269 +129,211 @@ func (s *StringLevel) Type() string {
...
-
-	for name, u := range raw {
-		// Unmarshal to a general type that ensure it capture all fields
-		var v map[string]any
-		if err := u.Unmarshal(&v); err != nil {
-			return fmt.Errorf("unable to unmarshal embedding model %q: %w", name, err)
+		var kind, name string
+		var ok bool
+		if kind, ok = resource["kind"].(string); !ok {
+			return nil, nil, nil, nil, nil, nil, fmt.Errorf("missing 'kind' field or it is not a string")
```
averikitsch: nit: can we tell the user which one failed?

internal/server/config.go
```diff
@@ -127,269 +129,211 @@ func (s *StringLevel) Type() string {
...
+	toolCfg, err := tools.DecodeConfig(ctx, typeStr, name, dec)
+	if err != nil {
+		return nil, err
+	}
+	return toolCfg, nil
+}
 
+func UnmarshalYAMLToolsetConfig(ctx context.Context, name string, r map[string]any) (tools.ToolsetConfig, error) {
+	var toolsetConfig tools.ToolsetConfig
+	justTools := map[string]any{"tools": r["tools"]}
```
averikitsch: It may be best practice to first validate r["tools]

cmd/root.go
Yuan325: @averikitsch it will still be supported during the convertToolsFile stage (authSources will be converted to authServices). Due to this, we can remove it from here. It won't be considered as breaking change for users.

Added `Deprecated:` to the comment of the code so that it could be easily searched in the future when we decide to remove them.

internal/server/config.go
Yuan325: Agree, we'll have to move ToolsFile out of the `cmd` package in order to do that. Or else there might be circular dependancy.

internal/server/config.go
Yuan325: will print the resource map

PR #2353, opened 2026-01-22
feat(cli/invoke): add support for direct tool invocation from CLI

helloeve:
## Description

This PR introduces a new subcommand, invoke, to the toolbox CLI. This feature allows developers to execute tools defined in their configuration directly from the command line.

- New Subcommand: Implemented invoke as subcommand, which handles tool lookup, parameter unmarshaling from JSON, and invocation.
- Persistent Configuration Flags: Updated cmd/root.go to make flags like --tools-file, --tools-folder, and --prebuilt persistent, allowing them to be used with subcommands.
- Testing: Added unit tests for various scenarios
- Documentation: Created a new "how-to" guide for CLI tool testing and updated the CLI reference documentation.

### DIFF
--- a/cmd/invoke_tool_test.go
+++ b/cmd/invoke_tool_test.go
@@ -1,0 +1,131 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestInvokeTool(t *testing.T) {
+	// Create a temporary tools file
+	tmpDir := t.TempDir()
+
+	toolsFileContent := `
+sources:
+  my-sqlite:
+    kind: sqlite
+    database: test.db
+tools:
+  hello-sqlite:
+    kind: sqlite-sql
+    source: my-sqlite
+    description: "hello tool"
+    statement: "SELECT 'hello' as greeting"
+  echo-tool:
+    kind: sqlite-sql
+    source: my-sqlite
+    description: "echo tool"
+    statement: "SELECT ? as msg"
+    parameters:
+      - name: message
+        type: string
+        description: message to echo
+`
+
+	toolsFilePath := filepath.Join(tmpDir, "tools.yaml")
+	if err := os.WriteFile(toolsFilePath, []byte(toolsFileContent), 0644); err != nil {
+		t.Fatalf("failed to write tools file: %v", err)
+	}
+
+	tcs := []struct {
+		desc    string
+		args    []string
+		want    string
+		wantErr bool
+		errStr  string
+	}{
+		{
+			desc: "success - basic tool call",
+			args: []string{"invoke", "hello-sqlite", "--tools-file", toolsFilePath},
+			want: `"greeting": "hello"`,
+		},
+		{
+			desc: "success - tool call with parameters",
+			args: []string{"invoke", "echo-tool", `{"message": "world"}`, "--tools-file", toolsFilePath},
+			want: `"msg": "world"`,
+		},
+		{
+			desc:    "error - tool not found",
+			args:    []string{"invoke", "non-existent", "--tools-file", toolsFilePath},
+			wantErr: true,
+			errStr:  `tool "non-existent" not found`,
+		},
+		{
+			desc:    "error - invalid JSON params",
+			args:    []string{"invoke", "echo-tool", `invalid-json`, "--tools-file", toolsFilePath},
+			wantErr: true,
+			errStr:  `params must be a valid JSON string`,
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			_, got, err := invokeCommandWithContext(context.Background(), tc.args)
+			if (err != nil) != tc.wantErr {
+				t.Fatalf("got error %v, wantErr %v", err, tc.wantErr)
+			}
+			if tc.wantErr && !strings.Contains(err.Error(), tc.errStr) {
+				t.Fatalf("got error %v, want error containing %q", err, tc.errStr)
+			}
+			if !tc.wantErr && !strings.Contains(got, tc.want) {
+				t.Fatalf("got %q, want it to contain %q", got, tc.want)
+			}
+		})
+	}
+}
+
+func TestInvokeTool_AuthUnsupported(t *testing.T) {
+	tmpDir := t.TempDir()
+	toolsFileContent := `
+sources:
+  my-bq:
+    kind: bigquery
+    project: my-project
+    useClientOAuth: true
+tools:
+  bq-tool:
+    kind: bigquery-sql
+    source: my-bq
+    description: "bq tool"
+    statement: "SELECT 1"
+`
+	toolsFilePath := filepath.Join(tmpDir, "auth_tools.yaml")
+	if err := os.WriteFile(toolsFilePath, []byte(toolsFileContent), 0644); err != nil {
+		t.Fatalf("failed to write tools file: %v", err)
+	}
+
+	args := []string{"invoke", "bq-tool", "--tools-file", toolsFilePath}
+	_, _, err := invokeCommandWithContext(context.Background(), args)
+	if err == nil {
+		t.Fatal("expected error for tool requiring client auth, but got nil")
+	}
+	if !strings.Contains(err.Error(), "client authorization is not supported") {
+		t.Fatalf("unexpected error message: %v", err)
+	}
+}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -34,6 +34,7 @@ import (
 	"github.com/fsnotify/fsnotify"
 	yaml "github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/internal/auth"
+	"github.com/googleapis/genai-toolbox/internal/cli/invoke"
 	"github.com/googleapis/genai-toolbox/internal/embeddingmodels"
 	"github.com/googleapis/genai-toolbox/internal/log"
 	"github.com/googleapis/genai-toolbox/internal/prebuiltconfigs"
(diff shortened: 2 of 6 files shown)


averikitsch (commented):
@Yuan325 can you do a first pass of this code review?

internal/cli/invoke/command.go
Yuan325, line 1: Might be leaning towards following the modular layout of organizing the subcommands since we will be adding more subcommand during the implementation of advance control plane. (https://cobra.dev/docs/how-to-guides/working-with-commands/ check out the "How to Organize Commands in Packages" section)

cmd/call_tool.go
```diff
@@ -0,0 +1,140 @@
...
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(c *cobra.Command, args []string) error {
+			return runCallTool(rootCmd, args)
+		},
+	}
+	return cmd
+}
+
+func runCallTool(cmd *Command, args []string) error {
+	ctx := cmd.Context()
```
Yuan325:
```suggestion
	ctx := context.WithCancel(cmd.Context())
	defer cancel()
```

cmd/root.go
```diff
@@ -360,26 +360,28 @@ func NewCommand(opts ...Option) *Command {
...
 	flags := cmd.Flags()
+	persistentFlags := cmd.PersistentFlags()
+
 	flags.StringVarP(&cmd.cfg.Address, "address", "a", "127.0.0.1", "Address of the interface the server will listen on.")
 	flags.IntVarP(&cmd.cfg.Port, "port", "p", 5000, "Port the server will listen on.")
 
-	flags.StringVar(&cmd.tools_file, "tools_file", "", "File path specifying the tool configuration. Cannot be used with --tools-files, or --tools-folder.")
+	persistentFlags.StringVar(&cmd.tools_file, "tools_file", "", "File path specifying the tool configuration. Cannot be used with --tools-files, or --tools-folder.")
 	// deprecate tools_file
 	_ = flags.MarkDeprecated("tools_file", "please use --tools-file instead")
```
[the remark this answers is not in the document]
Yuan325: @averikitsch Are we good that the new subcommands will not support the deprecated `tools_file` flag?

cmd/call_tool.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	ctx = util.WithInstrumentation(ctx, instrumentation)
+
+	// Load and merge tool configurations
+	if err := cmd.loadAndMergeConfig(ctx); err != nil {
+		return err
+	}
+
+	// Initialize Resources
+	sourcesMap, authServicesMap, embeddingModelsMap, toolsMap, toolsetsMap, promptsMap, promptsetsMap, err := server.InitializeConfigs(ctx, cmd.cfg)
```
Yuan325: With this implementation, it seems like Toolbox will initialize the whole server every time this subcommand it's called. Is this intended? Or do we want to use this subcommand as like a "HTTP Client" shim that will invoke the underlying endpoint of a deployed server?

docs/en/reference/cli.md
```diff
@@ -30,6 +30,21 @@ description: >
 |              | `--user-agent-extra`       | Appends additional metadata to the User-Agent.                                                                                                                                   |             |
 | `-v`         | `--version`                | version for toolbox                                                                                                                                                              |             |
 
+## Sub Commands
+
+### `call-tool`
+
+Executes a tool directly with the provided parameters. This is useful for testing tool configurations and parameters without needing a full client setup.
+
+**Usage:**
```
```suggestion
**Syntax**
```

docs/en/reference/cli.md
```diff
@@ -30,6 +30,21 @@ description: >
 |              | `--user-agent-extra`       | Appends additional metadata to the User-Agent.                                                                                                                                   |             |
 | `-v`         | `--version`                | version for toolbox                                                                                                                                                              |             |
 
+## Sub Commands
```
Yuan325, line 33: note: we can keep this for now, but eventually might want to move to a drop down of subcommand references when this increases.

cmd/call_tool.go
Yuan325: This is with the assumption that the config only consists of tools within the skill itself.

internal/cli/invoke/command.go
```diff
@@ -0,0 +1,140 @@
...
+		if err := json.Unmarshal([]byte(paramsInput), &params); err != nil {
+			return fmt.Errorf("params must be a valid JSON string: %w", err)
+		}
+	}
+
+	parsedParams, err := tool.ParseParams(params, nil)
+	if err != nil {
+		return fmt.Errorf("invalid parameters: %w", err)
+	}
+
```
Yuan325, line 123: Missing the embedding model code (ref: https://github.com/googleapis/genai-toolbox/commit/17b70ccaa754d15bcc33a1a3ecb7e652520fa600).

_note: Wondering if there's a way we can standardize these code, future update might be missing here._

cmd/root.go
Yuan325: Might want to allow update to user agent metadata?

docs/en/how-to/invoke_tool.md
```diff
@@ -0,0 +1,70 @@
...
+
+### 3. Using Prebuilt Configurations
+
+You can also use the `--prebuilt` flag to load prebuilt toolsets.
+
+```bash
+toolbox --prebuilt cloudsql-postgres call-tool cloudsql-postgres-list-instances
+```
+
+## Troubleshooting
```
Yuan325, line 70: Do we want to add a "notice tip" somewhere to recommend users to keep the tools.yaml configuration file as minimal as possible since this will initialize all tools during a call.

Yuan325 (changes requested):

cmd/root.go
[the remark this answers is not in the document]
helloeve: good call out - I am leaning towards to not introducing this in this new subcommand since this is a new capability so there is no previous usage this will break (modified to have this not included in persistentflags)

internal/cli/invoke/command.go
helloeve: Totally agree, updated this logic to be introduced within internal/cli/invoke/command.go

internal/cli/invoke/command.go
helloeve: I may have missed something but the referenced logic shall be captured in the shared function loadAndMergeConfig created in root.go.

Also to reduce the maintenance cost I have further extract out common logic such as logger, telemetry into a separate function.

docs/en/reference/cli.md
helloeve: noted

docs/en/how-to/invoke_tool.md
helloeve: good point, added one at the top of the doc - I think later on I can think about optimizing the resource loading a bit to only consider the relevant ones for that specific tool

internal/cli/invoke/command.go
Yuan325: ahh, not the tools file for embedding model but the code within api to resolve embeded parameters: https://github.com/googleapis/genai-toolbox/blob/2d5d33388cb638281b2eab351769332778cc0994/internal/server/api.go#L250

cmd/root.go
```diff
@@ -361,37 +362,49 @@ func NewCommand(opts ...Option) *Command {
...
 	cmd.RunE = func(*cobra.Command, []string) error { return run(cmd) }
 
+	// Register subcommands for tool invocation
+	baseCmd.AddCommand(invoke.NewCommand(invoke.Dependencies{
+		Cfg:        &cmd.cfg,
+		Out:        cmd.outStream,
+		Err:        cmd.errStream,
+		Version:    versionString,
+		LoadConfig: cmd.loadAndMergeConfig,
+		Setup:      cmd.Setup,
```
Yuan325: This dependencies might be tough to maintain in the long run. Plus we might need to add it to subsequent subcommands as well. Instead, we can use `PersistentPreRunE`.

cmd/root.go
```diff
@@ -361,37 +362,49 @@ func NewCommand(opts ...Option) *Command {
...
 	flags.BoolVar(&cmd.cfg.DisableReload, "disable-reload", false, "Disables dynamic reloading of tools file.")
 	flags.BoolVar(&cmd.cfg.UI, "ui", false, "Launches the Toolbox UI web server.")
 	// TODO: Insecure by default. Might consider updating this for v1.0.0
 	flags.StringSliceVar(&cmd.cfg.AllowedOrigins, "allowed-origins", []string{"*"}, "Specifies a list of origins permitted to access this server. Defaults to '*'.")
 	flags.StringSliceVar(&cmd.cfg.AllowedHosts, "allowed-hosts", []string{"*"}, "Specifies a list of hosts permitted to access this server. Defaults to '*'.")
-	flags.StringSliceVar(&cmd.cfg.UserAgentMetadata, "user-agent-metadata", []string{}, "Appends additional metadata to the User-Agent.")
+	persistentFlags.StringSliceVar(&cmd.cfg.UserAgentMetadata, "user-agent-metadata", []string{}, "Appends additional metadata to the User-Agent.")
 
 	// wrap RunE command so that we have access to original Command object
 	cmd.RunE = func(*cobra.Command, []string) error { return run(cmd) }
```
Yuan325, line 400:
```
	cmd.PersistentPreRunE = func(*cobra.Command, []string) error {return preRun(cmd) }
```
Within the `preRun()` function, we can run all the setup such as the `loadAndMergeConfig()`, and `Setup()`. Since the flag is persistent, all command/subcommand should be able to access it with example: `cmd.Flags().GetString("key-name")`. 

From what I understand, once we set the `PersistentPreRunE`, it will run before every subcommand's `RunE`.

The tricker part is the `shutdownFunc`, since we will need to place it in defer() to make sure that it is ran in the case when server crashes (`PersistentPostRunE` will not run if server crash during `RunE`)

```
type LifecycleManager struct {
    Shutdown func()
}

func (m *LifecycleManager) PreRun(*Command) error {
    // set up logger
    // set up telemetry
    // add the shutdownFunc into `m.Shutdown`
}

func (m *LifecycleManager) Shutdow(){
    // run `m.Shutdown()
}
```

And in `Execute()`
```
lm := lifecycleManager{}
baseCmd := NewCommand()

defer lm.Shutdown()

if err := baseCmd.Execute();err != nil.....
```

internal/cli/invoke/command.go
```diff
@@ -0,0 +1,131 @@
...
+	Err io.Writer
+	// LoadConfig is a function that loads and merges configuration
+	LoadConfig func(context.Context) error
+	// Setup initializes the environment (logger, telemetry, etc.)
+	Setup func(context.Context) (context.Context, func(context.Context) error, error)
+	// Version is the application version string
+	Version string
+}
+
+func NewCommand(deps Dependencies) *cobra.Command {
```
Yuan325: We might need to do `in := os.Stdin` and `cmd.SetIn()` for all the in, out, and err here too.

Or declare a new package (should be okay to place it in `internal/util/iostreams/iostreams.go` or `cmd/iostreams/iostreams.go`), and import it `func NewCommand(streams iostreams.IOStreams) *cobra.Command {}` with setting `cmd.SetIn(streams.In)`.

```
package iostreams

import (
    "io"
    "os"
)

type IOStreams struct {
    In     io.Reader
    Out    io.Writer
    ErrOut io.Writer
}

// System returns the standard OS streams (used in main.go)
func System() IOStreams {
    return IOStreams{
        In:     os.Stdin,
        Out:    os.Stdout,
        ErrOut: os.Stderr,
    }
}
```

With this, in the `Execute()` function, we can do the following:
```
streams := iostreams.System()
baseCmd := NewCommand(streams)
if err := baseCmd.Execute(); err != nil {
```

And then, we can do `baseCmd.AddCommand(invoke.NewCommand(streams))`

This should work and might be cleaner

cmd/invoke_tool_test.go
```diff
@@ -0,0 +1,133 @@
...
+			desc:    "error - invalid JSON params",
+			args:    []string{"invoke", "echo-tool", `invalid-json`, "--tools-file", toolsFilePath},
+			wantErr: true,
+			errStr:  `params must be a valid JSON string`,
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			_, got, err := invokeCommandWithContext(context.Background(), tc.args)
```
Yuan325, line 89: where is the `invokeCommandWithContext()` function?

Yuan325 (changes requested):
Please let me know if you need any clarifications! Thank you!

internal/cli/invoke/command.go
helloeve: gotcha, updated!

cmd/invoke_tool_test.go
helloeve: this is in root_test.go and used for other tests

cmd/root.go
Yuan325: optimization will be done on future iteration

internal/cli/invoke/command.go
```diff
@@ -0,0 +1,141 @@
...
+		},
+	}
+	return cmd
+}
+
+func runInvoke(cmd *cobra.Command, args []string, rootCmd RootCommand) error {
+	ctx, cancel := context.WithCancel(cmd.Context())
+	defer cancel()
+
+	ctx, shutdown, err := rootCmd.Setup(ctx)
```
Yuan325, line 71: are we planning to log errors with `logger.ErrorContext()`? Or is this not necessary if the error will just be displayed directly?

internal/cli/invoke/command.go
```diff
@@ -0,0 +1,141 @@
...
+// direct coupling to the root command's implementation.
+type RootCommand interface {
+	// Config returns a copy of the current server configuration.
+	Config() server.ServerConfig
+
+	// Out returns the writer used for standard output.
+	Out() io.Writer
+
+	// Err returns the writer used for error output.
+	Err() io.Writer
```
Yuan325: Doesn't seem like this is used? should we just remove it?

Yuan325 (approved):
some last clarifications~ otherwise LGTM

internal/cli/invoke/command.go
helloeve: Ah good point, updated to have errors logged

internal/cli/invoke/command.go
helloeve: right - removed

PR #2362, opened 2026-01-23
chore: update docs and root test

Yuan325:
Update docs and `cmd/root_test.go`.


README.md
averikitsch, line 1: nit: it would be helpful to add a notice to the readme to describe the changes or we could link to a migration faq

Yuan325: Will add in subsequent PR to keep things cleaner and ease review processes.

PR #2369, opened 2026-01-27
feat!: update configuration file v2

Yuan325:
This PR introduces a significant update to the Toolbox configuration file format, which is one of the primary **breaking changes** required for the implementation of the Advanced Control Plane.

# Summary of Changes
The configuration schema has been updated to enforce resource isolation and facilitate atomic, incremental updates.
* Resource Isolation: Resource definitions are now separated into individual blocks, using a distinct structure for each resource type (Source, Tool, Toolset, etc.). This improves readability, management, and auditing of configuration files.
* Field Name Modification: Internal field names have been modified to align with declarative methodologies. Specifically, the configuration now separates kind (general resource type, e.g., Source) from type (specific implementation, e.g., Postgres).

# User Impact
Existing tools.yaml configuration files are now in an outdated format. Users must eventually update their files to the new YAML format.

# Mitigation & Compatibility
Backward compatibility is maintained during this transition to ensure no immediate user action is required for existing files.
* Immediate Backward Compatibility: The source code includes a pre-processing layer that automatically detects outdated configuration files (v1 format) and converts them to the new v2 format under the hood.
* [COMING SOON] Migration Support: The new toolbox migrate subcommand will be introduced to allow users to automatically convert their old configuration files to the latest format. 

# Example
Example for config file v2:
```
kind: sources
name: my-pg-instance
type: cloud-sql-postgres
project: my-project
region: my-region
instance: my-instance
database: my_db
user: my_user
password: my_pass
---
kind: authServices
name: my-google-auth
type: google
clientId: testing-id
---
kind: tools
name: example_tool
type: postgres-sql
source: my-pg-instance
description: some description
statement: SELECT * FROM SQL_STATEMENT;
parameters:
- name: country
  type: string
  description: some description
---
kind: tools
name: example_tool_2
type: postgres-sql
source: my-pg-instance
description: returning the number one
statement: SELECT 1;
---
kind: toolsets
name: example_toolset
tools:
- example_tool
```

### DIFF
--- a/internal/tools/cloudsql/cloudsqlrestorebackup/cloudsqlrestorebackup.go
+++ b/internal/tools/cloudsql/cloudsqlrestorebackup/cloudsqlrestorebackup.go
@@ -26,7 +26,7 @@ import (
 	"google.golang.org/api/sqladmin/v1"
 )
 
-const kind string = "cloud-sql-restore-backup"
+const resourceType string = "cloud-sql-restore-backup"
 
 var _ tools.ToolConfig = Config{}
 
@@ -40,15 +40,15 @@ type compatibleSource interface {
 // Config defines the configuration for the restore-backup tool.
 type Config struct {
 	Name         string   `yaml:"name" validate:"required"`
-	Kind         string   `yaml:"kind" validate:"required"`
+	Type         string   `yaml:"type" validate:"required"`
 	Description  string   `yaml:"description"`
 	Source       string   `yaml:"source" validate:"required"`
 	AuthRequired []string `yaml:"authRequired"`
 }
 
 func init() {
-	if !tools.Register(kind, newConfig) {
-		panic(fmt.Sprintf("tool kind %q already registered", kind))
+	if !tools.Register(resourceType, newConfig) {
+		panic(fmt.Sprintf("tool type %q already registered", resourceType))
 	}
 }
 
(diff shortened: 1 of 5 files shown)


tests/mysql/mysql_integration_test.go
```diff
@@ -54,7 +54,7 @@ func getMySQLVars(t *testing.T) map[string]any {
 	}
 
 	return map[string]any{
-		"kind":     MySQLSourceKind,
+		"type":     MySQLSourceType,
```
shuzhou-gc, line 57: just to help myself to understand this change, so the old "kind" is now named "type", and the keyword "kind" is now a concept in a higher level such as "tool vs source", etc... And when we specify something, only "type" is needed, because "kind" is implict through the context, is it?

Yuan325: @shuzhou-gc Yes, the old "kind" is now "type", and "kind" is now a higher level. When specifying in config file, both the "kind" and "type" will be needed. However, user will not have to make any changes for now. We will update the config file under the hood. Soon we will push out a migration code to help user migrate from v1 to v2. :)

Yuan325: I've added an example in the description

shuzhou-gc: Thanks for the details.

PR #2387, opened 2026-01-29
feat(server): add Tool call error categories

duwenxin99:
Create Agent vs Server error types to distinguish between the two types.


internal/util/errors.go
Yuan325, line 1: We might want to add it in a separate package within util? `internal/util/errors` as `package errors`?

Also wondering if we want to move the the `var ErrUnauthorized` (https://github.com/googleapis/genai-toolbox/blob/3f1908a8225d22f29289441e768ea520b55ef347/internal/util/util.go#L192) into this package? (can also move it in the next PR though since we might need to update the import library)

internal/util/errors.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (e *AgentError) Error() string { return e.Msg }
+
+func (e *AgentError) Category() ErrorCategory { return CategoryAgent }
+
+func (e *AgentError) Unwrap() error { return e.Cause }
+
+func NewAgentError(msg string, cause error) *AgentError {
+	return &AgentError{Msg: msg, Cause: cause}
+}
```
Yuan325, line 53:
```suggestion
}

var _ ToolboxError = AgentError{}
```

To validate the interface during compile time, can we add this to the ClientServerError as well?

internal/util/errors.go
```diff
@@ -0,0 +1,61 @@
...
+type AgentError struct {
+	Msg   string
+	Cause error
+}
+
+func (e *AgentError) Error() string { return e.Msg }
+
+func (e *AgentError) Category() ErrorCategory { return CategoryAgent }
+
+func (e *AgentError) Unwrap() error { return e.Cause }
```
Yuan325, line 49: Should we add this function to the interface as well?

Yuan325 (changes requested):

internal/util/errors.go
duwenxin99: ErrUnauthorized will be removed since we are using ClientServerError instead.

internal/util/errors.go
duwenxin99: added

internal/util/errors.go
duwenxin99: I think it's better to keep errors.go in the util package for these reasons:
1. Idiomatic Go prefers packages named after what they provide, not what they are. In this case, errors.go is just part of the provided util.
2. if we named the sub-package errors, it conflicts with the standard library's errors package, making it harder for the developers to use it.

internal/util/errors.go
```diff
@@ -0,0 +1,69 @@
...
+
+func NewAgentError(msg string, cause error) *AgentError {
+	return &AgentError{Msg: msg, Cause: cause}
+}
+
+var _ ToolboxError = &AgentError{}
+
+// ClientServerError returns 4XX/5XX error code
```
Yuan325:
```suggestion
// ClientServerError returns 4XX/5XX error code
```

Duplicated~

PR #2392, opened 2026-01-30
feat(cli/skills): add support for generating agent skills from toolset

helloeve:
## Description

This PR introduces a new skills-generate command that enables users to generate standardized agent skills from their existing Toolbox tool configurations. This facilitates the integration of Toolbox tools into agentic workflows by automatically creating skill descriptions (SKILL.md) and executable wrappers.
- New Subcommand: Implemented skills-generate, which automates the creation of agent skill packages including metadata and executable scripts.
- Skill Generation: Added logic to generate SKILL.md files with parameter schemas and Node.js wrappers for cross-platform tool execution.
- Toolset Integration: Supports selective generation of skills based on defined toolsets, including support for both local files and prebuilt configurations.
- Testing: Added comprehensive unit tests for the generation logic and integration tests for the CLI command.
- Documentation: Created a new "how-to" guide for generating skills and updated the CLI reference documentation.


docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,109 @@
...
+
+```bash
+toolbox [--tools-file <path> | --prebuilt <name>] skills-generate \
+  --name <skill-name> \
+  --toolset <toolset-name> \
+  --description <description> \
+  --output-dir <output-directory>
+```
+
+> **Note:** The `<skill-name>` must follow the Agent Skill naming convention: it must contain only lowercase alphanumeric characters and hyphens, cannot start or end with a hyphen, and cannot contain consecutive hyphens (e.g., `my-skill`, `data-processing`).
```
Yuan325:
```suggestion
{{< notice note >}}
The `<skill-name>` must follow the Agent Skill naming convention: it must contain only lowercase alphanumeric characters and hyphens, cannot start or end with a hyphen, and cannot contain consecutive hyphens (e.g., `my-skill`, `data-processing`).
{{< /notice >}}
```

Will it be better to use a bullet point? Or link towards the agent skills docs with naming conventions?

docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,109 @@
...
+## Before you begin
+
+1. Make sure you have the `toolbox` executable in your PATH.
+2. Make sure you have [Node.js](https://nodejs.org/) installed on your system.
+
+## Generating a Skill from a Toolset
+
+A skill package consists of a `SKILL.md` file (with required YAML frontmatter) and a set of Node.js scripts. Each tool defined in your toolset maps to a corresponding script in the generated skill.
+
+### Command Signature
```
Yuan325:
```suggestion
### Command Usage
```

Will this be a better term? Might need to update the invoke ones too.

docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,109 @@
...
+
+The `skills-generate` command follows this signature:
+
+```bash
+toolbox [--tools-file <path> | --prebuilt <name>] skills-generate \
+  --name <skill-name> \
+  --toolset <toolset-name> \
+  --description <description> \
+  --output-dir <output-directory>
+```
```
Yuan325, line 31: Do we want to add a bullet point that explain each input similar to the invoke subcommand?
Is the toolset name required? By default, Toolbox will generate a toolset with empty string that includes all tools. Can user use that?

docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,109 @@
...
+
+   In this example, the skill contains two Node.js scripts (`tool_a.js` and `tool_b.js`), each mapping to a tool in the original toolset.
+
+### Example: Prebuilt Configuration
+
+You can also generate skills from prebuilt toolsets:
+
+```bash
+toolbox --prebuilt alloydb-postgres-admin skills-generate \
+  --name "alloydb-postgres-admin" \
```
Yuan325, line 96: Will we use the prebuilt name as the agent skill name if its's not provided? Are all these flags required? 

Also it will be great to have the bullet point of each input.

docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,109 @@
...
+  --description "skill for performing administrative operations on alloydb"
+```
+
+## Output Directory
+
+By default, skills are generated in the `skills` directory. You can specify a different output directory using the `--output-dir` flag.
+
+## Shared Node.js Scripts
+
+The `skills-generate` command generates shared Node.js scripts (`.js`) that work across different platforms (Linux, macOS, Windows). This ensures that the generated skills are portable.
```
Yuan325: Wondering if it's better to move this sentence above to "Generating a Skill from a Toolset" section. This feels too small to be a separate section.

docs/en/reference/cli.md
```diff
@@ -45,6 +46,30 @@ toolbox invoke <tool-name> [params]
...
+
+Generates a skill package from a specified toolset. Each tool in the toolset will have a corresponding Node.js execution script in the generated skill.
+
+**Syntax:**
+
+```bash
+toolbox [--tools-file <path> | --prebuilt <name>] skills-generate --name <name> --description <description> --toolset <toolset> --output-dir <output>
+```
+
+**Flags:**
```
Yuan325, line 66: can we add this to the invoke command too~ Thank you!

docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,109 @@
...
+
+```bash
+toolbox --prebuilt alloydb-postgres-admin skills-generate \
+  --name "alloydb-postgres-admin" \
+  --description "skill for performing administrative operations on alloydb"
+```
+
+## Output Directory
+
+By default, skills are generated in the `skills` directory. You can specify a different output directory using the `--output-dir` flag.
```
Yuan325: Might not need this section once we added the bullet point for input above? We can mentioned it in the description for the output-directory input value.

Yuan325: Where will the `skills` directory be? is it a directory in gemini-cli folder? If this `how-to` is specific to Gemini CLI, maybe we want to add that in the title "Generate Agent Skills for Gemini CLI"

docs/en/reference/cli.md
```diff
@@ -45,6 +46,30 @@ toolbox invoke <tool-name> [params]
...
+
+- `--name`: Name of the generated skill.
+- `--description`: Description of the generated skill.
+- `--toolset`: (Optional) Name of the toolset to convert into a skill. If not provided, all tools will be included.
+- `--output-dir`: (Optional) Directory to output generated skills (default: "skills").
+
+For more detailed instructions, see [Generate Agent Skills](../how-to/generate_skill.md).
```
Yuan325, line 73: can we please add this to the invoke ones above too

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+			return cmd.run(c)
+		},
+	}
+
+	cmd.Flags().StringVar(&cmd.name, "name", "", "Name of the generated skill.")
+	cmd.Flags().StringVar(&cmd.description, "description", "", "Description of the generated skill")
+	cmd.Flags().StringVar(&cmd.toolset, "toolset", "", "Name of the toolset (and generated skill folder). If provided, only tools in this toolset are generated.")
+	cmd.Flags().StringVar(&cmd.outputDir, "output-dir", "skills", "Directory to output generated skills")
+
+	cmd.MarkFlagRequired("name")
```
Yuan325: I'm assuming we are not allowing user to use this command with the default toolset then?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+// NewCommand creates a new Command.
+func NewCommand(rootCmd RootCommand) *cobra.Command {
+	cmd := &Command{
+		rootCmd: rootCmd,
+	}
+	cmd.Command = &cobra.Command{
+		Use:   "skills-generate",
+		Short: "Generate skills from tool configurations",
+		RunE: func(c *cobra.Command, args []string) error {
+			return cmd.run(c)
```
Yuan325, line 70:
```suggestion
			return run(c, cmd)
```

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+		errMsg := fmt.Errorf("error getting tools-file flag: %w", err)
+		logger.ErrorContext(ctx, errMsg.Error())
+		return errMsg
+	}
+	prebuiltConfigs, err := cmd.Flags().GetStringSlice("prebuilt")
+	if err != nil {
+		errMsg := fmt.Errorf("error getting prebuilt flag: %w", err)
+		logger.ErrorContext(ctx, errMsg.Error())
+		return errMsg
+	}
```
Yuan325: This might cause an error. tools-file could be empty if use is using prebuilt. Why do we need this here? The `LoadConfig()` should resolve all the configs that are needed and update the `cmd.cfg.SourceConfigs`, `cmd.cfg.AuthServiceConfigs`, etc. with the right configs~?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Load and merge tool configurations
+	if err := c.rootCmd.LoadConfig(ctx); err != nil {
+		return err
+	}
+
+	if len(prebuiltConfigs) == 0 && toolsFile == "" {
+		logger.InfoContext(ctx, "No configurations found to process. Use --tools-file or --prebuilt.")
+		return nil
+	}
```
Yuan325: This should be captured by the `LoadConfig()`?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+	if err := os.MkdirAll(c.outputDir, 0755); err != nil {
```
Yuan325, line 103: Will this work if the outputDir is a pre-existing dir?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, fmt.Errorf("failed to initialize resources: %w", err)
+	}
+
+	resourceMgr := resources.NewResourceManager(sourcesMap, authServicesMap, embeddingModelsMap, toolsMap, toolsetsMap, promptsMap, promptsetsMap)
+
+	result := make(map[string]Tool)
+
+	var toolsToProcess []string
+
+	if c.toolset != "" {
```
Yuan325: there will be a toolset with name "" that includes all the tool. shouldn't need to do an if-else here?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+				Default:     p.GetDefault(),
+				Required:    p.GetRequired(),
+			})
+		}
+
+		manifest := t.Manifest()
+		result[toolName] = Tool{
+			Name:        toolName,
+			Description: manifest.Description,
+			Parameters:  params,
```
Yuan325: why do we generate our own Tool? Why not utilizing the existing ParameterManifest?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+	Default     interface{} `json:"default"`
+	Required    bool        `json:"required"`
+}
+
+// Tool represents a tool.
+type Tool struct {
+	Name        string      `json:"name"`
+	Description string      `json:"description"`
+	Parameters  []Parameter `json:"parameters"`
+}
```
Yuan325: This might be tough to maintain in the long run? Is it possible to use existing structs?

internal/cli/skills/command.go
```diff
@@ -0,0 +1,299 @@
...
+
+// Config represents the structure of the tools.yaml file.
+type Config struct {
+	Sources map[string]interface{}            `yaml:"sources,omitempty"`
+	Tools   map[string]map[string]interface{} `yaml:"tools"`
+}
```
Yuan325: Why not reusing the existing ToolsFile?

Also the `yaml:"sources"` etc is now updated. The config file format had changed with the recent merge #2369

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,253 @@
...
+
+	schemaJSON, err := json.MarshalIndent(schema, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("error generating parameters schema: %w", err)
+	}
+
+	return fmt.Sprintf("## Parameters\n\n```json\n%s\n```", string(schemaJSON)), nil
+}
+
+func generateFilteredConfig(toolsFile, toolName string) ([]byte, error) {
```
Yuan325: Will be nice to have a comment to describe what each function is doing (for all of the functions here)

```suggestion
// generateFilteredConfig xyz...
func generateFilteredConfig(toolsFile, toolName string) ([]byte, error) {
```

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,253 @@
...
+	data, err := os.ReadFile(toolsFile)
+	if err != nil {
+		return nil, fmt.Errorf("error reading %s: %w", toolsFile, err)
+	}
+	var cfg Config
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("error parsing YAML from %s: %w", toolsFile, err)
+	}
```
Yuan325: The config file formatting had updated. Is there a way to reuse the config that are parsed instead of creating another unmarshal parser here?

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,253 @@
...
+
+	filteredCfg := Config{
+		Tools: map[string]map[string]interface{}{
+			toolName: cfg.Tools[toolName],
+		},
+	}
+
+	// Add relevant source if exists
+	if src, ok := cfg.Tools[toolName]["source"].(string); ok && src != "" {
```
Yuan325: Does Agent Skill need to know about sources?

internal/cli/skills/generator_test.go
Yuan325, line 1: Thanks for adding unit tests! :)

docs/en/how-to/generate_skill.md
helloeve: updated to follow the same style and naming - "Command Usage"

docs/en/how-to/generate_skill.md
helloeve: Makes sense - updated, and to your question, if no toolset is provided it will consider all tools (also noted in the flag description)

docs/en/how-to/generate_skill.md
helloeve: Updated with a link to agent skills spec

docs/en/how-to/generate_skill.md
helloeve: updated with flag description

docs/en/how-to/generate_skill.md
helloeve: yes, merged into previous section. For the naming, since it is intend to follow agent skills open standards, I am leaning to keep the naming, along with a section that describes how it can be integrated into Gemini CLI

docs/en/how-to/generate_skill.md
helloeve: Yes

docs/en/reference/cli.md
helloeve: Yes

docs/en/reference/cli.md
helloeve: Updated

internal/cli/skills/command.go
helloeve: I made `run` a function of cmd as we anyway declare the `Command` in this subcommand

internal/cli/skills/command.go
helloeve: We do -- I should have updated the flag description, which is confusing from reading

internal/cli/skills/command.go
helloeve: ah good point - updated

internal/cli/skills/command.go
helloeve: yes, removed

internal/cli/skills/command.go
helloeve: yes, if the path exists MkdirAll will just do nothing

internal/cli/skills/command.go
helloeve: didn't realize this - updated

internal/cli/skills/command.go
helloeve: updated

internal/cli/skills/generator.go
helloeve: sounds good!

internal/cli/skills/generator.go
helloeve: the agent skill itself won't but the filtered config will need to maintain it so the script can rely on it to make tool call

internal/cli/skills/command.go
helloeve: updated!

internal/cli/skills/command.go
helloeve: updated to remove these structs

docs/en/how-to/generate_skill.md
```diff
@@ -0,0 +1,111 @@
...
+
+A skill package consists of a `SKILL.md` file (with required YAML frontmatter) and a set of Node.js scripts. Each tool defined in your toolset maps to a corresponding script in the generated Node.js scripts (`.js`) that work across different platforms (Linux, macOS, Windows).
+
+
+### Command Usage
+
+The basic syntax for the command is:
+
+```bash
+toolbox [--tools-file <path> | --prebuilt <name>] skills-generate \
```
Yuan325: will we support `--tools-files` and `--tools-folder` here too? wonder if we want to add those as well.

docs/en/reference/cli.md
```diff
@@ -42,8 +43,36 @@ Executes a tool directly with the provided parameters. This is useful for testin
 toolbox invoke <tool-name> [params]
 ```
 
-- `<tool-name>`: The name of the tool to execute (as defined in your configuration).
-- `[params]`: (Optional) A JSON string containing the parameters for the tool.
+**Flags:**
+
+- `--tool-name`: The name of the tool to execute (as defined in your configuration).
```
```suggestion
- `tool-name`: The name of the tool to execute (as defined in your configuration).
```

docs/en/reference/cli.md
```diff
@@ -42,8 +43,36 @@ Executes a tool directly with the provided parameters. This is useful for testin
 toolbox invoke <tool-name> [params]
 ```
 
-- `<tool-name>`: The name of the tool to execute (as defined in your configuration).
-- `[params]`: (Optional) A JSON string containing the parameters for the tool.
+**Flags:**
```
```suggestion
**Arguments:**
```

Sorry for the confusion. These are probably arguments rather than flags.

internal/cli/skills/generator_test.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/googleapis/genai-toolbox/internal/sources"
+	"github.com/googleapis/genai-toolbox/internal/tools"
+	"github.com/googleapis/genai-toolbox/internal/util/parameters"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type MockToolConfig struct {
+	TypeVal   string
+	SourceVal string
+	OtherVal  string
```
Yuan325:
```suggestion
	Type   string "yaml:"type"`
	Source string "yaml:"source"`
	Other  string "yaml:"other"`
	Parameters   parameters.Parameters `yaml:"parameters"`
```

iirc, once declare this annotations, the `yaml.Marshal()` should work as expected? We might not need to declare the `MarshalYAML` function. This could be more similar to a tool config. It will be great if we can also add a parameter field here to test if the parameter marshal/unmarshaling works correctly.

internal/cli/skills/generator_test.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+// Custom marshaling to include fields in output YAML
+func (m MockToolConfig) MarshalYAML() (interface{}, error) {
+	return map[string]interface{}{
+		"source": m.SourceVal,
+		"other":  m.OtherVal,
+		"type":   m.TypeVal,
+	}, nil
+}
```
Yuan325: Shouldn't need this? if confirm, please remove from the mockSource as well :) Thank you!

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,301 @@
...
+// generateToolConfigYAML generates the YAML configuration for a single tool and its dependency (source).
+// It extracts the relevant tool and source configurations from the server config and formats them
+// into a YAML document suitable for inclusion in the skill's assets.
+func generateToolConfigYAML(cfg server.ServerConfig, toolName string) ([]byte, error) {
+	toolCfg, ok := cfg.ToolConfigs[toolName]
+	if !ok {
+		return nil, fmt.Errorf("error finding tool config: %s", toolName)
+	}
+
+	var documents [][]byte
```
Yuan325:
```suggestion
	var buf bytes.Buffer
	encoder := yaml.NewEncoder(&buf)
```

With this, we can do the following to add doc:
```
if err := encoder.Encode(input); err != nil {
  return nil, err
}
```
and when done, we can just do `return buf.Bytes(), nil`. It will add the `---` automatically. less error prone this way.

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,301 @@
...
+
+	toolMap["kind"] = "tools"
+	toolMap["name"] = toolName
+	toolMap["type"] = toolCfg.ToolConfigType()
+
+	toolDoc, err := yaml.Marshal(toolMap)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling tool doc: %w", err)
+	}
+	documents = append(documents, toolDoc)
```
Yuan325:
```suggestion
	if err := encoder.Encode(toolMap); err != nil {
    	return nil, err
    }
```

internal/cli/skills/generator_test.go
Yuan325: Or if we don't want to test Parameters here to keep it simple, we can add it in the `cmd/skill_generate_test.go`.

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,301 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling tool config: %w", err)
+	}
+	if err := yaml.Unmarshal(b, &toolMap); err != nil {
+		return nil, fmt.Errorf("error unmarshaling tool config map: %w", err)
+	}
+
+	toolMap["kind"] = "tools"
+	toolMap["name"] = toolName
+	toolMap["type"] = toolCfg.ToolConfigType()
```
Yuan325: Instead of marshal -> unmarshaling, we can do it directly from the struct.
Suggestion:
```
toolMap := struct {
  Kind string `yaml:"kind"`
  Config `yaml:",inline"`
}{
  Kind: "tools",
  Config: toolCfg,
}
```

We can encode this directly and it should have the right input :)

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,301 @@
...
+		return "", fmt.Errorf("error generating parameters schema: %w", err)
+	}
+
+	return fmt.Sprintf("#### Parameters\n\n```json\n%s\n```", string(schemaJSON)), nil
+}
+
+// generateToolConfigYAML generates the YAML configuration for a single tool and its dependency (source).
+// It extracts the relevant tool and source configurations from the server config and formats them
+// into a YAML document suitable for inclusion in the skill's assets.
+func generateToolConfigYAML(cfg server.ServerConfig, toolName string) ([]byte, error) {
```
Yuan325, line 244: Quick question, what about tools that require auth? Or are we assuming that users that use agent skills will not use auth?

docs/en/how-to/generate_skill.md
helloeve: we do, as the current implementation will inherent these flags from the root command, updated here and other places

docs/en/reference/cli.md
helloeve: updated

docs/en/reference/cli.md
helloeve: updated

internal/cli/skills/generator.go
helloeve: The skill-generation won't support client side auth, as the invoke command does not support that.

internal/cli/skills/generator.go
helloeve: Thank you for the recommendation! This is very neat - I tried to follow your recommendation in this and the following comments for a new implementation, PTAL!

internal/cli/skills/generator.go
helloeve: I have the marshal -> unmarshaling kept in the new implementation -- as this is needed to associate the tool back to the source

internal/cli/skills/generator_test.go
helloeve: good point, updated and added a new test case to cover parameters

internal/cli/skills/generator_test.go
helloeve: yes

internal/cli/skills/generator_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+func (m MockToolConfig) Initialize(map[string]sources.Source) (tools.Tool, error) {
+	return nil, nil
+}
+
+type MockSourceConfig struct {
+	TypeVal string
+	ConnVal string
+}
```
Yuan325, line 49: we can add the `yaml:""` tag here and remove the `MarshalYAML()` function too~

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, fmt.Errorf("error finding tool config: %s", toolName)
+	}
+
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
+
+	// Process Tool Config
+	toolWrapper := struct {
+		Kind   string           `yaml:"kind"`
+		Name   string           `yaml:"name"`
```
Yuan325: correct me if I'm wrong but name should be covered in the tool's config itself?

https://github.com/googleapis/genai-toolbox/blob/732eaed41d4c0e26c328b97576b214b3d4d2b276/internal/tools/alloydb/alloydbcreatecluster/alloydbcreatecluster.go#L52

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,297 @@
...
+	b, err := yaml.Marshal(toolCfg)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling tool config: %w", err)
+	}
+	if err := yaml.Unmarshal(b, &toolMap); err != nil {
+		return nil, fmt.Errorf("error unmarshaling tool config map: %w", err)
+	}
+
+	if sourceName, ok := toolMap["source"].(string); ok {
+		if sourceCfg, ok := cfg.SourceConfigs[sourceName]; ok {
```
Yuan325: I'm thinking if there's a source field in tool, and the source config is missing, we might want to throw an error; since this tool will not run successfully.

internal/cli/skills/generator.go
```diff
@@ -0,0 +1,297 @@
...
+			sourceWrapper := struct {
+				Kind   string               `yaml:"kind"`
+				Name   string               `yaml:"name"`
```
Yuan325: Same with tool, seems like source's Config does include the name (eg: https://github.com/googleapis/genai-toolbox/blob/732eaed41d4c0e26c328b97576b214b3d4d2b276/internal/sources/alloydbpg/alloydb_pg.go#L52).

internal/cli/skills/generator.go
Yuan325: Ideally we might want to add `ToolSource() string` to the ToolConfig interface, but this might be require quite some changes. I'm okay to keep the marshal -> unmarshal for source retrieval for now :)

internal/cli/skills/generator_test.go
helloeve: ah yes, updated

internal/cli/skills/generator.go
helloeve: you are right, I have updated the logic to remove this duplicate

internal/cli/skills/generator.go
helloeve: yes - added the logic for error handling here along with a new test case

internal/cli/skills/generator.go
helloeve: yes

PR #2402, opened 2026-02-02
refactor(server): standardize tool error handling and status code mapping

duwenxin99:
- Detect errors and return error codes accordingly in the tool call handler functions.
- Replace the old `util.ErrUnauthorized` with the new Toolbox error type.

### DIFF
--- a/tests/bigquery/bigquery_integration_test.go
+++ b/tests/bigquery/bigquery_integration_test.go
@@ -3023,12 +3023,23 @@ func runBigQuerySearchCatalogToolInvokeTest(t *testing.T, datasetName string, ta
 				t.Fatalf("expected 'result' field to be a string, got %T", result["result"])
 			}
 
-			if tc.isErr && (resultStr == "" || resultStr == "[]" || strings.Contains(resultStr, `"error":`)) {
+			var errorCheck map[string]any
+			if err := json.Unmarshal([]byte(resultStr), &errorCheck); err == nil {
+				if _, hasError := errorCheck["error"]; hasError {
+					if tc.isErr {
+						return
+					}
+					t.Fatalf("unexpected error object in result: %s", resultStr)
+				}
+			}
+
+			if tc.isErr && (resultStr == "" || resultStr == "[]") {
 				return
 			}
-			var entries []interface{}
+
+			var entries []any
 			if err := json.Unmarshal([]byte(resultStr), &entries); err != nil {
-				t.Fatalf("error unmarshalling result string: %v", err)
+				t.Fatalf("error unmarshalling result string: %v. Raw string: %s", err, resultStr)
 			}
 
 			if !tc.isErr {


internal/server/mcp/v20241105/method.go
```diff
@@ -124,7 +123,11 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
 	}
 	if clientAuth {
 		if accessToken == "" {
-			return jsonrpc.NewError(id, jsonrpc.INVALID_REQUEST, "missing access token in the 'Authorization' header", nil), util.ErrUnauthorized
+			return jsonrpc.NewError(id, jsonrpc.INVALID_REQUEST, "missing access token in the 'Authorization' header", nil), util.NewClientServerError(
+				"missing access token in the 'Authorization' header",
+				http.StatusUnauthorized,
+				nil,
+			)
```
Yuan325: wondering if we can prevent from writing the error message twice? Perhaps
```
err := util.NewClientServerError("message...", http.StatusUnauthoriezed, nil)
return jsonrpc.NewError(id, jsonrpc.INVALID_REQUEST, err.Error()), err
```

internal/server/mcp/v20241105/method.go
```diff
@@ -194,30 +201,44 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
...
+						} else {
+							rpcCode = jsonrpc.INTERNAL_ERROR
+						}
+					}
+				}
+				return jsonrpc.NewError(id, rpcCode, err.Error(), nil), err
 			}
-			// Auth error with ADC should raise internal 500 error
+		} else {
+			// Unknown error -> 500
```
Yuan325, line 240: Just to confirm, do we know this is 500 for sure?

internal/server/mcp.go
```diff
@@ -444,15 +443,17 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
-			} else if strings.Contains(errStr, "Error 403") {
-				w.WriteHeader(http.StatusForbidden)
+			var clientServerErr *util.ClientServerError
+			if errors.As(err, &clientServerErr) {
+				switch clientServerErr.Code {
+				case http.StatusUnauthorized:
+					w.WriteHeader(http.StatusUnauthorized)
+				case http.StatusForbidden:
+					w.WriteHeader(http.StatusForbidden)
+				}
```
Yuan325: Will it be simpler to add a function in ClientServerError that returns the status code? That way we can just do `w.WriteHeader(clientServerErr.Code())` without having to keep the switch case here and potentially having to increase more cases in the future? Or do we only want to write header for this 2 code specifically?

Second thought, we could even remove the `switch code { case jsonrpc.INTERNAL_ERROR...}`? Check if jsonrpc is INTERNAL_ERROR or INVALID_REQUEST, we can just write the status to header? Might wanna double check the specification to confirm why we only write header for this 2 cases~

Yuan325:
Might want to merge this into a branch first instead of main if the tests will fail before refactoring other parts of Toolbox? Seems like the err branch is merging into main, and once that is merged, this PR will be updated to merge into main.

internal/server/mcp/v20241105/method.go
duwenxin99: Yes because this technically shouldn't happen. All the errors should be either ClientServer or Agent error.

internal/server/mcp.go
duwenxin99: Nice catch! I updated this to propagate all the original error codes.

internal/server/mcp/v20241105/method.go
duwenxin99: updated.

duwenxin99:
> Might want to merge this into a branch first instead of main if the tests will fail before refactoring other parts of Toolbox? Seems like the err branch is merging into main, and once that is merged, this PR will be updated to merge into main.

 @Yuan325  I'm going to merge this refactor PR into this branch first: https://github.com/googleapis/genai-toolbox/pull/2403

internal/server/mcp/v20241105/method.go
Yuan325: nit: can we update this to be more consistent? 

```
			errMsg := "missing access token in the 'Authorization' header"
			return jsonrpc.NewError(id, jsonrpc.INVALID_REQUEST, errMsg, nil), util.NewClientServerError(
				errMsg,
				http.StatusUnauthorized,
				nil,
			)
```
vs 
```
		err = util.NewClientServerError(
			"unauthorized Tool call: Please make sure you specify correct auth headers",
			http.StatusUnauthorized,
			nil,
		)
		return jsonrpc.NewError(id, jsonrpc.INVALID_REQUEST, err.Error(), nil), err
```

Seems like we are using both formats.

internal/server/mcp.go
Yuan325, line 1: Are we planning to update all the `fmt.Errorf()` in this file?

PR #2403, opened 2026-02-02
refactor(tools): refactor all Tools to return ToolboxError

duwenxin99:
- Use a generic `util.ProcessGcpError` for GCP tools that return `googleapi.Error`.
- Use a generic `util.ProcessGeneralError` for all other tools.

Error categorization in Invoke():
- Query errors default to AgentError.
- Missing/malformed parameters are AgentError.
- config, ctx, logger, source fetching errors are ClientServerError.

### DIFF
--- a/tests/bigquery/bigquery_integration_test.go
+++ b/tests/bigquery/bigquery_integration_test.go
@@ -1010,7 +1010,7 @@ func runBigQueryExecuteSqlToolInvokeTest(t *testing.T, select1Want, invokeParamW
 			api:           "http://127.0.0.1:5000/api/tool/my-exec-sql-tool/invoke",
 			requestHeader: map[string]string{},
 			requestBody:   bytes.NewBuffer([]byte(`{}`)),
-			want:          `{"error":"parameter \"string_val\" is required"}`,
+			want:          `{"error":"parameter \"sql\" is required"}`,
 			isErr:         true,
 		},
 		{
@@ -1166,8 +1166,8 @@ func runBigQueryWriteModeBlockedTest(t *testing.T, tableNameParam, datasetName s
 		wantResult     string
 	}{
 		{"SELECT statement should succeed", fmt.Sprintf("SELECT id, name FROM %s WHERE id = 1", tableNameParam), http.StatusOK, `[{"id":1,"name":"Alice"}]`},
-		{"INSERT statement should fail", fmt.Sprintf("INSERT INTO %s (id, name) VALUES (10, 'test')", tableNameParam), http.StatusOK, "write mode is 'blocked', only SELECT statements are allowed"},
-		{"CREATE TABLE statement should fail", fmt.Sprintf("CREATE TABLE %s.new_table (x INT64)", datasetName), http.StatusOK, "write mode is 'blocked', only SELECT statements are allowed"},
+		{"INSERT statement should fail", fmt.Sprintf("INSERT INTO %s (id, name) VALUES (10, 'test')", tableNameParam), http.StatusOK, "{\"error\":\"write mode is 'blocked', only SELECT statements are allowed\"}"},
+		{"CREATE TABLE statement should fail", fmt.Sprintf("CREATE TABLE %s.new_table (x INT64)", datasetName), http.StatusOK, "{\"error\":\"write mode is 'blocked', only SELECT statements are allowed\"}"},
 	}
 
 	for _, tc := range testCases {


internal/tools/cloudsqlmssql/cloudsqlmssqlcreateinstance/cloudsqlmssqlcreateinstance.go
```diff
@@ -121,33 +123,33 @@ func (t Tool) ToConfig() tools.ToolConfig {
 }
 
 // Invoke executes the tool's logic.
-func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, params parameters.ParamValues, accessToken tools.AccessToken) (any, error) {
+func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, params parameters.ParamValues, accessToken tools.AccessToken) (any, util.ToolboxError) {
```
shuzhou-gc, line 126: Just a quick question, will there be some follow-up PR(s) to apply similar changes to community mysql branch?

duwenxin99: It has been added to this PR in later commits.

internal/tools/cloudsqlmysql/cloudsqlmysqlcreateinstance/cloudsqlmysqlcreateinstance.go
```diff
@@ -121,33 +123,33 @@ func (t Tool) ToConfig() tools.ToolConfig {
 }
 
 // Invoke executes the tool's logic.
-func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, params parameters.ParamValues, accessToken tools.AccessToken) (any, error) {
+func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, params parameters.ParamValues, accessToken tools.AccessToken) (any, util.ToolboxError) {
 	source, err := tools.GetCompatibleSource[compatibleSource](resourceMgr, t.Source, t.Name, t.Type)
 	if err != nil {
-		return nil, err
+		return nil, util.NewClientServerError("source used is not compatible with the tool", http.StatusInternalServerError, err)
```
shuzhou-gc, line 129: Sorry it is probably a naive question, but what is the rationale to have different error types tied to detailed reasons for such errors?

I imagine maybe there are automated handling based on error type? Maybe there are metrics built upon that? The trade-off, obviously, is that the developers will have to
   - be familiar with all these detailed error types, or
   - copy/paste what existed, or
   - write their own if the meaning/reason, based on individual developers interpretation, does not already exist.

duwenxin99: There are only two new error types - AgentError and ClientServerError. The main rationale to distinguish between them is to return 200 OK status code for all agent errors (with error message in the response body) to allow the agent to self-correct. Therefore, detailed error message for AgentError allows LLM to understand the cause of error; for ClientServerError, it allows developers to debug more easily. 
This refactor does increase the learning curve for the contributors, but I will add a more detailed guide in the documentation to make sure the categorization and expectations are clear.

shuzhou-gc: I see.

But interestingly, why do you want to return 200 for errors? Isn't it conventional to return 4xx-5xx for errors?

shuzhou-gc: And I am a little confused about the difference between those 2 errors... 

Say `ClientServerError` is for developers, does it mean that the error cannot be resolved automatically?  In the code above, the error is used in incompatible source. Is it possible to update the source type using prompt?

Another error example `NewAgentError`, it is used for missing password. This is cannot be resolve automatically, right?

I guess I must have missed something.

duwenxin99: The source type used is defined in the server configuration yaml file so that can only changed by the developers.
In this case, the password is passed in as a tool invocation parameter, then it should be an AgentError because the agent provides these params.

PR #2466, opened 2026-02-13
feat: add polling system to dynamic reloading

### LINKED ISSUE #2346: Add support for hot-reloading with NFS-based tools.yaml file
### Prerequisites

One challenge we've noted is that Toolbox's hot-reloading feature currently relies on inotify events and doesn't function well with NFS-based files. Is it possible to incorporate support for this directly into Toolbox? Perhaps we could add a different type of file watcher instead of solely relying on inotify events.

Our underlying infrastructure is K8 and tools.yaml file is hosted on NFS.

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

Yuan325:
Support polling system to existing Toolbox dynamic reloading feature. 
| flag | description | default |
| --- | --- | --- |
| `--poll-interval`          | Specifies the polling frequency (seconds) for configuration file updates.                                                                                                        | `0`         |

With `--poll-interval` = 0, it will not run the polling system at all. 

When polling is active, Toolbox checks for configuration file updates at every specified interval (in seconds).

Fixes #2346


cmd/root.go
```diff
@@ -248,6 +274,55 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
...
+				for f := range watchedFiles {
+					// We must explicitly check existence here because checkModTime
+					// swallows errors (returns false), making it impossible to
+					// distinguish between "no change" and "file deleted".
+					if _, err := os.Stat(f); err == nil {
+						currentDiskFiles[f] = true
+						if checkModTime(f, lastSeen) {
+							changed = true
+						}
+					}
```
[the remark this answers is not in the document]
Yuan325: the first os.Stat is to check for file existance as well. The ones in checkModTime just ignores deleted files and is used for both watching folders and watching files.

cmd/root.go
```diff
@@ -210,6 +225,17 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
...
+	if pollTickerSecond > 0 {
+		ticker := time.NewTicker(time.Duration(pollTickerSecond) * time.Second)
+		defer ticker.Stop()
+		pollTickerChan = ticker.C // Assign the channel
+		logger.DebugContext(ctx, fmt.Sprintf("NFS polling enabled every %v", pollTickerSecond))
+	} else {
+		logger.DebugContext(ctx, "NFS polling disabled (interval is 0)")
+	}
+
+	lastSeen := make(map[string]time.Time)
```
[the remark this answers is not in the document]
Yuan325: added

docs/en/reference/cli.md
```diff
@@ -136,6 +137,16 @@ used at a time.
 Toolbox enables dynamic reloading by default. To disable, use the
 `--disable-reload` flag.
 
+Use the `--poll-interval` flag to manually detect configuration file updates.
```
averikitsch: nit: please add more info explaining the difference between hot reload and the polling. If we don't want a super long discussion here, more info in the developer.md would be helpful

cmd/root.go
```diff
@@ -238,6 +253,42 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
...
+		ticker := time.NewTicker(time.Duration(pollTickerSecond) * time.Second)
+		defer ticker.Stop()
+		pollTickerChan = ticker.C // Assign the channel
+		logger.DebugContext(ctx, fmt.Sprintf("NFS polling enabled every %v", pollTickerSecond))
+
+		// Pre-populate lastSeen to avoid an initial spurious reload
+		if watchingFolder {
+			files, err := os.ReadDir(folderToWatch)
+			if err != nil {
+				logger.WarnContext(ctx, "error reading tools folder on initial scan %s", err)
```
averikitsch: Do we need to fmt.Sprintf this string?

cmd/root.go
```diff
@@ -248,6 +299,55 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
...
 			logger.DebugContext(ctx, "file watcher context cancelled")
 			return
+		case <-pollTickerChan:
+			changed := false
+			currentDiskFiles := make(map[string]bool)
+			// Get files that are currently on disk
+			if watchingFolder {
+				files, err := os.ReadDir(folderToWatch)
+				if err != nil {
+					logger.WarnContext(ctx, "error reading tools folder %s", err)
```
averikitsch: same here. Does warncontext expect a string and err or just the string?

cmd/root.go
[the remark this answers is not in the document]
averikitsch: This is the main concern I see. Since this feature is built specifically for NFS where os.Stat calls are relatively expensive network
  operations—you can still achieve both the existence check and the mod-time check using a single os.Stat (or f.Info()) by slightly refactoring the helper.

  If you update the helper to accept the time.Time directly:


   1 func checkModTime(path string, mTime time.Time, lastSeen map[string]time.Time) bool {
   2     if mTime.After(lastSeen[path]) {
   3         lastSeen[path] = mTime
   4         return true
   5     }
   6     return false
   7 }


  Then your explicitly watched files loop only needs one os.Stat instead of two:


   1 for f := range watchedFiles {
   2     // info gives us existence AND the timestamp in one call
   3     if info, err := os.Stat(f); err == nil {
   4         currentDiskFiles[f] = true
   5         if checkModTime(f, info.ModTime(), lastSeen) {
   6             changed = true
   7         }
   8     }
   9 }


  And your folder-watching loop can avoid os.Stat entirely by using f.Info() (since os.ReadDir returns fs.DirEntry):


    1 for _, f := range files {
    2     if !f.IsDir() && (strings.HasSuffix(f.Name(), ".yaml") || strings.HasSuffix(f.Name(), ".yml")) {
    3         fullPath := filepath.Join(folderToWatch, f.Name())
    4         currentDiskFiles[fullPath] = true
    5
    6         if info, err := f.Info(); err == nil {
    7             if checkModTime(fullPath, info.ModTime(), lastSeen) {
    8                 changed = true
    9             }
   10         }
   11     }
   12 }

docs/en/reference/cli.md
Yuan325: Updated the section at the bottom.

cmd/root.go
Yuan325: Updated

cmd/root.go
```diff
@@ -195,16 +196,58 @@ func validateReloadEdits(
...
+			}
+		}
+	} else {
+		for f := range watchedFiles {
+			if info, err := os.Stat(f); err == nil {
+				currentDiskFiles[f] = true
+				if checkModTime(f, info.ModTime(), lastSeen) {
+					changed = true
+				}
+			}
```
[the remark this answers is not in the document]
Yuan325: the path was provided at the starting of server. If it's not already in the server, attackers will not be able to add new path later.

PR #2468, opened 2026-02-13
fix(oracle): handle bracketed proxy usernames in go-ora URL

### LINKED ISSUE #2454: Oracle proxy username with brackets rejected in connection URL
Tracking from googleapis/mcp-toolbox-sdk-python#532. The issue seems to be strict URL parsing of Oracle proxy usernames like `user[client]` in connection URLs. Brackets are invalid in URL userinfo, so parsers reject; percent-encoding may pass parsing but not be decoded before reaching the Oracle driver, causing auth failure.

Suggested fix: decode percent-encoded usernames before passing to the driver, or avoid strict URL parsing for userinfo and construct DSN/username explicitly (e.g., use `URL.create()` equivalent in Go or split config fields).

Workaround: avoid URL strings for proxy users; pass proxy user separately via structured config/driver options if supported.

Deeven-Seru:
## Summary
- build go-ora DSNs using encoded userinfo so Oracle proxy usernames like `user[client]` are valid in URL form
- decode already-percent-encoded credentials before re-encoding to avoid double-encoding
- keep connect string path (`host:port/service`) intact and preserve wallet query handling
- add regression tests covering bracketed proxy users, pre-encoded users, and wallet/no-wallet forms

Fixes #2454


internal/sources/oracle/oracle.go
```diff
@@ -237,6 +238,31 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any) (an
...
+	base := fmt.Sprintf("oracle://%s@%s", userInfo, connectStringBase)
+	if strings.TrimSpace(walletLocation) == "" {
+		return base
+	}
+
+	q := url.Values{}
+	q.Set("ssl", "true")
+	q.Set("wallet", walletLocation)
+	return fmt.Sprintf("%s?%s", base, q.Encode())
+}
```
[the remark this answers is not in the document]
Deeven-Seru: In this Oracle path we intentionally preserve `connectStringBase` verbatim because parsing `host:port/service_name` through generic URL host/path handling can alter semantics for go-ora. We’re still safely encoding credentials via `url.UserPassword(...)` and wallet query params via `url.Values`, which addresses #2454 without changing Oracle connect-string behavior.

Yuan325:
Hi @DEVELOPER-DEEVEN, can you please rebase and clean the PR (seems like it's removing some already merged codes)? Thank you! :) Will take a look once this is ready

Deeven-Seru:
@Yuan325  yuan PTAL I think it works now

Yuan325:
Hi @Deeven-Seru, seems like the rebasing is still removing some existing code (e.g. internal/log/log.go). Please double-check that only your intended updates made it into this PR. Thank you!

Deeven-Seru:
Correction to prior note :

Addressed in latest force-push (6dfe250):
- Restored RunSQL(ctx, statement, params, readOnly bool) and reintroduced the readOnly=false DML ExecContext path with rows_affected.
- Updated go-ora DSN construction to preserve and merge existing query params safely when connectStringBase already contains a query, while still encoding userinfo and wallet params.
- Added/expanded regression tests in internal/sources/oracle/oracle_connstring_test.go for bracketed proxy usernames, pre-encoded userinfo (no double-encoding), wallet/no-wallet, and existing-query-plus-wallet merge cases.

This is now a clean single-commit diff with only Oracle source/test changes. PTAL.

Deeven-Seru:
@Yuan325  ptal

internal/sources/oracle/oracle.go
```diff
@@ -253,6 +254,57 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any, rea
 	return out, nil
 }
 
+func buildGoOraConnString(user, password, connectStringBase, walletLocation string) string {
+	userInfo := url.UserPassword(
+		decodePercentEncodedUserInfo(user),
```
Yuan325: Where is the user and password encoded? Is it before parsing through the config?

internal/sources/oracle/oracle.go
```diff
@@ -305,16 +357,11 @@ func initOracleConnection(ctx context.Context, tracer trace.Tracer, config Confi
...
 		driverName = "oracle"
 
-		user := config.User
-		password := config.Password
+		finalConnStr = buildGoOraConnString(config.User, config.Password, connectStringBase, config.WalletLocation)
 
 		if hasWallet {
-			finalConnStr = fmt.Sprintf("oracle://%s:%s@%s?ssl=true&wallet=%s",
-				user, password, connectStringBase, config.WalletLocation)
+			logger.DebugContext(ctx, fmt.Sprintf("Using go-ora driver (pure-Go) with wallet and serverString: %s\n", connectStringBase))
```
Yuan325, line 367: we might wanna move these logger to where it adds the wallet (within the `buildGoOraConnString()` instead of here.

internal/sources/oracle/oracle.go
```diff
@@ -253,6 +254,57 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any, rea
...
+		}
+	}
+
+	trimmedWalletLocation := strings.TrimSpace(walletLocation)
+	if rawExistingQuery != "" {
+		if trimmedWalletLocation == "" {
+			return fmt.Sprintf("%s?%s", dsnBase, rawExistingQuery)
+		}
+
+		walletQuery := url.Values{}
```
Yuan325, line 297: might wanna add a check for `hasWallet := strings.TrimSpace(config.WalletLocation) != ""` and add the logger here. We can either pass the logger here, or pass the context and retrieve the logger from context.

```
logger, err := util.LoggerFromContext(ctx)
	if err != nil {
		return err
	}
```

internal/sources/oracle/oracle_connstring_test.go
Yuan325, line 1: can we move all the tests into oracle_test.go please? TY

internal/sources/oracle/oracle.go
```diff
@@ -253,6 +254,57 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any, rea
 	return out, nil
 }
 
+func buildGoOraConnString(user, password, connectStringBase, walletLocation string) string {
+	userInfo := url.UserPassword(
+		decodePercentEncodedUserInfo(user),
+		decodePercentEncodedUserInfo(password),
+	).String()
+
+	baseConnStr, existingQuery, hasQuery := strings.Cut(connectStringBase, "?")
```
Yuan325, line 278: Thinking if we should move the whole connectStringBase building into this function? Feels inefficient (and potentially introducing bugs down the line) by building and then deconstructing it and rebuilding it again.

internal/sources/oracle/oracle.go
Deeven-Seru: I encoded inside buildGoOraConnString right before building the DSN: it first  decodes any percent‑encoded user/pass to avoid double‑encoding, then uses url.UserPassword(...) to encode safely for userinfo. It’s not in config parsing only at DSN construction time

internal/sources/oracle/oracle.go
Deeven-Seru: Done. The wallet decision and the connect‑string base now live in the helper, and  logs after building the DSN using the computed connectStringBase, so the log reflects the same wallet decision that went into the DSN as intended

internal/sources/oracle/oracle.go
Deeven-Seru: Handled via hasWallet := strings.TrimSpace(config.WalletL ocation) != "" at init; we keep logging in initOracleConnection using the same hasWallet and the base string returned by buildGoOraConnString so it’s consistent without pushing logger into pure string helppers.

internal/sources/oracle/oracle_connstring_test.go
Deeven-Seru: done .👍🏻

internal/sources/oracle/oracle.go
Deeven-Seru: Yes. Refactored to build connectStringBase inside buildGoOraConnString via a helper (buildConnectStringBase(config) ), so that we don’t build then deconstruct; reduces redundancy and avoids subtle drift.  Thanks for the catch ..!

Deeven-Seru:
colsing this as the issue isn fixed via #2469

PR #2478, opened 2026-02-16
oracle: support proxy usernames with brackets in go-ora DSN

### LINKED ISSUE #2454: Oracle proxy username with brackets rejected in connection URL

liuxiaopai-ai:
Fixes URL construction for the pure Go Oracle driver (`useOCI: false`) so usernames like `user[client]` work correctly.

What changed:
- Build go-ora DSN using `url.UserPassword(...).String()` for proper credential escaping
- Keep existing connection behavior for host/port/service, connectionString, and tnsAlias
- Use `url.Values` to build wallet query params
- Add unit tests to verify bracket usernames and wallet query handling

Tests run:
`go test -v ./internal/sources/oracle -count=1 -vet=off`

Fixes #2454

### DIFF
--- a/internal/sources/oracle/oracle.go
+++ b/internal/sources/oracle/oracle.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


Yuan325:
Hi @liuxiaopai-ai Thank you for your contributions, however, it seems like there's another existing PR for this issue. Closing due to duplication #2469

PR #2482, opened 2026-02-16
docs: optimize link checker

rahulpinto19:
Previous Functionality
The original workflow was designed to execute the link checker across every file in the repository during each run.

Planned Enhancements
To improve efficiency, the process will be bifurcated into three distinct components:

- Presubmit Check: The link checker will only target files that have been modified.
- Improved Reporting: Broken links are now reported via a single, automated comment in the PR conversation (which updates automatically on subsequent pushes).

Screenshot 2026-02-17 3 47 52 PM

- Cleaner Logs: Suppressed non-critical redirect warnings and PR comment now focuses strictly on broken URLs to make debugging faster.


.github/workflows/link_checker_workflow.yaml
```diff
@@ -23,8 +26,33 @@ jobs:
...
+            # Flatten newlines to spaces for the args list
+            FILES_FLAT=$(echo "$CHANGED_FILES" | tr '\n' ' ')
+            
+            echo "CHECK_FILES=$FILES_FLAT" >> $GITHUB_ENV
+            echo "HAS_CHANGES=true" >> $GITHUB_ENV
+          fi
+
 
       - name: Restore lychee cache
+        if: env.HAS_CHANGES == 'true'
```
twishabansal, line 59: What does this flag do?

twishabansal:
Is it possible to attach a screenshot to see how the output looks like?

.github/workflows/link_checker_workflow.yaml
```diff
@@ -23,8 +26,33 @@ jobs:
...
       - name: Checkout Repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+      - name: Identify Changed Files
+        id: changed-files
+        shell: bash
+        run: |
+          git fetch origin main
+          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT origin/main...HEAD)
```
twishabansal, line 37: Does this work for newly added md files?

.github/workflows/link_checker_workflow.yaml
```diff
@@ -42,27 +71,52 @@ jobs:
...
+          echo "Local links and directory changes work differently on GitHub than on the docsite." >> full-report.md
+          echo "You must ensure fixes pass the **GitHub check** and also work with **\`hugo server\`**." >> full-report.md
+          echo "See [Link Checking and Fixing with Lychee](https://github.com/googleapis/genai-toolbox/blob/main/DEVELOPER.md#link-checking-and-fixing-with-lychee) for more details." >> full-report.md
+          echo "---" >> full-report.md
+          echo "### Broken Links Found" >> full-report.md
+          
+          # Clean the report (remove redirects) AND append it to the file
+          sed -E '/(Redirect|Redirects per input)/d' lychee-report.md >> full-report.md
+
+      - name: Create PR Comment
```
twishabansal, line 115: What happens if the new run is a sucess? I am guessing that an old failure comment at that time might be confusing

.github/workflows/link_checker_workflow.yaml
```diff
@@ -23,8 +26,33 @@ jobs:
...
+          if [ -z "$CHANGED_FILES" ]; then
+            echo "No markdown files changed. Skipping checks."
+            echo "HAS_CHANGES=false" >> $GITHUB_ENV
+          else
+            echo "--- Changed Files to Scan ---"
+            echo "$CHANGED_FILES"
+            echo "-----------------------------"
+            
+            # Flatten newlines to spaces for the args list
+            FILES_FLAT=$(echo "$CHANGED_FILES" | tr '\n' ' ')
```
twishabansal: This might cause problems if there are filenames with spaces. It's an common case but can we find a better way to do things here?

.github/workflows/link_checker_workflow.yaml
rahulpinto19: A: Added
C: Copied
M: Modified
R: Renamed
T: Type changed 

yes

.github/workflows/link_checker_workflow.yaml
rahulpinto19: this tells that there are any changes in the files

.github/workflows/link_checker_workflow.yaml
rahulpinto19: wrapped the individual file names with the double quotes to avoid the conflict of handling the filenames with spaces.

.github/workflows/link_checker_workflow.yaml
rahulpinto19: If the new run is success , workflow will delete the exisiting comment if present.

twishabansal:
@rahulpinto19 do you mind keeping an eye on this link checker to ensure that we catch any issue and fix them as they arise?

PR #2488, opened 2026-02-17
feat(tools/looker): tools to list/create/delete directories

drstrangelooker:
within a LookML project. These tools only work on Looker 26.2 and later.


Yuan325:
Thank you! :)

PR #2501, opened 2026-02-18
docs: link check only on .md files

rahulpinto19:
Previously, the link checker ran on all changed files; it has now been restricted to changed .md files only.

### DIFF
--- a/.github/workflows/link_checker.yaml
+++ b/.github/workflows/link_checker.yaml
@@ -34,7 +34,7 @@ jobs:
         shell: bash
         run: |
           git fetch origin main
-          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT origin/main...HEAD -- '*.md' 'docs/**/*.md')
+          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT origin/main...HEAD -- '*.md')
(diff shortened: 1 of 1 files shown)


twishabansal:
nit: Can we rename this file to link_checker and remove the workflow suffix?

.github/workflows/link_checker.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
         shell: bash
         run: |
           git fetch origin main
-          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT origin/main...HEAD)
+          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT origin/main...HEAD -- '*.md' 'docs/**/*.md')
```
twishabansal: `docs/**/*.md` seems to be a subset of `*.md`

twishabansal:
```suggestion
          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT origin/main...HEAD -- '*.md')
```

rahulpinto19: I added as safety net however *.md works

PR #2541, opened 2026-02-20
doc: Fix ADK quickstart rendering with shortcodes

anubhav756:
## Description

Previously the code blocks nested under tab panes were not rendering properly. The code included with `regionInclude` shortcode was rendering with weird characters while that included with `include` shortcode rendered with extra blank lines.

This PR fixes both the issues.

### Before
image

### After
image

> [!NOTE]
> This PR also removes the license header from the Go sample code for pre/post processing, since no other language sample code (not even the Go sample code in quickstarts) were having license headers. Adjusted the header checker config for the same.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -489,8 +489,16 @@ The `regionInclude` shortcode reads a file, extracts content between `[START reg
 **Syntax:**
 `{{< regionInclude "path/to/file" "region_name" "language" >}}`
 
-**Example:**
+**Example Markdown:**
 `{{< regionInclude "samples/program.js" "program_setup" "javascript" >}}`
+
+**Example Code Snippet (`samples/program.js`):**
+```javascript
+// [START program_setup]
+import { Toolbox } from '@googleapis/genai-toolbox';
+const toolbox = new Toolbox();
+// [END program_setup]
+```
 
 **Source:** [.hugo/layouts/shortcodes/regionInclude.html](.hugo/layouts/shortcodes/regionInclude.html)
 


.hugo/layouts/shortcodes/include.html
```diff
@@ -1,2 +1,13 @@
 {{ $file := .Get 0 }}
-{{ (printf "%s%s" .Page.File.Dir $file) | readFile | replaceRE "^---[\\s\\S]+?---" "" | safeHTML }}
+{{ $lang := .Get 1 }}
+
+{{ $content := (printf "%s%s" .Page.File.Dir $file) | readFile | replaceRE "^---[\\s\\S]+?---" "" | replaceRE "\r\n"
+"\n" | strings.TrimRight "\n" }}
+
+{{ if $lang }}
+```{{ $lang }}
+{{ $content | safeHTML }}
```
twishabansal, line 9: Seems like this include tag change is messing up some other samples: https://screenshot.googleplex.com/C4KYnCdUgdiv8zX

It's adding some empty lines at the beginning of each. Can we fix this?

twishabansal: There might have been some caveats to these shortcodes. You can find the details here: https://docs.google.com/document/d/1JAX8o8Avx1RYk2zSsfhzVEiLk4UTqhf1uDCXhshBTPg/edit?usp=sharing&resourcekey=0-Y01aoSADhbVO_0jXdIrCmw

twishabansal: cc @rapid-killer-9 @AnmolShukla2002

anubhav756: Thanks for pointing that out! Fixed it https://screenshot.googleplex.com/34NKXMJeJUJM9Ko

The issue was that previously the `<include` shortcodes were being rendered as raw text so the code markers were required. Now they are being rendered as code blocks instead, so we have to remove the manual code markers and only specify the language to the shortcodes.

docs/en/samples/pre_post_processing/go/adk/agent.go
averikitsch, line 1: nit: we should keep the license if possible

.hugo/layouts/shortcodes/include.html
```diff
@@ -1,2 +1,13 @@
 {{ $file := .Get 0 }}
-{{ (printf "%s%s" .Page.File.Dir $file) | readFile | replaceRE "^---[\\s\\S]+?---" "" | safeHTML }}
+{{ $lang := .Get 1 }}
+
+{{ $content := (printf "%s%s" .Page.File.Dir $file) | readFile | replaceRE "^---[\\s\\S]+?---" "" | replaceRE "\r\n"
+"\n" | strings.TrimRight "\n" }}
+
```
averikitsch, line 6: It may be nice to document this syntax or link to a guide in the developer.md

docs/en/samples/pre_post_processing/go/adk/agent.go
anubhav756: I’d propose removing the license header from the Go pre/post-processing sample code.

Currently, it’s an outlier. Neither the [Python](https://screenshot.googleplex.com/5xkZw73vDAfjYXN) nor [JS](https://screenshot.googleplex.com/nemrexKxtAUT2Jw) examples use them, and other [Go quickstarts](https://screenshot.googleplex.com/5SXx6MMrheoN4Vm) skip them as well.

Since this code is intended for documentation, the header feels like extra noise. Removing it would ensure consistency across the docsite and allow readers to focus immediately on the logic.

Otherwise, do you think we should we be adding headers to the rest of the code samples instead?

I want to make sure we aren't missing a requirement. 🙂

.hugo/layouts/shortcodes/include.html
anubhav756: That's a great idea. Added the details in the [`DEVELOPER.md`](https://github.com/googleapis/genai-toolbox/pull/2541/changes#diff-35ea617bfbf0780bfbc554348314c5b11ba478a4d98fd92d33d45cdd54bf326a). Thanks!

DEVELOPER.md
```diff
@@ -314,6 +314,36 @@ preview link will be automatically added as a comment to your PR.
...
+
+**Example:**
+`{{< include "static/headers/license_header.txt" >}}`
+`{{< include "samples/program.js" "javascript" >}}`
+
+**Source:** [.hugo/layouts/shortcodes/include.html](.hugo/layouts/shortcodes/include.html)
+
+#### `regionInclude` Shortcode
+
+The `regionInclude` shortcode reads a file, extracts content between `[START region_name]` and `[END region_name]`, and optionally fences it.
```
twishabansal, line 487: nit: Can we add an example on how [START region_name] and [END region_name] are specified in the code snippet?

anubhav756: Good idea! Added. Thanks!

docs/en/samples/pre_post_processing/go/adk/agent.go
anubhav756: Closing this thread for now. We can pick this up in a separate PR if any changes are required per offline sync.

PR #2550, opened 2026-02-23
feat(cli): add serve subcommand

Yuan325:
The `serve` subcommand **starts a toolbox server** without any primitives. It WILL NOT run with config file. In the future, users could use `serve` with a backend storage.

To stop or shutdown the server, user can just terminate the port. The terminate signal will shutdown the server.

This new addition **WILL NOT** be a breaking change to existing users. Users can still run toolbox as is. 

**CLI command:**
```
toolbox serve
```

**Flags associated with the serve subcommand:**
| flag | description | default value |
| --- | --- | --- |
| address | Address of the interface the server will listen on. | 127.0.0.1 |
| port | Port the server will listen on. | 5000 |
| stdio | Listens via MCP STDIO instead of acting as a remote HTTP server. | false |
| ui | Launches the Toolbox UI web server. | false |
| allowed-origins | Specifies a list of origins permitted to access this server. | `*` |
| allowed-hosts | Specifies a list of hosts permitted to access this server. | `*` |

**This PR does the following:**
* Add a new `serve` subcommand. Including unit tests for the subcommand
* Rename the `cmd/internal/persistent_flags.go` to `cmd/internal/flags.go`, and refactored some flag definitions into dedicated functions. This change allows us to scope flags to specific subcommands as needed, rather than forcing all subcommands to inherit them globally via `PersistentFlags`.

### DIFF
--- a/cmd/internal/serve/command.go
+++ b/cmd/internal/serve/command.go
@@ -58,7 +58,7 @@ func runServe(cmd *cobra.Command, opts *internal.ToolboxOptions) error {
 		case syscall.SIGINT:
 			opts.Logger.DebugContext(sCtx, "Received SIGINT signal to shutdown.")
 		case syscall.SIGTERM:
-			opts.Logger.DebugContext(sCtx, "Sending SIGTERM signal to shutdown.")
+			opts.Logger.DebugContext(sCtx, "Received SIGTERM signal to shutdown.")
 		}
 		cancel()
 	}(ctx)
--- a/go.mod
+++ b/go.mod
@@ -230,7 +230,6 @@ require (
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/shirou/gopsutil/v4 v4.25.6 // indirect
 	github.com/sirupsen/logrus v1.9.3 // indirect
-	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/spiffe/go-spiffe/v2 v2.6.0 // indirect
 	github.com/stretchr/testify v1.11.1 // indirect
 	github.com/tklauser/go-sysconf v0.3.12 // indirect


cmd/internal/serve/command.go
```diff
@@ -0,0 +1,132 @@
...
+		defer cancel()
+		opts.Logger.WarnContext(shutdownContext, "Shutting down gracefully...")
+		err := s.Shutdown(shutdownContext)
+		if err == context.DeadlineExceeded {
+			return fmt.Errorf("graceful shutdown timed out... forcing exit")
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
Yuan325: Yes there will be a lot of duplication and this is expected.

cmd/internal/serve/command_test.go
```diff
@@ -0,0 +1,63 @@
...
+	// Inject the context into the Cobra command
+	parentCmd.SetContext(ctx)
+
+	err := parentCmd.Execute()
+	return buf.String(), err
+}
+
+func TestServe(t *testing.T) {
+	// context will automatically shutdown in 1 second.
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
averikitsch, line 47: This may be potentially flaky if it cancels without finishing the test, but that may be a rare scenario.

cmd/internal/serve/command.go
```diff
@@ -0,0 +1,138 @@
...
+
+func NewCommand(opts *internal.ToolboxOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "serve",
+		Short: "Deploy the toolbox server",
+		Long:  "Deploy the toolbox server",
+	}
+	flags := cmd.Flags()
+	internal.ServeFlags(flags, opts)
+	internal.ConfigFileFlags(flags, opts)
```
averikitsch: serve shouldn't have config flags

cmd/internal/serve/command.go
```diff
@@ -0,0 +1,138 @@
...
+	// start server
+	s, err := server.NewServer(ctx, opts.Cfg)
+	if err != nil {
+		errMsg := fmt.Errorf("toolbox failed to initialize: %w", err)
+		opts.Logger.ErrorContext(ctx, errMsg.Error())
+		return errMsg
+	}
+
+	// run server in background
+	srvErr := make(chan error)
```
averikitsch: recommending a buffered channel 
```suggestion
	srvErr := make(chan error, 1)
```

cmd/internal/serve/command_test.go
Yuan325: It's placed as 1 second purposefully so that the server would shut down~ If it becomes flaky in the future we can potentially increase this time.

cmd/internal/serve/command.go
```diff
@@ -0,0 +1,132 @@
...
+		case <-sCtx.Done():
+			// this should only happen when the context supplied when testing is canceled
+			return
+		case s = <-signals:
+		}
+		switch s {
+		case syscall.SIGINT:
+			opts.Logger.DebugContext(sCtx, "Received SIGINT signal to shutdown.")
+		case syscall.SIGTERM:
+			opts.Logger.DebugContext(sCtx, "Sending SIGTERM signal to shutdown.")
```
averikitsch:
```suggestion
			opts.Logger.DebugContext(sCtx, "Received SIGTERM signal to shutdown.")
```

averikitsch (approved):

PR #2551, opened 2026-02-24
docs: Add SECURITY.md

duwenxin99:
Update GitHub workflow security to align with [Google Open Source security requirements](https://opensource.google/documentation/reference/github/security).


SECURITY.md
```diff
@@ -0,0 +1,7 @@
+# Security Policy
+
+To report a security issue, please email
```
averikitsch: nit: I think we should create a public google group

PR #2556, opened 2026-02-24
ci: add SLSA build provenance to docker images

duwenxin99:
Ref: [Generate Cloud Build provenance](https://docs.cloud.google.com/build/docs/securing-builds/generate-validate-build-provenance)

- Remove the original binary upload steps.
- Let artifact registry config automatically upload binaries & dockers and generate provenance.
- Add a new final step to create versioned directories containing binaries to be automatically uploaded.

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,0 +1,1 @@
+../GEMINI.md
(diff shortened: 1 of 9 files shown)


.ci/versioned.release.cloudbuild.yaml
Yuan325, line 1: do we need to update the continuous release as well?

duwenxin99: I think we only need to update the artifacts that's available to the public users.

PR #2558, opened 2026-02-24
fix(server/mcp): scope defer span.End inside loop iteration

### LINKED ISSUE #2549: Bug: defer span.End() inside for loop in readInputStream() — spans never end until session closes
## Bug Description

`readInputStream()` in `internal/server/mcp.go` places `defer span.End()` inside a `for` loop. In Go, `defer` schedules execution at **function return**, not at the end of the enclosing block. This means every loop iteration creates a new span that is never ended until the entire function returns.

## Affected Code

**File:** `internal/server/mcp.go`, `readInputStream()` function

```go
func (s *stdioSession) readInputStream(ctx context.Context) error {
    for {
        // ...
        msgCtx, span := s.server.instrumentation.Tracer.Start(msgCtx, "toolbox/server/mcp/stdio",
            trace.WithSpanKind(trace.SpanKindServer),
        )
        defer span.End()   // BUG: defers accumulate, never run until function exits
        // ...
    }
}
```

## Impact

For long-running STDIO MCP sessions processing hundreds or thousands of messages:

1. **Memory leak**: Every span object accumulates in the tracer's active-span registry for the entire session lifetime
2. **Wrong span durations**: All spans show duration = (session end time - message receive time) instead of actual processing time
3. **OTel exporter flood**: On session disconnect, all accumulated spans are ended simultaneously, potentially overwhelming the exporter
4. **Deferred function stack growth**: Go accumulates all deferred calls on the goroutine's stack, growing linearly with message count

## Comparison with Correct Usage

The same file uses `defer span.End()` correctly **outside** loops:
- `sseHandler()` line ~325 — defer at function scope (correct)
- `processMcpMessage()` line ~575 — defer at function scope (correct)

The `cleanupRoutine()` (lines ~100-109) correctly uses a closure pattern to scope defers per iteration.

## Proposed Solution

**Option A (recommended):** Call `span.End()` explicitly at the end of the loop body
```go
for {
    msgCtx, span := s.server.instrumentation.Tracer.Start(...)
    // ... process message ...
    span.End()   // End span immediately after processing
}
```

**Option B:** Wrap the loop body in a closure to scope the defer
```go
for {
    func() {
        msgCtx, span := s.server.instrumentation.Tracer.Start(...)
        defer span.End()
        // ... process message ...
    }()
}
```

| Aspect | Option A | Option B |
|--------|----------|----------|
| Clarity | More explicit, easier to reason about | Familiar defer pattern but adds nesting |
| Error handling | Must ensure span.End() is called on all paths | defer handles all paths automatically |
| Performance | No closure overhead | Small closure allocation per iteration |

marlonbarreto-git:
## Description

In `readInputStream()`, `defer span.End()` is called inside a `for` loop. Since `defer` schedules execution when the enclosing **function** returns (not the current loop iteration), all spans created per message accumulate without ever being ended for the lifetime of the stdio session.

For a long-lived stdio server, this means:
- **Memory leak**: span objects accumulate indefinitely
- **Incorrect traces**: spans are never properly closed, producing misleading telemetry data

## Changes

Wrap the per-message processing logic in an immediately invoked function expression (IIFE), so `defer span.End()` correctly fires at the end of each iteration. This follows the same pattern already used in `InitializeConfigs` (`server.go` lines 85-98) for scoping spans inside loops.

## Test

All existing tests pass. The `TestStdioSession` test exercises `readLine` and `write` on the stdio session. The `readInputStream` loop is covered by the existing MCP endpoint tests that exercise the full message processing pipeline.

Fixes #2549


marlonbarreto-git:
@anubhav756 this PR is ready for review — single commit, CLA signed, conventional commit format.

Could you please trigger the CI by applying the `tests: run` label or commenting `/gcbrun`? As an external contributor I can't trigger the Cloud Build checks myself.

All tests pass locally:
```
$ go test ./internal/server/... -v
--- PASS: TestMcpEndpointWithoutInitialized (0.02s)
--- PASS: TestMcpEndpoint (0.03s)
--- PASS: TestStdioSession (0.00s)
PASS
ok  github.com/googleapis/genai-toolbox/internal/server  0.344s
```

Issue #2315, opened 2026-01-15
Views fail when allowedDatasets only includes the view's dataset, not the underlying table's dataset


pocky13:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### Description

When using `bigquery-execute-sql` with `allowedDatasets` restriction, queries against views fail if the view references tables from a different dataset that is not included in the `allowedDatasets` list, even when the view itself is in an allowed dataset.

### Environment
- MCP Toolbox version: 0.25.0
- Tool: `bigquery-execute-sql`
- Configuration: `allowedDatasets` restriction enabled

### Configuration Example
```yaml
sources:
  my-bigquery-source:
    kind: "bigquery"
    project: "my-project-id"
    allowedDatasets:
      - "dest_dataset"  # Only the destination dataset containing the view

tools:
  execute_sql_tool:
    kind: bigquery-execute-sql
    source: my-bigquery-source
```

### Dataset Structure
```
src_dataset.table         # Source table
dest_dataset.view         # View that references src_dataset.table
```

### Steps to Reproduce

1. Create a view in `dest_dataset` that references a table in `src_dataset`
```sql
   CREATE VIEW dest_dataset.view AS
   SELECT * FROM src_dataset.table;
```

2. Configure MCP Toolbox with `allowedDatasets: ["dest_dataset"]`

3. Execute a query against the view:
```sql
   SELECT * FROM dest_dataset.view;
```

### Expected Behavior

The query should succeed because:
- The directly referenced object (`dest_dataset.view`) is in the allowed datasets
- BigQuery's security model supports this pattern through [authorized views](https://cloud.google.com/bigquery/docs/authorized-views)
- Users with access to the view can query it without direct access to the underlying tables

### Actual Behavior

The query is rejected with an error indicating that `src_dataset.table` is not in the allowed datasets list.

According to the [documentation](https://googleapis.github.io/genai-toolbox/resources/tools/bigquery/bigquery-execute-sql/):

> Before execution, the tool performs a dry run to analyze the query. It will reject the query if it attempts to access any table outside the allowed datasets list.

The dry run exposes all table dependencies, including those referenced through views, causing the validation to fail.

### Current Workaround

Include both datasets in `allowedDatasets`:
```yaml
allowedDatasets:
  - "src_dataset"    # Must include the source dataset
  - "dest_dataset"
```

[...]

PR #2561, opened 2026-02-25
feat(bigquery): support authorized views with dataset restrictions

Genesis929:
Refactors BigQuery query validation to support Authorized Views by combining behavioral analysis (Dry Run) with explicit reference auditing.

### Working Logic:
The validation process uses a coordinated check:
1. Behavioral Detection (Dry Run): Identifies physical tables accessed by the query. If tables are found and all are allowed, the query is permitted immediately.
2. Explicit Reference Audit (Lexical Search): If the dry run identifies disallowed tables, a lexical scanner checks the SQL text for explicit mentions. Access is permitted if the disallowed tables are absent from the SQL, which correctly identifies Authorized View scenarios.
3. Fallback Intent Check: If no tables are found (e.g. DDL/Dynamic SQL) or to confirm the final intent, a local SQL parser validates the query to catch restricted operations and disallowed identifiers.

### Limitations:
- Structural vs. Lexical Gap: SQL parsing complexity makes it difficult to understand all structural nuances; lexical search provides a safer layer to detect user intent where the structural parser might fall short.
- False Positives: Queries may be blocked if non-table identifiers (like column aliases) exactly match a restricted table name. This prioritizes data security over permissive access.

### Changes:
- Added IsAnyTableExplicitlyReferenced using a lexical state machine.
- Reorganized ValidateQueryAgainstAllowedDatasets to optimize for performance (early return) and safety (coordinated checks).
- Simplified tool code by removing redundant mock indirection.
- Added tests for Authorized Views and complex table ID formats.

## Description

## PR Checklist

- [ ] Make sure you reviewed


internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -57,6 +57,100 @@ func DryRunQuery(ctx context.Context, restService *bigqueryrestapi.Service, proj
...
+		if len(parts) == 3 {
+			if !validator.IsDatasetAllowed(parts[0], parts[1]) {
+				violatingTables = append(violatingTables, tableID)
+			}
+		}
+	}
+
+	if len(tableIDSet) > 0 && len(violatingTables) == 0 {
+		return dryRunJob, nil
+	}
```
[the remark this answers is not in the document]
Genesis929: This should be secure enough.

internal/tools/bigquery/bigqueryanalyzecontribution/bigqueryanalyzecontribution.go
```diff
@@ -225,27 +245,12 @@ func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, para
 		// If not in protected mode, create a session for this invocation.
 		createModelQuery.CreateSession = true
 	}
+
 	if len(source.BigQueryAllowedDatasets()) > 0 {
-		createModelQuery.DryRun = true
-		dryRunJob, err := createModelQuery.Run(ctx)
+		_, err := bqutil.ValidateQueryAgainstAllowedDatasets(ctx, restService, source.BigQueryClient().Project(), source.BigQueryClient().Location, createModelSQL, nil, createModelQuery.ConnectionProperties, source, source.GetMaximumBytesBilled(), true)
 		if err != nil {
 			return nil, util.ProcessGcpError(err)
```
Yuan325: This should be `NewAgentError` or `NewClientServerErorr` instead? Are these errors fixable by the agent? Or require user to restart their Toolbox server with a new configuration? If there are different types within the function, we can also create the error within the function itself.

internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -148,20 +319,17 @@ func InitializeDatasetParameters(
...
+	return util.ProcessGcpError(err)
 }
 
-// ValidColumnParam returns true if s (stripped of leading/trailing single quotes) is a safe column name.
-func ValidColumnParam(s string) bool {
-	return ValidColumnName(StripSingleQuotes(s))
+// NewAgentError returns a new AgentError.
+func NewAgentError(message string, err error) util.ToolboxError {
+	return util.NewAgentError(message, err)
 }
```
Yuan325: This function seems redundant~

internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -148,20 +319,17 @@ func InitializeDatasetParameters(
...
-// StripSingleQuotes removes leading and trailing single quotes from a string if both are present.
-func StripSingleQuotes(s string) string {
-	if len(s) >= 2 && s[0] == '\'' && s[len(s)-1] == '\'' {
-		return s[1 : len(s)-1]
-	}
-	return s
+// ProcessGcpError converts a Google API error into a user-friendly error.
+func ProcessGcpError(err error) util.ToolboxError {
+	return util.ProcessGcpError(err)
 }
```
Yuan325: This function seems redundant. Seems like there's some import issues within the tools implementation. These functions are in `internal/util/errors.go`.

internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -148,20 +319,17 @@ func InitializeDatasetParameters(
...
+	return util.NewAgentError(message, err)
 }
 
-// ValidContributionMetricParam returns true if s (stripped of leading/trailing single quotes) is a safe contribution metric (does not contain single quotes).
-func ValidContributionMetricParam(s string) bool {
-	return !strings.ContainsRune(StripSingleQuotes(s), '\'')
+// NewClientServerError returns a new ClientServerError.
+func NewClientServerError(message string, statusCode int, err error) util.ToolboxError {
+	return util.NewClientServerError(message, statusCode, err)
 }
```
Yuan325: this as well.

internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -44,6 +45,24 @@ func ValidTableID(s string) bool {
...
+}
+
+// StripSingleQuotes removes leading and trailing single quotes from a string if both are present.
+func StripSingleQuotes(s string) string {
+	if len(s) >= 2 && s[0] == '\'' && s[len(s)-1] == '\'' {
+		return s[1 : len(s)-1]
+	}
+	return s
+}
+
```
Yuan325, line 67: Please add unit tests for these as well. I think we currently have the tests in `vaildator_test.go`? Feel free to rename that file to `util_test.go` and add the regression tests there. Thank you!

tests/common.go
```diff
@@ -1093,16 +1094,27 @@ func CleanupBigQueryDatasets(t *testing.T, ctx context.Context, client *bigquery
...
 		if err := ds.Delete(ctx); err != nil {
-			t.Errorf("INTEGRATION CLEANUP: Failed to delete dataset %s: %v", id, err)
+			if apiErr, ok := err.(*googleapi.Error); ok && apiErr.Code == 404 {
+				t.Logf("INTEGRATION CLEANUP: Dataset %s already deleted", id)
+			} else {
+				t.Errorf("INTEGRATION CLEANUP: Failed to delete dataset %s: %v", id, err)
+			}
 		} else {
 			t.Logf("INTEGRATION CLEANUP SUCCESS: Wiped dataset %s", id)
 		}
```
Yuan325: Do you mind moving this function into `bigquery_integration_test.go` since this function is only used by that source? Tyyy

Yuan325 (changes requested):

internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -84,6 +115,148 @@ func DryRunQuery(ctx context.Context, restService *bigqueryrestapi.Service, proj
...
+		return nil, util.NewAgentError("dry run failed to return query statistics", nil)
+	}
+
+	// Use a map to avoid duplicate table names from the dry run result.
+	tableIDSet := make(map[string]struct{})
+	queryStats := dryRunJob.Statistics.Query
+	if queryStats != nil {
+		for _, tableRef := range queryStats.ReferencedTables {
+			if tableRef != nil {
+				tableIDSet[fmt.Sprintf("%s.%s.%s", tableRef.ProjectId, tableRef.DatasetId, tableRef.TableId)] = struct{}{}
```
shobsi, line 156: can we simplify join and later split on '.' by storing structured (tuple/object of a custom class)? Also it misses a pattern where the project-id could have org prefixes containing '.', e.g. `google.com:project-id` (do consider adding a unit test with that special case).

structs are fully comparable and can be used as keys in maps. By defining a clean, typed key:

```go
type tableReference struct {
    ProjectID string
    DatasetID string
    TableID   string
}
```
And using it for the set:
```go
tableIDSet := make(map[tableReference]struct{})
```

internal/tools/bigquery/bigquerycommon/util.go
```diff
@@ -84,6 +115,148 @@ func DryRunQuery(ctx context.Context, restService *bigqueryrestapi.Service, proj
...
+			}
+			plural := ""
+			if len(violatingDatasets) > 1 {
+				plural = "s"
+			}
+			return nil, util.NewAgentError(fmt.Sprintf("access to dataset%s %s is not allowed", plural, strings.Join(violatingDatasets, ", ")), nil)
+		}
+	}
+
+	// Fall back to TableParser for final intent verification or if dry run was inconclusive.
```
shobsi, line 205: Can we write a comment why custom parsing is needed despite of dry run returning parsed tables in statistics, with examples? I know we were doing this before this PR but this feels counter-intuitive that we are writing custom SQL parsing in presence of dry run

internal/tools/bigquery/bigquerycommon/table_name_parser.go
```diff
@@ -255,39 +336,115 @@ func parseSQL(sql, defaultProjectID string, tableIDSet map[string]struct{}, visi
...
+				}
+				if !isKnownAlias && len(parts) > 1 {
+					if _, ok := aliases[strings.ToLower(parts[0])]; ok {
+						isKnownAlias = true
+					}
+				}
+
+				if expectingCTE {
+					aliases[fullID] = struct{}{}
+					aliases[strings.ToLower(parts[0])] = struct{}{}
```
shobsi, line 387: _[AI discovered, PTAL]_

Suppose `forbidden_dataset` is a restricted dataset. A user can write:
```sql
WITH forbidden_dataset.dummy AS (SELECT 1)
SELECT * FROM forbidden_dataset.highly_sensitive_table
```
1. When parsing `WITH forbidden_dataset.dummy`, `expectingCTE` is active.
2. `parts[0]` is `"forbidden_dataset"`, so `"forbidden_dataset"` is registered as an alias: `aliases["forbidden_dataset"] = struct{}{}`.
3. When parsing `forbidden_dataset.highly_sensitive_table`, `parts` is `["forbidden_dataset", "highly_sensitive_table"]`.
4. The parser checks if `parts[0]` is in `aliases` and marks `isKnownAlias = true` because `"forbidden_dataset"` was registered as an alias.
5. Because it is marked as an alias, `TableParser` completely ignores `forbidden_dataset.highly_sensitive_table` and does not add it to `tableIDSet`!

This allows a user to register any restricted dataset name as a CTE prefix and completely blind the `TableParser` fallback to any tables queried from that dataset.

#### **The Recommended Fix:**
We should **never** register `parts[0]` directly as an alias. Instead, only register the `fullID` as an alias, and when checking if an identifier is an alias, recursively inspect its dot-separated prefixes:

```go
isKnownAlias := false
// 1. Direct match on full identifier (e.g. "my_alias" or "derived.results")
if _, ok := aliases[fullID]; ok {
    isKnownAlias = true
}
// 2. Prefix match on dot-separated identifier paths (e.g. "derived.results" for column reference "derived.results.col")
if !isKnownAlias && len(parts) > 1 {
    for j := 1; j < len(parts); j++ {
        prefix := strings.ToLower(strings.Join(parts[:j], "."))
        if _, ok := aliases[prefix]; ok {
            isKnownAlias = true
            break
        }
    }
}
```
This correctly resolves dot-containing CTE references like `derived.results.double_score` while completely eliminating the security vulnerability of registering top-level dataset names as aliases.

pambroziak-rtbh:
Confirming this is still broken on **v1.8.0**, with a real-world authorized-view estate. Adding a concrete reproduction in case it helps move this along — the combination of this PR's gap and the startup metadata check currently makes `allowedDatasets` **unusable** for us, with no workaround.

### Setup

A chain of views ending in datasets we deliberately have no direct grants on (standard authorized-view governance):

```
proj-a.analytics.my_view                (view, the only thing we query)
  └─> proj-b.analytics.my_view_sample   (view)
        └─> proj-c.analytics_eu.*       (AUTHORIZED VIEWS — we can read results here)
              └─> proj-d.base_ds_a.base_table_a   (no access; datasets.get -> 403)
                  proj-d.base_ds_b.base_table_b   (no access; datasets.get -> 403)
```

Source config:

```yaml
sources:
  bq-guarded:
    kind: bigquery
    project: proj-a
    location: EU
    allowedDatasets:
      - analytics
      - proj-b.analytics
      - proj-c.analytics_eu
    maximumBytesBilled: 10737418240
    writeMode: protected
```

### What happens

Querying only the top view — the SQL never mentions anything in `proj-d`:

```sql
SELECT COUNT(*) FROM `proj-a.analytics.my_view`
WHERE partition_timestamp >= TIMESTAMP('2026-07-01')
  AND partition_timestamp <  TIMESTAMP('2026-07-02')
```

is rejected:

```
query accesses dataset 'proj-d.base_ds_a', which is not in the allowed list
```

The dry run resolves straight through the authorized views to the base tables and reports datasets the caller cannot see. This is exactly the scenario this PR's lexical-reference audit is designed to permit: the disallowed dataset appears nowhere in the SQL text.

Confirmation that the query itself is fine — removing `allowedDatasets` entirely (keeping the same credentials) makes both queries succeed:

```
totalBytesProcessed = 577,596,852   (0.54 GiB)
totalBytesProcessed = 1,047,193,326 (0.98 GiB)
```

So access is legitimate and BigQuery is happy; only the app-layer allowlist blocks it.

### Why there is no workaround

The obvious workaround — just add `proj-d.base_ds_a` / `proj-d.base_ds_b` to `allowedDatasets` — fails, because entries are verified with `datasets.get` at **startup** and the server refuses to boot:

```
toolbox failed to initialize: unable to initialize configs: unable to initialize source "bq-guarded":
failed to verify allowedDataset 'base_ds_a' in project 'proj-d': googleapi: Error 403:
Access Denied: Dataset proj-d:base_ds_a: Permission bigquery.datasets.get denied on dataset
proj-d:base_ds_a (or it may not exist)., accessDenied
```

So the two behaviors combine into a hard bind:

- **omit** the base datasets → queries rejected at runtime (this PR)
- **list** the base datasets → server won't start (startup verification)

`checkAtStartup: false` is rejected as an unknown field on v1.8.0 and there's no `--allow-partial-sources` flag (#2662 was closed unmerged, and explicitly noted BigQuery was the exception its approach didn't cover). I've filed the startup half separately so it isn't lost, but **this PR alone would fully unblock us**: with the lexical audit in place we'd keep the allowlist as written above and never need to name the inaccessible datasets at all.

Happy to test a build against this estate if that's useful.

Related: #2315 (same root cause, issue form), #2662 (closed; startup checks, BigQuery excluded).

internal/tools/bigquery/bigquerycommon/util.go
Genesis929: Good call — done. Introduced a typed `tableReference{ProjectID, DatasetID, TableID}` struct and
switched the set to `map[tableReference]struct{}`, so we no longer join-then-split on '.'.
This also fixes the domain-scoped project ID case; added unit tests with
`google.com:corp-proj` in `util_test.go` and `google.com:project.dataset.table` in
`table_name_parser_test.go`.

internal/tools/bigquery/bigquerycommon/util.go
Genesis929: Added a doc comment explaining why the lexical parser is still needed alongside dry run, with examples:
1. Dry run resolves authorized views down to their underlying base tables, so a legitimate query
   against an allowed view would look like it touches a restricted dataset. We need the parser to
   tell whether the restricted dataset was *explicitly* referenced in the user's SQL.
2. Dry run statistics alone don't surface every security boundary (e.g. dataset-level DDL,
   `SET @@dataset_id`, etc.).
So: dry run = source of truth for what is actually read; parser = source of truth for what the user
literally wrote. PTAL at the comment and let me know if you'd like more examples.

internal/tools/bigquery/bigquerycommon/table_name_parser.go
Genesis929: Great catch, fixed. We no longer register `parts[0]` as an alias — only the full identifier is
registered, and prefix matching is restricted to actual CTE names (`cteNames`), so a dataset name
used as a CTE prefix can no longer blind the parser.
Added a regression test: "CTE multi-part prefix does not shadow restricted dataset" with your exact
repro; it now correctly yields `myproj.forbidden_dataset.highly_sensitive_table`.

docs/en/integrations/bigquery/tools/bigquery-analyze-contribution.md
```diff
@@ -52,8 +52,11 @@ the `bigquery` source:
   `input_data` parameter only accesses tables within the allowed datasets.
   - If `input_data` is a table ID, the tool checks if the table's dataset is in
     the allowed list.
-  - If `input_data` is a query, the tool performs a dry run to analyze the query
-    and rejects it if it accesses any table outside the allowed list.
+  - If `input_data` is a query, the tool performs a dry run to analyze the query.
+    It will reject the query if it explicitly references any table outside the
+    allowed list. Like [`bigquery-execute-sql`](../bigquery-execute-sql), authorized views are supported:
```
shobsi, line 57: [nit] I see several wordings like

* allowed list
* allowed datasets
* allowed `datasets` list

Consider using "`allowedDatasets`" to directly link it to the config rather than leaving it on the user to infer it from the context.

shobsi (approved):

docs/en/integrations/bigquery/tools/bigquery-analyze-contribution.md
Genesis929: Done across all three pages this PR touches. Four other BigQuery pages still say "the allowed
list" in their prose — they're not part of this PR, so I'll fix them in a separate docs-only PR.

internal/tools/bigquery/bigqueryanalyzecontribution/bigqueryanalyzecontribution.go
Genesis929: Done — `NewAgentError` for anything the agent can fix by changing its request,
`NewClientServerError` for server/config failures, `ProcessGcpError` for BigQuery API errors.

internal/tools/bigquery/bigquerycommon/util.go
Genesis929: You're right, all three were pure pass-throughs. Removed — `bigquerycommon` now calls
`internal/util` directly.

internal/tools/bigquery/bigquerycommon/util.go
Genesis929: Removed along with the other two

internal/tools/bigquery/bigquerycommon/util.go
Genesis929: Removed along with the other two

tests/common.go
Genesis929: Moved — it's now `ensureTeardownDatasets` in `bigquery_integration_test.go`.

internal/tools/bigquery/bigquerycommon/util.go
Genesis929: Tests added (`TestStripSingleQuotes`, `TestValidColumnParam`,
`TestValidContributionMetricParam`).

I'd rather skip the rename though — `validators_test.go` already exists on `main`, so renaming
it would pull an unrelated file into this diff. The split is clean as is: validators there,
`ValidateQueryAgainstAllowedDatasets` in the new `util_test.go`. Happy to merge them if you
still prefer one file.

PR #2582, opened 2026-02-28
fix(oracle): allow successful exec without rows-affected metadata

### LINKED ISSUE #2044: Oracle raises error when the query uses "RETURN" keyword
### Prerequisites

### Toolbox version

0.21.0

### Environment

_No response_

### Client

toolbox_core with ADK

### Expected Behavior

Tool should return empty result.

### Current Behavior

Image

### Steps to reproduce?

1. Configure oracle-sql tool with "RETURN" keyword in a PL/SQL block
2. Start Oracle server
3. Invoke the tool
...

### Additional Details

_No response_

vishnukannaujia:
## Summary
- stop failing Oracle non-readOnly executions when `RowsAffected()` is unavailable
- return a success response without `rows_affected` metadata when the driver does not provide it
- add unit tests for both code paths (rows count available vs unavailable)

## Why
Issue #2044 reports failures when Oracle SQL includes `RETURN` in PL/SQL execution flows. In Oracle, some successful executions (including PL/SQL blocks) may not support `RowsAffected()`. The current code treats that as an error and fails the tool invocation.

## Fix
Refactor exec response formatting to treat missing rows-affected metadata as non-fatal. Successful execution now returns:
- `{ "status": "success", "rows_affected": <n> }` when available
- `{ "status": "success" }` when unavailable

## Testing
- Added `TestFormatExecResponseWithRowsAffected`
- Added `TestFormatExecResponseWithoutRowsAffected`

Fixes #2044

### DIFF
--- a/internal/sources/oracle/oracle.go
+++ b/internal/sources/oracle/oracle.go
@@ -143,15 +143,7 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any, rea
 			return nil, fmt.Errorf("unable to execute DML statement: %w", err)
 		}
 
-		rowsAffected, err := result.RowsAffected()
-		if err != nil {
-			return nil, fmt.Errorf("unable to get rows affected: %w", err)
-		}
-
-		return map[string]any{
-			"status":        "success",
-			"rows_affected": rowsAffected,
-		}, nil
+		return formatExecResponse(result), nil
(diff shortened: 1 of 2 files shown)


vishnukannaujia:
Superseded by #2585 to remove codex references from branch naming.

PR #2557, opened 2026-02-24
fix(server/mcp): guard nil dereference in sseManager.get

marlonbarreto-git:
## Description

`sseManager.get()` unconditionally accesses `session.lastActive` after a map lookup, even when the session ID doesn't exist. Since the zero value for a `*sseSession` pointer is `nil`, this causes a nil pointer dereference panic at runtime.

This can happen when a client sends a request with a stale or invalid `sessionId` query parameter — for example, after a network reconnection or if a session was cleaned up by `cleanupRoutine`.

## Changes

Add an `ok` guard before accessing `session.lastActive`:

```go
session, ok := m.sseSessions[id]
if ok {
    session.lastActive = time.Now()
}
```

## Test

Added `TestSseManagerGetNonExistentSession` that calls `get()` with a non-existent ID and verifies it returns `nil, false` without panicking.

Fixes #2548

PR #2583, opened 2026-02-28
fix(mcp): prevent panic when SSE session is missing

### LINKED ISSUE #2548: Bug: sseManager.get() dereferences nil session pointer — panics on unknown sessionId
## Bug Description

`sseManager.get()` in `internal/server/mcp.go` (lines ~60-65) unconditionally sets `session.lastActive = time.Now()` without checking if `session` is nil. When the session ID is not found in the map, Go returns the zero value (`nil`) for the pointer, and the field assignment panics.

## Affected Code

**File:** `internal/server/mcp.go`, lines 60-66

```go
func (m *sseManager) get(id string) (*sseSession, bool) {
    m.mu.Lock()
    defer m.mu.Unlock()
    session, ok := m.sseSessions[id]
    session.lastActive = time.Now()   // PANICS when session is nil
    return session, ok
}
```

## Trigger

Any HTTP POST to `/mcp` with a `?sessionId=<invalid-or-expired-id>` query parameter that doesn't match an active SSE session. The caller in `httpHandler` (line ~434) checks `ok` **after** calling `get()`, but the panic fires inside `get()` before the guard can execute:

```go
session, ok = s.sseManager.get(sessionId)
if !ok {   // <-- too late, already panicked
    s.logger.DebugContext(ctx, "sse session not available")
}
```

## Impact

- **Severity: Critical** — Any malformed, expired, or guessed session ID crashes the HTTP handler
- Any client reconnecting after a server restart with a stale session cookie triggers this
- Trivially reproducible: `curl -X POST "https://server/mcp?sessionId=nonexistent"`

## Steps to Reproduce

1. Start a genai-toolbox MCP server with SSE transport enabled
2. Send a POST request with an invalid session ID:
   ```bash
   curl -X POST "http://localhost:5000/mcp?sessionId=does-not-exist" \
     -d '{"jsonrpc":"2.0","method":"tools/list","id":1}'
   ```
3. Server panics with nil pointer dereference

## Expected Behavior

`get()` should guard against nil session before accessing fields.

## Proposed Solution

```go
func (m *sseManager) get(id string) (*sseSession, bool) {
    m.mu.Lock()
    defer m.mu.Unlock()
    session, ok := m.sseSessions[id]
    if ok {
        session.lastActive = time.Now()
    }
    return session, ok
}
```

This is a one-line fix that matches the existing pattern used in `delete()` (which already checks `ok` before accessing the session).

vishnukannaujia:
## Summary
- fix `sseManager.get` to only update `lastActive` when the session exists
- return `(nil, false)` cleanly for unknown `sessionId` lookups
- add a regression test for unknown session retrieval

## Why
Issue #2548 reports a nil pointer dereference in `sseManager.get()` when `sessionId` is unknown. The map lookup can return `nil`, and unguarded field assignment panics.

## Fix
Guard the `lastActive` update behind `ok`, preserving existing behavior for valid sessions and avoiding panic for invalid IDs.

## Testing
- Added `TestSseManagerGetUnknownSession` in `internal/server/mcp_test.go`

Fixes #2548

### DIFF
--- a/internal/server/mcp_test.go
+++ b/internal/server/mcp_test.go
@@ -1176,10 +1176,7 @@ func TestSseManagerGetUnknownSession(t *testing.T) {
 	manager := newSseManager(ctx)
 
 	session, ok := manager.get("unknown-session-id")
-	if ok {
-		t.Fatal("expected unknown session lookup to return ok=false")
-	}
-	if session != nil {
-		t.Fatal("expected unknown session lookup to return nil session")
+	if ok || session != nil {
+		t.Fatalf("manager.get() with unknown id returned (%v, %v), want (nil, false)", session, ok)
 	}
 }


vishnukannaujia:
Addressed review feedback from Gemini:
- updated `TestSseManagerGetUnknownSession` assertion to report both return values in a single diagnostic

Change pushed to the same branch: `fix-2548-ssemanager-get-nil-panic` (commit `c7059bc229`).

I will proceed with any additional maintainer feedback once CLA is complete.

averikitsch:
Thanks for the contribution. We have merged this fix in https://github.com/googleapis/genai-toolbox/pull/2557/changes. Thanks again!

PR #2584, opened 2026-02-28
fix(mcp): end stdio spans per message

### LINKED ISSUE #2549: Bug: defer span.End() inside for loop in readInputStream() — spans never end until session closes

vishnukannaujia:
## Summary
- replace `defer span.End()` inside the message loop with explicit per-iteration `span.End()`
- ensure spans are ended before returning on write errors
- add regression test that verifies spans are ended while the input loop is still running

## Why
Issue #2549 reports `defer span.End()` in `readInputStream` inside a `for` loop. In long-lived stdio sessions, deferred calls accumulate until function exit, causing memory growth and delayed span finalization.

## Fix
Call `span.End()` at the end of each loop iteration and before early return paths, so each message span is finalized immediately.

## Testing
- Added `TestStdioSessionReadInputStreamEndsSpansPerMessage` in `internal/server/mcp_test.go`

Fixes #2549

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -200,7 +200,6 @@ func (s *stdioSession) readInputStream(ctx context.Context) error {
 		msgCtx, span := s.server.instrumentation.Tracer.Start(msgCtx, "toolbox/server/mcp/stdio",
 			trace.WithSpanKind(trace.SpanKindServer),
 		)
-		defer span.End()
 
 		v, res, err := processMcpMessage(msgCtx, []byte(line), s.server, s.protocol, "", "", nil, "")
 		if err != nil {
@@ -216,9 +215,11 @@ func (s *stdioSession) readInputStream(ctx context.Context) error {
 		// no responses for notifications
 		if res != nil {
 			if err = s.write(msgCtx, res); err != nil {
+				span.End()
 				return err
 			}
 		}
+		span.End()
 	}
 }
 
--- a/internal/server/mcp_test.go
+++ b/internal/server/mcp_test.go
@@ -20,17 +20,22 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/googleapis/genai-toolbox/internal/log"
 	"github.com/googleapis/genai-toolbox/internal/server/mcp/jsonrpc"
 	"github.com/googleapis/genai-toolbox/internal/server/resources"
 	"github.com/googleapis/genai-toolbox/internal/telemetry"
+	"go.opentelemetry.io/otel"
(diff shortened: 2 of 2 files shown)


vishnukannaujia:
Reviewed current feedback on this PR. The span lifecycle updates called out in review are already in the branch and no additional code changes were requested.

Current implementation includes explicit `span.End()` on both normal and error paths in `readInputStream`.

I’ll address any maintainer-requested follow-ups after CLA is completed.

Yuan325:
Hi @vishnukannaujia, there's already another fix being submitted by the bug issuer, thank you for your contribution! Closing this due to duplication.

PR #2619, opened 2026-03-03
feat(auth): add generic `authService` type for MCP

duwenxin99:
Add new generic `authService` type as part of the MCP auth implementation.


internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+// validate interface
+var _ auth.AuthServiceConfig = Config{}
+
+// Auth service configuration
+type Config struct {
+	Name           string   `yaml:"name" validate:"required"`
+	Type           string   `yaml:"type" validate:"required"`
+	ClientID       string   `yaml:"clientId" validate:"required"`
+	McpEnabled     bool     `yaml:"mcpEnabled"`
+	AuthURL        string   `yaml:"authUrl" validate:"required"`
```
averikitsch: In theory, authurl is not required if using as generic tool level auth service.

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+const AuthServiceType string = "generic"
+
+// validate interface
+var _ auth.AuthServiceConfig = Config{}
+
+// Auth service configuration
+type Config struct {
+	Name           string   `yaml:"name" validate:"required"`
+	Type           string   `yaml:"type" validate:"required"`
+	ClientID       string   `yaml:"clientId" validate:"required"`
```
averikitsch: How are users generating client ids for generic auth services. I don't think this should be required.

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+
+	a := &AuthService{
+		Config: cfg,
+		kf:     kf,
+	}
+	return a, nil
+}
+
+func discoverJWKSURL(authURL string) (string, error) {
+	authURL = strings.TrimSuffix(authURL, "/")
```
averikitsch: We should probably use a URL parsing library

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// Validate 'aud' (audience) claim
+	aud, err := claims.GetAudience()
+	if err != nil {
+		return nil, fmt.Errorf("could not parse audience from token: %w", err)
+	}
+
+	isAudValid := false
+	for _, audItem := range aud {
+		if audItem == a.ClientID {
```
averikitsch: audience is not the same as client id

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+			isAudValid = true
+			break
+		}
+	}
+
+	// Some IDPs use 'client_id' instead of 'aud' or put it as a single string, checking that if aud not found or not matched
+	if !isAudValid {
+		if clientIDClaim, ok := claims["client_id"].(string); ok && clientIDClaim == a.ClientID {
```
averikitsch: client_id isn't a part of the JWT spec https://datatracker.ietf.org/doc/html/rfc7519. What IDPs use this?

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+		for _, requiredScope := range a.ScopesRequired {
+			if !scopeMap[requiredScope] {
+				return nil, fmt.Errorf("missing required scope: %s", requiredScope)
+			}
+		}
+	}
+
+	// Return claims dynamically
+	claimsMap := make(map[string]any)
+	for k, v := range claims {
```
averikitsch: can we just return claims?

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	a := &AuthService{
+		Config: cfg,
+		kf:     kf,
+	}
+	return a, nil
+}
+
+func discoverJWKSURL(authURL string) (string, error) {
```
averikitsch: If this fails (ie no /.well-known/openid-configuration?) how do we verify the token is valid?

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+	authURL = strings.TrimSuffix(authURL, "/")
+	oidcConfigURL := authURL + "/.well-known/openid-configuration"
+	client := &http.Client{Timeout: 10 * time.Second}
+	resp, err := client.Get(oidcConfigURL)
```
averikitsch: This slightly worries me about trying an unknown endpoint then passing it directly to the decoder

averikitsch (changes requested):

internal/auth/generic/generic.go
duwenxin99: We need the auth server URL to be able to get the public key for signature verification for tool-level auth.

internal/auth/generic/generic.go
duwenxin99: If it fails, we won't able to verify the token because we need the public key for verification.

internal/auth/generic/generic.go
duwenxin99: Updated client_id to audience

internal/auth/generic/generic.go
duwenxin99: Updated to the audience field.

internal/auth/generic/generic.go
duwenxin99: I agree. I added a custom dialer that validates IP addresses at the connection level, enforcing HTTPS, and disabling automatic redirects.

internal/auth/generic/generic.go
averikitsch: Do we need to do this conversion?

internal/auth/generic/generic.go
averikitsch: nit: authorizationServer maybe more clear of a name

averikitsch: This URL can also come from the JWT's decoded payload "iss" field

internal/auth/generic/generic.go
averikitsch: Did we want to update this?

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("failed to fetch OIDC config, status code %d", resp.StatusCode)
+	}
+
+	var config map[string]interface{}
+	if err := json.NewDecoder(resp.Body).Decode(&config); err != nil {
```
averikitsch: We should block potential malicious attacks by restricting decoding size
`resp.Body = http.MaxBytesReader(nil, resp.Body, 1<<20) // 1MB limit`

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+
+// Returns the auth service type
+func (cfg Config) AuthServiceConfigType() string {
+	return AuthServiceType
+}
+
+// Initialize a generic auth service
+func (cfg Config) Initialize() (auth.AuthService, error) {
+	// Discover the JWKS URL from the OIDC configuration endpoint
+	jwksURL, err := discoverJWKSURL(cfg.AuthURL)
```
averikitsch: we should first check for the use of https

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,209 @@
...
+	return a.Config
+}
+
+// Returns the name of the auth service
+func (a AuthService) GetName() string {
+	return a.Name
+}
+
+// Verifies generic JWT access token inside the Authorization header
+func (a AuthService) GetClaimsFromHeader(ctx context.Context, h http.Header) (map[string]any, error) {
```
averikitsch, line 166: This is only for the tool invocation not root level auth

internal/auth/generic/generic.go
averikitsch: IIRC, access tokens will have aud as the MCP server url and JWts with have aud a auth server urls

internal/auth/generic/generic.go
duwenxin99: added

internal/auth/generic/generic.go
duwenxin99: updated to AuthorizationServerUrl

internal/auth/generic/generic.go
duwenxin99: added check

internal/auth/generic/generic.go
duwenxin99: We are only supporting JWT access tokens for the root-level auth. The aud field should be the MCP server when the user is using root-level auth, and it should be the client ID if the user is using regular auth.

internal/auth/generic/generic.go
duwenxin99: Updated to be used only for tool invocation

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,239 @@
...
+// validate interface
+var _ auth.AuthServiceConfig = Config{}
+
+// Auth service configuration
+type Config struct {
+	Name                   string   `yaml:"name" validate:"required"`
+	Type                   string   `yaml:"type" validate:"required"`
+	Audience               string   `yaml:"audience" validate:"required"`
+	McpEnabled             bool     `yaml:"mcpEnabled"`
+	AuthorizationServerUrl string   `yaml:"AuthorizationServerUrl" validate:"required"`
```
averikitsch:
```suggestion
	AuthorizationServerUrl string   `yaml:"authorizationServerUrl" validate:"required"`
```

cmd/internal/tools_file_test.go
```diff
@@ -624,6 +625,16 @@ func TestParseToolFile(t *testing.T) {
 			name: my-google-auth
 			type: google
 			clientId: testing-id
+---
+			kind: authServices
+			name: my-generic-auth
+			type: generic
+			audience: testings
+			authorizationServerUrl: https://testings
```
averikitsch: nit: authorizationServerUrl is a long field name. I recommend authorizationServer or authzServer. Let me know if you think URL is necessary

internal/auth/generic/generic_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/MicahParks/jwkset"
+	"github.com/golang-jwt/jwt/v5"
+)
+
+func init() {
+	AllowInsecureForTest = true
```
averikitsch: I was wrong. We probably need to support http as well. A couple paths forward
1) use a flag or field to "allowInsecure" 
2) just log that http instead of https is being used.

docs/en/resources/authServices/generic.md
```diff
@@ -0,0 +1,65 @@
...
+  Use a Generic OpenID Connect (OIDC) provider for OAuth 2.0 flow and token
+  lifecycle.
+---
+
+## Getting Started
+
+The Generic Auth Service allows you to integrate with any OpenID Connect (OIDC)
+compliant identity provider (IDP). It discovers the JWKS (JSON Web Key Set) URL
+either through the provider's `/.well-known/openid-configuration` endpoint or
+directly via the provided `authUrl`.
```
averikitsch: need to update authUrl

internal/auth/generic/generic.go
averikitsch: How are we running root level auth checks?

internal/auth/generic/generic.go
```diff
@@ -0,0 +1,239 @@
...
+
+	isAudValid := false
+	for _, audItem := range aud {
+		if audItem == a.Audience {
+			isAudValid = true
+			break
+		}
+	}
+
+	if !isAudValid {
```
averikitsch, line 205: Where are we validating scopes?

internal/auth/generic/generic.go
averikitsch: Is this necessary?

duwenxin99: Removed this block.

cmd/internal/tools_file_test.go
duwenxin99: I think having Url is clearer to to the users what this field is expecting (instead of the auth server name like "Google"), but I agree that it's a long name. Replaced with `AuthorizationServer`.

internal/auth/generic/generic_test.go
duwenxin99: Got it. Changed to logging http for now.

internal/auth/generic/generic.go
duwenxin99: I was planning on adding it in a separate PR but added it here so that it is more self-contained.

internal/auth/generic/generic.go
duwenxin99: It is true if we want to allow the client to pass in whichever auth service they want. But the server may want to limit the auth provider it allows?

duwenxin99: A concern about getting the issuer directly from the jwt token is that an attacker could set up its own fake auth server and use it as the "iss" field to bypass any legit authorization.

Issue #2614, opened 2026-03-03
Oracle oracle-execute-sql tool broken in v0.28.0 — RunSQL interface mismatch after DML fix


iviireczech:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

0.28.0+binary.windows.amd64.81253a0

### Environment

1. OS: Windows 11 Enterprise 10.0.22631 (x64)
2. Running via npm: npx @toolbox-sdk/server@0.28.0 --stdio --tools-file tools.yaml

### Client

1. Client: Claude Code (Anthropic CLI) via MCP stdio
2. Version: 2.1.63
3. MCP protocol version 2024-11-05

### Expected Behavior

Invoking an oracle-execute-sql tool should successfully execute SQL against an Oracle source of type oracle, as it did in v0.27.0.

### Current Behavior

Every oracle-execute-sql tool invocation fails with:
source used is not compatible with the tool: invalid source for "oracle-execute-sql" tool: source "SOURCE_NAME" is not a compatible type

All sources initialize successfully (confirmed via --log-level DEBUG), but the type compatibility check fails at invocation time. Downgrading to v0.27.0 resolves the issue.

### Steps to reproduce?

1. Create a tools.yaml with an Oracle source and an oracle-execute-sql tool (v2 config format):
kind: sources
name: MY_ORACLE
type: oracle
host: my-oracle-host
port: 1521
serviceName: my_service
user: my_user
password: my_password
---
kind: tools
name: my-execute-sql
type: oracle-execute-sql
source: MY_ORACLE
description: Execute SQL on Oracle

2. Start toolbox v0.28.0: toolbox --stdio --tools-file tools.yaml
Send MCP tools/call for my-execute-sql with {"sql": "SELECT 1 FROM dual"}

3. Observe error: source used is not compatible with the tool: invalid source for "oracle-execute-sql" tool

### Additional Details

The DML fix in commit: 

fix(oracle): enable DML operations and resolve incorrect array type error (included in v0.28.0)

added a readOnly bool parameter to the Oracle source's RunSQL method:
// internal/sources/oracle/oracle.go (v0.28.0)
func (s *Source) RunSQL(ctx context.Context, statement string, params []any, readOnly bool) (any, error)

However, the oracle-execute-sql tool's compatibleSource interface was NOT updated to match:
// internal/tools/oracle/oracleexecutesql/oracleexecutesql.go (v0.28.0)
[...]

PR #2624, opened 2026-03-04
fix: Revert "fix(oracle): enable DML operations and resolve incorrect array type error"

anubhav756:
Reverts googleapis/genai-toolbox#2323

Seems to be introducing regression here: https://github.com/googleapis/genai-toolbox/issues/2614 which was also blocking presubmit integration tests in PRs.


anubhav756:
Since the integration tests break in the current revert PR and the fact that we are in a broken state, we decided to fix-forward.

https://github.com/googleapis/genai-toolbox/pull/2627 fix-forwards this issue. Closing this PR in favor of that.

PR #2627, opened 2026-03-04
fix(oracle): update oracle-execute-sql tool interface to match source signature

anubhav756:
## Description

This PR fixes a critical signature mismatch that caused all `oracle-execute-sql` tool invocations to fail with an incompatible source type error (`500` Internal Server Error).

In a previous PR (https://github.com/googleapis/genai-toolbox/pull/2323), the `RunSQL` method in the Oracle Source was updated to include a `readOnly` bool parameter to support DML operations for the `oracle-sql` tool:

However, the `oracle-execute-sql` tool's required interface (`compatibleSource`) was inadvertently left unchanged, still expecting the old 3-argument signature. Because Go interfaces are satisfied implicitly, the updated Oracle source no longer implemented the `oracle-execute-sql` tool's interface, causing `tools.GetCompatibleSource` to reject it at runtime during invocation.

🛠️ Fixes https://github.com/googleapis/genai-toolbox/issues/2614

### DIFF
--- a/internal/tools/oracle/oracleexecutesql/oracleexecutesql_test.go
+++ b/internal/tools/oracle/oracleexecutesql/oracleexecutesql_test.go
@@ -16,6 +16,9 @@ func TestParseFromYamlOracleExecuteSql(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %s", err)
 	}
+
+	valTrue := true
+	valFalse := false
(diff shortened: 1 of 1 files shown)


anubhav756:
> Can we add regression tests in a subsequent PR? Thanks!

Added regression tests in this PR for convenience.

PR #2635, opened 2026-03-04
feat(source/cloudsql)!: restructure prebuilt toolsets

averikitsch:
## Description

## PR Checklist

- [ ] Make sure you reviewed

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -154,6 +154,30 @@ func newMeterProvider(ctx context.Context, r *resource.Resource, telemetryOTLP s
 		metricOpts = append(metricOpts, metric.WithReader(metric.NewPeriodicReader(gcpExporter)))
 	}
 
+	// Configure custom histogram bucket boundaries for duration metrics as per MCP semantic conventions.
+	// Source: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#metric-mcpserversessionduration
+	durationBuckets := []float64{0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300}
+
+	// Create views for duration histograms
+	durationHistogramNames := []string{
+		mcpOperationDurationName,
+		mcpSessionDurationName,
+		toolExecutionDurationName,
+	}
+
+	views := make([]metric.View, 0, len(durationHistogramNames))
+	for _, name := range durationHistogramNames {
+		views = append(views, metric.NewView(
+			metric.Instrument{Name: name},
+			metric.Stream{
+				Aggregation: metric.AggregationExplicitBucketHistogram{
+					Boundaries: durationBuckets,
+				},
+			},
+		))
+	}
+	metricOpts = append(metricOpts, metric.WithView(views...))
+
 	meterProvider := metric.NewMeterProvider(metricOpts...)
 	return meterProvider, nil
 }


internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
```diff
@@ -13,46 +13,143 @@
...
-  cloud-sql-mysql-source:
-    kind: cloud-sql-mysql
-    project: ${CLOUD_SQL_MYSQL_PROJECT}
-    region: ${CLOUD_SQL_MYSQL_REGION}
-    instance: ${CLOUD_SQL_MYSQL_INSTANCE}
-    database: ${CLOUD_SQL_MYSQL_DATABASE}
-    user: ${CLOUD_SQL_MYSQL_USER:}
-    password: ${CLOUD_SQL_MYSQL_PASSWORD:}
-    ipType: ${CLOUD_SQL_MYSQL_IP_TYPE:PUBLIC}
+    cloud-sql-mysql-source:
```
shuzhou-gc: Are those 3 sources supposedly to be "fixed" - not sure if we want to keep all different dbs in sync here?

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
```diff
@@ -13,46 +13,143 @@
...
+        source: cloud-sql-admin-source
 toolsets:
-  cloud_sql_mysql_database_tools:
-    - execute_sql
-    - list_tables
-    - get_query_plan
-    - list_active_queries
-    - list_tables_missing_unique_indexes
-    - list_table_fragmentation
+    admin:
```
shuzhou-gc: Similar questions, are these "sets" fixed or can be updated independently (without keeping other db in sync)?

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
```diff
@@ -13,46 +13,143 @@
...
+        - create_instance
+        - get_instance
+        - list_instances
+        - create_database
+        - list_databases
+        - create_user
+        - wait_for_operation
+    data:
```
shuzhou-gc: What is the mapping between sources & toolset?

averikitsch: Each tool has a source. These prebuilt configs have a default toolset of all tools or named toolsets of these tools.

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
averikitsch: We do not need the same toolsets or tools across Cloud SQL engines

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
averikitsch: There are 3 sources but only one connects to the database. The monitoring and admin tools use an API client.

cmd/internal/tools_file_test.go
```diff
@@ -1960,6 +1996,15 @@ func TestPrebuiltTools(t *testing.T) {
 			if len(toolsFile.Prompts) != 0 {
 				t.Fatalf("expected empty prompts map for prebuilt config, got: %v", toolsFile.Prompts)
 			}
+
+			t.Run("check toolset sizes", func(t *testing.T) {
+				for tsName, ts := range toolsFile.Toolsets {
+					if len(ts.ToolNames) > 10 {
+						t.Logf("WARNING: Toolset %q in config %q has %d tools, which is larger than the recommended maximum of 10.", tsName, tc.name, len(ts.ToolNames))
+						fmt.Printf("WARNING: Toolset %q in config %q has %d tools, which is larger than the recommended maximum of 10.\n", tsName, tc.name, len(ts.ToolNames))
```
averikitsch:
```suggestion
```

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
shuzhou-gc: Got it, thanks for the clarification.

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
```diff
@@ -47,12 +52,107 @@ tools:
...
     - list_tables
     - get_query_plan
     - list_active_queries
-    - list_tables_missing_unique_indexes
+  monitor:
+    - get_query_plan
+    - list_active_queries
+    - get_query_metrics
+    - get_system_metrics
     - list_table_fragmentation
```
shuzhou-gc, line 150: Thanks for this change, I was about to ask.

Also not sure what's your take on get_query_plan, which might be more like data rather than monitor since the query is probably not in-flight.

Again, I guess I don't really know the exact definition of "data" vs "monitor"

Yuan325: @averikitsch will get back to you next week on this, we'll be merging this for now to unblock other features. TY for checking! :)

PR #2645, opened 2026-03-05
fix(skill): fix env variable propagation

helloeve:
This PR refactors the tool file parsing logic by introducing a new ToolsFileParser struct. This encapsulation allows for stateful parsing, specifically enabling the tracking of resolved environment variables during configuration loading. This is useful for skill generation, where we can now identify and omit default values that correspond to environment variables, ensuring more precise generated skill.


cmd/root.go
```diff
@@ -370,14 +370,16 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 
 			if watchingFolder {
 				logger.DebugContext(ctx, "Reloading tools folder.")
-				reloadedToolsFile, err = internal.LoadAndMergeToolsFolder(ctx, folderToWatch)
+				parser := internal.ToolsFileParser{}
```
averikitsch: nit: instantiate parse outside if/else

helloeve: sg, updated!

PR #2646, opened 2026-03-05
feat(skill): update skill generation logic

helloeve:
This PR refactors and improves the skill generation logic to make it more configurable and also simplifies the generated skill.

- Shift from generating individual tool-specific YAML files to a more centralized configuration approach using global toolbox flags.
- Add support for --license-header to prepend license information to generated Node.js scripts.
- Refactored tool parameter documentation from a JSON schema format to a more readable Markdown table.
- Update tool invocation log level to avoid unnecessary outputs

### DIFF
--- a/docs/en/how-to/generate_skill.md
+++ b/docs/en/how-to/generate_skill.md
@@ -27,7 +27,8 @@ toolbox <tool-source> skills-generate \
   --name <skill-name> \
   --toolset <toolset-name> \
   --description <description> \
-  --output-dir <output-directory>
+  --output-dir <output-directory> \
+  --license-header <license-header>
 ```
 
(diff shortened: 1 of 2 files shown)


cmd/internal/skills/command.go
```diff
@@ -33,10 +34,11 @@ import (
...
 	*cobra.Command
-	name        string
-	description string
-	toolset     string
-	outputDir   string
+	name          string
+	description   string
+	toolset       string
+	outputDir     string
+	licenseHeader string
```
twishabansal, line 41: nit: Should we add this flag to the docs too?

helloeve: good call out, updated the doc to reflect changes introduced in this PR

PR #2654, opened 2026-03-06
fix(http)!: sanitize non-2xx error output

### LINKED ISSUE #2617: security(http): non-2xx upstream response bodies are reflected in tool error output
### What I ran into

While using the HTTP source/tool for a third-party API, I noticed full upstream error bodies were being returned back through toolbox responses. In my case this included detailed provider diagnostics that I would not want exposed to model output or downstream clients.

From code reading, this appears to come from:

- `internal/sources/http/http.go`
  - on non-2xx, returns:
    - `fmt.Errorf("unexpected status code: %d, response body: %s", ...)`
- error then flows through generic handling and often ends up returned as an agent-visible error string.

### Why this matters

If upstream error payloads contain sensitive details (internal IDs, stack traces, request metadata, token hints, etc.), toolbox currently reflects that content back to callers.

This creates a data-leak path from upstream services into agent/client-visible outputs.

### Expected

For non-2xx upstream responses, toolbox should return a sanitized error message to callers and avoid reflecting raw response body content.

### Suggested fix

- Do not include raw upstream response body in returned errors.
- Return a normalized/sanitized message (status code + brief reason).
- Optionally log a redacted/truncated body only at debug level.
- Add tests verifying non-2xx responses do not leak body content in client-visible errors.

I can open a PR with a minimal patch if maintainers agree with this direction.

Deeven-Seru:
## Summary                                                                   
   - sanitize non-2xx response errors by default (no upstream body)             
   - add returnFullError opt-in for raw body in error                           
   - log truncated body at debug level when sanitized                           
   - add regression tests + docs                                                
                                                                                
   ## Breaking change                                                           
   - non-2xx errors now return sanitized messages unless returnFullError is enabled                                                                        
                                                                                
   ## Testing                                                                   
   - go test ./internal/sources/http                                            
                                                                                
   Fixes #2617

### DIFF
--- a/cmd/internal/invoke/command.go
+++ b/cmd/internal/invoke/command.go
@@ -18,10 +18,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/googleapis/genai-toolbox/cmd/internal"
 	"github.com/googleapis/genai-toolbox/internal/server"
 	"github.com/googleapis/genai-toolbox/internal/server/resources"
+	"github.com/googleapis/genai-toolbox/internal/util"
 	"github.com/googleapis/genai-toolbox/internal/util/parameters"
 	"github.com/spf13/cobra"
 )
@@ -87,7 +89,7 @@ func runInvoke(cmd *cobra.Command, args []string, opts *internal.ToolboxOptions)
 
 	params := make(map[string]any)
 	if paramsInput != "" {
-		if err := json.Unmarshal([]byte(paramsInput), &params); err != nil {
+		if err := util.DecodeJSON(strings.NewReader(paramsInput), &params); err != nil {
 			errMsg := fmt.Errorf("params must be a valid JSON string: %w", err)
 			opts.Logger.ErrorContext(ctx, errMsg.Error())
 			return errMsg
--- a/cmd/internal/invoke/command_test.go
+++ b/cmd/internal/invoke/command_test.go
@@ -68,6 +68,15 @@ tools:
       - name: message
         type: string
         description: message to echo
+  int-tool:
+    kind: sqlite-sql
+    source: my-sqlite
+    description: "int tool"
+    statement: "SELECT ? as val"
+    parameters:
+      - name: value
+        type: integer
+        description: int value
 `
 
 	toolsFilePath := filepath.Join(tmpDir, "tools.yaml")
@@ -91,6 +100,11 @@ tools:
 			desc: "success - tool call with parameters",
 			args: []string{"invoke", "echo-tool", `{"message": "world"}`, "--config", toolsFilePath},
 			want: `"msg": "world"`,
+		},
+		{
+			desc: "success - tool call with integer parameters",
+			args: []string{"invoke", "int-tool", `{"value": 42}`, "--tools-file", toolsFilePath},
+			want: `"val": 42`,
 		},
 		{
 			desc:    "error - tool not found",


docs/en/resources/sources/http.md
```diff
@@ -47,8 +48,9 @@ instead of hardcoding your secrets into the configuration file.
...
-| headers                | map[string]string |    false     | Default headers to include in the HTTP requests.                                                                                   |
-| queryParams            | map[string]string |    false     | Default query parameters to include in the HTTP requests.                                                                          |
-| disableSslVerification |       bool        |    false     | Disable SSL certificate verification. This should only be used for local development. Defaults to `false`.                         |
+| headers                     | map[string]string |    false     | Default headers to include in the HTTP requests.                                                                                   |
+| queryParams                 | map[string]string |    false     | Default query parameters to include in the HTTP requests.                                                                          |
+| includeResponseBodyInErrors |       bool        |    false     | Include raw upstream response bodies in error messages for non-2xx responses. Defaults to `false`.                                  |
```
kurtisvg: nit: This is a bit wordy -- maybe a more concise option? e.g. `returnFullError` or something? open to suggestoins

Deeven-Seru: Good point — I agree the current name is a bit verbose.
Would something like `includeErrorBody` or `returnFullError` make more sensse here? I'm happy to update the config name and docs if there's a preferred convention used elsewhere in the repo.

duwenxin99: returnFullError sounds good. Thanks!

Deeven-Seru: done ..👍🏻

Deeven-Seru:
@gemini-code-assist just making sure includeResponseBodyInErrors is replaced with returnFullError

internal/sources/http/http.go
```diff
@@ -49,13 +50,15 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
...
-	QueryParams            map[string]string `yaml:"queryParams"`
-	DisableSslVerification bool              `yaml:"disableSslVerification"`
+	Name                        string            `yaml:"name" validate:"required"`
+	Type                        string            `yaml:"type" validate:"required"`
+	BaseURL                     string            `yaml:"baseUrl"`
+	Timeout                     string            `yaml:"timeout"`
+	DefaultHeaders              map[string]string `yaml:"headers"`
+	QueryParams                 map[string]string `yaml:"queryParams"`
+	ReturnFullError             bool              `yaml:"returnFullError"`
+	IncludeResponseBodyInErrors bool              `yaml:"includeResponseBodyInErrors"`
```
Yuan325: What is the difference between this 2 new fields? Do we need both? If yes, can we please add this to the docs as well?

internal/sources/http/http.go
```diff
@@ -160,7 +163,24 @@ func (s *Source) RunRequest(req *http.Request) (any, error) {
 		return nil, err
 	}
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return nil, fmt.Errorf("unexpected status code: %d, response body: %s", resp.StatusCode, string(body))
+		returnFullError := s.ReturnFullError || s.IncludeResponseBodyInErrors
+		if returnFullError {
+			return nil, fmt.Errorf("unexpected status code: %d, response body: %s", resp.StatusCode, string(body))
+		}
+
+		if logger, err := util.LoggerFromContext(req.Context()); err == nil {
```
Yuan325: There might be no logger in the `req.Context()`. Instead, we can send the ctx from the tool. e.g. `func (s *Source) RunRequest(ctx context.Context, req *http.Request) (any, error)`


If it throws an error from retrieving logger, we can return an error. All the invocation context should have a logger.

internal/sources/http/http.go
Deeven-Seru: returnFullError is the legacy flag;                           
 includeResponseBodyInErrors is the new clearer name. We keep both for backward compatibility and treat them as equivalent (returnFullError || includeResponseBodyInErrors). Docs updated to reflect both, with            returnFullError marked legacy

internal/sources/http/http.go
Deeven-Seru: Agreed. I updated RunRequest to accept the invocation ctx and now fetches the logger from that (returning an error if missing). The tool passes its ctx through, so logger retrieval is guaranteed

Yuan325:
@Deeven-Seru When you say legacy flag do you meant legacy flag from the http library? Since this is a newly introduced flag in Toolbox there are no `legacy` fields, and I'd  think that we can just support one of the other. Correct me if I'm wrong, but it seems like based on previous discussion, we've decided to use the `returnFullError` field? It'll be great if you can remove the `IncludeResponseBodyInErrors` flag.

Deeven-Seru:
@Yuan325  Thanks for the clarification — you’re right. Since this flag is new to Toolbox (not inherited from the HTTP lib), I missed that point there’s no legacy field to preserve. I removed includeResponseBodyInErrors and kept only returnFullError, and updated the docs accordingly :)

internal/sources/http/http.go
```diff
@@ -49,12 +50,13 @@ func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (sources
 }
 
 type Config struct {
-	Name                   string            `yaml:"name" validate:"required"`
-	Type                   string            `yaml:"type" validate:"required"`
-	BaseURL                string            `yaml:"baseUrl"`
-	Timeout                string            `yaml:"timeout"`
-	DefaultHeaders         map[string]string `yaml:"headers"`
-	QueryParams            map[string]string `yaml:"queryParams"`
+	Name                        string            `yaml:"name" validate:"required"`
```
Yuan325: please run lint to fix the indentations here :)

internal/sources/http/http_request_test.go
Yuan325, line 1: Please move the tests here into `http_test.go`.

Deeven-Seru:
> Thank you @Deeven-Seru ! :)

happy to contribute @Yuan325 😀

Deeven-Seru:
@Yuan325 lint issue resolved

Yuan325:
Hi @Deeven-Seru can you please also update the http integration test? Since the output is altered, the integration is currently failing~

Deeven-Seru:
hi @Yuan325  I fixed the failing HTTP integration test by updating the expected error string for the “wrong params” case in tests/http/http_integration_test.go to match the new sanitized error format (unexpected status code: 400 (Bad Request) instead of including the upstream body.

Yuan325:
@Deeven-Seru Thank you again ! :) This will be included in our next release.

Deeven-Seru:
> @Deeven-Seru Thank you again ! :) This will be included in our next release.

thanks for reviewing the pr @Yuan325 I've learned a lot ...! 😄

PR #2655, opened 2026-03-06
security(http): prevent base URL host override

### LINKED ISSUE #2616: security(http): templated path can override base host via URL userinfo parsing
### What I ran into

I hit this while using the HTTP tool in a real workflow where path parameters come from model output / user input. I expected requests to stay pinned to my configured `baseUrl`, but I observed a request get routed to an unexpected host.

After digging, the root cause looks like URL construction in `internal/tools/http/http.go`:

- `getURL()` builds URL with string concatenation: `url.Parse(baseURL + templatedPath.String())`
- If the templated path starts with a crafted value (for example `@evil.com/v1`), the parser treats the `baseUrl` host as URL userinfo and resolves host to `evil.com`.

### Repro (minimal)

Configured base URL:

```yaml
baseUrl: "https://api.good.com"
```

Templated path parameter value:

```text
@evil.com/v1
```

Resulting parsed URL behaves as:

```text
https://api.good.com@evil.com/v1
host = evil.com
user = api.good.com
```

### Why this matters

In practice this can:

- route the request to an attacker-controlled host,
- and potentially forward configured/default headers (including auth headers) to that host.

### Expected

`baseUrl` host should remain authoritative. Path templating should not be able to change destination host or inject URL userinfo.

### Suggested fix

- Parse `baseUrl` once.
- Resolve only validated relative paths (e.g. via URL reference resolution).
- Reject path values that can alter authority/userinfo (e.g. absolute URLs, `@` authority tricks, host/scheme changes).
- Add tests that assert final host always equals configured base host.

Happy to send a PR with tests if helpful.

Deeven-Seru:
## Summary
- build request URLs by resolving validated relative paths against the base URL
- reject path templates that attempt to set scheme/host/userinfo
- add regression tests for host-override attempts

## Testing
- go test ./internal/tools/http

Fixes #2616


Deeven-Seru:
Hi! @duwenxin99 I intentionally disabled source connectivity checks because  startup resilience in environments where some sources are transiently unavailable (e.g., network blips, staged rollouts, dev/preview sources). Today the server hard‑fails on any single source ping, which blockks serving healthy sources.                                                             --allow-partial-sources and per‑source checkAtStartup: false lets operators start the server, keep the healthy sources online, and let the unavailable ones recover later. It’s opt‑in and preserves the current strict behavior by default.

Deeven-Seru:
🎉

PR #2662, opened 2026-03-07
feat(server): allow partial source startup checks

### LINKED ISSUE #2341: Skip connectivity checks on start up
### Prerequisites

### Suggested Solution(s)

### Alternatives Considered

### Additional Details

Deeven-Seru:
## Summary
- add `checkAtStartup` per-source config to skip startup connectivity checks
- add `--allow-partial-sources` flag to skip all source checks at startup
- make source initialization respect startup-check flag via context
- update docs + add regression test

## Testing
- go test ./cmd/internal -run TestParseToolFileCheckAtStartup
- go test ./internal/server
- go test ./internal/sources/...

Fixes #2341

### DIFF
--- a/docs/en/reference/cli.md
+++ b/docs/en/reference/cli.md
@@ -29,6 +29,7 @@ description: >
 |              | `--allowed-origins`        | Specifies a list of origins permitted to access this server for CORs access.                                                                                                     | `*`         |
 |              | `--allowed-hosts`          | Specifies a list of hosts permitted to access this server to prevent DNS rebinding attacks.                                                                                      | `*`         |
 |              | `--user-agent-metadata`    | Appends additional metadata to the User-Agent.                                                                                                                                   |             |
+|              | `--allow-partial-sources`  | Skip startup connectivity checks for sources to allow partial availability.                                                                                                      |             |
 |              | `--poll-interval`          | Specifies the polling frequency (seconds) for configuration file updates.                                                                                                        | `0`         |
 | `-v`         | `--version`                | version for toolbox                                                                                                                                                              |             |
 
--- a/cmd/internal/config_test.go
+++ b/cmd/internal/config_test.go
@@ -2297,3 +2297,83 @@ tools:
 		})
 	}
 }
+
+func TestParseToolFileCheckAtStartup(t *testing.T) {
+	ctx, err := testutils.ContextWithNewLogger()
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+
+	testCases := []struct {
+		name string
+		in   []byte
+		want bool
+	}{
+		{
+			name: "explicit false",
+			in: testutils.FormatYaml(`
+		kind: sources
+		name: http-source
+		type: http
+		baseUrl: https://example.com
+		checkAtStartup: false
+		`),
+			want: false,
+		},
+		{
+			name: "explicit true",
+			in: testutils.FormatYaml(`
+		kind: sources
+		name: http-source
+		type: http
+		baseUrl: https://example.com
+		checkAtStartup: true
+		`),
+			want: true,
+		},
+		{
+			name: "default true",
+			in: testutils.FormatYaml(`
+		kind: sources
+		name: http-source-default
+		type: http
+		baseUrl: https://example.com
+		`),
+			want: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			toolsFile, err := parseToolsFile(ctx, tc.in)
+			if err != nil {
+				t.Fatalf("failed to parse input: %v", err)
+			}
+			cfg := toolsFile.Sources["http-source"]
+			if cfg == nil {
+				cfg = toolsFile.Sources["http-source-default"]
+			}
+			if cfg == nil {
+				t.Fatalf("expected source to be parsed")
+			}
+			if sources.CheckAtStartup(cfg) != tc.want {
+				t.Fatalf("expected checkAtStartup=%v", tc.want)
+			}
+		})
+	}
+
+	invalidInput := testutils.FormatYaml(`
+	kind: sources
+	name: http-source-invalid
+	type: http
+	baseUrl: https://example.com
+	checkAtStartup: "nope"
+	`)
+	_, err = parseToolsFile(ctx, invalidInput)
+	if err == nil {
+		t.Fatalf("expected error for invalid checkAtStartup type")
+	}
+	if !strings.Contains(err.Error(), "checkAtStartup must be a boolean") {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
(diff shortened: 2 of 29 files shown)


Deeven-Seru:
Update: rebased this PR onto main to keep it focused on startup checks only. The HTTP tool host-override changes live in #2655, so the path-traversal comment is now out of scope here. Also updated valkey init to return an error instead of log.Fatalf, so allow-partial-sources can skip it cleanly.

Deeven-Seru:
Updates pushed:
- Expanded TestParseToolFileCheckAtStartup to cover explicit true + invalid type cases.
- Kept allowPartialSources as a parameter for now to avoid a broader Server struct refactor; happy to follow up if you’d prefer the config to live on server.Server.

averikitsch:
Thank you for the PR. We will review it shortly.

twishabansal:
@Deeven-Seru please fix the failing checks

docs/en/reference/cli.md
```diff
@@ -29,6 +29,7 @@ description: >
 |              | `--allowed-origins`        | Specifies a list of origins permitted to access this server for CORs access.                                                                                                     | `*`         |
 |              | `--allowed-hosts`          | Specifies a list of hosts permitted to access this server to prevent DNS rebinding attacks.                                                                                      | `*`         |
 |              | `--user-agent-metadata`    | Appends additional metadata to the User-Agent.                                                                                                                                   |             |
+|              | `--allow-partial-sources`  | Skip startup connectivity checks for sources to allow partial availability.                                                                                                      |             |
```
twishabansal, line 32: Can we document this behaviour better?  

- Do we need both the flag and the "checkAtStartup" field to skip connection checks? What are the behaviours for different combinations and does one of them override the another?
- Can we just use one instead?

twishabansal, line 32: "allow-partial-sources" might be a bit confusing? Maybe name it "skip-source-startup-checks" ? Feel free to add any other suggestions as well

twishabansal:
@Deeven-Seru it seems like some sources like bigquery might be missing from this list. I see that there might be some connectivity challenges there. Do you mind sharing the design and any potential issues so that we can document all this behaviour?

docs/en/reference/cli.md
Deeven-Seru: how about "bypass-startup-checks"

Deeven-Seru: fine with using one 👍

Deeven-Seru:
@twishabansal BigQuery is an exception in the current architecture: Initialize() does client setup/config validation, but there isn’t a single startup Ping path like SQL/Redis, so this PR’s CheckConnectivity(...) pattern doesn’t apply 1:1. As a workaround, I’ll try too document BigQuery’s current startup behavior explicitly in this PR (including allowedDatasets metadata checks), and then follow up with a BigQuery-specific startup-check hook so global/per-source skp semantics are consistent across sources. WDUT?

twishabansal:
> @twishabansal BigQuery is an exception in the current architecture: Initialize() does client setup/config validation, but there isn’t a single startup Ping path like SQL/Redis, so this PR’s CheckConnectivity(...) pattern doesn’t apply 1:1. As a workaround, I’ll try too document BigQuery’s current startup behavior explicitly in this PR (including allowedDatasets metadata checks), and then follow up with a BigQuery-specific startup-check hook so global/per-source skp semantics are consistent across sources. WDUT?

@Deeven-Seru it seems like we might need to ensure that a bunch of edits are in place before going ahead with this implementation.

As a platform team, we need to ensure that all our sources and tools work well with our features. The current state seems to a bit confusing since this PR does not update all sources. Apart from bigquery, some other sources seem to be missing. For a complete list of all resources, see https://mcp-toolbox.dev/integrations/.

I would love for us to investigate the codebase and point out any other such patterns that are not the same as our current implementations and find the best path forward. I'm currently investigating and will add more details to the issue itself.

It would be nice to have a design finalised before we jump in with the PR changes. What do you think?

Deeven-Seru:
> > @twishabansal BigQuery is an exception in the current architecture: Initialize() does client setup/config validation, but there isn’t a single startup Ping path like SQL/Redis, so this PR’s CheckConnectivity(...) pattern doesn’t apply 1:1. As a workaround, I’ll try too document BigQuery’s current startup behavior explicitly in this PR (including allowedDatasets metadata checks), and then follow up with a BigQuery-specific startup-check hook so global/per-source skp semantics are consistent across sources. WDUT?
> 
> @Deeven-Seru it seems like we might need to ensure that a bunch of edits are in place before going ahead with this implementation.
> 
> As a platform team, we need to ensure that all our sources and tools work well with our features. The current state seems to a bit confusing since this PR does not update all sources. Apart from bigquery, some other sources seem to be missing. For a complete list of all resources, see https://mcp-toolbox.dev/integrations/.
> 
> I would love for us to investigate the codebase and point out any other such patterns that are not the same as our current implementations and find the best path forward. I'm currently investigating and will add more details to the issue itself.
> 
> It would be nice to have a design finalised before we jump in with the PR changes. What do you think?

Hi @twishabansal I completely agree with the platform perspective here. The current approach of wrapping individual database pings inside `Initialize()` creates technical debt, especially for REST/RPC sources (BigQuery, Looker, Firestore) that don't have a traditional connection pool or `Ping` method.

To standardize this across all 40+ integrations, we need a unified interface contract. Here is the design I propose:

1. Expand the core `sources.Source` (or create a new `sources.Pingable`) interface with a `VerifyConnection(ctx context.Context) error` method.
2. Relational/Redis sources implement this by calling their native `.Ping()`. 
3. REST/HTTP sources (BigQuery, Looker) implement this by performing a lightweight metadata fetch (e.g., fetching datasets/dashboards) or return `nil` if a proactive check isn't feasible.
4. The server initialization loop handles the `--bypass-startup-checks` logic globally, iterating through the sources and calling `VerifyConnection()` only if the flag is disabled. 

This extracts the connectivity logic entirely out of the individual `Initialize()` functions and creates a single, consistent platform standard.

Deeven-Seru:
Regarding the architecture, I took a look at sources to find the inconsistency you mentioned. I see that the implementations are currently fractured in 2 ways 


The relational/standard NoSQL databases are actively sending network packets during `Initialize()` to verify the connection pool.

The REST, HTTP, or GCP RPC sources are allocating the client struct or HTTP client during `Initialize()` but dosen't verify network connectivity or credentials unless until a user actually runs a query. 

Because these two patterns are completely different, trying to inject `--bypass-startup-checks` directly into `Initialize()` like this PR does it  works for relational/standard NoSQL database, but completely mess up with REST, HTTP, or GCP RPC sources . 

This audit perfectly validates the interface design I proposed in the above. By introducing `type Pingable interface { VerifyConnection(ctx context.Context) error }`:
- lets remove the 'Active Ping' out of relational/standard NoSQL databases `Initialize()` functions entirely.
- implement `VerifyConnection()` for relational/standard NoSQL databases using their native `.Ping()`.
- We implement `VerifyConnection()` for REST, HTTP, or GCP RPC sources using a lightweight API fetch , that gives the startup validation they currently lack.
- The server loop handles the flag logic globally.

This will extract the connectivity logic entirely out of the individual `Initialize()` functions and unifies all 42 sources under one platform standard. 

I'm happy to formalize this into a design spec if we are aligned on the direction!

Deeven-Seru:
@gemini-code-assist is this feasible ?

Deeven-Seru:
@twishabansal I am looking forward  to hear your thoughts on standardization of data sources. 😊

twishabansal:
@Deeven-Seru apologies for the delayed reply. Would you mind moving these points to the issue instead? It keeps everything documented and easy to access in the future.

Deeven-Seru:
@twishabansal can I comment these points in issue 2341 itself ? or d u suggest opening a new one

twishabansal:
> @twishabansal can I comment these points in issue 2341 itself ? or d u suggest opening a new one

Please add to the original issue itself: 2341

twishabansal:
Moving to https://github.com/googleapis/mcp-toolbox/pull/2959/changes

PR #2679, opened 2026-03-09
feat: add migrate subcommand

Yuan325:
Add a `migrate` subcommand to help migrate all configuration files to flat format.

This PR also updates the `LoadConfig()` function in `options.go` by extracting the custom config path logic into a new function. This removes duplication and makes the logic reusable for the migrate subcommand.

### DIFF
--- a/.ci/integration.cloudbuild.yaml
+++ b/.ci/integration.cloudbuild.yaml
@@ -930,7 +930,7 @@ steps:
       - "GOPATH=/gopath"
       - "SERVICE_ACCOUNT_EMAIL=$SERVICE_ACCOUNT_EMAIL"
       - "ORACLE_SERVER_NAME=$_ORACLE_SERVER_NAME"
-    secretEnv: ["CLIENT_ID", "ORACLE_USER", "ORACLE_PASS", "ORACLE_HOST"]
+    secretEnv: ["CLIENT_ID", "ORACLE_USERNAME", "ORACLE_PASSWORD", "ORACLE_HOST"]
     volumes:
       - name: "go"
         path: "/gopath"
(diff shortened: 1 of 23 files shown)


cmd/internal/migrate/command.go
```diff
@@ -0,0 +1,152 @@
...
+	*cobra.Command
+	dryRun bool
+}
+
+func NewCommand(opts *internal.ToolboxOptions) *cobra.Command {
+	cmd := &migrateCmd{}
+	cmd.Command = &cobra.Command{
+		Use:   "migrate",
+		Short: "Migrate local configuration files",
+		Long:  "Migrate all configuration files provided to the v2 format, updating deprecated fields and ensuring compatibility.",
```
averikitsch: Are we using "v2" elsewhere? I think we should have a "more official" name. Maybe Flat Format or Manifest Format?

cmd/internal/migrate/command.go
```diff
@@ -0,0 +1,152 @@
...
+type migrateCmd struct {
+	*cobra.Command
+	dryRun bool
+}
+
+func NewCommand(opts *internal.ToolboxOptions) *cobra.Command {
+	cmd := &migrateCmd{}
+	cmd.Command = &cobra.Command{
+		Use:   "migrate",
+		Short: "Migrate local configuration files",
```
averikitsch: "Migrate .. files to X format."

cmd/internal/migrate/command.go
```diff
@@ -0,0 +1,152 @@
...
+	return cmd.Command
+}
+
+func runMigrate(cmd *migrateCmd, opts *internal.ToolboxOptions) error {
+	ctx, cancel := context.WithCancel(cmd.Context())
+	defer cancel()
+
+	// watch for sigterm / sigint signals
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGTERM, syscall.SIGINT)
```
averikitsch: Why do we need sigterm? We shouldn't need to spin up the server or any clients

cmd/internal/migrate/command.go
Yuan325: I like flat format. Will update to that!

PR #2683, opened 2026-03-09
feat: implement Secure Parameters via _meta

aseering:
## Description

Implement support for the Secure Parameters concept.  Provide isolation from regular Parameters such that a traditionally-configured AI Agent would not be able to see the Secure Parameters schema nor set Secure Parameter values.

## PR Checklist

- [X] Make sure you reviewed

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,123 @@
+# Secure Parameters Design Document
+
+## 1. Overview
+
+Secure parameters provide a mechanism for tools to accept sensitive data (like tokens or secrets) that should not be logged, exposed in standard parameter lists, or mixed with normal operational arguments.
+
+This document outlines the design decisions made to integrate secure parameters into the MCP Toolbox while maintaining strict backward compatibility for all existing tools.
+
+## 2. Core Requirements
+
+The design was guided by three primary requirements:
+
+1.  **YAML API Consistency:** Secure parameters must be defined as a property (`secure: true`) on standard parameters within the `parameters` list, rather than requiring a separate list or schema structure.
+2.  **Namespace Collisions:** A regular parameter and a secure parameter must be allowed to share the same name (e.g., a standard `id` for query routing and a secure `id` used within a database secure context).
+3.  **Zero Interface Changes (Backward Compatibility):** Existing tools must not require *any* code modifications. They must continue to receive standard parameters exactly as before, entirely shielded from the existence of secure parameters.
+
+## 3. Design Decisions
+
+### 3.1 Transport via Protocol `_meta`
+To separate secure parameters from standard operational parameters over the wire, we utilize the Model Context Protocol `_meta` extension fields.
+*   **Tool List (`tools/list`):** The schema for secure parameters is published under `_meta.toolbox/stateSchema`, rather than mixing it into the standard `inputSchema`.
+*   **Tool Execution (`tools/call`):** Secure parameters are passed under `_meta.toolbox/state` rather than within the standard `arguments` map.
+
+### 3.2 Parsing and Data Structures (`ParamValue`)
+The most significant challenge was passing these split namespaces into the `Tool.Invoke()` interface without altering its signature (`func Invoke(..., params ParamValues, ...)`).
+
+*   **Annotation:** We extended the `ParamValue` struct with an `IsSecure bool` annotation.
+*   **Encapsulation:** The MCP server parses both standard `arguments` and `_meta.toolbox/state` independently, flags them appropriately, and concatenates them into a single `ParamValues` slice. Because they are stored in a slice, name collisions (`id`: standard, `id`: secure) are natively supported.
+*   **Shielding:** We modified the standard `params.AsMap()` method to strictly **filter out** any parameter marked `IsSecure`.
+    *   *Result:* Because all existing tools call `AsMap()` to retrieve their bindings, they remain entirely unaffected by secure parameters.
+*   **Opt-In Access:** We introduced a new `params.AsSecureMap()` method. Tools that explicitly require secure parameters (like Spanner) can call this method to retrieve the separate namespace.
+
+## 4. API Examples
+
+### 4.1 YAML Configuration API
+Tools are configured via YAML. To mark a parameter as secure, simply add the `secure: true` property to its definition. 
+
+```yaml
+tools:
+  my-secure-tool:
+    kind: spanner-sql
+    source: my-instance
+    description: "Tool with colliding namespaces"
+    statement: "SELECT * FROM users WHERE id = @id AND token = SECURE_CONTEXT("id")"
+    parameters:
+      # This is the standard parameter used for @id
+      - name: id
+        type: integer
+        description: "The regular ID"
+      # This is the secure parameter used for SECURE_CONTEXT
+      - name: id
+        type: string
+        secure: true
+        description: "The secure ID"
+```
+
+### 4.2 JSONRPC API
+The MCP protocol relies on JSONRPC.
+
+**1. Discovery (`tools/list`)**
+The Toolbox exposes the secure parameter requirement in the `_meta` object:
+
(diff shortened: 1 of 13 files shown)


averikitsch:
Thanks @aseering ! I am headed OOO, but I will ensure we get a review on the PR next week. Thank you!

aseering:
Thanks!  Gentle _nudge_, wondering if you've had a chance to take a look and if you have any thoughts?

Yuan325:
Hi @aseering thanks for the proposal! To keep the PR clean, could you please open a dedicated issue for the design, move the design documentation there, and then remove `design.md` from this PR?

I also had a couple of questions regarding the implementation:

1. What is the specific rationale behind proposing `_meta.toolbox/stateSchema`? Does the agent currently have the logic to parse and utilize this value?
1. Have you considered resolving tools calls's `_meta.toolbox/state` values into a standard `var data map[string]any` parameter within `toolsCallHandler()` (in `internal/server/mcp/vxx/method.go`)? Since the distinction between "data" and "secure data" primarily affects visibility during tools/list—rather than server-side execution—merging them would allow us to leverage our existing `ParseParams()` logic without maintaining separate handling for data vs. secureData.

aseering:
Hi Yuan, thanks for the review!  And gentle _ping_ @averikitsch about taking a look at this (we've been discussing it offline).

Regarding the design doc -- oops!, sure, will do.  (I actually didn't intend to include it and missed that it snuck in...  I'll also review it to make sure it's still accurate.)

To provide context:  The immediate motivating use case (not the only use case) for this feature is support for Parameterized Secure Views in database engines.  The use case for PSVs is parameters that implement policy restrictions on data access.

The assumption for this use case is that the MCP client is linked into an application layer that is trusted; but the tool calls are constructed by a LLM that may interact with external users, so their content is untrusted, potentially not even trusted to know the list of secure parameters that the database system requires (which precludes exposing them as ordinary parameters and relying on the application layer to intercept and modify the tool call before we see it).

Another assumption is that the set of secure parameters that are applicable to any given SQL statement, is managed by the database engine.  Parameter validation is not controlled by the MCP server and not controlled by the tool -- the information needed to correctly validate whether the provided set of parameters is correct is only available inside the underlying database engine.  So, the MCP layer would act simply as a pass-through for these parameters.  (Of course, for other use cases, tools could perform validation if they want to and as applicable.)

Regarding your point:

> Since the distinction between "data" and "secure data" primarily affects visibility during tools/list

Given the above context, it turns out that's half-true :-)  It primarily affects (a) visibility during tools/list, and (b) routing of parameters into the database engine.  Since, if we get all of the parameters through a single API, we have to extract out the ones that the caller passed in via this other API so that we can pass them into the database engine for validation and consumption.

How would that latter bit work with this proposal?

Yuan325:
Got it, since it does differs when sending the param to the API, it make sense to separate it. Thank you for the clarification!

internal/util/parameters/parameters.go
```diff
@@ -66,11 +84,24 @@ func (p ParamValues) AsSlice() []any {
...
-		params[p.Name] = p.Value
+		if !p.IsSecure {
+			params[p.Name] = p.Value
+		}
+	}
+	return params
+}
+
+// AsSecureMap returns a map of secure ParamValue's names to values.
+func (p ParamValues) AsSecureMap() map[string]interface{} {
```
kurtisvg, line 99: I think this should probably be named something like "SecureParams" or "SecureParamsMap" -- it needs to be clear the "secure" is modifying the params, not the map

DESIGN.md
```diff
@@ -0,0 +1,123 @@
+# Secure Parameters Design Document
```
kurtisvg, line 1: nit: this should be removed from the PR

internal/util/parameters/parameters.go
```diff
@@ -513,6 +557,28 @@ func (ps Parameters) McpManifest() (McpToolsSchema, map[string][]string) {
 	}, authParam
 }
 
+func (ps Parameters) McpStateSchema() McpToolsSchema {
```
kurtisvg, line 560: McpManifest is already returning both the McpToolsSchema and the authParams, so it seems like maybe secure params should get folded into it as well?

internal/util/parameters/parameters.go
```diff
@@ -486,11 +532,9 @@ func (ps Parameters) McpManifest() (McpToolsSchema, map[string][]string) {
 	authParam := make(map[string][]string)
 
 	for _, p := range ps {
-		// If the parameter is sourced from another param, skip it in the MCP manifest
-		if p.GetValueFromParam() != "" {
+		if p.IsSecureParameter() || p.GetValueFromParam() != "" {
```
kurtisvg, line 535: We lost a comment here -- seems like we should update it instead of remove it?

internal/util/parameters/parameters.go
```diff
@@ -129,21 +160,35 @@ func CheckParamRequired(required bool, defaultV any) bool {
 
 // ParseParams is a helper function for parsing Parameters from an arbitraryJSON object.
 func ParseParams(ps Parameters, data map[string]any, claimsMap map[string]map[string]any) (ParamValues, error) {
-	params := make([]ParamValue, 0, len(ps))
+	return ParseParamsWithSecure(ps, data, nil, claimsMap)
+}
+
+// ParseParamsWithSecure is a helper function for parsing regular and secure Parameters separately.
+func ParseParamsWithSecure(ps Parameters, data map[string]any, secureData map[string]any, claimsMap map[string]map[string]any) (ParamValues, error) {
```
kurtisvg, line 167: I wonder if it would be better to represent Secure Parameters entirely separate -- it's not clear to me that this won't monkey up functions like AsMapWithDollarPrefix (which haven't been updated) and may process continue to process secure params. 

So maybe the right thing to do would be so sort Params into Normal, Secure, or Authenticated when they are parsed to avoid accidentally mixing them. 

@duwenxin99 fyi in case you have thoughts on that implementation affecting authenticated params

internal/server/mcp/v20241105/method.go
```diff
@@ -148,13 +148,19 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
...
 	var data map[string]any
 	if err = util.DecodeJSON(bytes.NewBuffer(aMarshal), &data); err != nil {
 		err = fmt.Errorf("unable to decode tools argument: %w", err)
 		return jsonrpc.NewError(id, jsonrpc.INTERNAL_ERROR, err.Error(), nil), err
 	}
 
+	var secureData map[string]any
+	if req.Params.Meta != nil {
+		if state, ok := req.Params.Meta["toolbox/state"].(map[string]any); ok {
+			secureData = state
```
kurtisvg, line 160: super-nit: for such a small scope, I think it's more idiomatic to Go to use a shorter var name
```suggestion
		if s, ok := req.Params.Meta["toolbox/state"].(map[string]any); ok {
			secureData = s
```

internal/util/parameters/common.go
```diff
@@ -109,15 +109,24 @@ func PopulateTemplateWithFunc(templateName, templateString string, data map[stri
 	return result.String(), nil
 }
 
-// CheckDuplicateParameters verify there are no duplicate parameter names
+// CheckDuplicateParameters verify there are no duplicate parameter names.
+// It allows a regular parameter and a secure parameter to have the same name.
```
kurtisvg, line 113: I'm not sure that allowing name collisions is desirable. For one, the secure parameters aren't passed to the LLM, so I'm unclear on what value it adds. 

Two, from the SDK perspective when a user tries to bind a value, the SDK won't know which one to bind (unless we separate out bind_param from bind_secure_param)

kurtisvg (changes requested):
It looks like both the integration checks and the lint are failing -- it would be good to address those.

PR #2697, opened 2026-03-10
feat(skill): attach user agent metadata for generated skill

helloeve:
This PR updates the agent skill script generation logic to extract and pass the user-agent-metadata during CLI invocation.

### DIFF
--- a/docs/en/how-to/deploy_docker.md
+++ b/docs/en/how-to/deploy_docker.md
@@ -34,7 +34,7 @@ services:
       - "5000:5000"
     volumes:
       - ./config:/config
-    command: [ "toolbox", "--tools-file", "/config/tools.yaml", "--address", "0.0.0.0"]
+    command: [ "--tools-file", "/config/tools.yaml", "--address", "0.0.0.0"]
     depends_on:
       db:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


cmd/internal/skills/generator.go
twishabansal, line 133: For now, we're not injecting the user agent metadata var for gemini cli. Do you think we should set it inside this if block?

helloeve: I may not fully understand this question... this changes introduced in this PR would inject user agent metadata for all invocations (independent from whether it is via gemini cli or not), as it appends --user-agent-metadata into toolboxArgs.

twishabansal: To clarify, we're currently not setting the env var: `env.USER_AGENT_METADATA` in the gemini cli env (Since env vars set in gemini-extension.json are not directly accessible to skills). So, if we get process.env.GEMINI_CLI == `1`, should we set the env var there itself? i.e.

```js
if (process.env.GEMINI_CLI === '1') {
       env.USER_AGENT_METADATA = 'skills-geminicli';
}
```

helloeve: Ah I see your point for tracking gemini-cli usage vs the remaining? Let me refactor this PR a bit

PR #2711, opened 2026-03-11
test: refactor integration test framework for full MCP validation

anubhav756:
## Overview
This PR refactors the underlying test methodology across MCP Toolbox to natively support integration testing using MCP JSON-RPC requests.

## Changes
* Introduced new native MCP test helpers (`RunMCPToolInvokeTest`, `RunMCPExecuteSqlToolInvokeTest`, `RunMCPToolInvokeSimpleTest`) in `tests/tool.go`.
* Added standard JSON-RPC request structures (`jsonrpc.JSONRPCRequest`) to simulate robust MCP payloads.
* Rerouted testing endpoints away from the `/api` HTTP implementation straight to the core `/mcp` pathway.

> [!NOTE]
> This PR is Part 1 of 5 to fully migrate MCP Toolbox onto MCP endpoints and remove the legacy `/api` REST endpoints.

### DIFF
--- a/tests/auth.go
+++ b/tests/auth.go
@@ -26,15 +26,6 @@ import (
 var ServiceAccountEmail = os.Getenv("SERVICE_ACCOUNT_EMAIL")
 var ClientId = os.Getenv("CLIENT_ID")
 
-func init() {
-	if ClientId == "" {
-		ClientId = "test-client-id"
-	}
-	if ServiceAccountEmail == "" {
-		ServiceAccountEmail = "test-service-account@example.com"
-	}
-}
-
 // GetGoogleIdToken retrieve and return the Google ID token
 func GetGoogleIdToken(audience string) (string, error) {
 	// For local testing - use gcloud command to print personal ID token


tests/http/http_integration_test.go
```diff
@@ -509,6 +365,11 @@ func getHTTPToolsConfig(sourceConfig map[string]any, toolType string) map[string
 	otherSourceConfig["headers"] = map[string]string{"X-Custom-Header": "unexpected", "Content-Type": "application/json"}
 	otherSourceConfig["queryParams"] = map[string]any{"id": 1, "name": "Sid"}
 
+	clientId := os.Getenv("CLIENT_ID")
+	if clientId == "" {
+		clientId = "test-client-id"
```
Yuan325: can we use a random client id here?

tests/auth.go
```diff
@@ -26,6 +26,15 @@ import (
 var ServiceAccountEmail = os.Getenv("SERVICE_ACCOUNT_EMAIL")
 var ClientId = os.Getenv("CLIENT_ID")
 
+func init() {
```
Yuan325: can we use these placeholder client id and service acc emails? woudlnt the integration test fail with these?

tests/tool.go
```diff
@@ -4816,3 +4816,971 @@ func RunStatementToolsTest(t *testing.T, tools map[string]string) {
...
+				err2 := json.Unmarshal([]byte(wantStr), &wantObj)
+				if err1 != nil || err2 != nil || !reflect.DeepEqual(gotObj, wantObj) {
+					t.Fatalf("unexpected value: got %q, want %q\nGOT HEX: %x\nWANT HEX: %x", got, wantStr, got, wantStr)
+				}
+			}
+		})
+	}
+}
+
+func RunMCPToolInvokeParametersTest(t *testing.T, name string, params []byte, simpleWant string) {
```
Yuan325, line 4926: this seems to be specific towards the http source? in that case let's move this into the `http_integration_test.go` and not export this function

tests/tool.go
```diff
@@ -4816,3 +4816,971 @@ func RunStatementToolsTest(t *testing.T, tools map[string]string) {
...
+
+			if !strings.Contains(got, wantStr) {
+				t.Fatalf("unexpected value: got %q, want %q", got, wantStr)
+			}
+		})
+	}
+}
+
+// RunMCPToolInvoke runs the tool invoke endpoint
+func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...MCPToolInvokeTestOption) {
```
Yuan325, line 5039: is this specific towards the http source?

tests/http/http_integration_test.go
anubhav756: Makes sense! Thanks for the suggestion :)

tests/auth.go
anubhav756: Fair enough. I think it is fine if we let the integration tests panic if these env vars are not supplied. Removed.

tests/tool.go
anubhav756: The Looker integration test rely on this in an upcoming PR (https://github.com/googleapis/genai-toolbox/pull/2713/changes). Keeping it here for now.

tests/tool.go
anubhav756: Because `RunMCPToolInvokeTest` relies heavily on `tests/common.go` mock tools and `tests/option.go` configs that are universal to testing setup for the `/mcp` endpoints across all DBs downstream (not just HTTP), I think it makes more architectural sense to keep it alongside those utilities in `tests/tool.go`. Let me know if you think otherwise! Thanks for bringing this up though :)

anubhav756:
> Why do we want to rename everything with a `MCP`? Since the entire Toolbox is being updated to MCP anyway, I'm wondering if keeping the existing names might keep things cleaner. I'm slightly concerned that adding 'MCP' to everything might end up being a bit redundant~

Agree that `MCP` will eventually be completely redundant! However I thought that during this transition phase through these multiple branches, it might be safer to temporarily have the prefix as a temporary scaffolding (eg. `RunToolInvokeTest` and `RunMCPToolInvokeTest` coexist in some branches). My plan is that once we reach the final state, where we drop the `/api` endpoints, I can drop the `MCP` prefixes and reclaim the clean names. WDYT?

Yuan325 (approved):
awesome thank you!!! LGTM :)

tests/tool.go
```diff
@@ -4816,3 +4816,971 @@ func RunStatementToolsTest(t *testing.T, tools map[string]string) {
...
+					return
+				}
+				t.Fatalf("response status code is not 200, got %d: %s", resp.StatusCode, string(respBody))
+			}
+
+			var body map[string]interface{}
+			err := json.Unmarshal(respBody, &body)
+			if err != nil {
+				t.Fatalf("error parsing response body: %v", err)
+			}
```
Yuan325, line 4871: Please refer to the comment in #2712. Might help to add a new file `mcp_types.go` that consists of the response structs. And we can just unmarshal this into the struct since all of them is of same formatting.

It's always better to use struct when possible than `map[string]any`. I think it will also help to reduce the checks for each fields here.

tests/tool.go
```diff
@@ -4816,3 +4816,971 @@ func RunStatementToolsTest(t *testing.T, tools map[string]string) {
...
+
+			resultMap, hasResult := body["result"].(map[string]interface{})
+			if !hasResult && !tc.isErr {
+				t.Fatalf("unable to find result in response body: %s", string(respBody))
+			}
+			contentList, hasContent := resultMap["content"].([]interface{})
+			if !hasContent {
+				t.Fatalf("unable to find result.content in response body: %s", string(respBody))
+			}
```
Yuan325, line 4895: we can just check for `if body.Result == nil || body.Result.Content == nil` then throw error.

tests/tool.go
```diff
@@ -4816,3 +4816,971 @@ func RunStatementToolsTest(t *testing.T, tools map[string]string) {
...
+				got = combined[0]
+			}
+
+			if !strings.Contains(got, wantStr) {
+				var gotObj, wantObj interface{}
+				err1 := json.Unmarshal([]byte(got), &gotObj)
+				err2 := json.Unmarshal([]byte(wantStr), &wantObj)
+				if err1 != nil || err2 != nil || !reflect.DeepEqual(gotObj, wantObj) {
+					t.Fatalf("unexpected value: got %q, want %q\nGOT HEX: %x\nWANT HEX: %x", got, wantStr, got, wantStr)
+				}
```
Yuan325, line 4920: why do we want to check the object again after checking the string? is checking string not sufficient?

tests/tool.go
```diff
@@ -4816,3 +4816,971 @@ func RunStatementToolsTest(t *testing.T, tools map[string]string) {
...
+				t.Fatalf("response status code is not 200, got %d: %s", resp.StatusCode, string(respBody))
+			}
+
+			var body map[string]interface{}
+			err := json.Unmarshal(respBody, &body)
+			if err != nil {
+				t.Fatalf("error parsing response body: %v", err)
+			}
+
```
Yuan325, line 4983: similar to above, the struct should be reuseable

tests/tool.go
anubhav756: We are moving to a new slim harness in the base branch in https://github.com/googleapis/genai-toolbox/pull/2828 which uses standard typed structs (e.g., tests.InvokeMCPTool from `mcp_execute.go`), rendering these manual map parsing helpers obsolete. Closing this to consolidate focus there.

Thanks for sharing your thoughts!

tests/tool.go
anubhav756: The new `InvokeMCPTool` (in https://github.com/googleapis/genai-toolbox/pull/2828/changes) handles this natively by unmarshaling into standard MCP response structs, so this manual check won't be necessary anymore! Closing this to consolidate focus there. Thanks!

tests/tool.go
anubhav756: The new slim harness in https://github.com/googleapis/genai-toolbox/pull/2828/changes handles this natively with typed structs, so we won't have manual string checks vs object checks in the new tests. Closing this to consolidate focus there. Thanks!

tests/tool.go
anubhav756: Completely agree! In the new slim harness (https://github.com/googleapis/genai-toolbox/pull/2828), we centralized this execution logic. Thanks anyways!

anubhav756:
Closing this PR in favor of the new slim test harness in https://github.com/googleapis/genai-toolbox/pull/2828/changes, which also incorporates all the changes suggested here.

PR #2712, opened 2026-03-11
test: migrate simple databases to MCP integration endpoints

anubhav756:
## Overview
This PR migrates integration tests for standard operational and analytical databases from using explicit REST API layouts to native MCP endpoints.

## Changes
* Transitioned tests for over 25+ database platforms (PostgreSQL, MySQL, AlloyDB, Spanner, Redis, etc.) to leverage the newly built MCP test framework functions.
* Resolved a bug inside SQL tool implementations that caused the MCP `tools/list` protocol schema to omit required template execution parameters by correctly returning mapped parameters via `GetParameters()`.
* Adjusted generic HTTP assertions to support deep evaluation of MCP JSON-RPC response structures and embedded error codes (like `-32603`).

> [!NOTE]
> This PR is Part 2 of 5 to migrate legacy endpoints to MCP endpoints. 
(Base PR: https://github.com/googleapis/genai-toolbox/pull/2711)

### DIFF
--- a/tests/alloydb/alloydb_integration_test.go
+++ b/tests/alloydb/alloydb_integration_test.go
@@ -47,16 +47,16 @@ var (
 
 func getAlloyDBVars(t *testing.T) map[string]string {
 	if AlloyDBProject == "" {
-		t.Skip("'ALLOYDB_PROJECT' not set")
+		t.Fatal("'ALLOYDB_PROJECT' not set")
 	}
 	if AlloyDBLocation == "" {
-		t.Skip("'ALLOYDB_REGION' not set")
+		t.Fatal("'ALLOYDB_REGION' not set")
 	}
 	if AlloyDBCluster == "" {
-		t.Skip("'ALLOYDB_CLUSTER' not set")
+		t.Fatal("'ALLOYDB_CLUSTER' not set")
 	}
 	if AlloyDBInstance == "" {
-		t.Skip("'ALLOYDB_INSTANCE' not set")
+		t.Fatal("'ALLOYDB_INSTANCE' not set")
 	}
 	if AlloyDBUser == "" {
 		t.Fatal("'ALLOYDB_USER' not set")
--- a/tests/alloydbainl/alloydb_ai_nl_integration_test.go
+++ b/tests/alloydbainl/alloydb_ai_nl_integration_test.go
@@ -46,19 +46,19 @@ var (
 func getAlloyDBAINLVars(t *testing.T) map[string]any {
 	switch "" {
 	case AlloyDBAINLProject:
-		t.Skip("'ALLOYDB_AI_NL_PROJECT' not set")
+		t.Fatal("'ALLOYDB_AI_NL_PROJECT' not set")
 	case AlloyDBAINLRegion:
-		t.Skip("'ALLOYDB_AI_NL_REGION' not set")
+		t.Fatal("'ALLOYDB_AI_NL_REGION' not set")
 	case AlloyDBAINLCluster:
-		t.Skip("'ALLOYDB_AI_NL_CLUSTER' not set")
+		t.Fatal("'ALLOYDB_AI_NL_CLUSTER' not set")
 	case AlloyDBAINLInstance:
-		t.Skip("'ALLOYDB_AI_NL_INSTANCE' not set")
+		t.Fatal("'ALLOYDB_AI_NL_INSTANCE' not set")
 	case AlloyDBAINLDatabase:
-		t.Skip("'ALLOYDB_AI_NL_DATABASE' not set")
+		t.Fatal("'ALLOYDB_AI_NL_DATABASE' not set")
 	case AlloyDBAINLUser:
-		t.Skip("'ALLOYDB_AI_NL_USER' not set")
+		t.Fatal("'ALLOYDB_AI_NL_USER' not set")
 	case AlloyDBAINLPass:
-		t.Skip("'ALLOYDB_AI_NL_PASS' not set")
+		t.Fatal("'ALLOYDB_AI_NL_PASS' not set")
 	}
 	return map[string]any{
 		"type":     AlloyDBAINLSourceType,
--- a/tests/cloudsqlmysql/cloud_sql_mysql_integration_test.go
+++ b/tests/cloudsqlmysql/cloud_sql_mysql_integration_test.go
@@ -46,17 +46,17 @@ var (
 func getCloudSQLMySQLVars(t *testing.T) map[string]any {
 	switch "" {
 	case CloudSQLMySQLProject:
-		t.Skip("'CLOUD_SQL_MYSQL_PROJECT' not set")
+		t.Fatal("'CLOUD_SQL_MYSQL_PROJECT' not set")
 	case CloudSQLMySQLRegion:
-		t.Skip("'CLOUD_SQL_MYSQL_REGION' not set")
+		t.Fatal("'CLOUD_SQL_MYSQL_REGION' not set")
 	case CloudSQLMySQLInstance:
-		t.Skip("'CLOUD_SQL_MYSQL_INSTANCE' not set")
+		t.Fatal("'CLOUD_SQL_MYSQL_INSTANCE' not set")
 	case CloudSQLMySQLDatabase:
-		t.Skip("'CLOUD_SQL_MYSQL_DATABASE' not set")
+		t.Fatal("'CLOUD_SQL_MYSQL_DATABASE' not set")
 	case CloudSQLMySQLUser:
-		t.Skip("'CLOUD_SQL_MYSQL_USER' not set")
+		t.Fatal("'CLOUD_SQL_MYSQL_USER' not set")
 	case CloudSQLMySQLPass:
-		t.Skip("'CLOUD_SQL_MYSQL_PASS' not set")
+		t.Fatal("'CLOUD_SQL_MYSQL_PASS' not set")
 	}
 
 	return map[string]any{


internal/tools/snowflake/snowflakesql/snowflakesql.go
```diff
@@ -97,11 +97,15 @@ type Tool struct {
 
 func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, params parameters.ParamValues, accessToken tools.AccessToken) (any, util.ToolboxError) {
 	source, err := tools.GetCompatibleSource[compatibleSource](resourceMgr, t.Source, t.Name, t.Type)
+
```
Yuan325: Please remove this extra line~ Thank you!

internal/tools/snowflake/snowflakesql/snowflakesql.go
```diff
@@ -97,11 +97,15 @@ type Tool struct {
...
 	if err != nil {
 		return nil, util.NewClientServerError("source used is not compatible with the tool", http.StatusInternalServerError, err)
 	}
 
 	paramsMap := params.AsMap()
+	fmt.Println("DEBUG: GetParameters() returned:", t.GetParameters())
+	fmt.Println("DEBUG: paramsMap:", paramsMap)
+
```
Yuan325: please remove these `fmt.Println` lines. If we want to add a debug log, please use the logger~ but it seems like ading a debug log is redundant for this PR?

tests/alloydb/alloydb_integration_test.go
```diff
@@ -47,16 +47,16 @@ var (
 
 func getAlloyDBVars(t *testing.T) map[string]string {
 	if AlloyDBProject == "" {
-		t.Fatal("'ALLOYDB_PROJECT' not set")
+		t.Skip("'ALLOYDB_PROJECT' not set")
```
Yuan325: why do we update it to skip?

internal/tools/snowflake/snowflakesql/snowflakesql.go
anubhav756: Thanks for pointing this out!

internal/tools/snowflake/snowflakesql/snowflakesql.go
anubhav756: Removed these. Thanks for the tip!

tests/alloydb/alloydb_integration_test.go
anubhav756: I had originally skipped them for easier local testing. Let me remove them. Thanks for pointing this out!

anubhav756:
> in general is it possible to separate the changes into each individual PRs to ease the review process?
> 
> For the GetParameters() bug it is resolved in another PR and should already been merged into the latest main.

I actually already proactively split the full migration effort into two distinct PR branches just to mitigate the size (this one here for the highly uniform SQL dialects, and an upcoming https://github.com/googleapis/genai-toolbox/pull/2713 that specifically handles the heavier custom integrations like Looker and BigQuery).

Because this branch serves as the foundation upon which that upcoming complex PR stack had to be done, fracturing it further into 15 smaller individual SQL PRs against `main` right now would break my downstream rebasing topology.

Since the test changes in this specific PR are extremely uniform (virtually identical `RunMCPToolInvoke...` substitutions repeated across perfectly matched SQL dialects), would it be acceptable to review this as a single structural migration? I promise the subsequent PRs in the stack are much more concise! :)

tests/alloydb/alloydb_integration_test.go
```diff
@@ -362,9 +311,16 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
+		Error *struct {
+			Message string `json:"message"`
+		} `json:"error"`
+		Result *struct {
+			Content []struct {
+				Text string `json:"text"`
+			} `json:"content"`
+			IsError bool `json:"isError"`
+		} `json:"result"`
 	}
```
Yuan325, line 323: Is this the MCP Tool Call Response? Might be worth adding a new `mcp_types.go` file that keeps all the response struct related to MCP there. This seems to be shared across multiple database. We can refer to the latest https://github.com/googleapis/genai-toolbox/blob/main/internal/server/mcp/jsonrpc/jsonrpc.go and https://github.com/googleapis/genai-toolbox/blob/main/internal/server/mcp/v20250618/types.go for structs.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -452,21 +421,45 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
 				}
 
-				var clustersData ListClustersResponse
-				if err := json.Unmarshal([]byte(body.Result), &clustersData); err != nil {
-					t.Fatalf("error parsing nested result JSON: %v", err)
-				}
+				if tc.expectAgentErr {
+					gotText := ""
+					if body.Error != nil {
+						gotText = body.Error.Message
```
Yuan325, line 427: can we just return here? that way we don't need to check for `gotText`.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -452,21 +421,45 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
+					} else if body.Result != nil && len(body.Result.Content) > 0 {
+						gotText = body.Result.Content[0].Text
```
Yuan325, line 429: similar as well, we can just return here.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -452,21 +421,45 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
+					}
 
-				var got []string
-				for _, cluster := range clustersData.Clusters {
-					got = append(got, cluster.Name)
-				}
+					if gotText == "" {
+						t.Fatalf("expected error response, but got none")
+					}
```
Yuan325, line 434: For this, we can just throw an error without having to check `gotText`.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -452,21 +421,45 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
+				} else {
```
Yuan325, line 435: we should be able to just remove the `else`. If the `if-` statement did not qualify, it will just come to here straight.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -452,21 +421,45 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
+					if body.Error != nil {
+						t.Fatalf("MCP error: %s", body.Error.Message)
+					}
+					if body.Result != nil && body.Result.IsError {
+						t.Fatalf("MCP result error")
```
Yuan325, line 440: Can we print the content as well? to help with debugging. Iirc, when `IsError = true`, it should have some error related text in `Content`.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -452,21 +421,45 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
+					}
+
+					if body.Result == nil || len(body.Result.Content) == 0 {
+						t.Fatalf("empty MCP result")
```
Yuan325, line 444:
```suggestion
						t.Fatalf("expected result content, but none found")
```

tests/alloydb/alloydb_integration_test.go
```diff
@@ -481,9 +474,16 @@ func runAlloyDBListUsersTest(t *testing.T, vars map[string]string) {
...
+		Error *struct {
+			Message string `json:"message"`
+		} `json:"error"`
+		Result *struct {
+			Content []struct {
+				Text string `json:"text"`
+			} `json:"content"`
+			IsError bool `json:"isError"`
+		} `json:"result"`
 	}
```
Yuan325, line 486: This should be the same as before right? I think we can just define it once and reuse it across all dbs since this formatting is the same.

tests/alloydb/alloydb_integration_test.go
Yuan325, line 1: Please update the remaining change according to the comments~ TY!

tests/alloydbainl/alloydb_ai_nl_integration_test.go
```diff
@@ -250,26 +209,44 @@ func runAINLToolInvokeTest(t *testing.T) {
...
+				Error *struct {
+					Message string `json:"message"`
+				} `json:"error"`
+				Result *struct {
+					Content []struct {
+						Text string `json:"text"`
+					} `json:"content"`
+					IsError bool `json:"isError"`
+				} `json:"result"`
+			}
```
Yuan325, line 224: Similar to before, this could be shared

tests/clickhouse/clickhouse_integration_test.go
```diff
@@ -753,9 +788,21 @@ func TestClickHouseEdgeCases(t *testing.T) {
 			t.Fatalf("error parsing response body")
 		}
 
-		got, ok := body["result"].(string)
+		resultObj, ok := body["result"].(map[string]interface{})
```
Yuan325, line 791: for all these, we can just unmarshal it into the mcp response struct.

tests/embedding.go
```diff
@@ -105,20 +105,6 @@ func RunSemanticSearchToolInvokeTest(t *testing.T, insertWant, mcpInsertWant, se
...
-			requestBody: `{"content": "The quick brown fox jumps over the lazy dog"}`,
-			want:        insertWant,
-		},
-		{
-			name:        "HTTP invoke search_docs",
-			api:         "http://127.0.0.1:5000/api/tool/search_docs/invoke",
-			isMcp:       false,
-			requestBody: `{"query": "fast fox jumping"}`,
-			want:        searchWant,
-		},
```
Yuan325, line 121: why are we removing this?

tests/clickhouse/clickhouse_integration_test.go
```diff
@@ -688,7 +711,7 @@ func TestClickHouseEdgeCases(t *testing.T) {
 		}
 		longQuery := "SELECT 1 WHERE " + strings.Join(conditions, " AND ")
 
-		api := "http://127.0.0.1:5000/api/tool/execute-sql-tool/invoke"
+		api := "http://127.0.0.1:5000/mcp"
```
Yuan325, line 714: This call might fail? since the request body is not in MCP format (currently using `bytes.NewBuffer([]byte(param))`)?

Alternatively, we can update the `tests.RunRequest()` function to reformat the request body to MCP formatting before sending the API request. We'll also need to include some default header values. (e.g. `MCP-Protocol-Version: <protocol-version>`) assuming that we are only testing the latest version in our integration test.

We can import `internal/server/mcp/mcp` and use the [LATEST_PROTOCOL_VERSION](https://github.com/googleapis/genai-toolbox/blob/4b074edbf5fdbd6df111ecf592a0cc3361556c9e/internal/server/mcp/mcp.go#L37) value.

tests/cloudsql/cloud_sql_create_database_test.go
```diff
@@ -165,8 +165,17 @@ func TestCreateDatabaseToolEndpoints(t *testing.T) {
...
-			req, err := http.NewRequest(http.MethodPost, api, bytes.NewBufferString(tc.body))
+			mcpReq := map[string]any{
+				"jsonrpc": "2.0",
+				"id":      "test-1",
+				"method":  "tools/call",
+				"params": map[string]any{
+					"name":      tc.toolName,
+					"arguments": json.RawMessage(tc.body),
+				},
+			}
```
Yuan325, line 176: Could be deduped. This could also be added into the `mcp_types.go` file and reuse across multiple databases.

tests/cloudsql/cloud_sql_create_database_test.go
```diff
@@ -190,15 +199,61 @@ func TestCreateDatabaseToolEndpoints(t *testing.T) {
...
+					Content []struct {
+						Type string `json:"type"`
+						Text string `json:"text"`
+					} `json:"content"`
+					IsError bool `json:"isError,omitempty"`
+				} `json:"result"`
+				Error *struct {
+					Code    int    `json:"code"`
+					Message string `json:"message"`
+				} `json:"error,omitempty"`
```
Yuan325, line 215: can dedup

tests/cloudsql/cloud_sql_create_database_test.go
anubhav756: We are moving to a unified `InvokeMCPTool` harness in https://github.com/googleapis/genai-toolbox/pull/2828, which doesn't duplicate these anymore! Thanks and closing this one.

tests/cloudsql/cloud_sql_create_database_test.go
anubhav756: Same as above, but thanks!

tests/clickhouse/clickhouse_integration_test.go
anubhav756: The new `InvokeMCPTool`in https://github.com/googleapis/genai-toolbox/pull/2828 handles request formatting and protocol headers automatically. Thanks for pointing this out!

tests/alloydb/alloydb_integration_test.go
anubhav756: The new tests written with the slim harness takes care of this. Thanks for the suggestion!

tests/alloydb/alloydb_integration_test.go
anubhav756: Good catch! I'll make sure the new tests written with the slim harness (https://github.com/googleapis/genai-toolbox/pull/2828) default to cleaner patterns like this.

tests/alloydb/alloydb_integration_test.go
anubhav756: Ack! Thanks!

tests/alloydb/alloydb_integration_test.go
anubhav756: We'll ensure the new tests directly fail on unexpected conditions. Thanks!

tests/alloydb/alloydb_integration_test.go
anubhav756: Thanks for pointing this out!

tests/alloydb/alloydb_integration_test.go
anubhav756: Good point! The new harness (https://github.com/googleapis/genai-toolbox/pull/2828) logs the full response on failure to help with debugging.

tests/alloydb/alloydb_integration_test.go
anubhav756: Thanks for the suggestion! We will use descriptive error messages in the new test suite!

tests/alloydb/alloydb_integration_test.go
anubhav756: Agree on deduplicating these structs. The new `InvokeMCPTool` helper will handles this centrally. Thanks!

tests/alloydb/alloydb_integration_test.go
anubhav756: Ack

tests/alloydbainl/alloydb_ai_nl_integration_test.go
anubhav756: Thanks for the suggestion!

tests/clickhouse/clickhouse_integration_test.go
anubhav756: Great point! The new harness unmarshals directly into MCP standard structs. Thanks for the suggestion!

tests/embedding.go
anubhav756: We removed these because they were targeting the legacy `/api` endpoints.

anubhav756:
Thank you for the extremely thorough review across these files! We have decided to pivot strategy and move to a consolidated architecture with a unified slim harness in https://github.com/googleapis/genai-toolbox/pull/2828.

To avoid redundant work, we are closing this PR and mapping all your specific feedback to the new implementation. Thank you! :)

PR #2717, opened 2026-03-11
feat(auth): support manual PRM override

duwenxin99:
Add flag to allow manual PRM override.

### DIFF
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -443,7 +443,11 @@ func TestPRMOverride(t *testing.T) {
 			fmt.Printf("Server serve error: %v\n", err)
 		}
 	}()
-	defer s.Shutdown(ctx)
+	defer func() {
+		if err := s.Shutdown(ctx); err != nil {
+			t.Errorf("failed to cleanly shutdown server: %v", err)
+		}
+	}()
 
 	// Perform the request to the well-known endpoint
 	url := fmt.Sprintf("http://%s:%d/.well-known/oauth-protected-resource", addr, port)


internal/server/mcp.go
```diff
@@ -773,6 +774,27 @@ type prmResponse struct {
 
 // prmHandler generates the Protected Resource Metadata (PRM) file for MCP Authorization.
 func prmHandler(s *Server, w http.ResponseWriter, r *http.Request) {
+	if s.mcpPrmFile != "" {
+		prmBytes, err := os.ReadFile(s.mcpPrmFile)
```
averikitsch: Can we unmarshal this file into a struct? We should know all the possible fields.

internal/server/mcp.go
```diff
@@ -773,6 +774,27 @@ type prmResponse struct {
 
 // prmHandler generates the Protected Resource Metadata (PRM) file for MCP Authorization.
 func prmHandler(s *Server, w http.ResponseWriter, r *http.Request) {
+	if s.mcpPrmFile != "" {
```
averikitsch: Let's not load this on the fly and on every request. Please have this preloaded and read to serve for the endpoint

averikitsch: We should fail fast

averikitsch (approved):

internal/server/prm.go
```diff
@@ -0,0 +1,69 @@
...
+
+package server
+
+// ProtectedResourceMetadata represents the OAuth 2.0 Protected Resource Metadata document as defined in RFC 9728.
+// Reference: https://datatracker.ietf.org/doc/html/rfc9728
+type ProtectedResourceMetadata struct {
+	// REQUIRED. The protected resource's resource identifier (a URL using the https scheme).
+	Resource string `json:"resource"`
+
+	// OPTIONAL. Array containing a list of OAuth authorization server issuer identifiers.
```
averikitsch: This is required for our use case

internal/server/server.go
```diff
@@ -419,8 +423,35 @@ func NewServer(ctx context.Context, cfg ServerConfig) (*Server, error) {
...
+	var cachedPrmBytes []byte
+	var prmConfig ProtectedResourceMetadata
+	if s.mcpPrmFile != "" {
+		var err error
+		cachedPrmBytes, err = os.ReadFile(s.mcpPrmFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read manual PRM file at startup: %w", err)
+		}
+		// Unmarshal into the struct to strictly validate the schema
+		if err := json.Unmarshal(cachedPrmBytes, &prmConfig); err != nil {
```
averikitsch, line 437: in the future we may wan to change to use decoder.DisallowUnknownFields(). This is good for now.

internal/server/prm.go
duwenxin99: updated.

PR #2713, opened 2026-03-11
test: migrate complex integrations to MCP

anubhav756:
## Overview
This PR migrates integration tests for highly complex enterprise APIs and pipeline systems (like Cloud Healthcare, Dataplex, BigQuery, Dgraph, Looker, Serverless Spark).

## Changes
* Switched hundreds of lines of complex API REST verifications across integrations to test precisely against unified `CallTool` generic wrappers.
* Streamlined and simplified Looker and Dataplex validation to properly interface natively with generic JSON representation payloads shipped over MCP.

> [!NOTE]
> This PR is Part 3 of 5 to migrate legacy endpoints to standard MCP endpoints.
(Base PR: https://github.com/googleapis/genai-toolbox/pull/2712)

PR #2725, opened 2026-03-12
feat(ui): adapt Web UI to native MCP endpoints

anubhav756:
## Overview
This PR adapts the frontend MCP Toolbox UI directly against the dynamically populated MCP specifications.

## Changes
* Overhauled underlying frontend engine files (`runTool.js`, `loadTools.js`).
* Updated the parser to read direct standard MCP interface Result structures and format content blocks using unified Markdown conversions onto the DOM.
* Stripped down proxy array mappers that were explicitly tailored for unwrapping old non-schema REST implementations.

> [!NOTE]
> This PR is Part 4 of 5 to migrate legacy endpoints to standard MCP endpoints.
(Base PR: https://github.com/googleapis/genai-toolbox/pull/2713)

### DIFF
--- a/internal/server/static/js/loadTools.js
+++ b/internal/server/static/js/loadTools.js
@@ -27,12 +27,31 @@ let toolDetailsAbortController = null;
 export async function loadTools(secondNavContent, toolDisplayArea, toolsetName) {
     secondNavContent.innerHTML = '<p>Fetching tools...</p>';
     try {
-        const response = await fetch(`/api/toolset/${toolsetName}`);
+        const rpcPayload = {
+            jsonrpc: "2.0",
+            id: crypto.randomUUID(),
+            method: "tools/list",
+            params: {}
+        };
+
+        const response = await fetch('/mcp', {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify(rpcPayload)
+        });
+
         if (!response.ok) {
             throw new Error(`HTTP error! status: ${response.status}`);
         }
         const apiResponse = await response.json();
-        renderToolList(apiResponse, secondNavContent, toolDisplayArea);
+        
+        if (apiResponse.error) {
+            throw new Error(`MCP error ${apiResponse.error.code}: ${apiResponse.error.message}`);
+        }
+
+        renderToolList(apiResponse.result, secondNavContent, toolDisplayArea);
     } catch (error) {
         console.error('Failed to load tools:', error);
         secondNavContent.innerHTML = `<p class="error">Failed to load tools: <pre><code>${escapeHtml(String(error))}</code></pre></p>`;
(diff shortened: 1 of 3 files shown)


internal/server/static/js/loadTools.js
```diff
@@ -27,12 +27,31 @@ let toolDetailsAbortController = null;
 export async function loadTools(secondNavContent, toolDisplayArea, toolsetName) {
```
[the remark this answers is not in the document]
Yuan325: /mcp endpoint could use toolset via the url.

internal/server/static/js/loadTools.js
```diff
@@ -27,12 +27,31 @@ let toolDetailsAbortController = null;
...
     try {
-        const response = await fetch(`/api/toolset/${toolsetName}`);
+        const rpcPayload = {
+            jsonrpc: "2.0",
+            id: crypto.randomUUID(),
+            method: "tools/list",
+            params: {}
+        };
+
+        const response = await fetch('/mcp', {
```
Yuan325, line 37:
```suggestion
        const response = await fetch('/mcp/${toolsetName}', {
```

mcp can also utilize toolset through the URL. would want to keep this here so that the toolset page does not break.

internal/server/static/js/runTool.js
```diff
@@ -74,22 +74,41 @@ export async function handleRunTool(toolId, form, responseArea, parameters, pret
...
+            method: "tools/call",
+            params: {
+                name: toolId,
+                arguments: typedParams
+            }
+        };
+
+        const response = await fetch(`/mcp`, {
             method: 'POST',
             headers: headers,
```
Yuan325, line 95: iirc, we will needa include the `MCP-Protocol-Version` header here with each `tools/call`. We can use the latest version (can import internal/server/mcp/mcp and use this[ LATEST_PROTOCOL_VERSION](https://github.com/googleapis/genai-toolbox/blob/4b074edbf5fdbd6df111ecf592a0cc3361556c9e/internal/server/mcp/mcp.go#L37)). The default if not provided will be the oldest MCP version.

internal/server/static/js/runTool.js
```diff
@@ -145,18 +164,30 @@ export function displayResults(results, responseArea, prettify) {
...
             responseArea.value = JSON.stringify(resultJson);
         }
     } catch (error) {
         console.error("Error parsing or stringifying results:", error);
-        if (typeof results.result === 'string') {
-            responseArea.value = results.result;
+        if (results.content && results.content.length > 0) {
+            responseArea.value = results.content.map(c => c.text).join('\n\n');
+        } else if (typeof results === 'string') {
+            responseArea.value = results;
```
Yuan325, line 188: this will always be results.content right since it's MCP format? in that case we should be able to remove this?

Or we can take some code from the gemini-code-assist recommendation to parse differently when the content length >1 and when it's 1. `resultJson = parsedContent.length === 1 ? parsedContent[0] : parsedContent;`?

internal/server/static/js/runTool.js
```diff
@@ -145,18 +164,25 @@ export function displayResults(results, responseArea, prettify) {
...
+        if (results.content && results.content.length > 0 && results.content[0].text) {
+             responseArea.value = results.content[0].text;
+        } else if (typeof results === 'string') {
+             responseArea.value = results;
         } else {
-            responseArea.value = "Error displaying results. Invalid format.";
+             responseArea.value = JSON.stringify(results, null, 2);
         }
     }
 }
```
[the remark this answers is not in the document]
anubhav756: Covering this in the new PRs.

internal/server/static/js/runTool.js
anubhav756: Handled this in https://github.com/googleapis/genai-toolbox/pull/2829. Thanks for pointing this out!

internal/server/static/js/runTool.js
anubhav756: Agree! The new implementation covers this. Thanks for the suggestion :)

internal/server/static/js/loadTools.js
anubhav756: Ack. Thanks for the suggestion!

anubhav756:
Closing this PR because the frontend migration to full MCP JSON-RPC is being completed as part of https://github.com/googleapis/genai-toolbox/pull/2829. Your points have been factored into the new implementation. Thank you for the detailed feedback! :)

PR #2734, opened 2026-03-13
fix: return AllParams for GetParameter()

### LINKED ISSUE #2720: bigquery-sql: templateParameters values silently dropped on invocation, causing 'missing parameter' error
## Bug description

Custom `bigquery-sql` tools that use `templateParameters` fail on every invocation with:

```
unable to extract template params: error getting template params missing parameter <name>
```

The MCP schema (`tools/list`) correctly advertises template parameters as required input, but when `tools/call` is received, the template parameter values are silently dropped before reaching `Invoke`.

## Steps to reproduce

1. Define a custom `bigquery-sql` tool with `templateParameters`:

```yaml
kind: tools
name: list_tables
type: bigquery-sql
source: bigquery-source
description: List all tables in a dataset.
statement: |
  SELECT t.table_id, t.row_count
  FROM `{{.dataset}}.__TABLES__` t
  ORDER BY t.row_count DESC
templateParameters:
  - name: dataset
    type: string
    description: "Project-qualified dataset, e.g. project.dataset_name"
```

2. Start toolbox with `--tools-file` pointing to the YAML above.
3. Call via MCP JSON-RPC:

```json
{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"list_tables","arguments":{"dataset":"my-project.my_dataset"}}}
```

4. Observe error: `unable to extract template params: error getting template params missing parameter dataset`

## Root cause

The disconnect is between schema generation and invocation parameter parsing:

### Schema generation (`bigquerysql.go:78-83`) — correct
```go
allParameters, paramManifest, err := parameters.ProcessParameters(cfg.TemplateParameters, cfg.Parameters)
mcpManifest := tools.GetMcpManifest(cfg.Name, cfg.Description, cfg.AuthRequired, allParameters, nil)
```
`allParameters` = templateParameters + parameters. The MCP schema correctly advertises all params.

### MCP handler (`method.go:198`) — bug
```go
params, err := parameters.ParseParams(tool.GetParameters(), data, claimsFromAuth)
```

### GetParameters (`bigquerysql.go:248-250`) — bug
```go
func (t Tool) GetParameters() parameters.Parameters {
    return t.Parameters  // only regular params, excludes templateParameters
}
```

`ParseParams` iterates over the parameter definitions list. Since `GetParameters()` returns only regular `Parameters` (not `AllParams`), any template parameter values present in `data` are silently ignored. The resulting `ParamValues` is missing the template param entries.

### Invoke (`bigquerysql.go:117-121`) — fails
```go
paramsMap := params.AsMap()  // empty for tools with only templateParameters
newStatement, err := parameters.ResolveTemplateParams(t.TemplateParameters, t.Statement, paramsMap)
// → "missing parameter dataset"
```

## Expected behavior

Template parameter values sent via `tools/call` should be preserved through `ParseParams` and available in the `paramsMap` passed to `ResolveTemplateParams`.

## Suggested fix

`GetParameters()` should return `t.AllParams` instead of `t.Parameters`, so that `ParseParams` in the MCP handler processes both regular and template parameters. Alternatively, the handler could call `ParseParams` with all parameters.

Note: this same pattern exists in the REST API handler (`api.go:234`), and likely affects all tool types that support `templateParameters` (not just `bigquery-sql`).

## Environment

- Toolbox version: v0.28.0 (commit `81253a0`)
- Tool type: `bigquery-sql`
- Invocation method: MCP stdio (`tools/call`)
- Also tested via REST API (`/api/tool/{name}/invoke`) and `toolbox invoke` CLI — same result

Yuan325:
Fixes for some tools that did not return `t.AllParams` for the `GetParameters()` when template parameter exists.

🛠️ Fixes #2720

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/googleapis/genai-toolbox
 
 go 1.25.7
 
-toolchain go1.26.0
+toolchain go1.25.8
 
 require (
 	cloud.google.com/go/alloydbconn v1.17.3


internal/tools/bigtable/bigtable.go
```diff
@@ -147,5 +147,5 @@ func (t Tool) GetAuthTokenHeaderName(resourceMgr tools.SourceProvider) (string,
 }
 
 func (t Tool) GetParameters() parameters.Parameters {
-	return t.Parameters
```
annguy3n, line 150: this is just renaming, right?

Yuan325: This will include all parameters including the template parameters. Or else parseParam will filter out template parameter and later throwing an error due to missing param.

PR #2737, opened 2026-03-13
chore!: update config `kind` from plural to singular

Yuan325:
Update all the `kind` value from plural to singular. This breaking change only affect users that are currently using the config file v2 format.

Config file with the v1 format will still be compatible since we manually convert the config file to the right format under the hood.


drstrangelooker:
Thje conventional commits error is that you need a space between `chore!:` and `update` in the title of the PR.

PR #2742, opened 2026-03-13
fix(test): address flaky healthcare integration test run

helloeve:
This PR introduces an auto-cleanup mechanism to address flakiness in the Cloud Healthcare integration tests caused by accumulation of orphaned resources (arising from abrupt CI terminations or local cancellations).


helloeve:
@Yuan325 the cleanup logic will attempt to delete stores with prefix `dicom-store-` or `fhir-store-` so the temp ones your created won't be deleted. I tested for a manual created one with `fhir-store-temporary-test` and it was deleted successfully.

PR #2771, opened 2026-03-17
chore(sources/cloudsqlmysql,sources/mysql): remove required from database name

Yuan325:
Tested locally.

To connect to MySQL, the database field is optional in the connection string. If not specified, the client will log into MySQL at root level. Tested that I am able to access object by full name `<database_name>.<table_name>` without providing the database field in source.

Update for mysql source to build DSN using the `go-sql-driver/mysql` Config struct (https://pkg.go.dev/github.com/go-sql-driver/mysql#Config.FormatDSN).

### DIFF
--- a/cmd/internal/skills/command.go
+++ b/cmd/internal/skills/command.go
@@ -91,123 +91,138 @@ func run(cmd *skillsCmd, opts *internal.ToolboxOptions) error {
 		return errMsg
 	}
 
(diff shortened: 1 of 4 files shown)


shuzhou-gc (commented):
why don't you update the mysql port also?

Yuan325:
@shuzhou-gc Updated~ also update mysql source to use the driver's Config struct to format DSN string.

PR #2777, opened 2026-03-18
test(cloudloggingadmin, serverlessspark): Fix flakey integration tests

duwenxin99:
Remove check for log names for flexibility

### DIFF
--- a/.github/workflows/deploy_dev_docs.yaml
+++ b/.github/workflows/deploy_dev_docs.yaml
@@ -56,7 +56,7 @@ jobs:
           node-version: "22"
 
       - name: Cache dependencies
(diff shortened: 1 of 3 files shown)


tests/cloudloggingadmin/cloud_logging_admin_integration_test.go
```diff
@@ -205,7 +205,7 @@ func getCloudLoggingAdminToolsConfig(sourceConfig map[string]any) map[string]any
 	}
 }
 
-func runListLogNamesTest(t *testing.T, expectedLogName string) {
+func runListLogNamesTest(t *testing.T) {
```
wangauone, line 208: nit: Do we want to update the test name if `expectedLogName` is not passed in?

PR #2798, opened 2026-03-20
feat: add PUT endpoint for admin api

Yuan325:
Support put endpoint for `/admin` api. 

This endpoint will create (if primitive not yet existed) or update primitive.
_Note: Updating name does not work. This endpoint will create a new primitive._

This PR also does the following: 
* move `internal/server/mock.go` to `internal/testutils/mock.go` so that it can be shared by other tests.
* create new mocks for each primitive configs.
* move the `Promptset` and `Toolset` initialize into a separate function. This will allow us to run Promptset and Toolset initialize (to update manifest) after updating the tools or prompts list

### DIFF
--- a/internal/server/admin.go
+++ b/internal/server/admin.go
@@ -15,9 +15,16 @@
 package server
 
 import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
+	"github.com/googleapis/genai-toolbox/internal/util"
 )
 
 // adminRouter creates a router that represents the routes under /admin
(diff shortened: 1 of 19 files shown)


internal/server/resources/resources.go
```diff
@@ -158,3 +164,214 @@ func (r *ResourceManager) GetPromptsMap() map[string]prompts.Prompt {
 	}
 	return copiedMap
 }
+
+// UpdateSource creates or update source.
+func (r *ResourceManager) UpdateSource(ctx context.Context, name string, config sources.SourceConfig) error {
```
averikitsch, line 184: We most likely need to mutex lock here as well

internal/server/admin.go
```diff
@@ -28,5 +35,92 @@ func adminRouter(s *Server) (chi.Router, error) {
...
+func createOrUpdatePrimitives(s *Server, w http.ResponseWriter, r *http.Request) {
+	kind := chi.URLParam(r, "kind")
+	name := chi.URLParam(r, "name")
+	ctx := r.Context()
+	ctx = util.WithInstrumentation(ctx, s.instrumentation)
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		err = fmt.Errorf("unable to read request body: %w", err)
+		s.logger.DebugContext(ctx, err.Error())
+		_ = render.Render(w, r, newErrResponse(err, http.StatusInternalServerError))
```
averikitsch: This should be a client error

internal/server/admin.go
```diff
@@ -28,5 +35,92 @@ func adminRouter(s *Server) (chi.Router, error) {
...
+		return
+	}
+	var req struct {
+		Config map[string]any `json:"config"`
+	}
+	if err := json.Unmarshal(body, &req); err != nil {
+		err = fmt.Errorf("unable to unmarshal request body: %w", err)
+		s.logger.DebugContext(ctx, err.Error())
+		_ = render.Render(w, r, newErrResponse(err, http.StatusInternalServerError))
```
averikitsch: this is also a client error

internal/server/resources/resources.go
```diff
@@ -158,3 +164,214 @@ func (r *ResourceManager) GetPromptsMap() map[string]prompts.Prompt {
...
+
+	defaultToolset := r.toolsets[""]
+	if exist {
+		curConfig := primitive.ToConfig()
+		// if config remains the same, return
+		// if diff, re-initialize the primitive
+		if reflect.DeepEqual(config, curConfig) {
+			return nil
+		}
+		defaultToolset = removeFirstFromToolset(defaultToolset, name)
```
averikitsch: What decision did we make for removign from other toolsets?

averikitsch (changes requested):

internal/testutils/mocks.go
```diff
@@ -12,27 +12,83 @@
...
+	MockToolConfig
 	Name                         string
 	Description                  string
 	Params                       []parameters.Parameter
-	manifest                     tools.Manifest
-	unauthorized                 bool
-	requiresClientAuthrorization bool
+	ToolManifest                 tools.Manifest
+	Unauthorized                 bool
+	RequiresClientAuthrorization bool
```
averikitsch: nit: fix typo "RequiresClientAuthorization"

internal/server/resources/resources.go
Yuan325: I think it remains that we do not update other toolset. Users have to explicitly make those changes. Hence this only remove the tool from the default toolset.

internal/server/admin.go
Yuan325: Updated to StatusBadRequest

internal/server/admin.go
Yuan325: Updated to bad request

internal/server/resources/resources.go
Yuan325: Removed this line. If the tool exist in the current toolset, we wouldn't have to remove and re-add them. We can just leave it. If the tool does not exist, only we append it to the toolset.

internal/server/resources/resources.go
Yuan325: added locks where the set happened. The get function will automatically use a read lock.

Yuan325:
@averikitsch following your comment, I've moved the promptset and toolset's initialize function. whenever any promptset or toolset is updated (including the default as well), we'll run the `Initialize()` to re-generate the manifests. This step was missing previously. Thanks for catching this.

PR #2826, opened 2026-03-24
feat(ui): migrate static ui from native API to MCP API endpoints

anubhav756:


internal/server/static/js/loadTools.js
```diff
@@ -27,10 +27,21 @@ let toolDetailsAbortController = null;
 export async function loadTools(secondNavContent, toolDisplayArea, toolsetName) {
     secondNavContent.innerHTML = '<p>Fetching tools...</p>';
     try {
-        const response = await fetch(`/api/toolset/${toolsetName}`);
+        const url = toolsetName ? `/mcp/${toolsetName}` : `/mcp`;
+        const response = await fetch(url, {
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
```
Yuan325, line 33: we might need the `MCP-Protocol-Version` header here as well
```suggestion
            headers: { 'Content-Type': 'application/json', 'Mcp-Protocol-Version': '2025-11-25' },
```

PR #2828, opened 2026-03-24
test(harness): implement MCP JSON-RPC test harness

anubhav756:
## Overview

This PR introduces the native MCP test harness, enabling us to run integration tests using the JSON-RPC protocol over `/mcp`.

## Changes
- Added `tests/mcp_execute.go` and `tests/mcp_types.go` as the core harness for invoking MCP tools and asserting responses.
- Updated `tests/server.go` to support dynamic flags like `--enable-api`.
- Added unit/integration tests in `internal/server/server_test.go` to verify the harness itself.

## Checklist

### DIFF
--- a/tests/mcp_tool.go
+++ b/tests/mcp_tool.go
@@ -26,7 +26,7 @@ import (
 	"github.com/google/uuid"
 
 	"github.com/googleapis/genai-toolbox/internal/server/mcp/jsonrpc"
-	v20250618 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20250618"
+	v20251125 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20251125"
 )
 
 // NewMCPRequestHeader takes custom headers and appends headers required for MCP.
@@ -36,7 +36,7 @@ func NewMCPRequestHeader(t *testing.T, customHeaders map[string]string) map[stri
 		headers[k] = v
 	}
 	headers["Content-Type"] = "application/json"
-	headers["MCP-Protocol-Version"] = v20250618.PROTOCOL_VERSION
+	headers["MCP-Protocol-Version"] = v20251125.PROTOCOL_VERSION
 	return headers
 }
 
@@ -70,7 +70,7 @@ func GetMCPToolsList(t *testing.T, requestHeader map[string]string) (int, []any,
 	req := MCPListToolsRequest{
 		Jsonrpc: jsonrpc.JSONRPC_VERSION,
 		Id:      uuid.New().String(),
-		Method:  v20250618.TOOLS_LIST,
+		Method:  v20251125.TOOLS_LIST,
 	}
 	reqBody, err := json.Marshal(req)
 	if err != nil {
--- a/tests/mcp_types.go
+++ b/tests/mcp_types.go
@@ -16,7 +16,7 @@ package tests
 
 import (
 	"github.com/googleapis/genai-toolbox/internal/server/mcp/jsonrpc"
-	v20250618 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20250618"
+	v20251125 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20251125"
 )
 
 // CallToolParams represents the internal payload of an MCP tool call request
@@ -45,7 +45,7 @@ type MCPCallToolRequest struct {
 type MCPCallToolResponse struct {
 	Jsonrpc string                   `json:"jsonrpc"`
 	Id      jsonrpc.RequestId        `json:"id"`
-	Result  v20250618.CallToolResult `json:"result,omitempty"`
+	Result  v20251125.CallToolResult `json:"result,omitempty"`
 	Error   *jsonrpc.Error           `json:"error,omitempty"`
 }
 
@@ -54,7 +54,7 @@ func NewMCPCallToolRequest(id jsonrpc.RequestId, toolName string, args map[strin
 	return MCPCallToolRequest{
 		Jsonrpc: jsonrpc.JSONRPC_VERSION,
 		Id:      id,
-		Method:  v20250618.TOOLS_CALL,
+		Method:  v20251125.TOOLS_CALL,
 		Params: CallToolParams{
 			Name:      toolName,
 			Arguments: args,


tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+	v20250618 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20250618"
+)
+
+func NewMCPRequestHeader(t *testing.T, customHeaders map[string]string) map[string]string {
+	headers := make(map[string]string)
+	for k, v := range customHeaders {
+		headers[k] = v
+	}
+	headers["Content-Type"] = "application/json"
+	headers["Mcp-Protocol-Version"] = v20250618.PROTOCOL_VERSION
```
Yuan325:
```suggestion
	headers["MCP-Protocol-Version"] = v20250618.PROTOCOL_VERSION
```

tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+	if headers["Mcp-Session-Id"] == "" {
+		headers["Mcp-Session-Id"] = RunInitialize(t, v20250618.PROTOCOL_VERSION)
+	}
```
Yuan325:
```suggestion
```

This is not anymore needed for newer versions. And it's only needed if the initialization request returned the header.

tests/mcp_tool.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	v20250618 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20250618"
+)
+
+func NewMCPRequestHeader(t *testing.T, customHeaders map[string]string) map[string]string {
```
Yuan325, line 33:
```suggestion
// NewMCPRequestHeader takes custom headers and append headers required for MCP
func NewMCPRequestHeader(t *testing.T, customHeaders map[string]string) map[string]string {
```

tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+	return headers
+}
+
+func ExecuteMCPToolCall(t *testing.T, toolName string, arguments map[string]any, requestHeader map[string]string) (string, error) {
```
Yuan325:
```suggestion
// ExecuteMCPToolCall is a helper function to send HTTP requests to MCP endpoint and return the response
func ExecuteMCPToolCall(t *testing.T, toolName string, arguments map[string]any, requestHeader map[string]string) (string, error) {
```

tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+	headers := NewMCPRequestHeader(t, requestHeader)
+
+	req := NewMCPCallToolRequest("1", toolName, arguments)
```
Yuan325:
```suggestion
	reqId := uuid.New().String()
	req := NewMCPCallToolRequest(reqId, toolName, arguments)
```

tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+	reqBody, _ := json.Marshal(req)
```
Yuan325:
```suggestion
	reqBody, err := json.Marshal(req)
	if err != nil {
		t.Fatalf("error marshalling request body")
	}
```

tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+
+	resp, respBody := RunRequest(t, http.MethodPost, "http://127.0.0.1:5000/mcp", bytes.NewBuffer(reqBody), headers)
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("response status code is not 200, got %d: %s", resp.StatusCode, string(respBody))
+	}
```
Yuan325: Might not want to check this. We have some test case that is expecting an error and check the exact error string.

tests/mcp_execute.go
```diff
@@ -0,0 +1,71 @@
...
+		return "null", nil
+	}
+
+	var contentText string
+	for _, c := range mcpResp.Result.Content {
+		if c.Type == "text" {
+			contentText += c.Text
+		}
+	}
+	return strings.TrimSpace(contentText), nil
```
Yuan325: For this, I might just
```
	var result map[string]any
	var err error
	isError := false
	if mcpResp.Result != nil {
		isError = mcpResp.Result.IsError
		err = json.Unmarshal(mcpResp.Result, &result)
	}
	if mcpResp.Error != nil {
		err := json.Unmarshal(mcpResp.Error, &result)
	}
	if err != nil {
		t.Fatalf("error unmarshaling body")
	}
	return resp.StatusCode, isError, result
```

With this, we will need to update the function header to return `(int, bool, map[string]any)`.

That way, the test cases for MCP endpoint can just check the following: 
1. check if the status code is correct.
2. check if the `IsError` is correct (want to make sure that this is false for expected successful returns)
3. check the content via `result`.

tests/mcp_execute.go
anubhav756: Thanks for catching that!

tests/mcp_execute.go
anubhav756: Got it. Thanks!

tests/mcp_tool.go
anubhav756: Thanks!

tests/mcp_execute.go
anubhav756: Thanks!

tests/mcp_execute.go
anubhav756: Thanks for the suggestion!

tests/mcp_execute.go
anubhav756: Good idea! I've updated it to support custom error codes. Thanks!

tests/mcp_execute.go
anubhav756: Thanks for pointing it out!

I have taken an approach to keep using `MCPCallToolResponse` rather than switching to a generic dictionary (`map[string]any`). This means the compiler already knows exactly how to extract the text from the JSON. If we just dump everything into a generic map, I think we'll have to write messy, error-prone code to guess where the fields are nested.

Additionally, the MCP protocol seems to specifically separate high-level system errors (like `"invalid parameters"`) from actual tool execution errors (like a database query failing). Our current approach handles this by surfacing protocol failures as standard Go errors while returning semantic tool errors directly. The snippet you shared seems to be squashing both of these scenarios together under a single generic dictionary and boolean flag. I feel it makes figuring out exactly what went wrong much harder?

Moreover, keeping it this way ensures our ~44 legacy database tests are already wired up to expect a simple string from this function. If we change the signature to return a complex dictionary instead, we would instantly break that translation bridge, and we'd be forced to manually rewrite thousands of lines of test assertions across the entire codebase.

tests/mcp_execute.go
anubhav756: Thanks for the suggestion!

tests/mcp_tool.go
```diff
@@ -0,0 +1,209 @@
...
+	if mcpResp.Result.IsError {
+		// If it's an application-level MCP tool failure, map it as an error text
+		var errText string
+		for _, c := range mcpResp.Result.Content {
+			if c.Type == "text" {
+				errText += c.Text
+			}
+		}
+		return resp.StatusCode, strings.TrimSpace(errText), nil
+	}
```
Yuan325: Just a heads up to take not of this when migrating the int tests. resp.Error will be returned as err, whereas resp.Result.IsError will be returned as regular text with nil error.

tests/mcp_execute.go
```diff
@@ -0,0 +1,209 @@
...
+
+	var textBlocks []string
+	for _, c := range mcpResp.Result.Content {
+		if c.Type == "text" {
+			textBlocks = append(textBlocks, strings.TrimSpace(c.Text))
+		}
+	}
+
+	if len(textBlocks) == 0 {
+		return resp.StatusCode, "null", nil
```
Yuan325:
```suggestion
		return resp.StatusCode, "", nil
```

Wondering if there's a reason to do `"null"`? Or just empty string seems to be more accurate?

tests/mcp_tool.go
```diff
@@ -0,0 +1,209 @@
...
+	if len(mcpResp.Result.Content) == 0 {
+		return resp.StatusCode, "null", nil
+	}
+
+	var textBlocks []string
+	for _, c := range mcpResp.Result.Content {
+		if c.Type == "text" {
+			textBlocks = append(textBlocks, strings.TrimSpace(c.Text))
+		}
```
Yuan325:
```suggestion
		textBlocks = append(textBlocks, strings.TrimSpace(c.Text))
```

We only support "text" now afaik, so there won't be a need to check the type :)

tests/mcp_execute.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	if len(textBlocks) == 1 {
+		return resp.StatusCode, textBlocks[0], nil
+	}
+
+	// For legacy assertions: if multiple blocks are returned and they look like JSON, wrap them into a JSON array
+	first := textBlocks[0]
+	if strings.HasPrefix(first, "{") || strings.HasPrefix(first, "[") || strings.HasPrefix(first, "\"") {
+		return resp.StatusCode, "[" + strings.Join(textBlocks, ",") + "]", nil
+	}
```
Yuan325: Do we need these? What does "legacy assertions" meant? Should we just do `strings.Join(textBlocks, "\n")` for all cases regardless?

Also, if we do need these, should we place this before the `if len(textBlocks) == 1` check?

tests/mcp_execute.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return resp.StatusCode, strings.Join(textBlocks, "\n"), nil
+}
+
+// InterceptLegacyDo intercepts a hardcoded HTTP request destined for /api/tool/.../invoke,
+// dynamically converts it into a local ExecuteMCPToolCall, and wraps the
+// MCP string (or JSON-RPC logic error) inside a standard Go *http.Response.
+func InterceptLegacyDo(t *testing.T, req *http.Request) (*http.Response, error) {
```
Yuan325: I believe we don't need this anymore? since we have the flag now. legacy API should still go through. I've added the `--enable-api` flag for integration test. so legacy /api should still work :) we just have to focus on migrating everyhting to /mcp, and at the last step, removing the `--enable-api` flag from integration test.

tests/mcp_tool.go
```diff
@@ -0,0 +1,209 @@
...
+
+	var mcpResp jsonrpc.JSONRPCResponse
+	if err := json.Unmarshal(respBody, &mcpResp); err != nil {
+		if resp.StatusCode != http.StatusOK {
+			return resp.StatusCode, nil, fmt.Errorf("%s", string(respBody))
+		}
+		t.Fatalf("error parsing tools/list response: %v\nraw body: %s", err, string(respBody))
+	}
+
+	return resp.StatusCode, &mcpResp, nil
```
Yuan325: Should we only response with the mcpResp.Results.Tools here? seems like thats the only thing that matters.

tests/mcp_tool.go
```diff
@@ -0,0 +1,209 @@
...
+	for k, v := range customHeaders {
+		headers[k] = v
+	}
+	headers["Content-Type"] = "application/json"
+	headers["MCP-Protocol-Version"] = v20250618.PROTOCOL_VERSION
+	return headers
+}
+
+// InvokeMCPTool is a transparent, native JSON-RPC execution harness for tests.
+func InvokeMCPTool(t *testing.T, toolName string, arguments map[string]any, requestHeader map[string]string) (int, *MCPCallToolResponse, error) {
```
Yuan325, line 44: What is the difference of this and the `ExecuteMCPToolCall`?

tests/mcp_tool.go
anubhav756: Good point! This should make assertions cleaner. Thanks!

tests/mcp_tool.go
anubhav756: Consolidated these into a single clean helper to avoid this confusion. Thanks for pointing this out!

tests/mcp_execute.go
```diff
@@ -0,0 +1,64 @@
...
+		return "null", nil
+	}
+
+	var contentText string
+	for _, c := range mcpResp.Result.Content {
+		if c.Type == "text" {
+			contentText += c.Text
+		}
+	}
+	return strings.TrimSpace(contentText), nil
```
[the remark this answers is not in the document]
anubhav756: Used an approach to wrap multiple blocks into a JSON array if they look like JSON, to avoid breaking legacy assertions.

tests/mcp_execute.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	requestHeader["Content-Type"] = "application/json"
+	if requestHeader["Mcp-Session-Id"] == "" {
+		requestHeader["Mcp-Session-Id"] = RunInitialize(t, v20250618.PROTOCOL_VERSION)
+	}
+
+	req := NewMCPCallToolRequest("1", toolName, arguments)
+	reqBody, _ := json.Marshal(req)
+
+	resp, respBody := RunRequest(t, http.MethodPost, "http://127.0.0.1:5000/mcp", bytes.NewBuffer(reqBody), requestHeader)
```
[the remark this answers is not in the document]
anubhav756: Agree. I will add t.Fatalf for marshaling errors.

tests/mcp_execute.go
anubhav756: Good point! I will change it.

tests/mcp_tool.go
anubhav756: Thanks for bringing that up. I'll verify if we can safely remove this check.

tests/mcp_execute.go
anubhav756: We will avoid using it. Thanks for pointing that!

tests/mcp_execute.go
anubhav756: Agree! I have simplified the logic to just use `strings.Join(textBlocks, "\n")` for all cases regardless of the count, removing the special handling and the `len(textBlocks) == 1` check entirely.

By 'legacy assertions', I was referring to some older tests that expected a JSON-like object format when multiple content blocks were returned. However, joining them simply with newlines is much cleaner. Thanks!

tests/mcp_tool.go
anubhav756: It's a great heads up. I will keep this distinction in mind when updating the assertions in the integration tests!

tests/mcp_tool.go
```diff
@@ -0,0 +1,229 @@
...
+				continue
+			}
+			if toolMap["name"] == expected.Name {
+				found = true
+				if expected.Description != "" && toolMap["description"] != expected.Description {
+					t.Fatalf("tool %s description mismatch: want %q, got %q", expected.Name, expected.Description, toolMap["description"])
+				}
+				break
+			}
+		}
```
Yuan325: We should probably check all aspect of Tools. not just the name and description. Might want to check the arguments as well. If we unmarshal the `resultMap["tools"] into `MCPToolManifest struct` instead, we can just do a `reflect.DeepEqual()` comparison.

tests/mcp_tool.go
anubhav756: We decided to remove the `ExecuteMCPToolCall` function entirely, since it wasn't very useful, this check has been eliminated. Thanks for pointing it out though!

tests/mcp_tool.go
anubhav756: Refactored `GetMCPToolsList` to return only the list part of the result rather than the full `jsonrpc.JSONRPCResponse`. Its return signature is now `(int, []any, error)`. Thanks for the suggestion!

tests/mcp_tool.go
anubhav756: Seems like I missed actually removing it. I have removed it entirely now. `InvokeMCPTool` is now the sole primitive for executing tools and returning the full structured response. This eliminates the confusion. Thanks for pointing this out!

tests/mcp_tool.go
anubhav756: Thanks for the great suggestion! Implemented.

tests/mcp_tool.go
```diff
@@ -0,0 +1,184 @@
...
+	"fmt"
+	"net/http"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/google/uuid"
+
+	"github.com/googleapis/genai-toolbox/internal/server/mcp/jsonrpc"
+	v20250618 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20250618"
```
Yuan325:
```suggestion
	v20251125 "github.com/googleapis/genai-toolbox/internal/server/mcp/v20251125"
```

just caught on that our latest mcp version is 20251125

Yuan325 (approved):

tests/mcp_tool.go
anubhav756: Great catch! Thanks :)

Updated that in https://github.com/googleapis/genai-toolbox/pull/2828/changes/4b41e0d2030266aa1812d5a034ceebb3b4c758d0 across both `mcp_types.go` and `mcp_tool.go`.

PR #2842, opened 2026-03-25
test(source/http): create MCP integration tests

anubhav756:
## Overview

This PR adds the mapped integration tests for HTTP tools using the new native MCP harness.

## Changes
- Created `tests/http/http_mcp_test.go` and ported all relevant tests from the legacy `http_integration_test.go` to use the `/mcp` pathway.
- Includes covered positive paths, negative paths, and optional query parameter variations.

## Checklist

### DIFF
--- a/tests/http/http_mcp_test.go
+++ b/tests/http/http_mcp_test.go
@@ -1,0 +1,314 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package http
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/MicahParks/jwkset"
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/googleapis/genai-toolbox/internal/testutils"
+	"github.com/googleapis/genai-toolbox/tests"
+)
+
+func getMCPHTTPSourceConfig(t *testing.T) map[string]any {
+	idToken, err := tests.GetGoogleIdToken(tests.ClientId)
+	if err != nil {
+		t.Logf("Warning: error getting ID token: %s. Using dummy token.", err)
+		idToken = "dummy-token"
+	}
+	idToken = "Bearer " + idToken
+
+	return map[string]any{
+		"type":    HttpSourceType,
+		"headers": map[string]string{"Authorization": idToken},
+	}
+}
+
+func TestHTTPListTools(t *testing.T) {
+	// Start a test server with multiTool handler
+	server := httptest.NewServer(http.HandlerFunc(multiTool))
+	defer server.Close()
+
+	sourceConfig := getMCPHTTPSourceConfig(t)
+	sourceConfig["baseUrl"] = server.URL
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// Set up generic auth mock server (copied from legacy test)
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf("failed to create RSA private key: %v", err)
+	}
+	jwksServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/.well-known/openid-configuration" {
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]interface{}{
+				"issuer":   "https://example.com",
+				"jwks_uri": "http://" + r.Host + "/jwks",
+			})
+			return
+		}
+		if r.URL.Path == "/jwks" {
+			options := jwkset.JWKOptions{
+				Metadata: jwkset.JWKMetadataOptions{
+					KID: "test-key-id",
+				},
+			}
+			jwk, _ := jwkset.NewJWKFromKey(privateKey.Public(), options)
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]interface{}{
+				"keys": []jwkset.JWKMarshal{jwk.Marshal()},
+			})
+			return
+		}
+		http.NotFound(w, r)
+	}))
+	defer jwksServer.Close()
+
+	toolsFile := getHTTPToolsConfig(sourceConfig, HttpToolType, jwksServer.URL)
+
+	// Start the toolbox server.
+	cmd, cleanup, err := tests.StartCmd(ctx, toolsFile)
+	if err != nil {
+		t.Fatalf("command initialization returned an error: %s", err)
+	}
+	defer cleanup()
+
+	// Wait for server ready
+	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	expectedTools := []tests.MCPToolManifest{
+		{
+			Name:        "my-simple-tool",
+			Description: "Simple tool to test end to end functionality.",
+			InputSchema: map[string]any{"type": "object", "properties": map[string]any{}, "required": []any{}},
+		},
+		{
+			Name:        "my-tool",
+			Description: "some description",
+			InputSchema: map[string]any{
+				"type":     "object",
+				"required": []any{"id", "name"},
+				"properties": map[string]any{
+					"id": map[string]any{
+						"type":        "integer",
+						"description": "user ID",
+					},
+					"name": map[string]any{
+						"type":        "string",
+						"description": "user name",
+					},
+				},
+			},
+		},
+	}
+
+	tests.RunMCPToolsListMethod(t, expectedTools)
+}
+
+func TestHTTPCallTool(t *testing.T) {
+	// Start a test server with multiTool handler
+	server := httptest.NewServer(http.HandlerFunc(multiTool))
+	defer server.Close()
+
+	sourceConfig := getMCPHTTPSourceConfig(t)
+	sourceConfig["baseUrl"] = server.URL
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// Set up generic auth mock server (needed for config generation)
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf("failed to create RSA private key: %v", err)
+	}
+	jwksServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/.well-known/openid-configuration" {
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]interface{}{
+				"issuer":   "https://example.com",
+				"jwks_uri": "http://" + r.Host + "/jwks",
+			})
+			return
+		}
+		if r.URL.Path == "/jwks" {
+			options := jwkset.JWKOptions{
+				Metadata: jwkset.JWKMetadataOptions{
+					KID: "test-key-id",
+				},
+			}
+			jwk, _ := jwkset.NewJWKFromKey(privateKey.Public(), options)
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]interface{}{
+				"keys": []jwkset.JWKMarshal{jwk.Marshal()},
+			})
+			return
+		}
+		http.NotFound(w, r)
+	}))
+	defer jwksServer.Close()
+
+	toolsFile := getHTTPToolsConfig(sourceConfig, HttpToolType, jwksServer.URL)
+
+	// Start the toolbox server.
+	cmd, cleanup, err := tests.StartCmd(ctx, toolsFile)
+	if err != nil {
+		t.Fatalf("command initialization returned an error: %s", err)
+	}
+	defer cleanup()
+
+	// Wait for server ready
+	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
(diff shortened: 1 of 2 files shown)


.ci/integration.cloudbuild.yaml
```diff
@@ -49,1475 +49,38 @@ steps:
...
-    - versionName: projects/$PROJECT_ID/secrets/singlestore_host/versions/latest
-      env: SINGLESTORE_HOST
-    - versionName: projects/$PROJECT_ID/secrets/mariadb_user/versions/latest
-      env: MARIADB_USER
-    - versionName: projects/$PROJECT_ID/secrets/mariadb_pass/versions/latest
-      env: MARIADB_PASS
-    - versionName: projects/$PROJECT_ID/secrets/mariadb_host/versions/latest
-      env: MARIADB_HOST
-    - versionName: projects/$PROJECT_ID/secrets/mongodb_uri/versions/latest
-      env: MONGODB_URI
```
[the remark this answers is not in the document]
anubhav756: NA

tests/http/http_integration_test.go
```diff
@@ -324,79 +323,137 @@ func TestHttpToolEndpoints(t *testing.T) {
 		t.Fatalf("toolbox didn't start successfully: %s", err)
 	}
 
-	// Run tests
-	tests.RunToolGetTest(t)
-	tests.RunToolInvokeTest(t, `"hello world"`, tests.DisableArrayTest())
+	// Native tools/list invocation test
+	statusCodeList, toolsResp, errList := tests.GetMCPToolsList(t, nil)
```
Yuan325: Should we just do a `RunMCPToolsListMethod()` in `tests/tools.go` similar to the `RunMCPToolCallMethod()`?

That way we can just do:

* In `mcp_types.go`:
```
type MCPListToolsResponse struct{
    ...
}

// MCPToolManifest is a copy of tools.McpManifest used for integration testing purposes
type MCPToolManifest struct {
	Name string `json:"name"`
	Description string           `json:"description,omitempty"`
	Annotations *ToolAnnotations `json:"annotations,omitempty"`
	InputSchema map[string]any `json:"inputSchema,omitempty"`
	Metadata    map[string]any            `json:"_meta,omitempty"`
}
```

In `tests/tools.go`:
```
func RunMCPToolsListMethod(t *testing, expectedOutput []McpToolManifest) {
    // run the tools list
    // marshal output to MCPListToolsResponse
    // compare the result of response.Results.Tools to expectedOutput (we'll can check the name, description, parameter, separately
}
```

With this, in each integration test we just have to do the following:
```
expectedOutput := []tests.MCPToolManifest{
    tests.MCPToolManifest{
        Name: "tool-1",
        Description: "description...",
        Annotations: "....",
        ...
    },
    tests.MCPToolManifest{
        Name: "tool-2",
        Description: "descirption...",
        ...  
    }
}
tests.RunMCPToolsListMethod(t, expectedOutput)
```

tests/http/http_integration_test.go
```diff
@@ -324,79 +323,137 @@ func TestHttpToolEndpoints(t *testing.T) {
...
-				t.Fatalf("unable to send request: %s", err)
+			if statusCode != http.StatusOK {
+				t.Fatalf("response status code is not 200, got %d", statusCode)
 			}
-			defer resp.Body.Close()
 
-			if resp.StatusCode != http.StatusOK {
-				bodyBytes, _ := io.ReadAll(resp.Body)
-				t.Fatalf("response status code is not 200, got %d: %s", resp.StatusCode, string(bodyBytes))
+			if tc.wantErrMsg != "" {
```
Yuan325: Wondering if we could also do a `RunMCPCustomToolCallMethod()` that runs the tool calling -> parse results -> compare with expected output.

That way, all the source that needs custom tool calls doesn't have to re-implement all these parsing.

Might make future migration easier.

tests/http/http_integration_test.go
Yuan325: Update: seems like we already do have a `GetMCPToolsList()` function. Should we update that function to take expected output and do the comparison as well? since all the tool will require the same format and comparison.

tests/http/http_integration_test.go
Yuan325: Seems like we have a toolCall method in `mcp_execute.go` maybe we can modify that?

tests/http/http_integration_test.go
anubhav756: I actually defined `RunMCPCustomToolCallMethod()` in `tests/mcp_execute.go` in #2828 [here](https://github.com/googleapis/genai-toolbox/pull/2828/changes#diff-bb0b446bf199d4e2bf95a53705c0cdf2bc231bcb69f8aa56d69124956b6d03c8R268) to handle this. It invokes the tool, verifies the successful response, and asserts that the output contains the expected substring. Using it in this PR to ensure consistent assertions!

Thanks for the suggestion!

tests/http/http_integration_test.go
anubhav756: Same as in the previous comment [above](https://github.com/googleapis/genai-toolbox/pull/2842#discussion_r2991125300).

Thanks! :)

tests/http/http_integration_test.go
```diff
@@ -324,79 +323,137 @@ func TestHttpToolEndpoints(t *testing.T) {
...
+					}
+				} else {
+					t.Fatalf("expected error containing %q, but got success result: %v", tc.wantErrMsg, mcpResp.Result)
+				}
+				
+				if !strings.Contains(errText, tc.wantErrMsg) {
+					t.Fatalf("expected error text containing %q, got %q", tc.wantErrMsg, errText)
+				}
+				return
 			}
```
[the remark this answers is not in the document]
anubhav756: Exported the error assertion logic as `tests.AssertMCPError` in `tests/mcp_execute.go` [here](https://github.com/googleapis/genai-toolbox/pull/2828/changes#diff-bb0b446bf199d4e2bf95a53705c0cdf2bc231bcb69f8aa56d69124956b6d03c8R205) to be shared across test packages and reduce deduplication.

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,235 @@
...
+			return
+		}
+		http.NotFound(w, r)
+	}))
+	defer jwksServer.Close()
+
+	toolsFile := getHTTPToolsConfig(sourceConfig, HttpToolType, jwksServer.URL)
+
+	// Start the toolbox server using our NEW helper!
+	cmd, cleanup, err := tests.StartCmd(ctx, toolsFile, "--enable-api=false")
```
Yuan325:
```suggestion
	cmd, cleanup, err := tests.StartCmd(ctx, toolsFile)
```

By default this should be false. I've created another PR (#2905) to remove the hardcoded `--enable-api` flag.

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,235 @@
...
+	defer cancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	// Test calling "my-simple-tool"
+	// From handleTool0, it expects POST and returns `"hello world"`
+	tests.RunMCPCustomToolCallMethod(t, "my-simple-tool", map[string]any{}, `"hello world"`)
```
Yuan325: We would probably want a similar tests as `RunToolInvokeTest` so that all edge cases are covered within the mcp tests since our goal is to eventually remove the `/api` endpoint. A lot of these tests are added after we caught some edge cases. Would recomment to update the existing `RunToolInvokeTest` instead to be /mcp compatible (or create a new `RunMCPToolInvokeTest()`).

For the other http-specific tests as below, we can then create these specific t.Run cases. 
```
	runAdvancedHTTPInvokeTest(t)
	runQueryParamInvokeTest(t)
	runGenericAuthInvokeTest(t, privateKey)
```

tests/http/http_mcp_test.go
Yuan325, line 1: I'm thinking we should be good to just replace the `http_integration_test.go` with this new codes? Since we are planning to remove the /api endpoint eventually anyway?

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,235 @@
...
+		{
+			Name:        "my-tool",
+			Description: "some description",
+		},
+	}
+
+	tests.RunMCPToolsListMethod(t, expectedTools)
+}
+
+func TestHTTPCallTool(t *testing.T) {
```
Yuan325, line 136: I like the clean separation of mcp method, but just wanted to make sure this will not cause longer integration run time? (having to start and run 2 separate servers for each integration test suites).

tests/http/http_mcp_test.go
anubhav756: I have updated `tests/server.go` to remove that fallback logic. Now the legacy API is disabled by default for all test servers started via `tests.StartCmd`, unless explicitly requested. Also removed the `--enable-api=false` argument in `http_mcp_test.go` as well. Will be removing it from the `anubhav-feat-native-mcp-alloydb` branch as well. Thanks for this simplification! :)

tests/http/http_mcp_test.go
anubhav756: I created a shared helper `RunMCPToolInvokeTest` in `tests/mcp_tool.go` that mirrors the legacy `RunToolInvokeTest` options using MCP calls. This will serve as a reusable harness for other DB sources too.

For the HTTP-specific edge cases, I added the missing positive test cases that you mentioned as specific `t.Run` blocks in `http_mcp_test.go` to ensure we cover everything before deleting the legacy files.

tests/http/http_mcp_test.go
anubhav756: We decided to keep `http_mcp_test.go` parallel to the legacy file for now to avoid disruptions while both the `/api` and `/mcp` endpoints are still supported (although `/api` is behind the flag), to ensure coverage without touching the legacy test suite during transition. We plan to delete the legacy file entirely once the transition to MCP is complete project-wide.

tests/http/http_mcp_test.go
anubhav756: You are right. However, I feel that for now, this isolation is a fair trade-off during transition! Once legacy tests are deleted, total execution time will go down. Moreover, we now have tests sharding implemented, so that all unrelated tests are skipped from a PR, which should also help with this. :)

tests/mcp_tool.go
```diff
@@ -182,3 +182,70 @@ func RunMCPCustomToolCallMethod(t *testing.T, toolName string, arguments map[str
...
+		supportClientAuth:        false,
+		supportSelect1Want:       true,
+		supportSelect1Auth:       true,
+	}
+
+	for _, option := range options {
+		option(configs)
+	}
+
+	t.Run("MCP invoke my-simple-tool", func(t *testing.T) {
```
Yuan325: Can we use the table-driven tests similar to our `RunToolInvokeTest` (e.g. define test cases and then run it all in a for loop). I think that way is more preferred for Go-way of writing tests to prevent duplication logic.

Yuan325 (changes requested):

tests/mcp_tool.go
anubhav756: That's a great idea! Thanks for the suggestion :)

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,299 @@
...
+	defer cancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	// Test calling "my-simple-tool"
+	// From handleTool0, it expects POST and returns `"hello world"`
+	tests.RunMCPCustomToolCallMethod(t, "my-simple-tool", map[string]any{}, `"hello world"`)
```
Yuan325: I think there are duplicated test cases here. This calling for `my-simple-tool` should already be covered within the `RunMCPToolInvokeTest`

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Auth Failure: Invoke generic auth tool without token
+	t.Run("invoke generic auth tool without token", func(t *testing.T) {
```
Yuan325, line 243: This seems to be included in the `runGenericAuthInvokeTest()` of the /api integration test. Do we also test invocation with valid auth header?

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,299 @@
...
+		if err != nil {
+			t.Fatalf("native error executing %s: %s", "my-auth-required-generic-tool", err)
+		}
+		if statusCode != http.StatusUnauthorized {
+			t.Fatalf("expected status 401, got %d", statusCode)
+		}
+	})
+
+	// Missing Required Param: Invoke query-param-tool with required param nil or absent
+	t.Run("invoke query-param-tool missing required param", func(t *testing.T) {
```
Yuan325: This seems to be covered within `RunMCPToolInvokeTest` under the "Invoke my-tool without parameters` test case.

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,299 @@
...
+		tests.RunMCPCustomToolCallMethod(t, "my-query-param-tool", arguments, `"reqId=test1"`)
+	})
+
+	t.Run("invoke query-param-tool some optional nil", func(t *testing.T) {
+		arguments := map[string]any{"reqId": "test2", "page": "5", "filter": nil}
+		tests.RunMCPCustomToolCallMethod(t, "my-query-param-tool", arguments, `"page=5\u0026reqId=test2"`) // 'filter' omitted!
+	})
+
+	// Advanced Tool Success
+	t.Run("invoke my-advanced-tool successfully", func(t *testing.T) {
```
Yuan325, line 304: This seems to be similar to `runAdvancedHTTPInvokeTest()` do we not need to test the "invoke my-advanced-tool with wrong params`?

tests/http/http_mcp_test.go
Yuan325: seems like it's at the bottom, can we move these right next to each other? Or use a separate function similar to the http_integration_test.go?

tests/http/http_mcp_test.go
```diff
@@ -0,0 +1,299 @@
...
+			t.Fatalf("native error executing %s: %s", "my-advanced-tool", err)
+		}
+		if statusCode != http.StatusOK {
+			t.Fatalf("expected status 200, got %d", statusCode)
+		}
+		tests.AssertMCPError(t, mcpResp, "unexpected status code")
+	})
+
+	// Query Parameter Variations: Tests with optional parameters omitted or nil
+	t.Run("invoke query-param-tool optional omitted", func(t *testing.T) {
```
Yuan325: Seems like this is similar to the `runQueryParamInvokeTest()` Do we not need the other test cases?

tests/http/http_mcp_test.go
anubhav756: Thanks for pointing that out! Removed.

tests/http/http_mcp_test.go
anubhav756: Thanks for the suggestion. Moved the test!

tests/http/http_mcp_test.go
anubhav756: Thanks for catching that! Deduped.

tests/http/http_mcp_test.go
anubhav756: The failure case of "invoke my-advanced-tool with wrong params" was actually present in the file before, but it was listed inline. I've now extracted it into `runAdvancedHTTPMCPInvokeTest()` so it matches the legacy structure.

tests/http/http_mcp_test.go
anubhav756: Agree! I have now extracted this into `runQueryParamMCPInvokeTest()` and added those cases that I missed. Thanks for catching that! :)

PR #2843, opened 2026-03-25
test(source/alloydb): create MCP integration tests

anubhav756:
## Overview

This PR adds the mapped integration tests for AlloyDB tools using the native MCP harness.

## Changes
- Created `tests/alloydb/alloydb_mcp_test.go` to test database listing and cluster operations over the native `/mcp` pathway.
- Achieves parity with coverage from the legacy `alloydb_integration_test.go`.

## Checklist

### DIFF
--- a/tests/alloydb/alloydb_mcp_test.go
+++ b/tests/alloydb/alloydb_mcp_test.go
@@ -1,0 +1,1165 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alloydb
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"reflect"
+	"regexp"
+	"sort"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/googleapis/genai-toolbox/internal/testutils"
+	"github.com/googleapis/genai-toolbox/tests"
+)
+
+func TestAlloyDBListTools(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	toolsFile := getAlloyDBToolsConfig()
+
+	// Start the toolbox server
+	cmd, cleanup, err := tests.StartCmd(ctx, toolsFile)
+	if err != nil {
+		t.Fatalf("command initialization returned an error: %v", err)
+	}
+	defer cleanup()
+
+	waitCtx, cancelWait := context.WithTimeout(ctx, 20*time.Second)
+	defer cancelWait()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %v", err)
+	}
+
+	// Verify list of tools
+	expectedTools := []tests.MCPToolManifest{
+		{
+			Name:        "alloydb-list-clusters",
+			Description: "Lists all AlloyDB clusters in a given project and location.",
+			InputSchema: map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"project": map[string]any{
+						"description": "The GCP project ID to list clusters for.",
+						"type":        "string",
+					},
+					"location": map[string]any{
+						"default":     "-",
+						"description": "Optional: The location to list clusters in (e.g., 'us-central1'). Use '-' to list clusters across all locations.(Default: '-')",
+						"type":        "string",
+					},
+				},
+				"required": []any{"project"},
+			},
+		},
+		{
+			Name:        "alloydb-list-users",
+			Description: "Lists all AlloyDB users within a specific cluster.",
+			InputSchema: map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"cluster": map[string]any{
+						"description": "The ID of the cluster to list users from.",
+						"type":        "string",
+					},
+					"location": map[string]any{
+						"description": "The location of the cluster (e.g., 'us-central1').",
+						"type":        "string",
+					},
+					"project": map[string]any{
+						"description": "The GCP project ID.",
+						"type":        "string",
+					},
+				},
+				"required": []any{"project", "location", "cluster"},
+			},
+		},
+	}
+
+	tests.RunMCPToolsListMethod(t, expectedTools)
+}
+
+func TestAlloyDBCallTool(t *testing.T) {
+	vars := getAlloyDBVars(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	toolsFile := getAlloyDBToolsConfig()
+
+	cmd, cleanup, err := tests.StartCmd(ctx, toolsFile)
+	if err != nil {
+		t.Fatalf("command initialization returned an error: %v", err)
+	}
+	defer cleanup()
+
+	waitCtx, cancelWait := context.WithTimeout(ctx, 20*time.Second)
+	defer cancelWait()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %v", err)
+	}
+
+	// Run tool-specific invoke tests
+	runAlloyDBListClustersMCPTest(t, vars)
+	runAlloyDBListInstancesMCPTest(t, vars)
+	runAlloyDBListUsersMCPTest(t, vars)
+	runAlloyDBGetClusterMCPTest(t, vars)
+	runAlloyDBGetInstanceMCPTest(t, vars)
+	runAlloyDBGetUserMCPTest(t, vars)
+
+	t.Run("MCP Invoke invalid tool", func(t *testing.T) {
+		statusCode, mcpResp, err := tests.InvokeMCPTool(t, "non-existent-tool", map[string]any{}, nil)
+		if err != nil {
+			t.Fatalf("native error executing %s: %s", "non-existent-tool", err)
+		}
+		if statusCode != http.StatusOK {
+			t.Fatalf("expected status 200, got %d", statusCode)
+		}
+		tests.AssertMCPError(t, mcpResp, `tool with name "non-existent-tool" does not exist`)
+	})
+}
+
+func runAlloyDBListClustersMCPTest(t *testing.T, vars map[string]string) {
+	type ListClustersResponse struct {
+		Clusters []struct {
+			Name string `json:"name"`
+		} `json:"clusters"`
+	}
+
+	wantForSpecificLocation := []string{
+		fmt.Sprintf("projects/%s/locations/us-central1/clusters/alloydb-ai-nl-testing", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-central1/clusters/alloydb-pg-testing", vars["project"]),
+	}
+
+	wantForAllLocations := []string{
+		fmt.Sprintf("projects/%s/locations/us-central1/clusters/alloydb-ai-nl-testing", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-central1/clusters/alloydb-pg-testing", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-east4/clusters/alloydb-private-pg-testing", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-east4/clusters/colab-testing", vars["project"]),
+	}
+
+	invokeTcs := []struct {
+		name           string
+		args           map[string]any
+		want           []string
+		wantContentErr string
+		expectError    bool
+		wantStatusCode int
+	}{
+		{
+			name:           "list clusters for all locations",
+			args:           map[string]any{"project": vars["project"], "location": "-"},
+			want:           wantForAllLocations,
+			wantStatusCode: http.StatusOK,
+		},
+		{
+			name:           "list clusters specific location",
+			args:           map[string]any{"project": vars["project"], "location": "us-central1"},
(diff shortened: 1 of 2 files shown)


tests/alloydb/alloydb_integration_test.go
```diff
@@ -428,32 +432,34 @@ func runAlloyDBListClustersTest(t *testing.T, vars map[string]string) {
...
-			req, err := http.NewRequest(http.MethodPost, api, tc.requestBody)
-			if err != nil {
-				t.Fatalf("unable to create request: %s", err)
+			bodyBytes, _ := io.ReadAll(tc.requestBody)
+			var args map[string]any
+			if len(bodyBytes) > 0 {
+				if err := json.Unmarshal(bodyBytes, &args); err != nil {
+					t.Fatalf("error parsing json body: %v", err)
+				}
 			}
```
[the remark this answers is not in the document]
anubhav756: Since this comment is on the legacy integration test, and we are focused on creating the new MCP test where error handling is handled properly by our custom helpers, ignoring this error for this specific PR.

tests/alloydb/alloydb_integration_test.go
```diff
@@ -658,32 +656,36 @@ func runAlloyDBListInstancesTest(t *testing.T, vars map[string]string) {
...
```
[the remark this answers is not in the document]
anubhav756: Since this comment is on the legacy integration test, and we are focused on creating the new MCP test where error handling is handled properly by our custom helpers, ignoring this error for this specific PR.

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,689 @@
...
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	tcs := []struct {
+		name           string
+		body           string
+		want           string
+		wantStatusCode int
```
Yuan325: can remove this since it doesnt seem like we are comparing these. (same with the other 2 Test function below)
```suggestion
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,689 @@
...
+		args := map[string]any{"location": vars["location"]}
+		statusCode, mcpResp, err := tests.InvokeMCPTool(t, "alloydb-list-clusters", args, nil)
+		if err != nil {
+			t.Fatalf("native error executing: %s", err)
+		}
+		if statusCode != http.StatusOK {
+			t.Fatalf("expected status 200, got %d", statusCode)
+		}
+		tests.AssertMCPError(t, mcpResp, `parameter "project" is required`)
+	})
```
Yuan325: can we change it back to the function-based test? to make sure we check all test cases.

tests/alloydb/alloydb_wait_for_operation_mcp_test.go
```diff
@@ -0,0 +1,232 @@
...
+				if !mcpResp.Result.IsError {
+					t.Fatalf("expected error result, got success")
+				}
+				var wantMap map[string]string
+				if err := json.Unmarshal([]byte(tc.want), &wantMap); err != nil {
+					t.Fatalf("failed to unmarshal want: %v", err)
+				}
+				if !strings.Contains(got, wantMap["error"]) {
+					t.Errorf("expected error containing %q, got: %q", wantMap["error"], got)
+				}
```
Yuan325: Is it possible to dedup this vs line 202-205 (string comparison in the else statement)? Seems to be similar comparison to me.

tests/alloydb/alloydb_mcp_test.go
anubhav756: Done.

tests/alloydb/alloydb_wait_for_operation_mcp_test.go
anubhav756: Deduped

tests/alloydb/alloydb_mcp_test.go
anubhav756: The server maps protocol-level JSON-RPC failures (such as GCP outages or schema drops) to explicit HTTP 500 codes, while reserving standard HTTP 200 for successful tool execution or tool-level logical errors.

I added `wantStatusCode` so we can strictly assert this deliberate distinction (matching how legacy tests did it), ensuring we track whether a failure was a standard tool error (200) or a platform-level drop (500).

Yuan325: Sounds good, it seems like we waren't checking these codes previoulsy. I saw the updates now :) Thank you!

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+			t.Fatalf("native error executing %s: %s", "alloydb-list-clusters", err)
+		}
+		if statusCode != http.StatusOK {
+			t.Fatalf("expected status 200, got %d", statusCode)
+		}
+		tests.AssertMCPError(t, mcpResp, `parameter "project" is required`)
+	})
+}
+
+func TestAlloyDBMCPReadEndpoints(t *testing.T) {
```
Yuan325: why are there `TestAlloyDBCallTool` vs `TestAlloyDBMCPReadEndpoints`?

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+				if mcpResp.Result.IsError {
+					t.Fatalf("returned error result: %v", mcpResp.Result)
+				}
+				if len(tc.want) > 0 {
+					got := mcpResp.Result.Content[0].Text
+					for _, want := range tc.want {
+						if !regexp.MustCompile(want).MatchString(got) {
+							t.Errorf("Expected substring not found: %q", want)
+						}
+					}
```
Yuan325: Are we expecting that all the want will be within got? 

Can we keep the original check?: I think in here, we want to make sure that got doesn't return more cluster than it originally should.

```
var got []string
for _, cluster := range clustersData.Clusters {
	got = append(got, cluster.Name)
}

sort.Strings(got)
sort.Strings(tc.want)

if !reflect.DeepEqual(got, tc.want) {
	t.Errorf("cluster list mismatch:\n got: %v\nwant: %v", got, tc.want)
}
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+		fmt.Sprintf("projects/%s/locations/us-central1/clusters/alloydb-pg-testing/instances/alloydb-pg-testing-instance", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-east4/clusters/alloydb-private-pg-testing/instances/alloydb-private-pg-testing-instance", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-east4/clusters/colab-testing/instances/colab-testing-primary", vars["project"]),
+	}
+
+	invokeTcs := []struct {
+		name           string
+		args           map[string]any
+		want           []string
+		wantContains   string
```
Yuan325:
```suggestion
		wantContentErr   string
```

want vs wantContains feels very confusing.
If you can update for the other tests as well that'll be great, ty!

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+		fmt.Sprintf("projects/%s/locations/us-central1/clusters/alloydb-pg-testing", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-east4/clusters/alloydb-private-pg-testing", vars["project"]),
+		fmt.Sprintf("projects/%s/locations/us-east4/clusters/colab-testing", vars["project"]),
+	}
+
+	invokeTcs := []struct {
+		name           string
+		args           map[string]any
+		want           []string
+		wantContains   string
```
Yuan325:
```suggestion
		wantContentErr   string
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+				t.Fatalf("expected status %d, got %d", tc.wantStatusCode, statusCode)
+			}
+
+			if tc.expectError {
+				tests.AssertMCPError(t, mcpResp, tc.wantContains)
+			} else {
+				if mcpResp.Result.IsError {
+					t.Fatalf("returned error result: %v", mcpResp.Result)
+				}
+				if len(tc.want) > 0 {
```
Yuan325: we might be able to just remove this check since tc.want will always be >0.

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
```
Yuan325: we should be able to remove this check since tc.want is always >0

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+	type UsersResponse struct {
+		Users []struct {
+			Name string `json:"name"`
+		} `json:"users"`
+	}
+
+	invokeTcs := []struct {
+		name           string
+		args           map[string]any
+		wantContains   string
```
Yuan325:
```suggestion
		want   string
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+			if err != nil {
+				t.Fatalf("native error executing: %s", err)
+			}
+
+			if statusCode != tc.wantStatusCode {
+				t.Fatalf("expected status %d, got %d", tc.wantStatusCode, statusCode)
+			}
+
+			if tc.expectError {
+				tests.AssertMCPError(t, mcpResp, tc.wantContains)
```
Yuan325: will this be a problem for cases that does not have tc.wantContains?

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+		})
+	}
+}
+
+func runAlloyDBGetClusterMCPTest(t *testing.T, vars map[string]string) {
+	invokeTcs := []struct {
+		name           string
+		args           map[string]any
+		want           map[string]any
+		wantContains   string
```
Yuan325:
```suggestion
		wantContentErr   string
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+				t.Fatalf("expected status %d, got %d", tc.wantStatusCode, statusCode)
+			}
+
+			if tc.expectError {
+				tests.AssertMCPError(t, mcpResp, tc.wantContains)
+			} else {
+				if mcpResp.Result.IsError {
+					t.Fatalf("returned error result: %v", mcpResp.Result)
+				}
+				if tc.want != nil {
```
Yuan325: might not need to check this based on the test cases that we have.

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+		})
+	}
+}
+
+func runAlloyDBGetInstanceMCPTest(t *testing.T, vars map[string]string) {
+	invokeTcs := []struct {
+		name           string
+		args           map[string]any
+		want           map[string]any
+		wantContains   string
```
Yuan325:
```suggestion
		wantContentErr   string
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
+				t.Fatalf("expected status %d, got %d", tc.wantStatusCode, statusCode)
+			}
+
+			if tc.expectError {
+				tests.AssertMCPError(t, mcpResp, tc.wantContains)
+			} else {
+				if mcpResp.Result.IsError {
+					t.Fatalf("returned error result: %v", mcpResp.Result)
+				}
+				if tc.want != nil {
```
Yuan325: won't need to check this.
```suggestion
```

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1215 @@
...
```
Yuan325: won't need to check this.

Yuan325 (changes requested):
Could you double-check the comments across all test cases? Just want to make sure the updates carried over to all of them. Thank you!

tests/alloydb/alloydb_mcp_test.go
anubhav756: I took the unique negative test case of "invalid tool" check from it and moved it as a subtest into `TestAlloyDBMCPReadEndpoints` and deleted `TestAlloyDBCallTool` entirely. Thanks for suggesting this one!

tests/alloydb/alloydb_mcp_test.go
anubhav756: Done! Thanks for the suggestion!

tests/alloydb/alloydb_mcp_test.go
anubhav756: That does make sense. Done!

tests/alloydb/alloydb_mcp_test.go
anubhav756: Sounds good! Thanks :)

tests/alloydb/alloydb_mcp_test.go
anubhav756: Removed this redundant condition as well. Thanks!

tests/alloydb/alloydb_mcp_test.go
anubhav756: Removed. Thanks!

tests/alloydb/alloydb_mcp_test.go
anubhav756: Thanks!

tests/alloydb/alloydb_mcp_test.go
anubhav756: Done. Thanks

tests/alloydb/alloydb_mcp_test.go
anubhav756: Good catch! You are right. Since `strings.Contains` matches everything when the search string is empty, we are not strictly validating the error message content for cases that omit `wantContains` (now `wantContentErr`).

This behavior is the same as legacy tests (in `alloydb_integration_test.go`), which only checked the HTTP status code 500 and completely ignored the response body for these failure cases.

Nevertheless, in this new MCP tests, we are slightly safer because `AssertMCPError` at least verifies that the response has a valid MCP error structure.

tests/alloydb/alloydb_mcp_test.go
```diff
@@ -0,0 +1,1165 @@
...
+				},
+				"required": []any{"project", "location", "cluster"},
+			},
+		},
+	}
+
+	tests.RunMCPToolsListMethod(t, expectedTools)
+}
+
+func TestAlloyDBMCPReadEndpoints(t *testing.T) {
```
Yuan325:
```suggestion
func TestAlloyDBCallTool(t *testing.T) {
```

Will try to be consistent with the http tool and all other tool.

anubhav756: Makes total sense! Thanks for the suggestion! :)

PR #2845, opened 2026-03-25
Fix: Prevent integer to float conversion for tool parameters invoked via agent skills

pavankrishna13:
## Description
This PR addresses a bug where tools fail to execute due to a type mismatch error when integer parameters are provided. Crucially, this issue only occurs when the tool is invoked through agent skills; it does not happen when the tools are executed directly. When an agent skill routes a payload via the Gemini CLI, JSON integers are incorrectly unmarshalled as floats before being passed down to the underlying tool.

**Steps to Reproduce**:

Activate an agent skill (e.g., cloud-sql-postgres-data from the cloud-sql-postgresql extensions) via the Gemini CLI.

Invoke a tool that requires an integer parameter (like list_indexes) using a natural language prompt, such as:

"List all the indexes in the database, max limit of rows is 2"

The Gemini CLI agent internally constructs and executes the following command:

```node /usr/local/google/home/pavanrampalli/.gemini/extensions/cloud-sql-postgresql/skills/cloud-sql-postgres-data/scripts/list_indexes.js '{"limit": 2}'```
The execution fails because the tool expects an integer for the limit parameter, but the agent's request parser has converted and passed it as a float.

**Root Cause**:
Upon investigation of the agent skill invocation path, the standard json.Unmarshal function defaults to parsing all JSON numbers as float64. When this parsed request is handed off to a tool that strictly enforces an integer type, it triggers a validation error.

**Proposed Solution**:
Replaced the standard json.Unmarshal with util.DecodeJSON specifically within the request unmarshalling logic for agent skills. This ensures that numeric types are decoded accurately and integers are preserved without being implicitly converted to floats before reaching the tools.

## PR Checklist

- [ ] Make sure you reviewed

### DIFF
--- a/cmd/internal/invoke/command.go
+++ b/cmd/internal/invoke/command.go
@@ -18,10 +18,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/googleapis/genai-toolbox/cmd/internal"
 	"github.com/googleapis/genai-toolbox/internal/server"
 	"github.com/googleapis/genai-toolbox/internal/server/resources"
+	"github.com/googleapis/genai-toolbox/internal/util"
 	"github.com/googleapis/genai-toolbox/internal/util/parameters"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


pavankrishna13:
Code changes committed and merged as part of this PR : https://github.com/googleapis/genai-toolbox/pull/2847

PR #2855, opened 2026-03-26
feat(dataplex): improve dataplex params

sumedhdk22:
## Description

feat(dataplex): refine dataplex tools for better agent usability

Refines the parameters and descriptions for Dataplex tools to reduce redundancy
and improve tool selection accuracy for AI agents.

Key changes:
- Removed the redundant 'name' parameter from 'lookup_entry' and 'lookup_context' tools.
- Implemented automatic derivation of the parent resource path (project and location)
  from the 'entry' or 'resources' arguments.
- Added validation logic to ensure provided resource names follow the correct format
  and that all resources in a batch request belong to the same project and location.
- Enhanced tool and parameter descriptions across 'search_entries', 'lookup_entry',
  and 'lookup_context' to provide clearer instructions and search syntax examples.
- Updated integration tests and YAML configurations to align with the new schema.

## PR Checklist

- [ ] Make sure you reviewed

### DIFF
--- a/cmd/internal/config.go
+++ b/cmd/internal/config.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/goccy/go-yaml"
 	"github.com/google/go-cmp/cmp"
+	"github.com/googleapis/genai-toolbox/internal/auth/generic"
 	"github.com/googleapis/genai-toolbox/internal/server"
 )
 
@@ -307,6 +308,18 @@ func mergeConfigs(files ...Config) (Config, error) {
 	// If conflicts were detected, return an error
 	if len(conflicts) > 0 {
 		return Config{}, fmt.Errorf("resource conflicts detected:\n  - %s\n\nPlease ensure each source, authService, tool, toolset and prompt has a unique name across all files", strings.Join(conflicts, "\n  - "))
+	}
+
+	// Ensure only one authService has mcpEnabled = true
+	var mcpEnabledAuthServers []string
+	for name, authService := range merged.AuthServices {
+		// Only generic type has McpEnabled right now
+		if genericService, ok := authService.(generic.Config); ok && genericService.McpEnabled {
+			mcpEnabledAuthServers = append(mcpEnabledAuthServers, name)
+		}
+	}
+	if len(mcpEnabledAuthServers) > 1 {
+		return Config{}, fmt.Errorf("multiple authServices with mcpEnabled=true detected: %s. Only one MCP authorization server is currently supported", strings.Join(mcpEnabledAuthServers, ", "))
 	}
 
 	return merged, nil
--- a/cmd/internal/config_test.go
+++ b/cmd/internal/config_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/googleapis/genai-toolbox/internal/auth/generic"
 	"github.com/googleapis/genai-toolbox/internal/auth/google"
 	"github.com/googleapis/genai-toolbox/internal/embeddingmodels/gemini"
 	"github.com/googleapis/genai-toolbox/internal/prebuiltconfigs"
@@ -616,38 +617,48 @@ func TestParseConfig(t *testing.T) {
 			type: google
 			clientId: testing-id
 ---
-			kind: embeddingModel
-			name: gemini-model
-			type: gemini
-			model: gemini-embedding-001
-			apiKey: some-key
-			dimension: 768
+      kind: authService
+      name: my-generic-auth
+      type: generic
+      audience: testings
+      authorizationServer: https://testings
+      mcpEnabled: true
+      scopesRequired:
+        - read:files
+        - write:files
 ---
-			kind: tool
-			name: example_tool
-			type: postgres-sql
-			source: my-pg-instance
-			description: some description
-			statement: |
-				SELECT * FROM SQL_STATEMENT;
-			parameters:
-			- name: country
-			  type: string
-			  description: some description
+      kind: embeddingModel
+      name: gemini-model
+      type: gemini
+      model: gemini-embedding-001
+      apiKey: some-key
+      dimension: 768
 ---
-			kind: toolset
-			name: example_toolset
-			tools:
-			- example_tool
+      kind: tool
+      name: example_tool
+      type: postgres-sql
+      source: my-pg-instance
+      description: some description
+      statement: |
+        SELECT * FROM SQL_STATEMENT;
+      parameters:
+      - name: country
+        type: string
+        description: some description
 ---
-			kind: prompt
-			name: code_review
-			description: ask llm to analyze code quality
-			messages:
-			- content: "please review the following code for quality: {{.code}}"
-			arguments:
-			- name: code
-			  description: the code to review
+      kind: toolset
+      name: example_toolset
+      tools:
+      - example_tool
+---
+      kind: prompt
+      name: code_review
+      description: ask llm to analyze code quality
+      messages:
+      - content: "please review the following code for quality: {{.code}}"
+      arguments:
+      - name: code
+        description: the code to review
 			`,
 			wantConfig: Config{
 				Sources: server.SourceConfigs{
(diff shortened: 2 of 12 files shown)


internal/tools/dataplex/dataplexlookupcontext/dataplexlookupcontext.go
```diff
@@ -65,9 +66,13 @@ func (cfg Config) ToolConfigType() string {
 }
 
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
-	name := parameters.NewStringParameter("name", "The project to which the request should be attributed in the following form: projects/{project}/locations/{location}.")
-	resources := parameters.NewArrayParameter("resources", "A list of up to 10 resources names for which metadata is needed.", parameters.NewStringParameter("resource", "Name of a resource in the following format: projects/{project}/locations/{location}/entryGroups/{group}/entries/{entry}."))
-	params := parameters.Parameters{name, resources}
+	resources := parameters.NewArrayParameter("resources",
+		"Required. A list of up to 10 resource names from same project and location.",
+		parameters.NewStringParameter("resource",
+			"Name of a resource in the following format: projects/{project_number}/locations/{location}/entryGroups/{group}/entries/{entry}."+
```
estshy: It can be either project_id or project_number. I would call it either `{project}` or `{project_id_or_number}`

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -64,7 +64,12 @@ func (cfg Config) ToolConfigType() string {
 }
 
 func (cfg Config) Initialize(srcs map[string]sources.Source) (tools.Tool, error) {
-	query := parameters.NewStringParameter("query", "The query against which entries in scope should be matched.")
+	query := parameters.NewStringParameter("query",
+		"A query string for searching entries, following Dataplex search syntax. " +
+			"Supports logical operators (AND, OR, NOT) and grouping. " +
+			"For example, to find a table that might have been renamed, you could use 'type:table (name:books OR fiction)'. " +
```
estshy: nit: This sounds like a skill. There is a lot of edge cases and details about search syntax that might cause LLM to come up with non existing syntax. 

I would either describe all the complexity in a skill or keep it high level as "natural language query"

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
```diff
@@ -64,7 +64,12 @@ func (cfg Config) ToolConfigType() string {
...
+			"This can be more efficient than multiple separate calls." +
+			"Warning: Performing broad searches without specific filters (e.g., type:table) can be slow and consume significant resources. When performing exploratory searches, always use the pageSize parameter to limit the number of results returned.")
```
estshy, line 72: Is performance of tool call a decision factor for LLM? I think agent/system prompt would need to explicitly ask to look for performant solution.

internal/prebuiltconfigs/tools/dataplex.yaml
```diff
@@ -21,19 +21,19 @@ tools:
   search_entries:
     kind: dataplex-search-entries
     source: dataplex-source
-    description: Use this tool to search for entries in Dataplex Catalog based on the provided search query.
+    description: Searches the entries in Dataplex Catalog based on the provided search query.
```
estshy: I am not sure if we want to surface the concept of "entries". We do not use this terminology in the user interface, therefore users will most likely not use it either.

I would consider something generic that can be matched by generic knowledge of LLM like "Searches for data asset (e.g. table/view) in Catalog based on the provided search query.

internal/prebuiltconfigs/tools/dataplex.yaml
```diff
@@ -21,19 +21,19 @@ tools:
...
   lookup_entry:
     kind: dataplex-lookup-entry
     source: dataplex-source
-    description: Use this tool to retrieve a specific entry from Dataplex Catalog.
+    description: Retrieves a specific entry from Dataplex Catalog.
```
estshy: Here as well I would suggest to avoid technical terminology and focus on user intention.

For example:
```
Retrieves a specific metadata regarding a data asset (e.g. table/view) from Catalog
```

internal/prebuiltconfigs/tools/dataplex.yaml
```diff
@@ -21,19 +21,19 @@ tools:
...
   search_aspect_types:
     kind: dataplex-search-aspect-types
     source: dataplex-source
-    description: Use this tool to find aspect types relevant to the query.
+    description: Find aspect types relevant to the query.
```
estshy: I think "find" implies a single result but we are providing multiple candidates. Should we call this operation "search"?

sumedhdk22: Done.

internal/prebuiltconfigs/tools/dataplex.yaml
sumedhdk22: Done

internal/prebuiltconfigs/tools/dataplex.yaml
sumedhdk22: Updated.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
sumedhdk22: It would increase the latency if filters are not specified and agent needs to make multiple calls to get the desired output. To minimize this, added this comment so agents can use the filter in query for most of the scenarios.

internal/tools/dataplex/dataplexsearchentries/dataplexsearchentries.go
sumedhdk22: Skill will be to use this tool if user wants to search for any data asset. It will go in Agent configuration.

This change in tool param description will be helpful for agent to construct the query. Current behavior is it takes multiple calls to find the result and even struggle to construct the right query. Adding the details in description makes it clear and agent can get the desired result in minimum number of calls.

internal/tools/dataplex/dataplexlookupcontext/dataplexlookupcontext.go
sumedhdk22: Done, updated to project_id_or_number. Agent was getting confused with what to use, hence added project_id_or_number to indicate any of project id or project number can be used.

internal/tools/dataplex/dataplexlookupentry/dataplexlookupentry.go
```diff
@@ -119,10 +119,16 @@ func (t Tool) Invoke(ctx context.Context, resourceMgr tools.SourceProvider, para
...
+
+	parts := strings.Split(entry, "/")
+	if len(parts) < 4 || parts[0] != "projects" || parts[2] != "locations" {
+		err = fmt.Errorf("invalid entry format: must be in the form projects/{project}/locations/{location}/entryGroups/{entryGroup}/entries/{entry}")
+		return nil, util.NewAgentError(err.Error(), err)
+	}
+	name := strings.Join(parts[:4], "/")
 	if err != nil {
 		return nil, util.NewAgentError(fmt.Sprintf("can't convert aspectTypes to array of strings: %s", err), err)
 	}
```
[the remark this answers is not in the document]
sumedhdk22: Done

internal/tools/dataplex/dataplexlookupcontext/dataplexlookupcontext_test.go
Yuan325: Shouldn't need to define these, since these are added during tool's initialization?

internal/tools/dataplex/dataplexlookupentry/dataplexlookupentry_test.go
Yuan325: Same with this, shouldn't need to add these since it is defined during initialization.

Yuan325:
@sumedhdk22 seems like dataplex integration tests is failing~

internal/tools/dataplex/dataplexlookupcontext/dataplexlookupcontext_test.go
sumedhdk22: Ack. Removed the 'advanced example' test.

internal/tools/dataplex/dataplexlookupentry/dataplexlookupentry_test.go
sumedhdk22: Ack. Removed the 'advanced example' test.

PR #2871, opened 2026-03-27
feat: add DELETE endpoint for admin endpoint

Yuan325:
New DELETE endpoint for `/admin/{kind}/{name}`. 

Deletions are made per primitive and will not be made in batches. Successful deletion will return a `200 OK` to the client. If the source is not found, Toolbox will return a `404 NOT FOUND`. After deletion, the primitive will not be discoverable (GET SHOULD NOT return the deleted primitive).

Deletion of primitives **WILL NOT** cascade towards other primitives.
* **Deletion of Source:** Tools that are linked to the Source will still be discoverable. However, when the user invokes the tool, an error will be thrown.
* **Deletion of Tool:** 
    * Default toolset: The tool is automatically removed from default toolset (name: `""`).
    * If custom toolset is updated but tool is not removed - initialization will throw a "tool does not exist" error.
    * If custom toolset is not updated - the deleted tool will still appear when calling tools/list, but any attempt to invoke it will fail.

### DIFF
--- a/internal/server/admin.go
+++ b/internal/server/admin.go
@@ -36,8 +36,31 @@ func adminRouter(s *Server) (chi.Router, error) {
 	r.Use(render.SetContentType(render.ContentTypeJSON))
 
 	r.Put("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { createOrUpdatePrimitives(s, w, r) })
+	r.Delete("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { deletePrimitives(s, w, r) })
 
 	return r, nil
+}
+
+// deletePrimitives handles the deletion of primitives
+// Invalid primitive kind will result in http.StatusBadRequest
+// Primitive not found will result in http.StatusNotFound
+func deletePrimitives(s *Server, w http.ResponseWriter, r *http.Request) {
+	kind := chi.URLParam(r, "kind")
+	name := chi.URLParam(r, "name")
+
+	if err := s.ResourceMgr.Delete(kind, name); err != nil {
+		s.logger.DebugContext(r.Context(), err.Error())
+
+		// Determine status code based on error message/type
+		status := http.StatusNotFound
+		if strings.Contains(err.Error(), "invalid primitive kind provided") {
+			status = http.StatusBadRequest
+		}
+
+		_ = render.Render(w, r, newErrResponse(err, status))
+		return
+	}
+	w.WriteHeader(http.StatusOK)
 }
 
 // createOrUpdatePrimitives handles the creation or updating of primitives
--- a/internal/server/resources/resources.go
+++ b/internal/server/resources/resources.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 
 	"github.com/googleapis/genai-toolbox/internal/auth"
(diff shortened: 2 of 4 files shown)


internal/server/resources/resources.go
```diff
@@ -375,3 +376,82 @@ func (r *ResourceManager) UpdatePrompt(ctx context.Context, name string, config
 	r.prompts[name] = p
 	return nil
 }
+
+// DeleteSource deletes a source.
+func (r *ResourceManager) DeleteSource(name string) bool {
```
averikitsch: nit: we could DRY this up like "func deleteResource[T any](m map[string]T, name
     string) bool"

internal/server/resources/resources.go
averikitsch, line 1: when/how do we update the MCPManifest and ToolsManifest?

internal/server/resources/resources.go
```diff
@@ -375,3 +376,82 @@ func (r *ResourceManager) UpdatePrompt(ctx context.Context, name string, config
...
+	defer r.mu.Unlock()
+	_, ok := r.tools[name]
+	if !ok {
+		return false
+	}
+	delete(r.tools, name)
+
+	// Also remove the tool from the default toolset
+	// This function do not remove tool from other toolset. User have to update
+	// toolset explicitly
```
averikitsch: Can we add this comment on the put PR?

internal/server/admin.go
```diff
@@ -36,10 +36,51 @@ func adminRouter(s *Server) (chi.Router, error) {
...
 }
 
+// deletePrimitives handles the deletion of primitives
+// Invalid primitive kind will result in http.StatusBadRequest
+// Primitive not found will result in http.StatusNotFound
+func deletePrimitives(s *Server, w http.ResponseWriter, r *http.Request) {
+	kind := chi.URLParam(r, "kind")
+	name := chi.URLParam(r, "name")
+	ctx := r.Context()
+	ctx = util.WithInstrumentation(ctx, s.instrumentation)
```
averikitsch: Do we want to pass this ctx?

internal/server/resources/resources.go
```diff
@@ -260,6 +260,7 @@ func (r *ResourceManager) UpdateEmbeddingModel(ctx context.Context, name string,
 	return nil
 }
 
+// removeFirstFromToolset removes a target tool from toolset
 func removeFirstFromToolset(ts tools.Toolset, target string) tools.Toolset {
 	for i, tool := range ts.Tools {
 		if tool != nil && (*tool) != nil && (*tool).McpManifest().Name == target {
```
averikitsch: Why do we need both "tool != nil && (*tool) ..."?

internal/server/resources/resources.go
Yuan325: I’ve removed the tool-deletion logic from the PUT PR, as it's unnecessary for updates. Since the toolset tracks tools by name regardless of content changes, a 'remove and re-add' cycle isn't required for an update to take effect.

internal/server/resources/resources.go
Yuan325: Updated the toolsets and promptsets to have a separate initialize function to re-initialize all the manifests. The tools manifests will run during their respectives `Initialize` function.

internal/server/resources/resources.go
Yuan325: Refactored to be simpler by updating the ToolsetConfig.ToolNames --> initialize Toolset.

PR #2879, opened 2026-03-28
feat: add apply subcommand

Yuan325:
**Description:** apply configuration to the server

To load or update configuration files, after making changes, users can run toolbox apply. Dynamic reloading capability within Toolbox will be triggered when toolbox apply is run.

```
toolbox apply --config my-tool-file.yaml # to load a file
toolbox apply --configs my-tool-file1.yaml -f my-tool-file2.yaml # to load multiple file
toolbox apply --config-folder config/ # to load a folder
```

If a user modifies their configuration file directly, it will not trigger dynamic reloading of the Toolbox server. This is different from the existing Toolbox implementation that will automatically update the server when changes are detected in the configuration file. With the control plane, users have to run `toolbox apply` after making changes in order for it to take effect.

The introduction of the `toolbox apply` command changes how server updates are handled; simply editing the configuration file will no longer cause the server to reload dynamically. Instead, users must now explicitly initiate the dynamic reloading process by executing `toolbox apply` after modifying their configuration. This new workflow ensures that server updates remain seamless and do not require a full restart.

**This subcommand will run the following:**
1. Retrieve the Current State
Before making any changes, the CLI should send a GET request to the active server's Admin API.
This retrieves the "current state," which is the configuration of all existing primitives currently running on the server.
Store this state in an in-memory dictionary or map, keyed by the primitive's Kind and Name.

2. Parse the Desired State
Read and parse the local YAML configuration files or directories passed via the toolbox apply -f command.
Consolidate all parsed primitives into a "desired state" map, also keyed by Kind and Name.

3. Calculate the Delta (The Union Strategy)
The design dictates that the CLI must correctly calculate the union of the newly applied files and the existing primitives on the server.
Iterate through the desired state and current state maps to categorize the required HTTP requests:
New primitives (PUT): If a primitive exists in the desired state (local files) but not in the current state (server), mark it for creation via a PUT request.
Updating primitives (PUT): If a primitive exists in both states, perform a deep comparison (such as a field-by-field check or a configuration hash). If the configuration has changed, mark it for an update via a PUT request.
Ignore updates (Prevent Unintended Deletions): If a primitive exists in the current state (server) but is missing from the desired state (local files), take no action. Do not mark it for deletion. Ignoring missing files allows users to apply a single new configuration file without wiping out the rest of the server's active primitives. All deletions should be strictly deferred to the explicit toolbox delete <kind> <name> command. This operational logic mirrors the Kubernetes apply subcommand; for scenarios where explicit deletion is required, users may specify a `--prune` flag.

4. Execute API Requests for Dynamic Reloading
Once the delta is calculated, act as an HTTP client and push the necessary PUT requests to the deployed server's endpoint. If there are multiple primitives being updated at once, always run the source before tools and toolsets.
To prevent full server restarts or downtime, the server-side logic must handle these incoming requests by dynamically reloading only the specific primitive being updated.


cmd/internal/apply/command.go
```diff
@@ -0,0 +1,271 @@
...
+			localJSON, err := json.Marshal(doc)
+			if err != nil {
+				return fmt.Errorf("failed to marshal local config for %s: %w", name, err)
+			}
+			remoteJSON, err := json.Marshal(remoteBody)
+			if err != nil {
+				return fmt.Errorf("failed to marshal remote config for %s: %w", name, err)
+			}
+
+			if bytes.Equal(localJSON, remoteJSON) {
```
averikitsch, line 255: This is brittle. We most likely need to use reflect.DeepEqual

cmd/internal/apply/command.go
```diff
@@ -0,0 +1,271 @@
...
+		bodyReader = bytes.NewReader(b)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, url, bodyReader)
+	if err != nil {
+		return nil, fmt.Errorf("request creation failed: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{
```
averikitsch, line 112: We should save this client instead of remaking it each time

cmd/internal/apply/command.go
```diff
@@ -0,0 +1,271 @@
...
+// using this type allow O(1) lookups and deletes
+type primitives map[string]map[string]struct{}
+
+func NewPrimitives() primitives {
+	return make(primitives)
+}
+
+// Helper to fetch and populate the map
+func (p primitives) Load(ctx context.Context, address string, port int) error {
+	kinds := []string{"source", "authservice", "embeddingmodel", "tool", "toolset", "prompt"}
```
averikitsch, line 64: Is there a way we dont need to hardcode these here? I worry long term about missing this

cmd/internal/apply/command.go
```diff
@@ -0,0 +1,271 @@
...
+			remoteJSON, err := json.Marshal(remoteBody)
+			if err != nil {
+				return fmt.Errorf("failed to marshal remote config for %s: %w", name, err)
+			}
+
+			if bytes.Equal(localJSON, remoteJSON) {
+				logger.DebugContext(ctx, "skipping: no changes detected", "kind", kind, "name", name)
+				continue
+			}
+			logger.DebugContext(ctx, "change detected, updating resource", "kind", kind, "name", name)
```
averikitsch, line 259: do we want to have a log for creating new resources?

cmd/internal/apply/command.go
```diff
@@ -0,0 +1,271 @@
...
+	url := fmt.Sprintf("http://%s:%d/admin/%s/%s", address, port, kind, name)
+
+	respBody, err := adminRequest(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var body map[string]any
+	if err := json.Unmarshal(respBody, &body); err != nil {
+		return nil, fmt.Errorf("erorr unmarshaling response body: %w", err)
```
averikitsch, line 158:
```suggestion
		return nil, fmt.Errorf("error unmarshaling response body: %w", err)
```

cmd/internal/apply/command.go
averikitsch, line 1: RE: "If there are multiple primitives being updated at once, always run the source before tools and toolset" - is this PR handling this?

cmd/internal/apply/command.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+			logger.DebugContext(ctx, "change detected, updating resource", "kind", kind, "name", name)
+		}
+
+		// TODO: check --prune flag: if prune, delete primitives that are left
+		// in the primitive list
+		// TODO: check for --dry-run flag.
+
+		if err := applyPrimitive(ctx, address, port, kind, name, doc); err != nil {
+			return err
```
averikitsch, line 267: We should probably only fail at the end

averikitsch (changes requested):

Issue #2415, opened 2026-02-04
Support all vector databases


averikitsch:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

We need to ensure that all vector-enable databases support `embeddedBy:`. Currently the blog only mentions PostgreSQL being supported.

### Suggested Solution(s)

Add tests for all vector-enabled databases and ensure the vector is formatted and inserted correctly.

## **Databases Supporting Vectors**

According to the source documentation and the nature of these integrations for GenAI, the following sources specifically support vector search or storage:

* **AlloyDB for PostgreSQL:** Includes built-in support via the google\_ml\_integration and pgvector extensions for high-performance vector search.  
* **Cloud SQL for PostgreSQL:** Supports vector storage and search through the pgvector extension.  
* **PostgreSQL:** Standard PostgreSQL support via the pgvector extension.  
* **Spanner:** Supports vector search using the VECTOR\_COSINE\_DISTANCE and VECTOR\_L2\_DISTANCE functions and vector indexing.  
* **BigQuery:** Supports vector search via the VECTOR\_SEARCH function and vector indexes.  
* **Firestore:** Supports vector search and K-nearest neighbor (KNN) queries on document fields.  
* **MongoDB:** Supports vector search through MongoDB Atlas Vector Search.  
* **Elasticsearch:** A native vector database supporting dense and sparse vector types.  
* **Neo4j:** Supports vector indexing and search via Cypher procedures.  
* **SingleStore:** Built-in vector data types and functions for dot product and Euclidean distance.  
* **Redis / Valkey:** Supports vector search using the RediSearch module (HNSW/Flat indexing).  
* **ClickHouse:** Supports vector search using specialized distance functions and experimental ANN (Approximate Nearest Neighbor) indexes.  
* **Cassandra:** Supports vector data types (since v5.0) and SAI (Storage-Attached Indexing).

---

## **Requirements for Enabling Vectors**

To use vectors with these sources within the GenAI Toolbox, the following general and specific requirements must be met:

#### **1\. Database-Specific Requirements**

* **PostgreSQL-based (AlloyDB, Cloud SQL, Self-hosted):**  
  * **Extension:** You must enable the pgvector extension by running CREATE EXTENSION IF NOT EXISTS vector; in your database.  
  * **Column Type:** Use the VECTOR(dimensions) data type for your embedding columns.  
[...]

PR #2889, opened 2026-03-30
test(spannersql): add semantic search integration tests and docs

duwenxin99:
Part of https://github.com/googleapis/genai-toolbox/issues/2415


tests/spanner/spanner_integration_test.go
averikitsch, line 1: Can we test against both googlesql and postgresql sources?

duwenxin99: added tests.

PR #2906, opened 2026-03-30
feat: add GET endpoint by primitive kind and name

Yuan325:
New GET endpoint for `/admin/{kind}/{name}`.

Created a new `type Secret string` type for all the secret value. Sensitive information with this type will not be shown with the GET endpoint. It will be replaced with`***`.

### DIFF
--- a/internal/server/admin.go
+++ b/internal/server/admin.go
@@ -15,9 +15,16 @@
 package server
 
 import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/render"
+	"github.com/googleapis/genai-toolbox/internal/util"
 )
 
 // adminRouter creates a router that represents the routes under /admin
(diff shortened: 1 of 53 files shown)


internal/server/admin.go
```diff
@@ -165,3 +166,77 @@ func createOrUpdatePrimitives(s *Server, w http.ResponseWriter, r *http.Request)
 		return
 	}
 }
+
+func getPrimitiveByName(s *Server, w http.ResponseWriter, r *http.Request) {
+	kind := chi.URLParam(r, "kind")
+	name := chi.URLParam(r, "name")
+	ctx := r.Context()
```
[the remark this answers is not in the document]
Yuan325: Not needed since ctx are not required in the function that is called within this endpoint.

internal/util/util.go
```diff
@@ -214,3 +214,19 @@ func GenAIMetricAttrsFromContext(ctx context.Context) *GenAIMetricAttrs {
 	}
 	return nil
 }
+
+type Secret string
+
+// MarshalJSON masks the value when converting to JSON
+func (s Secret) MarshalJSON() ([]byte, error) {
```
duwenxin99, line 221: We should create a new file util_test.go and add unit tests. admin_test only tests for yaml and not json marshaler.

duwenxin99 (changes requested):

internal/server/admin.go
```diff
@@ -165,3 +166,77 @@ func createOrUpdatePrimitives(s *Server, w http.ResponseWriter, r *http.Request)
...
+
+	switch strings.ToLower(kind) {
+	case "source":
```
averikitsch, line 176: nit: this could be refactor to not be so repetitive

averikitsch (commented):
where are we worried about secrets being exposed in API responses, in logs, etc?

averikitsch:
> where are we worried about secrets being exposed in API responses, in logs, etc?
We may be missing:
   * Auth Services Ignored: You completely missed internal/auth. ClientSecret, PersonalAccessToken, and generic APIKeys are still string and will be returned in plaintext.
   * Embedding Models Ignored: internal/embeddingmodels/gemini/gemini.go uses ApiKey string. This will be leaked.
   * MongoDB Ignored: MongoDB uses a connection string URI (mongodb://user:password@host). Because this is a single string field, the password is leaked.
   * HTTP Source Ignored: The HTTP source stores credentials in DefaultHeaders and QueryParams (both map[string]string). Your util.Secret wrapper only works on individual struct properties, meaning Bearer tokens
     and API keys in headers will be returned unmasked.

internal/util/util.go
```diff
@@ -214,3 +214,19 @@ func GenAIMetricAttrsFromContext(ctx context.Context) *GenAIMetricAttrs {
...
+	return []byte("\"***\""), nil
+}
+
+// MarshalYAML masks the value when converting to YAML
+func (s Secret) MarshalYAML() (interface{}, error) {
+	return "***", nil
+}
+
+func (s Secret) String() string {
```
averikitsch, line 230: nit: maybe unmask() and use strings to return *** just incase?

averikitsch (changes requested):
If we pass back "***" and the users

PR #2907, opened 2026-03-31
feat: add GET endpoint by primitive kind

Yuan325:
This will be the path to get a list of primitives that are configured within Toolbox. 
For example, /admin/source will return a list of sources. The available api includes the following (non-case sensitive):

* /admin/source
* /admin/authService or /admin/authservice
* /admin/tool
* /admin/toolset

An invalid primitive name will result in a bad request error.

### DIFF
--- a/internal/server/admin.go
+++ b/internal/server/admin.go
@@ -37,7 +37,10 @@ func adminRouter(s *Server) (chi.Router, error) {
 
 	r.Put("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { createOrUpdatePrimitives(s, w, r) })
 	r.Delete("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { deletePrimitives(s, w, r) })
-	r.Get("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { getPrimitiveByName(s, w, r) })
+	r.Route("/{kind}", func(r chi.Router) {
+		r.Get("/", func(w http.ResponseWriter, r *http.Request) { getPrimitive(s, w, r) })
+		r.Get("/{name}", func(w http.ResponseWriter, r *http.Request) { getPrimitiveByName(s, w, r) })
+	})
 
 	return r, nil
 }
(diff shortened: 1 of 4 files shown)


duwenxin99 (approved):

duwenxin99 (approved):

internal/server/resources/resources.go
```diff
@@ -76,6 +78,42 @@ func (r *ResourceManager) SetResources(sourcesMap map[string]sources.Source, aut
 	r.promptsets = promptsetsMap
 }
 
+func (r *ResourceManager) GetSources() []string {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return slices.Sorted(maps.Keys(r.sources))
```
averikitsch, line 84: this may return nil if sorting empty list. we should probably ensure we are returning an empty list in that case.

internal/server/admin.go
```diff
@@ -37,7 +37,10 @@ func adminRouter(s *Server) (chi.Router, error) {
 
 	r.Put("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { createOrUpdatePrimitives(s, w, r) })
 	r.Delete("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { deletePrimitives(s, w, r) })
-	r.Get("/{kind}/{name}", func(w http.ResponseWriter, r *http.Request) { getPrimitiveByName(s, w, r) })
+	r.Route("/{kind}", func(r chi.Router) {
```
averikitsch, line 40: Any reason not to stack these like
   1     r.Route("/{kind}", func(r chi.Router) {
   2         r.Get("/", ...)
   3         r.Get("/{name}", ...)
   4         r.Put("/{name}", ...)
   5         r.Delete("/{name}", ...)
   6     })

internal/server/resources/resources.go
```diff
@@ -76,6 +78,42 @@ func (r *ResourceManager) SetResources(sourcesMap map[string]sources.Source, aut
...
+	return slices.Sorted(maps.Keys(r.tools))
+}
+
+func (r *ResourceManager) GetToolsets() []string {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return slices.Sorted(maps.Keys(r.toolsets))
+}
+
+func (r *ResourceManager) GetPrompts() []string {
```
averikitsch, line 111: Remind me why we do prompts and not promptset as well?

averikitsch (changes requested):

PR #2912, opened 2026-03-31
feat: add dry-run and prune flag for apply command

Yuan325:
**New flags that will be introduced in this subcommand:**
The `--dry-run` Flag: Before sending GET/PUT/DELETE requests to the Admin API, allow users to preview the execution plan. This will print exactly which primitives will be created or updated, acting as a safeguard against misconfigurations.
The `--prune` Flag: Users must explicitly pass --prune to tell the CLI to delete missing resources.

### DIFF
--- a/cmd/internal/apply/command.go
+++ b/cmd/internal/apply/command.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/goccy/go-yaml"
 	"github.com/googleapis/genai-toolbox/cmd/internal"
-	"github.com/googleapis/genai-toolbox/internal/log"
 	"github.com/spf13/cobra"
 )
 
@@ -35,6 +34,8 @@ type applyCmd struct {
 	*cobra.Command
 	port    int
 	address string
+	dryRun  bool
+	prune   bool
 }
 
 func NewCommand(opts *internal.ToolboxOptions) *cobra.Command {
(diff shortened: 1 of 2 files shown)


cmd/internal/apply/command.go
```diff
@@ -196,18 +220,36 @@ func runApply(cmd *applyCmd, opts *internal.ToolboxOptions) error {
...
 	opts.Logger.InfoContext(ctx, "starting apply sequence", "count", len(filePaths))
 	for _, filePath := range filePaths {
-		if err := processFile(ctx, opts.Logger, filePath, p, cmd.address, cmd.port); err != nil {
+		if err := processFile(ctx, opts, filePath, p, cmd); err != nil {
 			opts.Logger.ErrorContext(ctx, err.Error())
 			return err
 		}
 	}
+	if cmd.prune {
+		remaining := p.GetRemaining()
```
averikitsch, line 234: nit: if we need this outptu to be deterministic we will need to sort first

cmd/internal/apply/command.go
```diff
@@ -253,17 +295,17 @@ func processFile(ctx context.Context, logger log.Logger, path string, p primitiv
...
+			opts.Logger.DebugContext(ctx, "change detected, updating resource", "kind", kind, "name", name)
 		}
 
-		// TODO: check --prune flag: if prune, delete primitives that are left
-		// in the primitive list
-		// TODO: check for --dry-run flag.
-
-		if err := applyPrimitive(ctx, address, port, kind, name, doc); err != nil {
+		if cmd.dryRun {
+			fmt.Fprintf(opts.IOStreams.Out, "[dry-run] would update %s: %s\n", kind, name)
```
averikitsch, line 305: Do we have the ability to distinguish update from  "[dry-run] would create"?

cmd/internal/apply/command.go
```diff
@@ -196,18 +220,36 @@ func runApply(cmd *applyCmd, opts *internal.ToolboxOptions) error {
...
+		opts.Logger.InfoContext(ctx, "starting prune sequence", "count", len(remaining))
+		for _, resource := range remaining {
+			if cmd.dryRun {
+				fmt.Fprintf(opts.IOStreams.Out, "[dry-run] would delete %s: %s\n", resource.kind, resource.name)
+				continue
+			}
+
+			// Perform the actual deletion
+			if err := deletePrimitive(ctx, cmd.address, cmd.port, resource.kind, resource.name); err != nil {
+				return fmt.Errorf("failed to prune %s/%s: %w", resource.kind, resource.name, err)
```
averikitsch, line 244: I'm not sure we should fail fast here ie return right away. We probably want to delete everything possible then fail.

PR #2921, opened 2026-04-01
test(source/alloydbainl): create MCP integration tests

anubhav756:
This PR adds the mapped integration tests for AlloyDB AINL tools using the native MCP harness.

### DIFF
--- a/tests/alloydbainl/alloydb_ai_nl_mcp_test.go
+++ b/tests/alloydbainl/alloydb_ai_nl_mcp_test.go
@@ -47,29 +47,57 @@ func TestAlloyDBAINLListTools(t *testing.T) {
 		t.Fatalf("toolbox didn't start successfully: %s", err)
 	}
 
-	statusCode, toolsList, err := tests.GetMCPToolsList(t, nil)
-	if err != nil {
-		t.Fatalf("native error executing tools/list: %s", err)
-	}
-	if statusCode != http.StatusOK {
-		t.Fatalf("expected status 200, got %d", statusCode)
+	// Verify list of tools
+	expectedTools := []tests.MCPToolManifest{
+		{
+			Name:        "my-simple-tool",
+			Description: "Simple tool to test end to end functionality.",
+			InputSchema: map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"question": map[string]any{
+						"description": "The natural language question to ask.",
+						"type":        "string",
+					},
+				},
+				"required": []any{"question"},
+			},
+		},
+		{
+			Name:        "my-auth-tool",
+			Description: "Tool to test authenticated parameters.",
+			InputSchema: map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"question": map[string]any{
+						"description": "The natural language question to ask.",
+						"type":        "string",
+					},
+					"email": map[string]any{
+						"description": "user email",
+						"type":        "string",
+					},
+				},
+				"required": []any{"question", "email"},
+			},
+		},
+		{
+			Name:        "my-auth-required-tool",
+			Description: "Tool to test auth required invocation.",
+			InputSchema: map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"question": map[string]any{
+						"description": "The natural language question to ask.",
+						"type":        "string",
+					},
+				},
+				"required": []any{"question"},
+			},
+		},
 	}
 
-	// Verify that my-simple-tool is in the list
-	found := false
-	for _, tool := range toolsList {
-		toolMap, ok := tool.(map[string]any)
-		if !ok {
-			continue
-		}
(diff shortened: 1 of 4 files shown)


tests/alloydbainl/alloydb_ai_nl_mcp_test.go
```diff
@@ -0,0 +1,269 @@
...
+			"my-instance": sourceConfig,
+		},
+		"authServices": map[string]any{
+			"my-google-auth": map[string]any{
+				"type":     "google",
+				"clientId": tests.ClientId,
+			},
+		},
+		"tools": map[string]any{
+			"my-simple-tool": map[string]any{
```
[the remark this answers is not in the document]
anubhav756: The existing legacy integration tests use kebab-case for these mock test tools. To maintain consistency with the rest of the test suite, I've left them as kebab-case for now.

tests/alloydbainl/alloydb_ai_nl_mcp_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+			if tc.isErr {
+				if !mcpResp.Result.IsError {
+					t.Fatalf("expected error result, got success")
+				}
+			} else {
+				if mcpResp.Result.IsError {
+					t.Fatalf("expected success result, got error: %v", mcpResp.Result)
+				}
+				got := mcpResp.Result.Content[0].Text
```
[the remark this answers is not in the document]
anubhav756: Done

tests/alloydbainl/alloydb_ai_nl_mcp_test.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	runAINLToolGetMCPTest(t)
+	runAINLToolInvokeMCPTest(t)
+}
+
+func runAINLToolGetMCPTest(t *testing.T) {
```
Yuan325: can we do `TestAlloyDBAINLListTools()` and similar with the other tools?

Similarly, we can have the `TestAlloyDBAINLCallTool()` as well with both the `runAINLToolInvokeTest(t)` and `runAINLMCPToolCallMethod(t)` test cases.

anubhav756:
> Let's try to keep all the mcp int test update to the 2 method-based functions. :)

That's a good idea. Thanks!

tests/alloydbainl/alloydb_ai_nl_mcp_test.go
anubhav756: That would be neat! Thanks for the suggestion :)

tests/alloydbainl/alloydb_ai_nl_mcp_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	statusCode, toolsList, err := tests.GetMCPToolsList(t, nil)
+	if err != nil {
+		t.Fatalf("native error executing tools/list: %s", err)
+	}
+	if statusCode != http.StatusOK {
+		t.Fatalf("expected status 200, got %d", statusCode)
+	}
+
+	// Verify that my-simple-tool is in the list
```
Yuan325: Can we check the full list of tools just like how we did for `alloydb`? 

Moving forward, this check would be nice to also confirm all the parameters that are created during the tool initialization. (i remember we had a bug earlier regarding parameter that was spelled wrongly and we didn't catch that since we did not check the parameters in our integration tests).

Yuan325 (changes requested):
Quick feedback on the list tools test, otherwise LGTM! :)

tests/alloydbainl/alloydb_ai_nl_mcp_test.go
anubhav756: That's a great suggestion! :) Updated! Now using `RunMCPToolsListMethod`, which checks the full list of tools and verifies the entire `InputSchema` against what we expect.

I also added the length check to ensure only the expected tools are present, in `RunMCPToolsListMethod`.

anubhav756:
> Is this PR supposed to contain changes from the alloydb admin tests and https tests?

The migrations were done in https://github.com/googleapis/mcp-toolbox/pull/2842 and https://github.com/googleapis/mcp-toolbox/pull/2843 for http and alloydb respectively. However, post their merge, in this PR, specifically around [this comment](https://github.com/googleapis/mcp-toolbox/pull/2921#discussion_r3055510515), we realised we can improve the common harness that was being used in this PR and also in the http and alloydb tests as well.

Earlier this method was verifying just one tool out of a list of tools. But we've made it stricter to check all the required tools, and verifies if only those tools are present. But for this change, assertions of those previous tests had to be updated as well. So this PR adds the full list of tools in their assertions.

PR #2922, opened 2026-04-01
test(source/cloud-sql-mysql): create MCP integration tests

dishaprakash:
## Description

This PR migrates the Cloud SQL MySQL integration tests to use the /mcp endpoint on the MCP Toolbox Server.

### Key Changes

- `tests/mcp_tool.go`: 
  - Added modular builder functions (Base, ExecuteSQL, TemplateParams) to dynamically generate expected MCP manifests.
  - Added concatenation of individual MCP text blocks into a fully formatted JSON array whenever the test expects one, preserving compatibility with previous test assertions.

- `cloud_sql_mysql_mcp_test.go`: Migrated legacy `/api` tests to `/mcp`

- `cloud_sql_mysql_create_instance_mcp_test.go`: Migrated create-instance tests to validate MCP JSON-RPC handling.

- `tests/options.go`: Added functional options for the various MySQL test fixtures to provide an MCP tests specific path

- `tests/tool.go`: Used the above mentioned options to allow calling MCP Endpoints in the test fixtures of MySQL


tests/cloudsqlmysql/cloud_sql_mysql_mcp_test.go
```diff
@@ -0,0 +1,250 @@
...
+		got := mcpResp.Result.Content[0].Text
+		if !strings.Contains(got, tableNameParam) || !strings.Contains(got, tableNameAuth) {
+			t.Errorf("list_tables missing expected tables. Got: %s", got)
+		}
+
+		go func() {
+			_ = pool.PingContext(ctx)
+			_, _ = pool.ExecContext(ctx, "SELECT sleep(5);")
+		}()
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
dishaprakash: done

tests/mcp_tool.go
```diff
@@ -177,7 +177,24 @@ func RunMCPCustomToolCallMethod(t *testing.T, toolName string, arguments map[str
...
+	}
+
+	var got string
+	// If the test is expecting a JSON array, format the stitched blocks as a JSON array
+	if want != "" && strings.HasPrefix(strings.TrimSpace(want), "[") {
+		got = "[" + strings.Join(blocks, ",") + "]"
+	} else {
+		// Otherwise, just concatenate them normally (for single responses or errors)
+		got = strings.Join(blocks, "")
+	}
```
Yuan325: I think anubhav created a function to convert it to []any in #2992. We might be able to use that, then marshal the result into string for comparison here.

tests/mcp_tool.go
```diff
@@ -191,7 +208,7 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
 		myToolId3NameAliceWant:   "[{\"id\":1,\"name\":\"Alice\"},{\"id\":3,\"name\":\"Sid\"}]",
 		myToolById4Want:          "[{\"id\":4,\"name\":null}]",
 		myArrayToolWant:          "[{\"id\":1,\"name\":\"Alice\"},{\"id\":3,\"name\":\"Sid\"}]",
-		nullWant:                 "null",
+		nullWant:                 "",
```
Yuan325: Might need to double check but I vaguely remember we intentionally add these  to cover cases where it returns the `"null"` string specifically. is the response different now with MCP? or will it response with `["null"]`?

tests/mcp_tool.go
```diff
@@ -276,12 +293,214 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
 				t.Fatalf("%s returned error result: %v", tc.toolName, mcpResp.Result)
 			}
 			if len(mcpResp.Result.Content) == 0 {
+				if tc.wantResult == "null" || tc.wantResult == "[]" || tc.wantResult == "" {
```
Yuan325: Just wanna double check if this is expected in MCP?

tests/mcp_tool.go
```diff
@@ -276,12 +293,214 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
...
+			}
+
+			var got string
+			// If the test is expecting a JSON array, format the stitched blocks as a JSON array
+			if tc.wantResult != "" && strings.HasPrefix(strings.TrimSpace(tc.wantResult), "[") {
+				got = "[" + strings.Join(blocks, ",") + "]"
+			} else {
+				// Otherwise, just concatenate them normally (for single responses or errors)
+				got = strings.Join(blocks, "")
+			}
```
Yuan325: Might be able to utilize anubhav's function here too~ with `json.Marshal`

tests/mcp_tool.go
```diff
@@ -276,12 +293,214 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
...
+
 			if !strings.Contains(got, tc.wantResult) {
 				t.Fatalf(`expected %q to contain %q`, got, tc.wantResult)
 			}
 		})
 	}
 }
+
+// GetBaseMCPExpectedTools returns the MCP manifests for the base tools loaded by GetToolsConfig.
+func GetBaseMCPExpectedTools() []MCPToolManifest {
```
Yuan325, line 394: I like this, we can use this moving forward~ alot of the source might only consist of the BaseExepctedTools.

tests/tool.go
```diff
@@ -420,12 +420,79 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
 				return
 			}
-			// Send Tool invocation request
-			resp, respBody := RunRequest(t, http.MethodPost, tc.api, tc.requestBody, tc.requestHeader)
+
+			reqBytes, _ := io.ReadAll(tc.requestBody)
+			var got string
+			var actualStatusCode int
+
+			if configs.IsMCP {
```
Yuan325: This is a good idea. 

Wondering if we can update instead of `api string`, we can do `toolName string`. And then build the `/api`'s url with `fmt.Sprintf("http://127.0.0.1:5000/api/tool/%s/invoke", tc.toolName)`.

With this, we won't need to split the `tc.api` string.

Yuan325: Similarly for the `args`. Can we update the format to be `map[string]any` and then for `/api` endpoint, we can just marshal it to string? Might make things less error-prone.

tests/tool.go
```diff
@@ -420,12 +420,79 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
+				if args == nil {
+					args = make(map[string]any)
+				}
+
+				// Invoke the tool via MCP protocol
+				mcpStatusCode, mcpResp, _ := InvokeMCPTool(t, toolName, args, tc.requestHeader)
+				actualStatusCode = mcpStatusCode
+
+				// Translate MCP response back to string for unified assertion
+				if actualStatusCode == http.StatusOK && tc.wantBody != "" {
```
Yuan325: this feels like a lot of if-else, can we use similar comparison logic in `mcp_tool.go`? ref: https://github.com/googleapis/mcp-toolbox/blob/49e097df06711abbd1bb993a78023593ed824bc6/tests/mcp_tool.go#L341

I think it will be okay to completely separate the MCP vs API endpoint checks just with a single if-else and have some code duplication within it. It will make future removal easier too.

tests/tool.go
```diff
@@ -420,12 +420,79 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
+							}
+						}
+						if strings.HasPrefix(strings.TrimSpace(tc.wantBody), "[") {
+							got = "[" + strings.Join(blocks, ",") + "]"
+						} else if len(blocks) == 0 {
+							got = "null"
+						} else {
+							got = strings.Join(blocks, "")
+						}
+					}
```
Yuan325: we can use anubhav's function for this. If you want to prevent merge issues, feel free to copy the function over to this PR, during merge, git should catch that its the same code.

Yuan325: Another thing that differs between /mcp and /api is the `content.IsError`. for some errors, mcp will parse it as a successful request but with IsError = true and the content.Text will have the error message.

Yuan325 (changes requested):
I like the adding "options" method and reusing the functions. That way we don't need a lot of duplicated code. If we can use that moving forward that'll be great!

tests/mcp_tool.go
dishaprakash: Rebased onto the alloydb pg tests branch and reused the new mcp harness!

tests/mcp_tool.go
dishaprakash: Rebased onto the alloydb pg tests branch and reused the new mcp harness!

tests/mcp_tool.go
dishaprakash: Rebased onto the alloydb pg tests branch and reused the new mcp harness!

tests/tool.go
dishaprakash: You're right! I've simplified it as you've mentioned!

tests/tool.go
dishaprakash: Done!

tests/tool.go
dishaprakash: Got it! Added that case and reused the mcp testing harness

tests/tool.go
```diff
@@ -420,33 +419,125 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
-			if !ok {
-				t.Fatalf("unable to find result in response body")
-			}
+				// Extract error text if any
+				var errText string
+				if mcpResp.Error != nil {
+					errText = mcpResp.Error.Message
+				} else if mcpResp.Result.IsError {
+					for _, content := range mcpResp.Result.Content {
+						if content.Type == "text" {
```
Yuan325: probably don't need to check this since we only support text.

tests/tool.go
```diff
@@ -420,33 +419,125 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
 
-			// Check response body
-			var body map[string]interface{}
-			err = json.Unmarshal(respBody, &body)
-			if err != nil {
-				t.Fatalf("error parsing response body: %s", err)
-			}
+				if tc.wantBody == "" {
+					return
+				}
```
Yuan325: for mcp's cases, we might not want to skip these? seems like these are the ones that have `content.IsError = true` and might have a `content.Text` value? if needed we can add more field to the test cases struct such as `wantContentErr string` and set those as the string value.

tests/option.go
```diff
@@ -28,10 +28,18 @@ type InvokeTestConfig struct {
 	supportArrayParam        bool
 	supportClientAuth        bool
 	supportSelect1Auth       bool
+	IsMCP                    bool
```
Yuan325:
```suggestion
	isMCP                    bool
```

shouldn't need to be exported. User should always set it with the function.

tests/tool.go
```diff
@@ -420,33 +419,125 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
-			}
+				// Check status code
+				wantStatus := tc.wantStatusCode
+				// MCP might return 200 OK for some error cases that REST returns 401
+				if wantStatus == http.StatusUnauthorized && mcpStatusCode == http.StatusOK {
+					wantStatus = http.StatusOK
+				}
+				if mcpStatusCode != wantStatus {
+					t.Errorf("StatusCode mismatch: got %d, want %d", mcpStatusCode, wantStatus)
+				}
```
Yuan325: Seems like the mcp status is always StatusOK. Lets just check if its statusOK. If there are test cases thats not statusOK, we can add a `tc.wantStatusCodeMCP`.

```suggestion
				if mcpStatusCode != http.StatusOK {
					t.Error("expected status ok")
				}
```

tests/tool.go
```diff
@@ -420,33 +419,125 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
-				t.Fatalf("unexpected value: got %q, want %q", got, tc.wantBody)
+					if diff := cmp.Diff(wantJSON, gotJSON); diff != "" {
+						t.Fatalf("unexpected JSON value mismatch (-want +got):\n%s\nRaw got: %s\nRaw want: %s", diff, gotStr, tc.wantBody)
+					}
+				} else {
+					// Plain string, use strings.Contains as suggested by user
+					if !strings.Contains(gotStr, tc.wantBody) {
+						t.Fatalf(`expected %q to contain %q`, gotStr, tc.wantBody)
+					}
+				}
```
Yuan325, line 493: This comparison should be the same as the api endpoint? since they compare direct `got != tc.wantBody`, we should be able to do the same here too? rather than converting it to type any.

tests/tool.go
```diff
@@ -420,33 +419,125 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
-			}
+				if tc.wantBody == "" {
+					return
+				}
 
-			got, ok := body["result"].(string)
-			if !ok {
-				t.Fatalf("unable to find result in response body")
-			}
+				// Extract error text if any
```
Yuan325: For the checks under here, we can check the Content.Text together regardless if it's "IsError". 

E.g. 
```
// first parse the content.Text to string
// next check if isError is correct
isError := tc.wantContentErr != ""
if isError != mcpResp.Result.IsError {
    // throw t.Fatalf here
}

// check the content.Text string.
if isError {
    // check content.Text with tc.wantContentErr
} else {
    // check content.Text with tc.wantBody
}
```

tests/tool.go
```diff
@@ -420,33 +419,125 @@ func RunToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTestOp
...
 				return
 			}
-			// Send Tool invocation request
-			resp, respBody := RunRequest(t, http.MethodPost, tc.api, tc.requestBody, tc.requestHeader)
 
-			// Check status code
-			if resp.StatusCode != tc.wantStatusCode {
-				t.Errorf("StatusCode mismatch: got %d, want %d. Response body: %s", resp.StatusCode, tc.wantStatusCode, string(respBody))
-			}
+			if configs.IsMCP {
```
Yuan325: did a quick draft -- 
we'll need to add `wantContentErr string` to the test case struct.

```
if configs.IsMCP {
				// Invoke the tool via MCP protocol
				mcpStatusCode, mcpResp, err := InvokeMCPTool(t, tc.toolName, tc.args, tc.requestHeader)
				if err != nil {
					t.Fatalf("native error executing %s: %s", tc.toolName, err)
				}
				if mcpStatusCode != http.StatusOK {
					t.Fatalf("expected status ok")
				}

				if tc.wantContentErr != "" {
					AssertMCPError(t, mcpResp, tc.wantContentErr)
					return
				}
				if mcpResp.Result.IsError {
					t.Fatalf("%s returned error result: %v", tc.toolName, mcpResp.Result)
				}
				gotObj := getMCPResultText(t, mcpResp)
				gotBytes, err := json.Marshal(gotObj)
				if err != nil {
					t.Fatalf("error marshaling result object")
				}
				if string(gotBytes) != tc.wantBody {
					var gotJSON, wantJSON any
					errGot := json.Unmarshal(gotBytes, &gotJSON)
					errWant := json.Unmarshal([]byte(tc.wantBody), &wantJSON)

					if errGot == nil && errWant == nil {
						if diff := cmp.Diff(wantJSON, gotJSON); diff != "" {
							t.Fatalf("unexpected JSON value mismatch (-want +got):\n%s\nRaw got: %s\nRaw want: %s", diff, got, tc.wantBody)
						}
					} else {
						t.Fatalf("unexpected value: got %q, want %q", got, tc.wantBody)
					}
				}
			} else {
```

Yuan325 (approved):

tests/cloudsqlmysql/cloud_sql_mysql_mcp_test.go
```diff
@@ -0,0 +1,226 @@
...
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	select1Want, mcpMyFailToolWant, createTableStatement, mcpSelect1Want := tests.GetMySQLWants()
+
+	tests.RunToolInvokeTest(t, select1Want, tests.DisableArrayTest(), tests.WithMCP(), tests.WithNullWant("[]"))
+	tests.RunMCPToolCallMethod(t, mcpMyFailToolWant, mcpSelect1Want)
```
Yuan325, line 217: final nit, do we still need this? Might be able to remove it since we have the `RunToolInvokeTest()` that's refactored to MCP version.

Might be able to remove it from the non-mcp int test file (`cloud_sql_mysql_test.go`) as well~

PR #2924, opened 2026-04-01
refactor!: update gcs bucket name for binary

Yuan325:
Update GCS bucket name to `mcp-toolbox-for-databases` and `mcp-toolbox-for-databases-dev`.

TBD: pre-v1 binaries will remain in the OG `genai-toolbox` bucket.

### DIFF
--- a/.ci/continuous.release.cloudbuild.yaml
+++ b/.ci/continuous.release.cloudbuild.yaml
@@ -91,6 +91,7 @@ steps:
     script: |
         #!/usr/bin/env bash
         gcloud storage cp toolbox.linux.amd64 gs://$_BUCKET_NAME/$REF_NAME/linux/amd64/toolbox
+        gcloud storage cp toolbox.linux.amd64 gs://$_OLD_BUCKET_NAME/$REF_NAME/linux/amd64/toolbox
 
   - id: "build-linux-amd64-geminicli"
     name: golang:1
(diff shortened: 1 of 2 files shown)


.ci/sample_tests/run_tests.sh
```diff
@@ -66,7 +66,7 @@ setup_toolbox() {
   echo "Setting up Toolbox server..."
   TOOLBOX_YAML="/tools.yaml"
   echo "${TOOLS_YAML_CONTENT}" > "$TOOLBOX_YAML"
-  wget -q "https://storage.googleapis.com/genai-toolbox/v${VERSION}/linux/amd64/toolbox" -O "/toolbox"
+  wget -q "https://storage.googleapis.com/mcp-toolbox-for-databases/v${VERSION}/linux/amd64/toolbox" -O "/toolbox"
```
averikitsch, line 69: Do we have an estimate on the timeline for the new bucket exemption approval? We may need to continue to support both buckets just in case people have built automation around this

.ci/continuous.release.cloudbuild.yaml
```diff
@@ -291,7 +291,7 @@ steps:
     script: |
         #!/usr/bin/env bash
         export VERSION=$(cat ./cmd/version.txt)
-        go build -ldflags "-X github.com/googleapis/genai-toolbox/cmd.commitSha=$(git rev-parse --short HEAD)" -o toolbox.geminicli.windows.amd64
+        go build -ldflags "-X github.com/googleapis/mcp-toolbox/cmd.commitSha=$(git rev-parse --short HEAD)" -o toolbox.geminicli.windows.amd64
```
averikitsch, line 298: Depending on timeline, we may want to switch over the build flags before the bucket. Though GitHub should auto redirect for us.

.ci/sample_tests/run_tests.sh
Yuan325: It's already approved :)

Yuan325: I'll update to support both buckets for now 👍

.ci/continuous.release.cloudbuild.yaml
Yuan325: Updated the build flags. The repo name will be updated before the release. Kept both bucket and run release to both buckets instead until post-v1 release.

PR #2936, opened 2026-04-02
refactor!: update docsite with repo name change

Yuan325:
Update all the `googleapis.github.io/genai-toolbox/*` docsite.
Update files within `.hugo`.


.hugo/hugo.toml
averikitsch, line 1: @dishaprakash to review

dishaprakash (approved):
This would cause a breaking link, as the previous version's documentations would still point at the `genai-toolbox` repo.

@Yuan325 @averikitsch is that okay for us to let the previous version links in the docsite break (and add a P1 task into the repo name change task list to update it to the new link)?

Issue #2872, opened 2026-03-27
MCP Connection is not working in Claude for Toolbox


cb-salamonm:
<img width="1044" height="283" alt="Image" src="https://github.com/user-attachments/assets/11d30358-03d2-4a3c-b21d-0eedfb7ed900" />

### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### Question

I have deploy the application in gcloud run and when i try to add the mcp server url(cloud run app url) in claude and when i try to connect, its not working. Is it supported to connect in claude ? 

<img width="1044" height="283" alt="Image" src="https://github.com/user-attachments/assets/84c98dbc-2f4c-484d-8a09-24875395c043" />

### Code

kind: authService
name: my-mcp-auth
type: generic
mcpEnabled: true
authorizationServer: "https://accounts.google.com" 
audience: "****"
scopesRequired:
    - "email"

cb-salamonm:
@Yuan325 I deployed the application in Google Cloud Run and if i give requires permission and set IAP, its not working. When i try to connect with Claude it say 404 page not found. Can you please help me to resolve the issue

duwenxin99:
Hi @cb-salamonm, thanks for opening the issue. Let's keep communicating over discord and I'll update our resolution here.

PR #2944, opened 2026-04-03
feat(auth): support opaque token validation for `generic` authService

duwenxin99:
Add opaque token support to MCP auth to generic authService
- Add opaque token validator
- Add unit and integration tests
- Improve doc to outline the two different usages of generic authService

ref: [OAuth Token Introspection](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2)

fix: https://github.com/googleapis/genai-toolbox/issues/2872

### DIFF
--- a/docs/en/documentation/configuration/authentication/generic.md
+++ b/docs/en/documentation/configuration/authentication/generic.md
@@ -104,9 +104,10 @@ When a request is received in this mode, the service will:
    - Verifies expiration (`exp`) and audience (`aud`).
    - Verifies required scopes in `scope` claim.
 4. For **Opaque Tokens**:
-   - Calls the introspection endpoint (`<authorizationServer>/introspect`).
+   - Calls the introspection endpoint (as listed in the `authorizationServer`'s
+     OIDC configuration).
    - Verifies that the token is `active`.
-   - Verifies expiration (`exp`) and audience (`client_id`).
+   - Verifies expiration (`exp`) and audience (`aud`).
    - Verifies required scopes in `scope` field.
 
 #### Example


internal/auth/generic/generic.go
```diff
@@ -86,20 +103,7 @@ func discoverJWKSURL(AuthorizationServer string) (string, error) {
...
-			IdleConnTimeout:       90 * time.Second,
-			TLSHandshakeTimeout:   5 * time.Second,
-			ExpectContinueTimeout: 1 * time.Second,
-		},
-		// Prevent redirect loops or redirects to internal sites
-		CheckRedirect: func(req *http.Request, via []*http.Request) error {
-			return http.ErrUseLastResponse
-		},
-	}
+	client := newSecureHTTPClient()
```
averikitsch: Can we initialize this in initialize method and save it on the authservice so we can reuse it?

internal/auth/generic/generic.go
```diff
@@ -230,7 +234,20 @@ func (a AuthService) ValidateMCPAuth(ctx context.Context, h http.Header) error {
...
+	tokenStr := headerParts[1]
+
+	if isJWTFormat(tokenStr) {
+		return a.validateJwtToken(ctx, tokenStr)
+	}
+	return a.validateOpaqueToken(ctx, tokenStr)
+}
+
+func isJWTFormat(token string) bool {
+	return strings.Count(token, ".") == 2
```
averikitsch, line 250: nit: is there a better way to confirm this?

internal/auth/generic/generic.go
```diff
@@ -280,3 +297,91 @@ func (a AuthService) ValidateMCPAuth(ctx context.Context, h http.Header) error {
...
 	return nil
 }
+
+func (a AuthService) validateOpaqueToken(ctx context.Context, tokenStr string) error {
+	logger, err := util.LoggerFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get logger from context: %w", err)
+	}
+
+	introspectionURL, err := url.JoinPath(a.AuthorizationServer, "introspect")
```
averikitsch: It looks like we should first look from "introspection_endpoint" in  the well-known file. 
Does this work with the accounts.google.com?

internal/auth/generic/generic.go
```diff
@@ -280,3 +297,91 @@ func (a AuthService) ValidateMCPAuth(ctx context.Context, h http.Header) error {
...
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
+	if err != nil {
+		return fmt.Errorf("failed to read introspection response: %w", err)
+	}
+
+	var introspectResp struct {
+		Active   bool   `json:"active"`
+		Scope    string `json:"scope"`
+		ClientId string `json:"client_id"`
```
averikitsch: Can you remind me why we are using ClientId vs Aud here?

internal/auth/generic/generic.go
```diff
@@ -280,3 +297,91 @@ func (a AuthService) ValidateMCPAuth(ctx context.Context, h http.Header) error {
...
+		Exp      int64  `json:"exp"`
+	}
+
+	if err := json.Unmarshal(body, &introspectResp); err != nil {
+		return fmt.Errorf("failed to parse introspection response: %w", err)
+	}
+
+	if !introspectResp.Active {
```
averikitsch, line 330: nit: is there an easy way for refactor the token validation into one method for both opaque and jwt tokens?

internal/auth/generic/generic.go
duwenxin99: We can also try to parse the JWT token and if it fails, fall back to validating access token, but it will create extra overhead.

internal/auth/generic/generic.go
duwenxin99: Updated to look for the introspection_endpoint first

internal/auth/generic/generic.go
duwenxin99: I refactored to use a helper to validate the claims for both token types.

internal/auth/generic/generic.go
duwenxin99: updated

internal/auth/generic/generic.go
duwenxin99: We should check aud instead of client ID. Updated.

PR #2961, opened 2026-04-06
feat(arcadedb): add source and tools

### LINKED ISSUE #2689: Add ArcadeDB source (multi-model database with Bolt protocol support)
### Prerequisites

- [x] I have searched existing issues to verify this is not a duplicate

### Problem

ArcadeDB is a multi-model database that natively supports **graph** (OpenCypher/Gremlin), **document** (SQL), **time-series**, and **key-value** data models in a single engine. It's open source (Apache 2.0) and designed for high performance with a small footprint.

There is currently no ArcadeDB source in the Toolbox, despite growing adoption in AI/LLM pipelines — particularly for knowledge graphs and RAG workflows that benefit from its multi-model capabilities.

### Proposed Solution

ArcadeDB implements the **Neo4j Bolt wire protocol**, so the implementation can reuse the existing `neo4j-go-driver` — similar to how SingleStore reuses the MySQL driver (#1333).

The ArcadeDB source would add value beyond the existing Neo4j source by:

- **Multi-model queries**: ArcadeDB supports SQL alongside Cypher, allowing agents to query documents and graphs in one database
- **Discoverability**: Users searching for ArcadeDB in the Toolbox docs would find it listed as a supported source
- **ArcadeDB-specific defaults**: Different default port, auth patterns, and connection URI scheme

#### Proposed configuration

```yaml
kind: sources
name: my-arcadedb-source
type: arcadedb
uri: bolt://localhost:7687
user: root
password: ${PASSWORD}
database: "mydb"
```

#### Proposed tools

- `execute-cypher` — Run OpenCypher queries (reusing Neo4j tool patterns)
- `execute-sql` — Run SQL queries (ArcadeDB's native query language, multi-model advantage)

### Implementation Notes

- The source implementation would be mostly based on the **Neo4j source**, using the same `neo4j-go-driver` via Bolt protocol
- Minimal new code required — primarily configuration, SQL tool support, and documentation
- I'm willing to contribute the implementation and testing

### References

- ArcadeDB: https://arcadedb.com
- ArcadeDB GitHub: https://github.com/ArcadeData/arcadedb
- Bolt protocol docs: https://docs.arcadedb.com/#bolt-driver
- ArcadeDB SQL reference: https://docs.arcadedb.com/#SQL

Deeven-Seru:
## Summary

This PR implements native ArcadeDB support in the GenAI Toolbox by introducing a new `arcadedb` source and two associated tools for executing Cypher and SQL queries.

ArcadeDB is a multi-model database supporting graph, document, key-value, and time-series workloads through the Bolt protocol. Since it is compatible with the Neo4j driver, this implementation reuses the existing `neo4j-go-driver`, enabling seamless integration with minimal additional complexity.

This change resolves the feature request to add ArcadeDB as a supported source in the Toolbox.

Fixes: #2689

---

### 1) New Source: `arcadedb`

A fully functional source implementation that:

- Registers a new source type: `arcadedb`
- Uses the Neo4j Bolt protocol driver
- Verifies connectivity during initialization
- Supports configurable database selection
- Integrates with OpenTelemetry tracing
- Follows existing Toolbox source architecture patterns

Key capabilities:

- Bolt protocol connectivity
- Standard authentication support
- Config validation
- Structured query execution
- Telemetry-compatible initialization

---

### 2) New Tool: `arcadedb-execute-cypher`

Executes OpenCypher queries against ArcadeDB.

Features:

- Read-only enforcement mode
- Dry-run validation mode
- Query classification (read vs write)
- Execution plan inspection support

Example:

```yaml
kind: tool
name: query_arcadedb
type: arcadedb-execute-cypher
source: my-arcadedb-source
readOnly: true
description: Execute Cypher queries against ArcadeDB

### CODE REVIEW

**Comment by Deeven-Seru** [comment]
@stenalpjolly /gcburn is what maintainers do :)

**Comment by stenalpjolly** [comment]
@Deeven-Seru I'm from Google only. I was testing my maintainer access for the org

**Comment by lvca** [comment]
Thanks @Deeven-Seru for picking this up, and @duwenxin99 for looping me in. The overall shape matches what we agreed in #2689: a dedicated `arcadedb` source reusing `neo4j-go-driver`, with separate Cypher and SQL tools. Happy to commit to long-term maintenance of this code on the ArcadeDB side.

A few items I'd like to see addressed before merge, and a couple of things worth verifying.

### Blocking

**1. `arcadedb-execute-sql` has no `readOnly` mode.**
The Cypher tool supports `readOnly` and `dry_run`, but the SQL tool passes statements straight through. ArcadeDB SQL can `DELETE`, `DROP TYPE`, `CREATE VERTEX`, etc., so for parity with the Cypher tool and to match the "developer assistant with human-in-the-loop" framing in the docs, the SQL tool should at minimum support `readOnly`. A simple first-token check (allow `SELECT`, `MATCH`, `TRAVERSE`) covers the common case; we can iterate on a proper classifier later.

**2. Query parameters are hardcoded to `nil`.**
Both `Invoke` methods call `source.RunCypher(ctx, cypherStr, nil, ...)` and `source.RunSQL(ctx, sqlStr, nil)`. The `params map[string]any` argument is plumbed through the source layer but dead-ended at the tool layer, so callers have no way to pass parameterized queries. That forces string concatenation for any dynamic value and defeats Bolt's built-in parameterization. Either expose a `params` tool input or remove the unused argument from the interface.

**3. Driver is never closed.**
`Source.Driver` has no teardown path. Worth confirming the toolbox runtime calls a close hook on source shutdown/reload; if it does, we should implement it here to avoid leaking Bolt connections. The existing Neo4j source should be the reference.

**4. Redundant `Database` defaulting.**
`newConfig` sets `Database: "arcadedb"`, the struct has `validate:"required"` on `Database`, and `Initialize` re-checks `if r.Database == ""`. Two of those three paths are unreachable. The docs also list `database` as required while the code defaults it, which is contradictory. Pick one behavior (I'd suggest: make it optional with a documented default) and align docs, validator, and code.

### Nits

**5. Inconsistent JSON key casing in execution plan output.**
Top-level plan uses `childrenCount`, recursive children use `children_count`. Should be consistent (camelCase to match the rest of the payload).

**6. Unrelated conversationalanalytics changes.**
The diff includes ~11 lines of context-propagation and HTTP-timeout changes in `internal/tools/conversationalanalytics/**` that aren't related to ArcadeDB. Probably a rebase artifact. Worth splitting or rebasing cleanly so the ArcadeDB PR stands alone.

### Worth verifying

**7. `EXPLAIN` behavior for Cypher dry-run.**
The code prefixes queries with `EXPLAIN` and then reads `summary.Plan()` via the Neo4j driver. ArcadeDB supports `EXPLAIN` in its Cypher implementation, but I want to confirm the plan summary returned over Bolt is shaped compatibly with `plan.Operator()`, `plan.Children()`, etc. Otherwise dry-run will silently return empty data. I can run this against a local ArcadeDB once the other items are addressed, and report back with concrete output.

**8. Classifier coverage for ArcadeDB-specific Cypher.**
The `classifier.QueryClassifier` was built for Neo4j Cypher. ArcadeDB has some dialect differences. I'll sanity-check that writes are classified correctly across common ArcadeDB Cypher forms and open a follow-up issue if we find gaps, so it doesn't block this PR.

### Offer

Once these are addressed, I'll test the branch end-to-end against a live ArcadeDB (Cypher + SQL, read-only enforcement, dry-run plan output) and post results here. Happy to also contribute a basic integration-test harness in a follow-up if the maintainers want one.

Thanks again for the work on this.

**Comment by lvca** [comment]
Thanks @Deeven-Seru, nice turnaround on the blocking items. Quick takes on both open questions:

### 1. SQL readOnly allow-list

I'd keep it narrow for v1 and **not** add `RETURN` or `LET` to the first-token allow-list.

The reason is that in ArcadeDB SQL both keywords are almost always used inside script-mode bodies, not as standalone read statements, and a first-token classifier can't see past them into what they actually do:

- `LET $x = SELECT ...` is read-only, but `LET $x = INSERT ...` / `LET $x = UPDATE ...` / `LET $x = DELETE ...` are all valid and write.
- `RETURN $x` is safe only if the preceding script that produced `$x` was read-only, which the classifier has no way to verify.

Adding these to a first-token allow-list would silently let writes slip through readOnly mode. Safer to stay strict, let real user query samples tell us which read-only patterns are being blocked, and then either extend the list case-by-case or swap in a proper parser-based classifier. A strict default is easy to loosen; a permissive one that let a `DELETE` through under `readOnly` is much harder to walk back.

Current list (`SELECT`, `MATCH`, `TRAVERSE`, `WITH`, `EXPLAIN`) looks right to ship.

### 2. Cypher dry-run with partial plan output

My preference: return the partial payload with an explicit `plan_incomplete: true` flag (plus a short human-readable `warning` string), rather than fail fast. Reasoning:

- Even a partial plan confirms the query parsed and the optimizer ran, which is still useful signal in a developer-assistant / human-in-the-loop flow.
- Failing the whole dry-run over a server-version quirk is a worse UX than surfacing the degradation and letting the calling client decide what to render.
- The check is cheap to emit at driver level: if `plan.Operator()` is empty, or `plan.Children()` is nil on a node whose `OperatorType` implies children, set the flag.

One carve-out: if the plan payload is **completely empty** (zero operators at all), I'd treat that as an error rather than partial. An empty plan almost certainly means the server-side EXPLAIN didn't produce anything usable on that version, and returning a silent empty object is more misleading than helpful. So: empty → error, partial → flagged payload.

### Next step from my side

Once you push the updated PR, I'll run the end-to-end test against a live ArcadeDB (Cypher + SQL, readOnly enforcement across the cases above, dry-run plan output on a current server build) and post results here. If the partial-plan case shows up in practice I'll include concrete output so we can calibrate the warning threshold.

**Comment by Deeven-Seru** [comment]
@lvca thanks for answering , PTAL at the pr

**Review by Deeven-Seru — COMMENTED** [review]
/gemini review

**Comment by Deeven-Seru** [comment]
@lvca do we have an update on this ?

**Comment by lvca** [comment]
@Deeven-Seru I didn't know you were waiting for me. How can I help with this PR?

**Comment by Deeven-Seru** [comment]
@lvca  please proceed with your tests against the live  Arcade db like you mentioned above and let me know if something is not working as intended here 😊

**Comment by lvca** [comment]
Tested PR HEAD (623372a) against a live ArcadeDB 26.5.1-SNAPSHOT instance over Bolt (port 7687). Most things work, with one blocker on Cypher `dry_run` and a couple of nits on the source config.

### Test setup

Spun up `arcadedata/arcadedb:latest` in Docker with the Neo4j-Bolt plugin enabled, created `testdb`, seeded a `Person` vertex type, then drove the toolbox over MCP JSON-RPC.

```yaml
sources:
  test-arcadedb:
    kind: arcadedb
    uri: bolt://localhost:7687
    user: root
    password: ${PASSWORD}
    database: testdb
```

### What works

`arcadedb-execute-cypher`:
- Read queries (`MATCH ... RETURN`) return rows correctly.
- Parameter binding (`$name`, `$minAge`) works.
- `readOnly: true` rejects writes with a clean error.
- `readOnly: true` correctly rejects writes that try to hide behind a leading `// comment` (classifier handles this).
- `CALL db.labels()` and similar procedures work.

`arcadedb-execute-sql`:
- `SELECT`, `MATCH {type: ...}`, parameterized queries (`:minAge`) work.
- `readOnly: true` allow-list (`SELECT`, `MATCH`, `TRAVERSE`, `WITH`, `EXPLAIN`) blocks `INSERT`, `UPDATE`, `DELETE`, `CREATE` cleanly. Leading whitespace is handled correctly.
- `dry_run: true` returns ArcadeDB's `explainPlan` and `executionPlanAsString` from the HTTP `EXPLAIN` response. Verified that `INSERT ... dry_run: true` does **not** actually write (count stays 0).
- INSERT/DELETE/UPDATE in read-write mode all work, return the expected document or count.
- 500s from ArcadeDB (bad type, syntax errors) are surfaced as `isError: true` with the full server error.

### Blocker: `arcadedb-execute-cypher` `dry_run` is broken on ArcadeDB

Every Cypher `dry_run: true` invocation fails with:

```
error processing request: dry-run produced no execution plan
```

Reproduced on three different queries (simple `MATCH`, `count(n)`, parameterized lookup). Root cause: the source prepends `EXPLAIN ` and runs it over Bolt, then reads `summary.Plan()`. ArcadeDB's Bolt implementation does not populate the `Plan` field of the Bolt result summary; the execution plan is exposed via the HTTP `/api/v1/command` response (`explainPlan` / `explain` fields), which is what the SQL tool already consumes.

Suggested fix: when the Cypher tool is in `dry_run` mode, route the `EXPLAIN` through the same HTTP path that `RunSQL` already uses (with `language: cypher`), instead of going through Bolt. That keeps a single code path for plans and matches what works for SQL today. Alternatively, fall back to the HTTP path only when `summary.Plan()` is nil.

### Smaller issues worth fixing before merge

**1. SQL tool hardcodes HTTP endpoint to `:2480` over plain HTTP.**
`arcadeHTTPCommandURL` in `internal/sources/arcadedb/arcadedb.go` builds `http://<host-from-bolt-uri>:2480/api/v1/command/<db>`. This breaks for any deployment that:
- runs HTTP on a non-default port (we ship installs that move it to e.g. 2481),
- runs HTTPS only (the cloud and any TLS-fronted deployment),
- terminates Bolt and HTTP behind different hostnames (proxied setups).

Suggest exposing two more optional fields on the source config, e.g. `httpUri` (full base URL, takes precedence) or `httpPort` + `httpScheme`, with the current behavior as the default. The same applies to the future Cypher `dry_run` HTTP fallback above.

**2. Default `database: "arcadedb"` doesn't exist.**
`newConfig` sets `Database: "arcadedb"` as the default, but ArcadeDB has no built-in database with that name. If a user omits `database`, the source initializes (Bolt verifies connectivity at the server level) but every query then fails on `testdb` -ish access. Either make `database` required (`validate:required`, like the Neo4j source) or drop the silent default.

**3. Cypher empty result serializes as `null` instead of `[]`.**
`MATCH (n:NoSuchLabel) RETURN n` returns `{"content":[{"type":"text","text":"null"}]}`. Likely `formatRecords` returns a nil slice when there are zero records. Cosmetic, but `[]` would match the SQL tool's empty-content behavior.

### Verdict

Cypher (read/write/readOnly/params) and the full SQL surface are solid. The only thing standing between this PR and a clean merge from my side is the Cypher `dry_run` blocker; the hardcoded `:2480` and the silent `database` default are good follow-ups (the port one will bite real deployments). Happy to retest once the Cypher dry-run path is updated.

lvca:
I've just fixed the issue on ArcadeDB about `arcadedb-execute-cypher dry_run`, it's in:

https://github.com/ArcadeData/arcadedb/issues/4129

Using latest ArcadeDB 26.5.1-SNAPSHOT solves the problem.

lvca:
ArcadeDB 26.5.1 is out, if you use this release the issue is already fixed.

internal/tools/arcadedb/arcadedbexecutesql/arcadedbexecutesql.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (t Tool) GetAuthTokenHeaderName(resourceMgr tools.SourceProvider) (string, error) {
+	return "Authorization", nil
+}
+
+func (t Tool) GetParameters() parameters.Parameters {
+	return t.Parameters
+}
+
+func isReadOnlySQL(sql string) bool {
```
duwenxin99: I don't think this is a solid check. If a user passes multiple stacked statements separated by a semicolon (e.g., SELECT * FROM Person; DROP TABLE Person;), the first token SELECT is verified as read-only, but subsequent write statements will still execute.

internal/sources/arcadedb/arcadedb.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, httpURL, bytes.NewReader(payload))
+	if err != nil {
+		return nil, fmt.Errorf("unable to create SQL request: %w", err)
+	}
+	req.SetBasicAuth(s.User, s.Password)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
```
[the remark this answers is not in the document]
duwenxin99: Have we considered using drivers?

internal/tools/arcadedb/arcadedbexecutesql/arcadedbexecutesql.go
```diff
@@ -0,0 +1,189 @@
...
+	ArcadeDBDatabase() string
+	RunSQL(context.Context, string, map[string]any) (any, error)
+}
+
+type Config struct {
+	Name         string   `yaml:"name" validate:"required"`
+	Type         string   `yaml:"type" validate:"required"`
+	Source       string   `yaml:"source" validate:"required"`
+	Description  string   `yaml:"description" validate:"required"`
+	ReadOnly     bool     `yaml:"readOnly"`
```
duwenxin99: I would suggest removing this config field if we can't ensure the read-only constraint.

duwenxin99 (changes requested):
Hi @Deeven-Seru @lvca, thanks for the contribution! I left some comments and there are also some suggestions from gemini. The PR is on the right track. Feel free to request review again once the comments are addressed. Also please just make sure to add integration tests.

duwenxin99 (changes requested):
Hi @Deeven-Seru @lvca, thanks for the contribution! I left some comments and there are also some suggestions from gemini. The PR is on the right track. Feel free to request review again once the comments are addressed. Also please just make sure to add integration tests.

lvca:
Thanks @duwenxin99 for the re-review, and @gemini-code-assist for the additional pass. Going through each item below with concrete fixes so @Deeven-Seru has a clean punch list to work from.

### From @duwenxin99

**1, 3. SQL `readOnly` enforcement (`arcadedbexecutesql.go:57, 176`)**

These two are best resolved together by dropping client-side parsing entirely and using ArcadeDB's two HTTP endpoints, which mirror exactly this distinction:

* `POST /api/v1/query/{database}` is idempotent / read-only. The server rejects any non-idempotent statement.
* `POST /api/v1/command/{database}` is general execution (read or write).

The SQL tool routes by mode:

* `readOnly: true` → `/api/v1/query/{db}`. Enforcement is server-side. No parser, no allow-list, no stacked-statement, comment, or string-literal evasion to worry about.
* `readOnly: false` (default) → `/api/v1/command/{db}` as today.

That's a hard guarantee instead of a heuristic, which also answers point 3: `readOnly` is worth keeping in the config as-is, and `isReadOnlySQL` plus the whole allow-list go away.

`dry_run` routes EXPLAIN to the same endpoint the underlying statement would use, so `readOnly: true` + `dry_run: true` on a write statement gets rejected by the server consistently with how the bare statement would be rejected. Cypher stays on Bolt with the existing classifier, since Bolt doesn't expose the same endpoint split.

**2. "Have we considered using drivers?" (`arcadedb.go:192`)**

The reason the SQL path is HTTP and the Cypher path is Bolt:

* `neo4j-go-driver/v6` only speaks Cypher. ArcadeDB SQL is not a Cypher dialect.
* ArcadeDB's Bolt endpoint does support a `language` hint, but the Neo4j Go driver doesn't expose a per-statement language switch cleanly, and pinning to a private fork would be worse than what we have.
* `/api/v1/query` and `/api/v1/command` are the documented, first-class multi-language entry points on ArcadeDB and are what production deployments use.

There isn't a community Go driver for ArcadeDB's HTTP API today; a standalone driver is a sensible follow-up on the ArcadeData side. For the scope of this PR the inline client is reasonable as long as it's hardened, which the gemini items below cover.

**4. Integration tests**

Done. Opened against @Deeven-Seru's branch: https://github.com/Deeven-Seru/mcp-toolbox/pull/1

`tests/arcadedb/arcadedb_integration_test.go` mirrors the `tests/neo4j` pattern, env-var driven (`ARCADEDB_URI` / `ARCADEDB_USER` / `ARCADEDB_PASS` / `ARCADEDB_DATABASE`, plus an optional `ARCADEDB_HTTP_URL` override). Covers manifest shape, Cypher read / parameterized read / `dry_run` plan / `readOnly` rejection, SQL read / parameterized read / `dry_run` plan / `readOnly` rejection / stacked-statement rejection with a direct-HTTP side-effect check, and a write-then-read round trip. Setup and teardown go through ArcadeDB's HTTP API directly so the data path doesn't depend on the tools under test.

`.ci/integration.cloudbuild.yaml` is deliberately not modified in that PR: adding the shard needs Google infra to provision an ArcadeDB test instance plus `arcadedb_user` / `arcadedb_pass` secrets, otherwise the `secretEnv` mount fails the build. Happy to add the cloudbuild stanza in a follow-up once those secrets exist, or to switch the shard to a docker-compose / TestContainers approach instead if that's how upstream CI prefers to handle DB-backed tests.

### From gemini-code-assist

**5. `_index.md` and `tools/_index.md` frontmatter-only**

Agreed. Move the body of `docs/en/integrations/arcadedb/_index.md` into `source.md` (already exists in this PR), and drop the `{{< list-tools >}}` shortcode from `docs/en/integrations/arcadedb/tools/_index.md` so both index files are pure frontmatter.

**6. Tool title casing**

Agreed. Apply both suggested edits:

* `docs/en/integrations/arcadedb/tools/arcadedb-execute-cypher.md` → `title: "arcadedb-execute-cypher Tool"`
* `docs/en/integrations/arcadedb/tools/arcadedb-execute-sql.md` → `title: "arcadedb-execute-sql Tool"`

**7. `http.DefaultClient` in `arcadedb.go:192`**

Agreed. Package-level client with a timeout:

```go
var arcadeHTTPClient = &http.Client{Timeout: 60 * time.Second}
```

…and replace `http.DefaultClient.Do(req)` with `arcadeHTTPClient.Do(req)`. 60s is a reasonable default; can be promoted to a source-config field later if a deployment needs to tune it.

**8. HTTP URL construction (`arcadedb.go:287`)**

Same surface I raised on 2026-05-07 (hardcoded `:2480`, hardcoded `http://`). Combined fix, now also parameterized on the endpoint name so `/query` and `/command` share one builder:

* URL-encode the database name with `url.PathEscape`.
* Add two optional source-config fields: `httpUri` (full base URL, takes precedence) and `httpScheme` + `httpPort` (override the defaults derived from the Bolt URI).
* Default behavior stays as today: `http://<bolt-host>:2480`.

```yaml
sources:
  test-arcadedb:
    kind: arcadedb
    uri: bolt://localhost:7687
    httpUri: https://arcade.example.com:2480  # optional; takes precedence
    user: root
    password: ${PASSWORD}
    database: testdb
```

```go
// endpoint is "query" or "command"
func (s *Source) arcadeHTTPEndpointURL(endpoint string) (string, error) {
    base := s.HTTPUri
    if base == "" {
        parsed, err := url.Parse(s.Uri)
        if err != nil {
            return "", fmt.Errorf("unable to parse arcade uri %q: %w", s.Uri, err)
        }
        host := parsed.Hostname()
        if host == "" {
            return "", fmt.Errorf("unable to parse host from arcade uri %q", s.Uri)
        }
        scheme := s.HTTPScheme
        if scheme == "" {
            scheme = "http"
        }
        port := s.HTTPPort
        if port == 0 {
            port = 2480
        }
        base = fmt.Sprintf("%s://%s:%d", scheme, host, port)
    }
    return fmt.Sprintf("%s/api/v1/%s/%s",
        strings.TrimRight(base, "/"),
        endpoint,
        url.PathEscape(s.ArcadeDBDatabase())), nil
}
```

Callers pass `"query"` or `"command"` based on `readOnly`. URL-encoding, scheme, and port are all closed off in one pass.

### Next step

Once the in-tree changes above land on Deeven's branch (the integration-tests PR will rebase cleanly on top) I'll run the live end-to-end suite against an ArcadeDB 26.5.1 instance and post results.

internal/tools/arcadedb/arcadedbexecutesql/arcadedbexecutesql.go
Deeven-Seru: I've completely removed the client-side parsing checks and now route readOnly: true queries to ArcadeDB's native /api/v1/query endpoint, which guarantees server-side enforcement and blocks stacked write statements.

internal/tools/arcadedb/arcadedbexecutesql/arcadedbexecutesql.go
Deeven-Seru: I ve kept the config field because we now route queries to ArcadeDB's native read-only HTTP endpoint (/api/v1/query), fully enforcing the constraint server-side.

Deeven-Seru:
Hi @lvca ,@duwenxin99 The feedback items from the review passes have been fully addressed, verified, and squashed into a single clean commit on this branch (`Deeven-Seru:fix-2689-arcadedb`). 

Here is a summary of the changes:

### 1 & 3. SQL `readOnly` Enforcement & Config Parameters
- **Server-Side Enforcement:** Removed all client-side query string-parsing and regex heuristics (`isReadOnlySQL`) entirely. 
- **Dynamic Endpoint Routing:** We now route SQL queries depending on the query mode:
  - `readOnly: true` -> `POST /api/v1/query/{database}` (server-side read-only endpoint, natively blocks any writes or stacked write statements).
  - `readOnly: false` (default) -> `POST /api/v1/command/{database}` (general execution).
- `dry_run` validations route `EXPLAIN` queries through the corresponding read-only or read-write endpoints so validation matches execution semantics.

### 2. Bolt/REST Driver Architecture
- Clarified the separation: Neo4j Go driver only speaks Cypher. Since ArcadeDB SQL uses a distinct dialect, we communicate via ArcadeDB's native HTTP REST API for SQL commands/queries, while Cypher statements leverage the Bolt protocol.

### 4. Integration Tests
- Rebased and squashed the live integration tests (`tests/arcadedb/arcadedb_integration_test.go`) directly into this branch.
- The tests run in an env-var driven manner (`ARCADEDB_URI`, `ARCADEDB_USER`, `ARCADEDB_PASS`, `ARCADEDB_DATABASE`) and cover Cypher/SQL, parameterized queries, `dry_run` plan generation, and native `readOnly` rejection (including stacked write attacks).

### 5. Frontmatter-Only Index Files
- Moved the documentation body out of `docs/en/integrations/arcadedb/_index.md` into `source.md`.
- Stripped all body text and shortcodes from both `_index.md` files under the `arcadedb/` integration paths so they consist strictly of YAML frontmatter.

### 6. Tool Title Casing
- Appended the `" Tool"` suffix to titles in the tool documentation files:
  - `arcadedb-execute-cypher.md` -> `title: "arcadedb-execute-cypher Tool"`
  - `arcadedb-execute-sql.md` -> `title: "arcadedb-execute-sql Tool"`

### 7. Hardened HTTP Client
- Swapped out the unconfigured `http.DefaultClient` for a package-level, timeout-hardened `http.Client` configured with a `60s` timeout.

### 8. HTTP URL Construction
- Implemented a unified `arcadeHTTPEndpointURL(endpoint string)` builder using `url.PathEscape` on database names.
- Added support for new optional connection parameter overrides (`httpUri`, `httpScheme`, `httpPort`) in the sources configuration.

lvca:
@Deeven-Seru and @duwenxin99  It looks all good to me!

duwenxin99 (approved):

Deeven-Seru:
@duwenxin99 I've fixed the lint, please re initiate the checks again thanks :)

duwenxin99 (approved):

Deeven-Seru:
🎉

PR #2830, opened 2026-03-24
feat(tools/looker): Looker agent management from MCP

drstrangelooker:
## Description

Looker agent management from MCP

## PR Checklist

> Thank you for opening a Pull Request! Before submitting your PR, there are a
> few things you can do to make sure it goes smoothly:

- [x] Make sure you reviewed
  [CONTRIBUTING.md](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md)
- [x] Make sure to open an issue as a
  [bug/issue](https://github.com/googleapis/genai-toolbox/issues/new/choose)
  before writing your code! That way we can discuss the change, evaluate
  designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)
- [x] Make sure to add `!` if this involve a breaking change

🛠️ Fixes #2827

PR #2967, opened 2026-04-06
fix(tools/looker): enforce critical safety annotations for agent tools

hiracky16:
## Description
This PR addresses the security and safety concerns raised in #2830 regarding critical annotations in Looker agent tools.

### Changes
- **Annotation Protection**: Hardcoded `readOnlyHint` and `destructiveHint` in the `Initialize` method for the following tools to ensure they cannot be overridden by user configuration:
  - `create_agent` (readOnly: false)
  - `update_agent` (readOnly: false)
  - `delete_agent` (readOnly: false, destructive: true)
  - `get_agent` (readOnly: true)
  - `list_agents` (readOnly: true)
- **Verification Tests**: Added `TestAnnotations` to each tool's test suite to verify that these hints are strictly enforced even when a user attempts to override them in the configuration.
- **Go Environment Consistency**: Kept the Go version at `1.25.7` to match the current `main` branch state while ensuring no unwanted toolchain updates were introduced.

## Verification Results
Ran unit tests for all 5 tools and confirmed that all tests, including the new annotation enforcement checks, passed successfully.


hiracky16:
@duwenxin99 Updated the code to avoid direct modification of the `cfg.Annotations` pointer by creating a local copy in `Initialize`. This ensures no side effects while still enforcing the mandatory safety annotations. PTAL!

PR #2992, opened 2026-04-08
test(source/alloydbpg): create MCP integration tests

anubhav756:
This PR adds the mapped integration tests for AlloyDB PG tools using the native MCP harness.

### DIFF
--- a/tests/alloydbpg/alloydb_pg_mcp_test.go
+++ b/tests/alloydbpg/alloydb_pg_mcp_test.go
@@ -26,8 +26,8 @@ import (
 
 	"cloud.google.com/go/alloydbconn"
 	"github.com/google/uuid"
-	"github.com/googleapis/genai-toolbox/internal/testutils"
-	"github.com/googleapis/genai-toolbox/tests"
+	"github.com/googleapis/mcp-toolbox/internal/testutils"
+	"github.com/googleapis/mcp-toolbox/tests"
 	"github.com/jackc/pgx/v5/pgxpool"
 )
 
--- a/tests/alloydbpg/alloydb_pg_mcp_test.go
+++ b/tests/alloydbpg/alloydb_pg_mcp_test.go
@@ -178,6 +178,8 @@ func TestAlloyDBPgCallTool(t *testing.T) {
 	tests.RunMCPPostgresListViewsTest(t, ctx, pool)
 	tests.RunMCPPostgresListSchemasTest(t, ctx, pool, AlloyDBPostgresUser, uniqueID)
 	tests.RunMCPPostgresListActiveQueriesTest(t, ctx, pool)
+	tests.RunMCPPostgresListTablesTest(t, ctx, pool, AlloyDBPostgresUser)
+	tests.RunMCPPostgresListQueryStatsTest(t, ctx, pool)
 	tests.RunMCPPostgresListAvailableExtensionsTest(t)
 	tests.RunMCPPostgresListInstalledExtensionsTest(t)
 	tests.RunMCPPostgresDatabaseOverviewTest(t, ctx, pool)
--- a/tests/alloydbpg/alloydb_pg_mcp_test.go
+++ b/tests/alloydbpg/alloydb_pg_mcp_test.go
@@ -83,7 +83,7 @@ func getAlloyDBDialOpts(ipType string) ([]alloydbconn.DialOption, error) {
 	}
 }
 
-func initAlloyDBPgConnectionPool(project, region, cluster, instance, ipType, user, pass, dbname string) (*pgxpool.Pool, error) {
+func initAlloyDBPgConnectionPool(ctx context.Context, project, region, cluster, instance, ipType, user, pass, dbname string) (*pgxpool.Pool, error) {
 	dsn := fmt.Sprintf("user=%s password=%s dbname=%s sslmode=disable", user, pass, dbname)
 	config, err := pgxpool.ParseConfig(dsn)
 	if err != nil {
@@ -94,7 +94,7 @@ func initAlloyDBPgConnectionPool(project, region, cluster, instance, ipType, use
 	if err != nil {
 		return nil, err
 	}
-	d, err := alloydbconn.NewDialer(context.Background(), alloydbconn.WithDefaultDialOptions(dialOpts...))
+	d, err := alloydbconn.NewDialer(ctx, alloydbconn.WithDefaultDialOptions(dialOpts...))
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse connection uri: %w", err)
 	}
@@ -104,7 +104,7 @@ func initAlloyDBPgConnectionPool(project, region, cluster, instance, ipType, use
 		return d.Dial(ctx, i)
 	}
 
-	pool, err := pgxpool.NewWithConfig(context.Background(), config)
+	pool, err := pgxpool.NewWithConfig(ctx, config)
 	if err != nil {
 		return nil, err
 	}
@@ -148,7 +148,7 @@ func TestAlloyDBPgCallTool(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 
-	pool, err := initAlloyDBPgConnectionPool(AlloyDBPostgresProject, AlloyDBPostgresRegion, AlloyDBPostgresCluster, AlloyDBPostgresInstance, "public", AlloyDBPostgresUser, AlloyDBPostgresPass, AlloyDBPostgresDatabase)
+	pool, err := initAlloyDBPgConnectionPool(ctx, AlloyDBPostgresProject, AlloyDBPostgresRegion, AlloyDBPostgresCluster, AlloyDBPostgresInstance, "public", AlloyDBPostgresUser, AlloyDBPostgresPass, AlloyDBPostgresDatabase)
 	if err != nil {
 		t.Fatalf("unable to create AlloyDB connection pool: %s", err)
 	}
--- a/tests/alloydbpg/alloydb_pg_mcp_test.go
+++ b/tests/alloydbpg/alloydb_pg_mcp_test.go
@@ -157,7 +157,8 @@ func TestAlloyDBPgCallTool(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to create AlloyDB connection pool: %s", err)
 	}
-	defer pool.Close()
+	// Note: Don't defer pool.Close() here - the pool is only used for test setup/teardown.
+	// Closing it causes indefinite hangs due to background goroutines in the driver.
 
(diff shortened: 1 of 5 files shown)


tests/mcp_tool.go
```diff
@@ -348,13 +385,1706 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
...
+func buildPostgresURL(host, port, user, pass, dbname string) *url.URL {
+	return &url.URL{
+		Scheme: "postgres",
+		User:   url.UserPassword(user, pass),
+		Host:   fmt.Sprintf("%s:%s", host, port),
+		Path:   dbname,
+	}
+}
+
+// InitPostgresConnectionPool initializes a connection pool for Postgres.
```
Yuan325: Seem like we are not using this function?

tests/mcp_tool.go
```diff
@@ -136,6 +143,39 @@ func InvokeMCPTool(t *testing.T, toolName string, arguments map[string]any, requ
...
+}
+
+// unmarshalMCPResult unmarshals a JSON string into a slice of Ts.
+func unmarshalMCPResult[T any](data string) ([]T, error) {
+	var list []T
+	if err := json.Unmarshal([]byte(data), &list); err == nil {
+		return list, nil
+	}
+	var obj T
+	if err := json.Unmarshal([]byte(data), &obj); err != nil {
```
Yuan325: These 2 functions feels very "hacky" and specific towards these postgres use cases. Is there a more general way to compare these? Would be cleaner to try to prevent any source-specific parsing functions~ WDYT about converting MCPCallToolResponse to string by marshaling it directly. Or we can have the `wants` to be in the format of MCPCallToolResponse.

If we need the `getMCPResultText()` to help with comparison it might be okay, but i'm not too sure about the `unmarshalMCPResult()`.

Yuan325 (changes requested):

tests/mcp_tool.go
Yuan325: Another thought: can we utilize `RunMCPCustomToolCallMethod()` for running these test cases? Or modify that function a little to compare list if needed.

tests/mcp_tool.go
anubhav756: This will be used in a future PR that would port the alloydbomni and postgres tests, since those would use it.

tests/mcp_tool.go
anubhav756: Thanks for the suggestion, and I agree with the concern about this helper being specific and potentially hiding structure issues.

For these PRs, our primary focus was simply to migrate the tests to use the new `/mcp` endpoint while maintaining strict assertion parity with the legacy tests. To keep the PR focused, we largely ported the existing string-matching assertions as they were.

However, I did take a pragmatic step to make `GetMCPResultText` safe and robust by using proper JSON marshaling/unmarshaling instead of string manipulation, so it doesn't mask errors or fail on valid JSON. It also now takes `*testing.T` so it can fail loudly if JSON processing fails.

Moving to full structured comparison is definitely the right direction for long-term maintainability. But it would be a large change in volume right now because we would need to rewrite dozens of test cases to use expected Go objects instead of strings, and add logic to ignore dynamic fields (like timestamps or PIDs).

I've added a `TODO` in `mcp_tool.go` and we would be adding this to our integration test best practices guidelines going forward, to recommend structured comparison for any new tests.

tests/alloydbpg/alloydb_pg_mcp_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	pool, err := pgxpool.NewWithConfig(context.Background(), config)
+	if err != nil {
+		return nil, err
+	}
+	return pool, nil
+}
+
+func TestAlloyDBPgListTools(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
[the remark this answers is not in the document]
anubhav756: Done.

tests/alloydbpg/alloydb_pg_mcp_test.go
```diff
@@ -0,0 +1,206 @@
...
+	defer cancel()
+
+	pool, err := initAlloyDBPgConnectionPool(AlloyDBPostgresProject, AlloyDBPostgresRegion, AlloyDBPostgresCluster, AlloyDBPostgresInstance, "public", AlloyDBPostgresUser, AlloyDBPostgresPass, AlloyDBPostgresDatabase)
+	if err != nil {
+		t.Fatalf("unable to create AlloyDB connection pool: %s", err)
+	}
+	defer pool.Close()
+
+	uniqueID := strings.ReplaceAll(uuid.New().String(), "-", "")
```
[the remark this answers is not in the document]
anubhav756: `RunMCPPostgresListSchemasTest` in `mcp_tool.go` manages its own cleanup.

It calls the helper `setupMCPPostgresSchemas` which creates the schema and returns a cleanup function that executes `DROP SCHEMA <name> CASCADE`. `RunMCPPostgresListSchemasTest` defers this cleanup function immediately.

tests/mcp_tool.go
```diff
@@ -136,6 +143,49 @@ func InvokeMCPTool(t *testing.T, toolName string, arguments map[string]any, requ
 	return resp.StatusCode, &mcpResp, nil
 }
 
+// getMCPResultText safely extracts the text from content blocks, reconstructing a valid JSON array if there are multiple items.
+//
+// TODO: This helper and unmarshalMCPResult rely on extracting text content.
+// For tests that need to strictly validate the exact schema or structure of the output,
+// consider avoiding these helpers and instead unmarshal the raw JSON directly into expected Go structs for comparison.
+func getMCPResultText(t *testing.T, resp *MCPCallToolResponse) string {
```
Yuan325:
```suggestion
func getMCPResultText(t *testing.T, resp *MCPCallToolResponse) []any {
```

tests/mcp_tool.go
```diff
@@ -352,13 +399,1706 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
...
+			}
+			if mcpResp.Result.IsError {
+				t.Fatalf("list_views returned error result: %v", mcpResp.Result)
+			}
+
+			got := getMCPResultText(t, mcpResp)
+			gotObj, err := unmarshalMCPResult[map[string]any](got)
+			if err != nil {
+				t.Fatalf("failed to unmarshal nested result string: %v", err)
+			}
```
Yuan325: let's just try to get the getMCPResultText to return the format that its looking for directly instead of unmarshaling it twice. we will be able to remove the second marshaling function this way.

tests/mcp_tool.go
```diff
@@ -136,6 +143,49 @@ func InvokeMCPTool(t *testing.T, toolName string, arguments map[string]any, requ
...
+		} else {
+			combined = append(combined, item)
+		}
+	}
+
+	respBytes, err := json.Marshal(combined)
+	if err != nil {
+		t.Fatalf("failed to marshal combined result: %v", err)
+	}
+	return string(respBytes)
```
Yuan325:
```suggestion
	var res []any
	for _, content := range resp.Result.Content {
		var item map[string]any
		if err := json.Unmarshal([]byte(content.Text), &item); err != nil {
			res = append(combined, content.Text)
		} else {
			res = append(combined, item)
		}
	}
	return res
```

This way, if Content have length 0, it will return `nil` (seems like thats what the results are looking for), and if there's content, it will return []any (seems like this is what all the wants format are e..g `[]map[string]any`. We won't have to reunmarshal it again.

tests/mcp_tool.go
```diff
@@ -352,13 +399,1706 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
 			if mcpResp.Result.IsError {
 				t.Fatalf("%s returned error result: %v", tc.toolName, mcpResp.Result)
 			}
-			if len(mcpResp.Result.Content) == 0 {
-				t.Fatalf("%s returned empty content field", tc.toolName)
-			}
-			got := mcpResp.Result.Content[0].Text
+			got := getMCPResultText(t, mcpResp)
 			if !strings.Contains(got, tc.wantResult) {
```
Yuan325:
```suggestion
			if !strings.Contains(got[0], tc.wantResult) {
```

tests/mcp_tool.go
```diff
@@ -352,13 +399,1706 @@ func RunMCPToolInvokeTest(t *testing.T, select1Want string, options ...InvokeTes
...
+			if err != nil {
+				t.Fatalf("failed to unmarshal nested result string: %v", err)
+			}
+
+			var wantObj []map[string]any
+			if err := json.Unmarshal([]byte(tc.want), &wantObj); err != nil {
+				t.Fatalf("failed to unmarshal want string: %v", err)
+			}
+
+			if !reflect.DeepEqual(gotObj, wantObj) {
```
Yuan325:
```suggestion
			if diff := cmp.Diff(gotObj, wantObj); diff != "" {
				t.Fatalf("mismatch (-want +got): \n%s", diff)
			}
```

cmp might be better if you're facing type issue with reflect.DeepEqual (deep equal is more strict and check specific types).

Yuan325 (changes requested):
Here's a suggestion to simplify it and prevent double-unmarshaling~ :)

tests/mcp_tool.go
Yuan325: Can we add these functions in `common.go` instead? since they are not technically "tool call" related. Theres where most of the common functions etc. were defined (e.g. setup table etc.)

tests/mcp_tool.go
anubhav756: Done. Thanks!

tests/mcp_tool.go
anubhav756: That's a good point. That way any new tests could directly benefit from this and existing tests can marshal it back to check string contains.

tests/mcp_tool.go
anubhav756: Makes sense. Thanks for pointing that out!

tests/mcp_tool.go
anubhav756: That's a good standard to follow. Thanks!

tests/mcp_tool.go
anubhav756: Ack! We now handle this by marshaling the entire got slice to a JSON string (`gotStr`) and searching within that. Thanks anyways :)

tests/mcp_tool.go
anubhav756: Thanks for these suggestions! I have updated this to return `[]any` directly, but with a few additional improvements. Instead of unmarshaling into `map[string]any`, we unmarshal into `any`. This allows it to handle both JSON objects and JSON arrays seamlessly. Also, we return `[]any{}` instead of `nil` when the content length is `0`, to avoid false failures as `cmp.Diff` treats `nil` and `[]any{}` as different.

PR #3036, opened 2026-04-12
fix: enforce toolset/promptset boundary on tools/call and prompts/get

### LINKED ISSUE #2755: MCP toolset boundary bypass: tools/call can invoke tools outside toolset (IDOR)

sjhddh:
## Summary
- **Fix toolset boundary bypass (IDOR)**: `tools/call` previously resolved tools via `resourceMgr.GetTool()` without verifying the requested tool belongs to the current toolset, allowing clients to invoke any tool by name regardless of their toolset scope.
- **Fix promptset boundary bypass**: Same issue existed for `prompts/get` with `resourceMgr.GetPrompt()`.
- Add `ContainsTool()` / `ContainsPrompt()` membership checks and enforce them in all four MCP protocol versions (v20241105, v20250326, v20250618, v20251125) before delegating to the global resource manager.

Fixes #2755

## Test plan
- [x] Added unit tests for `Toolset.ContainsTool()` and `Promptset.ContainsPrompt()` (both positive and negative cases, including empty sets)
- [x] All existing tests pass (`go test ./internal/tools/... ./internal/prompts/... ./internal/server/mcp/...`)
- [ ] Manual verification: configure two toolsets with different tools, connect to one, and confirm `tools/call` rejects tools from the other toolset


internal/server/mcp/v20241105/method.go
```diff
@@ -114,6 +114,12 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
 		attribute.String("gen_ai.operation.name", "execute_tool"),
 	)
 
+	// Verify tool belongs to the current toolset before resolving globally.
+	if !toolset.ContainsTool(toolName) {
+		err = fmt.Errorf("tool %q is not part of the current toolset", toolName)
+		return jsonrpc.NewError(id, jsonrpc.INVALID_PARAMS, err.Error(), nil), err
```
duwenxin99, line 120: We should return the same error message as non-existent tool call (invalid tool name: tool with name %q does not exist) . Otherwise if the tool doesn't exist at all, we will still return "tool %q is not part of the current toolset" which could be misleading.

internal/server/mcp/v20250326/method.go
```diff
@@ -114,6 +114,12 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
 		attribute.String("gen_ai.operation.name", "execute_tool"),
 	)
 
+	// Verify tool belongs to the current toolset before resolving globally.
+	if !toolset.ContainsTool(toolName) {
```
duwenxin99, line 118: Same for the other errors.

internal/server/mcp/v20241105/method.go
```diff
@@ -361,6 +367,12 @@ func promptsGetHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *r
 	span.SetName(fmt.Sprintf("%s %s", PROMPTS_GET, promptName))
 	span.SetAttributes(attribute.String("gen_ai.prompt.name", promptName))
 
+	// Verify prompt belongs to the current promptset before resolving globally.
+	if !promptset.ContainsPrompt(promptName) {
```
duwenxin99, line 371: Similar for prompt-let's treat it the same way as a non-existent prompt.

duwenxin99 (changes requested):
Hi @sjhddh , the change LGTM overall except for the error messages. Please let me know once you fix them and I'll handle the merge process for you. Thank you for contributing!

duwenxin99:
You may want to squash the commits to get rid of the agent account that's missing CLA.

sjhddh:
@duwenxin99 Thanks! Fixed both:

- Boundary violations on tools/call and prompts/get now return the same error message as genuinely missing tool/prompt (so clients can't distinguish the two cases) — applied across all four protocol versions.
- Amended and force-pushed to drop the co-author trailer so the commit is signed by just my account for CLA.

Let me know if anything else needs polishing.

internal/tools/toolsets.go
```diff
@@ -35,6 +35,16 @@ func (t Toolset) ToConfig() ToolsetConfig {
...
 
+// ContainsTool reports whether the toolset includes a tool with the given name.
+func (t Toolset) ContainsTool(name string) bool {
+	for _, n := range t.ToolNames {
+		if n == name {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
duwenxin99: @sjhddh Can you also fix this comment? Thanks!

duwenxin99 (changes requested):

internal/prompts/promptsets.go
```diff
@@ -36,6 +36,16 @@ func (p Promptset) ToConfig() PromptsetConfig {
...
 
+// ContainsPrompt reports whether the promptset includes a prompt with the given name.
+func (p Promptset) ContainsPrompt(name string) bool {
+	for _, n := range p.PromptNames {
+		if n == name {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
duwenxin99: @sjhddh Can you also fix this comment? Thanks!

sjhddh:
@duwenxin99 Thanks for the review! Pushed 1e5a7f2.

Switched `ContainsTool` / `ContainsPrompt` to O(1) map lookups, per the gemini-bot suggestion:

- Added unexported `toolNameSet` / `promptNameSet` (`map[string]struct{}`) fields
- Populated during `Initialize` alongside the existing slice/manifest builds
- Kept a linear-scan fallback for when the struct is constructed directly (tests that bypass `Initialize` — zero-value contract preserved)
- Updated the doc comments on both methods

Also updated three test files that diffed full `Toolset`/`Promptset` structs with `cmp.Diff` — switched to `cmp.AllowUnexported` / `cmpopts.IgnoreUnexported` so the new private cache doesn't break equality. `go test ./internal/tools/... ./internal/prompts/... ./internal/server/... ./cmd/...` is green.

duwenxin99 (approved):

PR #3049, opened 2026-04-13
feat(auth): implement MCP auth tool-level scopes validation

duwenxin99:


internal/auth/generic/generic.go
```diff
@@ -351,11 +355,21 @@ func (a AuthService) validateOpaqueToken(ctx context.Context, tokenStr string) e
...
+			return nil, &MCPAuthError{Code: http.StatusUnauthorized, Message: "invalid aud claim", ScopesRequired: a.ScopesRequired}
 		}
 	}
 
-	return a.validateClaims(ctx, aud, introspectResp.Scope)
+	err = a.validateClaims(ctx, aud, introspectResp.Scope)
+	if err != nil {
+		return nil, err
+	}
+	claims := map[string]any{
```
averikitsch, line 413: do we need to validate the issuer too?

averikitsch (approved):
See question about issuer

internal/auth/generic/generic.go
duwenxin99: added issuer validation.

PR #3050, opened 2026-04-14
fix: allow converting string literal block with list

### LINKED ISSUE #3023: migration add `\n` instead of separating lines for description
### Prerequisites

### Toolbox version

latest

### Environment

_No response_

### Client

_No response_

### Expected Behavior

migration should convert into flat format without `\n`

### Current Behavior

The descriptions got written as a string with embedded `\n` newlines instead of 
```
description: |
  Description line 1
  description line 2
```

### Steps to reproduce?

migrate subcommand

### Additional Details

_No response_

Yuan325:
This PR updates the following: 
* Fix YAML String Block Conversion: Resolved an issue where multiline strings (like tool descriptions) containing list syntax were being re-encoded as double-quoted strings with explicit \n characters. They now correctly use the | literal block style as expected.
```
description: |
  this is the description
  this tool uses the following parameter:
  - param_1
  - param_2

# will turn into 

description: "this is the description\nthis tool uses the following parameter:\n-param_1\n-param_2"
```
* Updated the converter to identify and retain initial comment lines/license headers at the top of configuration files.
* Updated migration completion status to reflect "ended" state.
* Update to use "v1" -> "nested format" and "v2" -> "flat format"
* Remove "authSources" when checking for keys. We had previously removed support for "authSources".

Fixes #3023

### DIFF
--- a/cmd/internal/config.go
+++ b/cmd/internal/config.go
@@ -15,6 +15,7 @@
 package internal
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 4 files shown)


cmd/internal/config.go
```diff
@@ -142,15 +157,15 @@ func ConvertConfig(raw []byte) ([]byte, error) {
...
 			}
-			// check if the key is config file v1's key
-			if slices.Contains(v1keys, key) {
+			// check if the key is config file nested format's key
+			if slices.Contains(nestedFormatKey, key) {
 				// check if value conversion to yaml.MapSlice successfully
 				// fields such as "tools" in toolsets might pass the first check but
 				// fail to convert to MapSlice
 				if slice, ok := item.Value.(yaml.MapSlice); ok {
 					// Deprecated: convert authSources to authServices
```
duwenxin99:
```suggestion

```

cmd/internal/config.go
```diff
@@ -142,15 +157,15 @@ func ConvertConfig(raw []byte) ([]byte, error) {
 			if !ok {
 				return nil, fmt.Errorf("doc %d: unexpected non-string key in input: %v", docIndex, item.Key)
 			}
-			// check if the key is config file v1's key
-			if slices.Contains(v1keys, key) {
+			// check if the key is config file nested format's key
```
duwenxin99: If a document is already in flat format but happens to contain these nested format keys (which is unlikely but still possible?), this unction will still attempt to transform it. Should we check `hasKindField(input)` here first?

duwenxin99 (approved):

cmd/internal/config.go
Yuan325: Simplified code to check for hasKindField() before the type check. Thank you for the suggestion.

PR #3054, opened 2026-04-14
fix: remove hardcoded * allowed origin for sse

### LINKED ISSUE #3053: [SECURITY] Hardcoded CORS bypass on SSE endpoint in internal/server/mcp.go
### Summary
The `mcp-toolbox` (formerly `genai-toolbox`) implements a global CORS (Cross-Origin Resource Sharing) middleware in `internal/server/server.go` but overrides it with a hardcoded `Access-Control-Allow-Origin: *` in the SSE handler for the Model Context Protocol (MCP).

### Technical Details
In `internal/server/mcp.go:370`, the `sseHandler` manually sets the header:
```go
w.Header().Set("Access-Control-Allow-Origin", "*")
```
This hardcoded value overrides the global CORS policy, allowing any website to establish a connection to the toolbox, hijack session IDs, and execute arbitrary tools on behalf of the user.

### Impact
- **Security Policy Bypass:** Renders the `--allowed-origins` flag ineffective for the SSE endpoint.
- **CSRF / Session Hijacking:** Malicious sites can use the toolbox as a proxy to exfiltrate data from databases (Postgres, BigQuery, etc.) configured in the toolbox.

### Recommended Mitigation
Remove the hardcoded header at line 370 of `internal/server/mcp.go`. The global CORS middleware in `internal/server/server.go` is already correctly configured to handle this based on administrator flags.

CC: @averikitsch

Yuan325:
Remove the hardcoded "Access-Control-Allow-Origin" header in sseHandler. This are only affecting users that are using SSE via MCP specs from v2024-11-05.

Tested with MCP inspector that removing this does not break SSE connection. Added regression test for checking security with allowedOrigin and allowedHost values.

🛠️ Fixes #3053

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -367,7 +367,6 @@ func sseHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/event-stream")
 	w.Header().Set("Cache-Control", "no-cache")
 	w.Header().Set("Connection", "keep-alive")
-	w.Header().Set("Access-Control-Allow-Origin", "*")
 
 	// Define attributes for session metrics
 	networkProtocolVersion := fmt.Sprintf("%d.%d", r.ProtoMajor, r.ProtoMinor)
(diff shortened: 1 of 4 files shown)


kurtisvg (changes requested):
Can we ensure we add a regression test?

kurtisvg (approved):

PR #3056, opened 2026-04-14
fix(tools/bigquery-execute-sql): avoid surfacing invalid queries as MCP 500s

### LINKED ISSUE #3055: fix(bigquery-execute-sql): invalid BigQuery query errors are surfaced as MCP 500s
## Description

When the `bigquery-execute-sql` tool receives a user/query validation error from BigQuery during the dry-run phase, it currently wraps that error as a `ClientServerError` with `http.StatusInternalServerError`.

That classification causes the MCP layer to return a JSON-RPC internal error and an HTTP 500 instead of a normal tool error result. In practice, this makes LLM clients treat normal BigQuery SQL mistakes as transport failures.

## Repro

In our sandbox, this showed up when an MCP client called `execute_sql` with invalid-but-common BigQuery SQL such as:

```sql
SELECT *
FROM `project.dataset.table`
WHERE created_at >= DATE_SUB(CURRENT_DATE(), INTERVAL 6 MONTH)
```

against a `TIMESTAMP` column, or with `DATE_TRUNC(timestamp_col, WEEK)` instead of `TIMESTAMP_TRUNC(...)`.

BigQuery correctly returns a 400-series query validation error like `invalidQuery` / `No matching signature ...`, but the toolbox currently turns that into an MCP HTTP 500.

## Root cause

`internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go` currently does:

```go
if err != nil {
    return nil, util.NewClientServerError("query validation failed", http.StatusInternalServerError, err)
}
```

The similar `bigquery-sql` tool already uses `util.ProcessGcpError(err)` instead, which preserves auth failures as transport errors but treats normal GCP request/query failures as agent/tool errors.

## Expected behavior

For invalid BigQuery SQL submitted through MCP:

- the toolbox should not emit HTTP 500
- the MCP response should be a normal tool result with `isError: true`
- the BigQuery error message should stay visible to the client so the model/user can correct the query

## Proposed fix

Use `util.ProcessGcpError(err)` in `bigquery-execute-sql` for dry-run validation failures, matching the existing `bigquery-sql` path.

I have a small PR ready with that change plus a focused regression test for the GCP error classifier.

jack-larch:
## Description

Route `bigquery-execute-sql` dry-run validation failures through the shared GCP error processor instead of hardcoding them as internal server errors.

This keeps normal BigQuery query mistakes in the tool-error path rather than escalating them into MCP transport failures.

## PR Checklist

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/mcp-toolbox/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

Notes:
- I could not run `go test` locally in this environment because the Go toolchain is not installed on this machine.
- I added a focused regression test around `ProcessGcpError` so the intended classification is explicit in CI.

🛠️ Fixes #3055

Made with [Cursor](https://cursor.com)


duwenxin99 (approved):
@jack-larch Thanks for opening the PR! Could you make sure to sign the [CLA](https://cla.developers.google.com/)? Thanks!

duwenxin99 (approved):

PR #3060, opened 2026-04-15
feat(server): add /healthz endpoint for container health checks

### LINKED ISSUE #2644: Add /healthz endpoint or healthcheck CLI command for Docker health checks
## Problem

The Toolbox Docker image uses `gcr.io/distroless/cc-debian12:nonroot` as the base image, which contains no shell, curl, wget, or any other utilities. This makes it impossible to configure Docker `HEALTHCHECK` for the container.

```yaml
# None of these work - distroless has no shell or HTTP clients
healthcheck:
  test: ["CMD-SHELL", "curl -sf http://localhost:5000/healthz"]  # no shell, no curl
  test: ["CMD", "wget", "--spider", "http://localhost:5000/"]     # no wget
```

The only option is `healthcheck: { disable: true }`, which means Docker/Compose cannot track whether the service is actually healthy.

## Proposed solution

One or both of:

1. **`/healthz` endpoint** - lightweight GET endpoint returning 200 when the server is ready to accept requests. Many Go servers already do this (e.g. Kubernetes MCP server exposes `/healthz`).

2. **`toolbox healthcheck` CLI subcommand** - a built-in command that checks if the main process is responding, usable as `HEALTHCHECK CMD ["toolbox", "healthcheck"]`. This is a common pattern for distroless images (e.g. [Traefik](https://doc.traefik.io/traefik/operations/cli/#healthcheck), [Loki](https://grafana.com/docs/loki/latest/operations/storage/wal/)).

## Why this matters

- `docker compose ps` shows no health status, making monitoring harder
- Orchestrators (Docker Swarm, Nomad) rely on healthchecks for restarts and rolling updates
- Kubernetes users can use `httpGet` probes directly, but Docker/Compose users have no workaround without rebuilding the image

## Current workaround

```yaml
healthcheck:
  disable: true
```

This disables health monitoring entirely. The container will restart on crash (`restart: unless-stopped`) but won't be detected if it hangs or stops accepting connections.

## Environment

- Toolbox version: 0.26.0
- Image: `us-central1-docker.pkg.dev/database-toolbox/toolbox/toolbox:0.26.0`
- Base: `gcr.io/distroless/cc-debian12:nonroot`
- Runtime: Docker Compose

saivedant169:
Fixes #2644

## Summary

Adds a `/healthz` endpoint to the Toolbox HTTP server so container orchestrators (Kubernetes liveness/readiness probes, Docker `HEALTHCHECK`, Cloud Run startup probes) have a dedicated, lightweight path to hit. The response is `HTTP 200` with a JSON body of `{"status":"ok"}`, so probes can check either the status code or the payload depending on their configuration.

## Why a separate endpoint

The existing `/` handler is a landing page that returns a greeting string. Reusing it for health checks is fine today but couples probe behavior to a user-facing route, and the non-JSON body makes it awkward for tooling that parses health responses. Giving probes their own path follows the convention most Go services already use and keeps `/` free to evolve as a human-facing entry point.

## Implementation

- Registered `r.Get("/healthz", ...)` in `internal/server/server.go` right after the default `/` handler, so it inherits the same CORS and host-check middleware already applied at the router level.
- Returns `Content-Type: application/json` with body `{"status":"ok"}`.
- No new dependencies.

## Testing

Added `TestHealthz` in `internal/server/server_test.go`. It follows the same pattern as `TestServe`: spins up a real server on a free port, sends a GET to `/healthz`, and verifies the status code, the `Content-Type` header, and the JSON body. Runs on port `5004` to avoid collisions with other tests in the package.

```
$ go test ./internal/server/ -run "TestServe|TestHealthz" -count=1
ok  	github.com/googleapis/mcp-toolbox/internal/server	1.283s

$ go test ./internal/server/ -count=1
ok  	github.com/googleapis/mcp-toolbox/internal/server	2.220s
```

Also verified `go vet ./internal/server/...` and `gofmt -l` are clean.


saivedant169:
Pushed a fix for both items in commit 27d8d45.

1. `/healthz` and the host check: I went with the in-middleware bypass rather than a separate router group. The handler now short-circuits the host validation only for `/healthz`, so DNS rebinding protection still applies to every other path. A new `TestHealthzBypassesHostCheck` confirms `/healthz` returns 200 and `/` returns 403 when `AllowedHosts` does not include the request host.

2. Race on `err`: replaced the goroutine's reassignment with a shadowed `serveErr` local. Verified clean with `go test -race`.

duwenxin99 (changes requested):
Hi @saivedant169, thanks for opening the PR! Sorry for the delayed review as we got many PRs coming in lately. The /healthz endpoint addition looks good to me overall, but could you fix the gemini review comments and the test failures? After that, I'll help you trigger the tests and merge it. Thank you!

saivedant169:
Yes I will do it in an hour.

saivedant169:
@duwenxin99 pushed 57add88 addressing the Gemini review:

- /healthz handler now uses render.JSON instead of writing the JSON body and Content-Type by hand, matching the rest of api.go.
- Both healthz tests switched off the hardcoded 5004/5005 ports onto port 0 with s.Addr() so the OS assigns a free port. Avoids the CI port-collision flake the bot flagged.
- Also bumped the s.Listen calls in the two tests to the (ctx, certFile, keyFile) signature that landed on main after the rebase, so the file builds again.

Ran locally:
```
go test -race ./internal/server/...
```
All green. Ready for another look + CI trigger.

saivedant169:
Hi @duwenxin99, the review comments are addressed in the latest push: the /healthz handler uses render.JSON, and both tests use port 0 with s.Addr() instead of the fixed 5004/5005 ports. Local run of go test ./internal/server/... passes. Whenever you get a chance to trigger CI and take another look, I think it is ready.

duwenxin99 (approved):

saivedant169:
Thanks @duwenxin99. Quick heads up: the `/gcbrun` above ran on `ac35096`, and I pushed `e07f29c` right after that fixes the lint failure. Two `SetupOTel` calls in `server_test.go` were still missing the `telemetryServiceName` argument that the recent main merge added, so lint (and the unit tests) failed on the typecheck. The latest commit adds it to both, matching the other call sites. `go vet` and the server tests pass locally.

Whenever you get a chance, could you approve and re-run the workflows on `e07f29c`? Thank you!

saivedant169:
Thanks @duwenxin99. Both red checks look unrelated to this change:

- macOS unit failed on `TestSingleEdit` (`cmd` package) with a 60s file-watcher startup timeout. ubuntu and windows unit both passed the same commit, and this PR only touches `internal/server`.
- The `/healthz` change is scoped to that one path, so the integration failure looks like a flake on the Cloud Build side rather than something this diff affects.

Could you re-run the macOS job and `/gcbrun` when you get a chance? Happy to dig in if either reproduces.

saivedant169:
Thanks for the review and the merge, @duwenxin99. Appreciate the quick re-runs.

duwenxin99:
@saivedant169 Thank you for the contribution!

PR #3067, opened 2026-04-15
fix(mcp): implement router-level logger injection for MCP auth

### LINKED ISSUE #3076: Open authentication bypass in MCP middleware
### Summary
A critical authentication bypass exists in `internal/server/server.go` within `mcpAuthMiddleware`. If an external Authorization Server using generic opaque tokens experiences an outage, timeout, or DNS failure (`Code: 500`), the middleware fails open, granting an unauthenticated attacker full access to the MCP SSE session and backend tools.

### Technical Details
In `internal/server/server.go:mcpAuthMiddleware`, the authorization error is handled using a switch statement over `mcpErr.Code`:
```go
if err := mcpSvc.ValidateMCPAuth(r.Context(), r.Header); err != nil {
	var mcpErr *generic.MCPAuthError
	if errors.As(err, &mcpErr) {
		switch mcpErr.Code {
		case http.StatusUnauthorized:
			// Returns 401
			return
		case http.StatusForbidden:
			// Returns 403
			return
		}
	}
	// BUG: Exits the if block and unconditionally falls through!
}
next.ServeHTTP(w, r)
```
If an attacker deliberately sends a fake opaque token and the external introspection endpoint fails or times out, `ValidateMCPAuth` returns an `*MCPAuthError` with `Code: http.StatusInternalServerError`. The `switch` statement misses the `500` HTTP status code, exits the error block, and immediately calls `next.ServeHTTP(w, r)`. This completely bypasses authentication and allows the unauthenticated request to proceed.

### Impact
* **Fail-Open Authentication Bypass**: Unauthenticated actors can bypass security checks if the introspection server is temporarily degraded or offline. 
* **Privilege Escalation**: Allows unauthorized external actors to execute internal database tools (SQL querying, BigQuery execution, etc.) without valid tokens.
* **Exploitability**: Attackers can actively trigger this by performing an HTTP/DNS DoS against the Authorization Server, forcing a `500` introspection error, and subsequently unlocking the Toolbox framework.

### Recommended Mitigation
Change `mcpAuthMiddleware` to fail-closed immediately by applying a `default` case in the switch block, and a catch-all `http.Error` for any non-MCPAuthError standard errors:

```go
		default:
			http.Error(w, mcpErr.Message, mcpErr.Code)
			return
		}
	}
	http.Error(w, "Internal Server Error during authentication", http.StatusInternalServerError)
	return
}
next.ServeHTTP(w, r)
```

CC: @averikitsch @Yuan325

duwenxin99:
- Any unexpected errors during token validation now explicitly deny access rather than bypassing security checks.
- Moved the injection of the logger into the request context to the router level. This ensures that the logger is available to all subsequent middlewares and handlers.

Fix: https://github.com/googleapis/mcp-toolbox/issues/3076


averikitsch (commented):
@Yuan325 can you be final approver.

internal/server/mcp.go
```diff
@@ -333,6 +333,13 @@ func mcpRouter(s *Server) (chi.Router, error) {
 	r.Use(middleware.AllowContentType("application/json", "application/json-rpc", "application/jsonrequest"))
 	r.Use(middleware.StripSlashes)
 	r.Use(render.SetContentType(render.ContentTypeJSON))
+	// Inject logger into ctx
+	r.Use(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := util.WithLogger(r.Context(), s.logger)
```
Yuan325, line 339: Do you know which part require the logger to be set in the context? We did set it in the httpHandler.

For the `r.Use()` that adds log handler, we already have this: https://github.com/googleapis/mcp-toolbox/blob/03deb66b077aaa3282582244864421f024ca45df/internal/server/server.go#L366

What do we need to re-add logger to the router here?

internal/auth/generic/generic.go
Yuan325, line 1: Do we need to update the error at https://github.com/googleapis/mcp-toolbox/blob/0917e5bf75ca4d29261bf630a7ce4cdab3d9949b/internal/auth/google/google.go#L77 too?

duwenxin99: No, that is a separate feature and the error is not checked in the `mcpAuthMiddleware`. The claims error will be checked in tool handlers.

internal/server/mcp.go
duwenxin99: This logger is retrievable via `httplog.LogEntry(ctx)`. It does not inject the logger using the key that `util.LoggerFromContext` expects. It seems like we are not retrieving this logger anywhere. Should we replace it with `util.WithLogger` instead?

internal/auth/generic/generic.go
```diff
@@ -324,7 +324,7 @@ func (a AuthService) validateOpaqueToken(ctx context.Context, tokenStr string) e
 	}
 
 	if err := json.Unmarshal(body, &introspectResp); err != nil {
-		return fmt.Errorf("failed to parse introspection response: %w", err)
+		return &MCPAuthError{Code: http.StatusInternalServerError, Message: fmt.Sprintf("failed to parse introspection response: %v", err), ScopesRequired: a.ScopesRequired}
```
averikitsch: Why do we need to update this? The server doesn't parse 500s  https://github.com/googleapis/mcp-toolbox/blob/bc8092e932f22544ce82a698549d666a2bd4a089/internal/server/server.go#L513.either need to add it to the switch statement or have a catch all for all other errors

internal/server/mcp.go
```diff
@@ -333,6 +333,13 @@ func mcpRouter(s *Server) (chi.Router, error) {
 	r.Use(middleware.AllowContentType("application/json", "application/json-rpc", "application/jsonrequest"))
 	r.Use(middleware.StripSlashes)
 	r.Use(render.SetContentType(render.ContentTypeJSON))
+	// Inject logger into ctx
+	r.Use(func(next http.Handler) http.Handler {
```
averikitsch, line 337: Is there any other way to get this from context? This leads me to believe there is a performance issue constructing this on every request

internal/auth/generic/generic.go
duwenxin99: The error parsing logic is already updated in this PR. But you are right it is not necessary to make this change so I removed it. I originally changed it in case if we want to add other specific error code parsing in the future, but it seems not necessary for now.

internal/server/mcp.go
duwenxin99: `context.WithValue` is a very lightweight operation. It simply creates a small struct that points to the parent context so it shouldn't impact the performance. And also we are already doing it for the regular tool call endpoint: https://github.com/googleapis/mcp-toolbox/blob/bc8092e932f22544ce82a698549d666a2bd4a089/internal/server/api.go#L116

hits313:
hi @duwenxin99  and team

Thanks for landing this  the fail-closed default is correct, 
and both error paths from the original report are blocked.

One small suggestion before merge: the new TestMCPAuthMiddleware 
covers the malformed-introspection-JSON case but not the 
unreachable-introspection-endpoint case (which produced 
MCPAuthError{Code: 500} in the original PoC). A short variant 
with mockOIDC.Close() or a closed port would close the gap.

Happy to open a follow-up PR if helpful.

Hitarth

internal/server/server.go
```diff
@@ -527,6 +527,10 @@ func mcpAuthMiddleware(s *Server) func(http.Handler) http.Handler {
 						return
 					}
 				}
+				// Fail closed on unexpected errors
+				s.logger.ErrorContext(r.Context(), "unexpected error during MCP auth validation", "error", err)
+				http.Error(w, "Internal Server Error", http.StatusInternalServerError)
```
Deeven-Seru: Using http.Error hardcodes Content-Type: text/plain, is breaking the JSON-RPC contract expected by all MCP clients. this is  trading an auth bypass for a guaranteed client-side parser crash; I think its better if it returns a properly formatted JSON error payload instead.

internal/server/server_test.go
```diff
@@ -547,3 +549,173 @@ func TestLegacyAPIGone(t *testing.T) {
...
+			req.Header.Set("Authorization", "Bearer "+tc.token)
+			req.Header.Set("Content-Type", "application/json")
+
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatalf("request failed: %v", err)
+			}
+			defer resp.Body.Close()
+
+			if resp.StatusCode != tc.wantStatusCode {
```
Deeven-Seru, line 1119: This assertion only validates the HTTP status code, acting as a false positive that completely masks the text/plain regression introduced above. A robust API test must also assert the Content-Type header and validate the response payload i guess WDUT?.

Yuan325 (approved):

internal/server/server_test.go
duwenxin99: Thanks for the suggestion. Updated.

internal/server/server.go
duwenxin99: Great point. Updated.

internal/server/server.go
```diff
@@ -520,14 +521,21 @@ func mcpAuthMiddleware(s *Server) func(http.Handler) http.Handler {
...
 					case http.StatusForbidden:
 						w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer error="insufficient_scope", scope="%s", resource_metadata="%s", error_description="%s"`, strings.Join(mcpErr.ScopesRequired, " "), s.toolboxUrl+"/.well-known/oauth-protected-resource", mcpErr.Message))
-						http.Error(w, mcpErr.Message, http.StatusForbidden)
+						render.Status(r, http.StatusForbidden)
+						render.JSON(w, r, jsonrpc.NewError(nil, jsonrpc.INVALID_REQUEST, mcpErr.Message, nil))
 						return
 					}
 				}
+				// Fail closed on unexpected errors
+				s.logger.ErrorContext(r.Context(), "unexpected error during MCP auth validation", "error", err)
```
averikitsch, line 535: Why not use "util.LoggerFromContext(r.Context())"?

internal/server/server.go
```diff
@@ -520,14 +521,21 @@ func mcpAuthMiddleware(s *Server) func(http.Handler) http.Handler {
 							scopesArg = fmt.Sprintf(`, scope="%s"`, strings.Join(mcpErr.ScopesRequired, " "))
 						}
 						w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer resource_metadata="%s"%s`, s.toolboxUrl+"/.well-known/oauth-protected-resource", scopesArg))
-						http.Error(w, mcpErr.Message, http.StatusUnauthorized)
+						render.Status(r, http.StatusUnauthorized)
+						render.JSON(w, r, jsonrpc.NewError(nil, jsonrpc.INVALID_REQUEST, mcpErr.Message, nil))
```
averikitsch: INVALID_REQUEST is a different error than StatusUnauthorized

averikitsch (approved):

internal/server/server.go
averikitsch: I would recommend something like "render.JSON(w, r, jsonrpc.NewResponse(r.Context(), "1", result))"

internal/server/server.go
duwenxin99: Because we already save this logger to server in the `NewServer()` function, so we don't need to get this from context and check erro ragain.

internal/server/server.go
duwenxin99: Shouldn't we return jsonrpc error instead of a regular response? How about creating our custom jsonrpc error code for auth errors?

PR #3068, opened 2026-04-15
feat(auth/generic): support custom introspection endpoints

duwenxin99:
This PR extends the generic authentication service to support custom introspection formats, specifically enabling integration with Google's tokeninfo endpoint for validating opaque access tokens. Added e2e integration test with Google OAuth.


docs/en/documentation/configuration/authentication/generic.md
```diff
@@ -124,6 +123,22 @@ scopesRequired:
...
+
+```yaml
+kind: authServices
+name: google-auth
+type: generic
+audience: "YOUR_GOOGLE_CLIENT_ID.apps.googleusercontent.com"
+authorizationServer: https://accounts.google.com
+introspectionEndpoint: https://www.googleapis.com/oauth2/v1/tokeninfo
+introspectionMethod: GET
+introspectionParamName: access_token
```
averikitsch, line 138: Can we not get these from the initial request?

duwenxin99: We do attempt to discover the `introspection_endpoint` automatically by querying the provider's `/.well-known/openid-configuration`, which works for many OAuth providers. 

However, we must explicitly configure them for Google because Google OAuth does not host the endpoint: Google's OIDC discovery document does not include an `introspection_endpoint` field.

Google OAuth also has a non-standard implementation for opaque tokens. The OAuth2 Token Introspection standard ([RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662)) specifies using a `POST` request with a parameter named `token`. Google's legacy tokeninfo endpoint requires a `GET` request and the parameter name `access_token`.

These fields are not required for most other OAuth providers that provide standard OAuth2 introspection endpoints though.

docs/en/documentation/configuration/authentication/generic.md
```diff
@@ -124,6 +123,22 @@ scopesRequired:
...
 
+#### Google Opaque Access Token Validation Example
+
+To use Google's `tokeninfo` endpoint for validating opaque access tokens, configure the service to use the `GET` method and `access_token` parameter name:
+
+```yaml
+kind: authServices
+name: google-auth
+type: generic
+audience: "YOUR_GOOGLE_CLIENT_ID.apps.googleusercontent.com"
```
averikitsch, line 134: We may want to add a note like "The only time the aud claim matches the client_id is inside an ID Token (a concept from OpenID Connect used to verify a user's identity). Because an ID token is intended to be consumed by the client application itself, the client is the audience."

docs/en/documentation/configuration/authentication/generic.md
```diff
@@ -124,6 +123,22 @@ scopesRequired:
...
+authorizationServer: https://accounts.google.com
+introspectionEndpoint: https://www.googleapis.com/oauth2/v1/tokeninfo
```
averikitsch:
```suggestion
introspectionEndpoint: https://www.googleapis.com/oauth2/v3/tokeninfo
```

docs/en/documentation/configuration/authentication/generic.md
```diff
@@ -124,6 +123,22 @@ scopesRequired:
   - write
 ```
 
+#### Google Opaque Access Token Validation Example
```
averikitsch, line 126: In this PR or separate can we add the Okta example?

averikitsch (approved):

docs/en/documentation/configuration/authentication/generic.md
duwenxin99: Added Okta example.

PR #3081, opened 2026-04-16
feat(source/cloud-storage): add Cloud Storage source with list_objects and read_object tools

huangjiahua:
## Description

Adds Google Cloud Storage as a first-class source in MCP Toolbox, enabling LLM agents to work with objects across buckets in a GCP project. The source is project-scoped and authenticates via Application Default Credentials, mirroring Firestore/Bigtable.

This first PR ships the source plus two read-only tools from the approved design (14 total):

- **`cloud-storage-list-objects`** — prefix filter, delimiter-based grouping (returns `prefixes`), and pagination via `max_results` / `page_token`. Passes through whatever metadata the GCS client returns (`*storage.ObjectAttrs`) so we don't have to plumb new fields later.
- **`cloud-storage-read-object`** — reads an object's bytes, textual data only, with optional HTTP-style byte ranges (`bytes=0-999`, `bytes=-500`, `bytes=500-`).

GCS-aware error categorization (per [DEVELOPER.md](../blob/main/DEVELOPER.md#tool-invocation--error-handling)) is implemented in a new `cloudstoragecommon` helper that maps GCS sentinels and `*googleapi.Error` codes to Agent errors (missing bucket/object, bad request, unsatisfiable range) vs. Server errors (auth, IAM denial, quota, 5xx, context cancellation). This replaces the coarse `util.ProcessGcpError` for the two new tools.

Remaining 12 tools from the design doc (`list_buckets`, `create_bucket`, `copy/move/delete_object`, etc.) will land in follow-up PRs.

**CI note:** the `cloud-storage` shard in `.ci/integration.cloudbuild.yaml` expects `CLOUD_STORAGE_PROJECT=$PROJECT_ID` and requires the test service account to have a Cloud Storage admin role in the test project. Integration test self-manages its own UUID-suffixed bucket with defer-based cleanup.

## PR Checklist

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/mcp-toolbox/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea (communicated internally)

## What's included

- New source: `internal/sources/cloudstorage/` (+ YAML-parse unit tests)
- Two tools: `internal/tools/cloudstorage/cloudstoragelistobjects/`, `.../cloudstoragereadobject/` (+ YAML-parse + range-parser unit tests)
- New `cloudstoragecommon` error classifier (+ 17-case unit test covering sentinels, HTTP statuses, `context.Canceled`/`DeadlineExceeded`, and fallback)
- Integration test: `tests/cloudstorage/cloud_storage_integration_test.go` — 12 sub-tests against a real bucket (self-created, self-cleaned)
- Docs: `docs/en/integrations/cloud-storage/` (source + both tool pages; passes `.ci/lint-docs-{source,tool}-page.sh`)
- CI shard: `cloud-storage` in `.ci/integration.cloudbuild.yaml`
- Dependency: `cloud.google.com/go/storage v1.62.1`

Opening as **draft** for initial review — happy to split the error-classifier refactor into a separate commit if reviewers prefer.


docs/en/integrations/cloud-storage/tools/cloud-storage-list-objects.md
```diff
@@ -0,0 +1,56 @@
+---
+title: "cloud-storage-list-objects"
```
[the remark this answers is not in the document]
huangjiahua: Thanks for the catch — I checked the repo convention against existing tool docs (e.g. `docs/en/integrations/firestore/tools/firestore-list-collections.md`, `bigquery-list-dataset-ids.md`, `bigtable-sql.md`) and they all use the bare kebab-case name without a ` Tool` suffix. Keeping the title consistent with the rest of the repo; happy to change if maintainers would prefer a project-wide rename.

docs/en/integrations/cloud-storage/tools/cloud-storage-read-object.md
```diff
@@ -0,0 +1,50 @@
+---
+title: "cloud-storage-read-object"
```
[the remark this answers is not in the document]
huangjiahua: Same as the list-objects thread — matching the existing `firestore-*`, `bigquery-*`, `bigtable-*` tool-page title convention in the repo.

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// pageSize returns the effective page size for pagination. The GCS API defaults
+// to 1000 when a non-positive value is supplied.
+func pageSize(maxResults int) int {
+	if maxResults <= 0 {
+		return 1000
+	}
+	return maxResults
+}
```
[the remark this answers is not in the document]
huangjiahua: Good catch — done in 483c70fb. `pageSize` now clamps both `<= 0` and `> 1000` to the GCS 1000-per-page max, matching the tool doc.

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -0,0 +1,181 @@
...
+// length == -1 means "read to end of object"; a negative offset means "suffix
+// from end" (in which case length must be -1).
+func (s *Source) ReadObject(ctx context.Context, bucket, object string, offset, length int64) (map[string]any, error) {
+	reader, err := s.Client.Bucket(bucket).Object(object).NewRangeReader(ctx, offset, length)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open object %q in bucket %q: %w", object, bucket, err)
+	}
+	defer reader.Close()
+
+	data, err := io.ReadAll(reader)
```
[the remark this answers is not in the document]
huangjiahua: Good catch — done in 483c70fb, updated to 8 MiB in f5fdac8e. `ReadObject` now rejects reads > 8 MiB via the `cloudstoragecommon.ErrReadSizeLimitExceeded` sentinel, which the classifier maps to an Agent error so the LLM can narrow the `range` and retry. Two integration sub-tests (`oversize_read_returns_agent_error`, `oversize_read_narrowed_by_range_succeeds`) cover both the rejection and recovery paths. Doc updated accordingly. Subsequently refactored in 91a459cc: the cap now lives as a private `defaultMaxReadBytes` in the source package; only the sentinel stays in `cloudstoragecommon`.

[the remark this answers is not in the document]
huangjiahua: actually changed to 8 MiB

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -0,0 +1,209 @@
...
+		Client: client,
+	}
+	return s, nil
+}
+
+var _ sources.Source = &Source{}
+
+type Source struct {
+	Config
+	Client *storage.Client
```
Yuan325:
```suggestion
	client *storage.Client
```

Probably don't need to export this~

internal/sources/cloudstorage/pagesize_internal_test.go
Yuan325, line 1: can we move this test function into `cloudstorage_test.go` instead?

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -0,0 +1,209 @@
...
+		"contentType": reader.Attrs.ContentType,
+		"size":        len(data),
+	}, nil
+}
+
+// pageSize returns the effective page size for pagination. The GCS API caps
+// results at 1000 per page; we enforce the same cap here so callers don't
+// pre-allocate larger buffers and so the contract matches the tool's
+// 'max_results' documentation.
+func pageSize(maxResults int) int {
```
Yuan325, line 186: Should we trigger an AgentError in the tool during parameter extraction if the value exceeds 1,000? This makes the limit explicit to the agent/user, preventing confusion when the returned page count is lower than requested.

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -0,0 +1,209 @@
...
+// ListObjects lists objects in a bucket with optional prefix and delimiter filtering.
+// maxResults == 0 means return up to one page as returned by the GCS API. A non-empty
+// pageToken resumes listing from a prior call. The returned map contains "objects"
+// (raw *storage.ObjectAttrs entries as returned by the GCS client), "prefixes"
+// (common prefixes when a delimiter is set), and "nextPageToken" (empty when
+// there are no more results).
+func (s *Source) ListObjects(ctx context.Context, bucket, prefix, delimiter string, maxResults int, pageToken string) (map[string]any, error) {
+	it := s.Client.Bucket(bucket).Objects(ctx, &storage.Query{
+		Prefix:    prefix,
+		Delimiter: delimiter,
```
Yuan325, line 112: just confirming, will be okay if these 2 values are ""?

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -0,0 +1,209 @@
...
+	})
+	pager := iterator.NewPager(it, pageSize(maxResults), pageToken)
```
Yuan325, line 114: will this be okay if `pageToken` is `""`

internal/tools/cloudstorage/cloudstoragereadobject/cloudstoragereadobject_parse_range_test.go
Yuan325, line 1: Let's just move this test to `cloudstoragereadobject_test.go`

tests/cloudstorage/cloud_storage_integration_test.go
```diff
@@ -0,0 +1,462 @@
...
+		t.Fatal("'CLOUD_STORAGE_PROJECT' not set")
+	}
+	return map[string]any{
+		"type":    CloudStorageSourceType,
+		"project": CloudStorageProject,
+	}
+}
+
+func initStorageClient(ctx context.Context) (*storage.Client, error) {
+	return storage.NewClient(ctx, option.WithUserAgent("genai-toolbox-integration-test"))
```
Yuan325:
```suggestion
	return storage.NewClient(ctx)
```

Can we just init without the user agent option for int test?

tests/cloudstorage/cloud_storage_integration_test.go
```diff
@@ -0,0 +1,462 @@
...
+
+	waitCtx, waitCancel := context.WithTimeout(ctx, 10*time.Second)
+	defer waitCancel()
+	out, err := testutils.WaitForString(waitCtx, regexp.MustCompile(`Server ready to serve`), cmd.Out)
+	if err != nil {
+		t.Logf("toolbox command logs: \n%s", out)
+		t.Fatalf("toolbox didn't start successfully: %s", err)
+	}
+
+	runCloudStorageToolGetTest(t)
```
Yuan325: Probably wouldn't need this function. We can just utilize this existing function https://github.com/googleapis/mcp-toolbox/blob/2375ffcff198cda45886b9f912cdc11506381cbe/tests/tool.go#L90

ref: https://github.com/googleapis/mcp-toolbox/blob/2375ffcff198cda45886b9f912cdc11506381cbe/tests/looker/looker_integration_test.go#L358-L366

tests/cloudstorage/cloud_storage_integration_test.go
```diff
@@ -0,0 +1,462 @@
...
+	first, _ := params[0].(map[string]any)
+	if first["name"] != "bucket" {
+		t.Fatalf("expected first parameter to be 'bucket', got %v", first["name"])
+	}
+	if required, _ := first["required"].(bool); !required {
+		t.Fatalf("expected 'bucket' parameter to be required, got %v", first)
+	}
+}
+
+func runCloudStorageListObjectsTest(t *testing.T, bucket string) {
```
Yuan325, line 222: For this function and `runCloudStorageReadObjectTest()`, is it possible to use the golang's table-driven tests? There's probably alot of duplication here in each `t.Run()`.

Reference: https://github.com/googleapis/mcp-toolbox/blob/2375ffcff198cda45886b9f912cdc11506381cbe/tests/tool.go#L231

Yuan325 (changes requested):
Hi @huangjiahua Thank you for the contribution! Please let me know if you need any clarifications

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: Can confirm. I've also add integration test with these 2 values being "".

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: Yes. Also added an integration test case.

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: Yes, I did that in internal/tools/cloudstorage/cloudstoragelistobjects/cloudstoragelistobjects.go

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Done

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Do you mean we can remove this test?

internal/sources/cloudstorage/pagesize_internal_test.go
huangjiahua: Done

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: Done

internal/tools/cloudstorage/cloudstoragereadobject/cloudstoragereadobject_parse_range_test.go
huangjiahua: Done

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Done

internal/sources/cloudstorage/cloudstorage.go
Yuan325: got it, in this case we probably do not have to check it here again? since page >1000 will throw an error so it will not get to the client.

huangjiahua: Done

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: I've already removed that test

tests/cloudstorage/cloud_storage_integration_test.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	runCloudStorageListObjectsTest(t, bucketName)
+	runCloudStorageReadObjectTest(t, bucketName)
+}
+
+func getCloudStorageToolsConfig(sourceConfig map[string]any) map[string]any {
+	return map[string]any{
+		"sources": map[string]any{
+			"my-instance": sourceConfig,
```
Yuan325:
```suggestion
			"my_instance": sourceConfig,
```

Sorry some nit. source and tool name's naming convention is underscore

tests/cloudstorage/cloud_storage_integration_test.go
```diff
@@ -0,0 +1,400 @@
...
+		},
+		"tools": map[string]any{
+			"my-list-objects": map[string]any{
```
Yuan325:
```suggestion
			"my_list_objects": map[string]any{
```

tests/cloudstorage/cloud_storage_integration_test.go
```diff
@@ -0,0 +1,400 @@
...
+				"type":        "cloud-storage-list-objects",
+				"source":      "my-instance",
+				"description": "List objects in a Cloud Storage bucket.",
+			},
+			"my-read-object": map[string]any{
```
Yuan325:
```suggestion
			"my_read_object": map[string]any{
```

tests/cloudstorage/cloud_storage_integration_test.go
Yuan325: sorry I meant that we can instead do the folllowing:
```
tests.RunToolGetTestByName(t, "my_list_objects", map[string]any{...})
tests.RunToolGetTestByName(t, "my_read_object", map[string]any{...})
```

This would be the same as your previous `runCloudStorageToolGetTest()`.

Yuan325 (changes requested):

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Done.

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Done

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Done

tests/cloudstorage/cloud_storage_integration_test.go
huangjiahua: Thanks. I added these tests

Yuan325 (approved):

PR #3094, opened 2026-04-17
docs: add read-only configuration guide

duwenxin99:


averikitsch (approved):
In the future we should move this to a top level doc under configuration section @anubhav756 .

PR #3095, opened 2026-04-17
chore: avoid silent exit by printing server error

### LINKED ISSUE #3063: `toolbox serve` exits silently with status 1 when `--config` is passed on darwin-arm64 (regression since ~v0.17)
### Summary
Passing `--config` (or `--config-folder`) to `toolbox serve` causes the process to exit immediately with status 1 and zero output on stdout or stderr — at any `--log-level` and with any config, including a minimal SQLite-only config and an empty YAML file. Calling `toolbox serve` without any config flag works normally. The same configs work correctly with `toolbox invoke`.

### Versions tested (darwin-arm64)

| Version | `serve --config <file>` | `invoke <tool> --config <file>` |
|---------|-------------------------|---------------------------------|
| v1.1.0 (2026-04-13) | ❌ silent exit 1 | ✅ works |
| v1.0.0 (2026-04-10) | ❌ silent exit 1 | ✅ works |
| v0.32.0 (2026-04-08) | ❌ silent exit 1 (with legacy `--tools-file`) | — |
| v0.16.0 (2025-11-19) | ✅ works (with legacy `--tools-file`) | — |

So the regression was introduced somewhere between v0.17 and v0.32 and has persisted through v1.1.0. It does not appear to have been reported previously — searched issues for "exit 1", "darwin", and "serve silent" with no matches.

### Repro on v1.1.0

```bash
# 1. Install (direct download from GCS, not npm)
curl -sSL https://storage.googleapis.com/genai-toolbox/v1.1.0/darwin/arm64/toolbox \
  -o /usr/local/bin/toolbox && chmod +x /usr/local/bin/toolbox
toolbox --version
# => toolbox version 1.1.0+binary.darwin.arm64.da6f5f8

# 2. Minimal config
sqlite3 /tmp/t.db "CREATE TABLE IF NOT EXISTS t(id INTEGER);"
cat > /tmp/tb.yaml <<'EOF'
sources:
  s1:
    kind: sqlite
    database: /tmp/t.db
tools:
  t1:
    kind: sqlite-sql
    source: s1
    description: test
    statement: "SELECT 1"
toolsets:
  default:
    - t1
EOF

# 3. invoke works:
toolbox invoke t1 --config /tmp/tb.yaml
# => [{"?column?": 1}]

# 4. serve fails silently:
toolbox --log-level DEBUG serve --config /tmp/tb.yaml --port 5050
# => (no output, exit status 1, no port bound)
echo $?  # => 1

# 5. serve without --config works:
toolbox serve --port 5051 &
# => 2026-04-15... INFO "Initialized 0 sources: "
#    ... "Initialized 1 toolsets: default"
# Port 5051 binds correctly.
```

### Variations tried (all identical: silent exit 1)

- `serve --config` with sources + tools + toolsets
- `serve --config` with sources + tools, no toolsets
- `serve --config` with sources only (no tools)
- `serve --config` with empty YAML file
- `serve --config-folder <dir>` containing a single YAML
- Postgres source (valid host), Postgres source (invalid host), SQLite source
- `--log-level DEBUG`, `--logging-format JSON`
- `nohup`, `script(1)`, direct fg invocation, stdin redirected from `/dev/null`
- `od -c` of captured output confirms truly zero bytes

Only the no-`--config` default case works.

### Expected
`toolbox serve --config <file>` either starts the HTTP server and logs initialization, or prints a clear error message to stderr explaining what's wrong with the config.

### Actual
Process exits with status 1 and produces no observable output on any stream.

### Environment
- macOS 25.1 (darwin-arm64, Apple Silicon)
- Binary source: `storage.googleapis.com/genai-toolbox/v1.1.0/darwin/arm64/toolbox`
- No env vars set that should affect toolbox
- Same behavior in fresh shell with minimal PATH

### Workaround
Use `toolbox invoke` CLI mode — reads the same config correctly. But this is not a standard MCP host integration path, so it is not a substitute for a working `serve` mode.

### Impact
Blocks MCP stdio / HTTP integration on macOS for all Toolbox versions shipped since ~v0.17 when a user provides a config. Given Toolbox's extensive Google-cloud DB catalog (Firestore, BigQuery, Spanner, AlloyDB, etc.), `serve` is the primary way developers would integrate Toolbox with MCP hosts like Claude Desktop / Cursor on their development machines.

### Bisectable
v0.16.0 works, v0.32.0 does not — 16 releases to bisect. Happy to help verify a fix.

Yuan325:
printing server startup error if an error was discovered. Previously it will just exit silently if theres an unknown flag caught by Cobra.

🛠️ Fixes #3063

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -87,6 +87,7 @@ func Execute() {
 	opts := internal.NewToolboxOptions()
 
 	if err := NewCommand(opts).Execute(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
 		exit := 1
 		os.Exit(exit)
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -87,7 +87,7 @@ func Execute() {
 	opts := internal.NewToolboxOptions()
 
 	if err := NewCommand(opts).Execute(); err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintf(opts.IOStreams.ErrOut, "Error: %v\n", err)
 		exit := 1
 		os.Exit(exit)
 	}


duwenxin99 (approved):
Can we use `opts.IOStreams.ErrOut` instead

PR #3112, opened 2026-04-22
feat(mcp): add URL parameter binding for HTTP transport

duwenxin99:
This PR implements URL Parameter Binding for the HTTP transport. 

- Filters out connection-level URL parameters from the client-facing schemas during tools/list and automatically injects them during tools/call.
- Performs automatic type conversion of incoming string query parameters to simple types (integer, boolean, number) and unmarshals complex types (array, object) from JSON-encoded values.


internal/tools/tools.go
Yuan325, line 1: This is updated. The McpManifest is now only generated when list method is called (tools list or pormpts list). It will be generated as needed, so we probbaly will not need to clone it anymore. 

The manifest generation code live in `internal/server/mcp/<version>/manifest.go`.

internal/server/mcp/v20251125/method.go
```diff
@@ -157,6 +164,39 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
 		return jsonrpc.NewError(id, jsonrpc.INTERNAL_ERROR, err.Error(), nil), err
 	}
 
+	// Auto-populate arguments from URL parameters
+	urlParams, ok := util.UrlParamsFromContext(ctx)
```
Yuan325: Should we add a function in `internal/util/parameters.go` for this?

docs/en/documentation/configuration/tools/url_parameter_binding.md
```diff
@@ -0,0 +1,84 @@
...
+title: "URL Parameter Binding"
+type: docs
+weight: 10
+description: >
+  How to bind tool arguments at the transport level using URL query parameters.
+---
+
+## About
+
+URL Parameter Binding is a transport-level feature for the SSE (HTTP) transport that allows you to bind specific arguments to tools at the connection level. This is useful for creating scoped connections for generic MCP clients where you want to restrict the client to a specific database, project, or instance without hardcoding these values in the server configuration for all users, or requiring the client to provide them.
```
Yuan325: This is not specific to SSE right?

docs/en/documentation/configuration/tools/url_parameter_binding.md
```diff
@@ -0,0 +1,84 @@
...
+
+## How It Works
+
+When an MCP client connects to the server via SSE with query parameters (e.g., `http://localhost:5000/mcp/sse?project=my-project`):
```
Yuan325: Is it just with SSE? Seems like we are making it work for all http transport in general (not sse specific).

Yuan325 (commented):
i'll take another look once the conflicts are resolved. Thank you! :)

internal/server/mcp.go
```diff
@@ -464,6 +468,20 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	ctx = util.WithLogger(ctx, s.logger)
 
+	queryParams := r.URL.Query()
```
Yuan325, line 498: Since this is within the `httpHandler()`, it will apply URL params for all http transport, not just SSE.

internal/server/mcp/v20251125/method.go
duwenxin99: It's fetching from context so I think it would be better to put it together with the rest of the functions fetching info from context (such as `InstrumentationFromContext` and `GenAIMetricAttrsFromContext`)  in `internal/util/util.go`.

internal/server/mcp.go
duwenxin99: Sorry, miscommunicated previously. It is for all HTTP transport and not just SSE. Updated the PR title.

docs/en/documentation/configuration/tools/url_parameter_binding.md
duwenxin99: Right. Sorry for the misinformation previously. It is for all http transport.

internal/tools/tools.go
duwenxin99: Updated.

docs/en/documentation/configuration/tools/url_parameter_binding.md
duwenxin99: No it is not.

docs/en/documentation/configuration/tools/url_parameter_binding.md
averikitsch: Can we update the wording here?

internal/server/mcp/v20241105/manifests.go
```diff
@@ -68,6 +68,13 @@ func generateParamManifest(ps parameters.Parameters) (InputSchema, map[string][]
 		}
 
 		name := p.GetName()
+		if urlParams != nil {
```
averikitsch, line 71: Do we have any helper logs for 
1) same param name but different type 
2) params that are in URL but aren't used?

averikitsch (approved):

internal/server/mcp/util/url_binding.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"strconv"
+
+	"github.com/googleapis/mcp-toolbox/internal/util"
+	"github.com/googleapis/mcp-toolbox/internal/util/parameters"
+)
+
+// PopulateUrlParams injects bound URL query parameters into the data arguments
+// and performs automatic type conversion for integer, boolean, and float parameters.
+func PopulateUrlParams(ctx context.Context, data map[string]any, toolParams parameters.Parameters) map[string]any {
```
Yuan325, line 28: can we add unit tests to this function?

internal/server/mcp/v20241105/manifests_test.go
```diff
@@ -188,7 +188,7 @@ func TestParamManifest(t *testing.T) {
 	}
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			gotSchema, gotAuthParam := generateParamManifest(tc.in)
+			gotSchema, gotAuthParam := generateParamManifest(tc.in, nil)
```
Yuan325: can we add a test case here where the `urlParams` is not nil?

Yuan325 (approved):

docs/en/documentation/configuration/tools/url_parameter_binding.md
duwenxin99: updated

PR #3126, opened 2026-04-24
feat: add support for HTTPS/TLS listener

Yuan325:
This PR introduces the ability to run the Toolbox server over HTTPS. While the server still defaults to HTTP for local development, users can now enable TLS encryption via command-line flags. This is essential for secure communication when the Toolbox is exposed over a network or used in production-like environments.

**New Flags:**
* `--tls`: Boolean flag to enable HTTPS.
* `--tls-cert`: String flag specifying the path to the PEM-encoded certificate file.
* `--tls-key`: String flag specifying the path to the PEM-encoded private key file.

**Use Case: How the Server Obtains .pem Files**
In a typical deployment, the server does not generate these files itself; it expects them to be provided by the environment. 

1. Local Development: Users can use tools like mkcert to generate a locally-trusted cert.pem and key.pem.
2. Production (Manual): Users obtain certificates from a Certificate Authority (CA) like Let's Encrypt via Certbot. Certbot handles the domain validation and saves the .pem files to a specific directory (e.g., /etc/letsencrypt/live/).
3. Execution: The user starts the Toolbox and points it to those specific paths:

    ```
    ./toolbox --tls --tls-cert=cert.pem --tls-key=key.pem
    ```

4. Loading: The server uses tls.LoadX509KeyPair to read these files from the disk and injects them into the listener before the HTTP server starts processing requests.

🛠️ Related https://github.com/googleapis/mcp-toolbox/issues/3113

### DIFF
--- a/cmd/internal/flags.go
+++ b/cmd/internal/flags.go
@@ -61,6 +61,8 @@ func ConfigFileFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 func ServeFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 	flags.StringVarP(&opts.Cfg.Address, "address", "a", "127.0.0.1", "Address of the interface the server will listen on.")
 	flags.IntVarP(&opts.Cfg.Port, "port", "p", 5000, "Port the server will listen on.")
+	flags.StringVar(&opts.Cfg.CertFile, "tls-cert", "", "Path to TLS certificate file")
+	flags.StringVar(&opts.Cfg.KeyFile, "tls-key", "", "Path to TLS key file")
 	flags.BoolVar(&opts.Cfg.Stdio, "stdio", false, "Listens via MCP STDIO instead of acting as a remote HTTP server.")
 	flags.BoolVar(&opts.Cfg.UI, "ui", false, "Launches the Toolbox UI web server.")
 	flags.BoolVar(&opts.Cfg.EnableAPI, "enable-api", false, "Enable the /api endpoint.")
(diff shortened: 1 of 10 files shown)


cmd/internal/flags.go
```diff
@@ -61,6 +61,9 @@ func ConfigFileFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 func ServeFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 	flags.StringVarP(&opts.Cfg.Address, "address", "a", "127.0.0.1", "Address of the interface the server will listen on.")
 	flags.IntVarP(&opts.Cfg.Port, "port", "p", 5000, "Port the server will listen on.")
+	flags.BoolVar(&opts.Cfg.UseTLS, "tls", false, "Enable HTTPS")
```
kurtisvg: Why do I need a flag to use tls and 2 more? can't we just say that if one of these flags is flipped we should use tls?

cmd/internal/flags.go
```diff
@@ -61,6 +61,9 @@ func ConfigFileFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
...
+	flags.StringVar(&opts.Cfg.CertFile, "tls-cert", "cert.pem", "Path to TLS certificate file")
+	flags.StringVar(&opts.Cfg.KeyFile, "tls-key", "key.pem", "Path to TLS key file")
```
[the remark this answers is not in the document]
kurtisvg: +1

cmd/internal/flags.go
kurtisvg: I think we should avoid hardcoding a default as well

kurtisvg (changes requested):

cmd/internal/flags.go
Yuan325: Removed the useTLS boolean

kurtisvg (approved):

PR #3129, opened 2026-04-24
feat(source/cloud-storage): add bucket and object management tools

huangjiahua:
## Description

Adds four native Cloud Storage tools:

- `cloud-storage-list-buckets` for project bucket discovery
- `cloud-storage-get-object-metadata` for object metadata inspection without reading payloads
- `cloud-storage-download-object` for downloading Cloud Storage objects to the Toolbox server filesystem
- `cloud-storage-upload-object` for uploading server-local files to Cloud Storage objects

This also wires the tool registrations, adds path validation and Cloud Storage error classification for local file operations, expands unit and integration coverage, and updates the Cloud Storage integration docs for the new tool surface.

## PR Checklist


docs/en/integrations/cloud-storage/tools/cloud-storage-get-object-metadata.md
```diff
@@ -0,0 +1,53 @@
+---
+title: "cloud-storage-get-object-metadata"
```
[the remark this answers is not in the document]
huangjiahua: Other tool doc does have this suffix.

docs/en/integrations/cloud-storage/tools/cloud-storage-list-buckets.md
```diff
@@ -0,0 +1,59 @@
+---
+title: "cloud-storage-list-buckets"
```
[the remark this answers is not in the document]
huangjiahua: Other tool doc does not have this suffix

internal/tools/cloudstorage/cloudstoragecommon/paths.go
```diff
@@ -0,0 +1,49 @@
...
+	if p == "" {
+		return "", fmt.Errorf("path is empty")
+	}
+	// Reject any ".." segment in the raw input. We check the raw input
+	// (not just the cleaned output) so that escapes like
+	// "/legit/../../etc/passwd" — which filepath.Clean collapses to an
+	// innocuous-looking absolute path — are still rejected. Legitimate
+	// names that happen to *contain* two dots (e.g. "foo..bar") are fine;
+	// only a standalone ".." segment is disallowed.
+	for _, seg := range strings.Split(p, string(filepath.Separator)) {
```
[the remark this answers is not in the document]
huangjiahua: Done.

docs/en/integrations/cloud-storage/tools/cloud-storage-download-object.md
```diff
@@ -0,0 +1,67 @@
...
+destination already exists.
+
+## Parameters
+
+| **parameter** | **type** | **required** | **description**                                                                                         |
+|---------------|:--------:|:------------:|---------------------------------------------------------------------------------------------------------|
+| bucket        |  string  |     true     | Name of the Cloud Storage bucket containing the object.                                                 |
+| object        |  string  |     true     | Full object name (path) within the bucket, e.g. `path/to/file.txt`.                                     |
+| destination   |  string  |     true     | Absolute local filesystem path where the object will be written. Relative paths and paths containing `..` are rejected. |
+| overwrite     | boolean  |    false     | If true, overwrite the destination when it already exists. If false, return an error when it exists.     |
```
Yuan325:
```suggestion
| overwrite     | boolean  |    false     | If true, overwrite the destination when it already exists. If false (default), return an error when it exists.     |
```

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -185,6 +189,132 @@ func (s *Source) ReadObject(ctx context.Context, bucket, object string, offset,
...
+		}
+		return nil, fmt.Errorf("failed to open destination %q: %w", destination, err)
+	}
+
+	n, err := io.Copy(f, reader)
+	if err != nil {
+		_ = f.Close()
+		return nil, fmt.Errorf("failed to write object %q to %q: %w", object, destination, err)
+	}
+	if err := f.Close(); err != nil {
```
Yuan325, line 264:
```suggestion
	if err := f.Sync(); err != nil {
		return fmt.Errorf("flush to disk failed: %w", err)
	}
	if err := f.Close(); err != nil {
```

Should we do `f.Sync()`?

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -185,6 +189,132 @@ func (s *Source) ReadObject(ctx context.Context, bucket, object string, offset,
...
+func (s *Source) ListBuckets(ctx context.Context, project, prefix string, maxResults int, pageToken string) (map[string]any, error) {
+	if project == "" {
+		project = s.Project
+	}
+	it := s.client.Buckets(ctx, project)
+	if prefix != "" {
+		it.Prefix = prefix
+	}
+	ps := maxResults
+	if ps <= 0 {
```
Yuan325, line 206: wondering if value < 0 should be accepted? If we want to accept it, we might want to modify the description to `A value under or equals to 0 uses the API default (1000)` to be consistent.

If we decide to throw an error, we can handle it at the tool level within `Invoke()` along when we are checking `if maxResults > maxResultsLimit`

internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -185,6 +189,132 @@ func (s *Source) ReadObject(ctx context.Context, bucket, object string, offset,
...
+	if contentType != "" {
+		w.ContentType = contentType
+	}
+
+	n, err := io.Copy(w, f)
+	if err != nil {
+		_ = w.Close()
+		return nil, fmt.Errorf("failed to copy %q to object %q in bucket %q: %w", source, object, bucket, err)
+	}
+	if err := w.Close(); err != nil {
```
Yuan325, line 301: similar question as above if we need to do `w.Sync()` before closing.

Yuan325 (commented):
Some clarifications needed, thank you!

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: There is no `w.Sync()`.  [cloud.google.com/go/storage@v1.62.1/writer.go](https://pkg.go.dev/cloud.google.com/go/storage#Writer)

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: I don't think we need it. Close already gives us page-cache durability (survives process crashes); Sync would only add protection against kernel panic / power loss, at the cost of a blocking fsync per download. Other similar tools (cp, gsutil cp, wget) don't fsync, and the operation is idempotent so a crashed download can just be retried. Happy to add it if you'd rather, but I'd lean toward leaving it off

internal/sources/cloudstorage/cloudstorage.go
huangjiahua: I added verification at tool level for both `list_objects` and `list_buckets` for negative values. I'm keeping this line unchanged just to be defensive and we need to map 0 to 1000 anyway.

docs/en/integrations/cloud-storage/tools/cloud-storage-download-object.md
huangjiahua: Done

Yuan325 (approved):

PR #3147, opened 2026-04-29
fix: mcpManifest return empty tools list instead of nil

Yuan325:
When there are no tools, return an empty list in McpManifest instead of null.

This PR also refactor to move all mocks to `testutils`. And added new `toolset_tests.go` file.

```
curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc": "2.0","id": 1,  "method": "tools/list"}' http://127.0.0.1:5000/mcp


{"jsonrpc":"2.0","id":1,"result":{"tools":[]}}
```

Previously UI will response with `Error: Invalid response format from toolset API.`. After the fix, UI will run with `no tools found`

---

Go nil slice vs empty slice --

`var tools []tools.Tool` (Nil Slice):
This creates a slice header where the internal pointer to the underlying array is nil. In Go logic, the length is 0 and the capacity is 0, but the slice itself is technically nil.

`tools := []tools.Tool{}` (Empty Slice):
This initializes the slice header. Even though the length is 0, the internal pointer points to a special "zerobase" memory address. It is not nil.

This is probably due to some memory state of variable between `nil` vs allocated.

### DIFF
--- a/internal/testutils/mocks.go
+++ b/internal/testutils/mocks.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/internal/testutils/mocks.go
+++ b/internal/testutils/mocks.go
@@ -1,4 +1,4 @@
-// Copyright 2026 Google LLC
+// Copyright 2025 Google LLC
 //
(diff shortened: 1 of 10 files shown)


internal/testutils/mocks.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 Google LLC
+// Copyright 2025 Google LLC
```
averikitsch: Please revert copyright changes

duwenxin99 (approved):

PR #3216, opened 2026-05-12
fix: bound MCP HTTP body size

### LINKED ISSUE #3103: Security: Unbounded request body on MCP HTTP handler enables memory exhaustion DoS
## Description

The MCP HTTP handler at `internal/server/mcp.go:469` reads the entire request body into memory using `io.ReadAll(r.Body)` with **no size limit**:

```go
// httpHandler handles all mcp messages.
func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
    // ...
    body, err := io.ReadAll(r.Body)  // No MaxBytesReader, no LimitReader
```

This allows an attacker to send a multi-gigabyte POST request to `/mcp/` and crash the server via OOM (Out of Memory).

### Impact

- **Denial of Service**: A single crafted HTTP request can crash the Toolbox server
- **No authentication required**: The body is read *before* any auth validation occurs
- **Affects all transport modes**: The HTTP/Streamable HTTP transport is affected

### Reproduction

```bash
# Generate a ~1GB payload and send it to the MCP endpoint
dd if=/dev/zero bs=1M count=1024 | curl -X POST -H "Content-Type: application/json" --data-binary @- http://localhost:5000/mcp/
```

### Suggested Fix

Apply `http.MaxBytesReader` before reading the body:

```go
const maxMCPBodySize = 10 << 20 // 10 MB
r.Body = http.MaxBytesReader(w, r.Body, maxMCPBodySize)
body, err := io.ReadAll(r.Body)
```

Note: The auth module already correctly uses `io.LimitReader(resp.Body, 1<<20)` for OIDC discovery responses — the same pattern should be applied here.

### Versions Affected

All versions (current `main` branch confirmed).

---

**PR Checklist**

Deeven-Seru:
## Summary
- guard the MCP HTTP transport with http.MaxBytesReader capped at 10 MiB
- translate over-sized bodies into HTTP 413 + JSON-RPC parse errors so clients see a deterministic failure
- cover the body limit with an HTTP test to prevent regressions

## Testing
- go test ./internal/server -run TestMcpRequestBodyLimit
fixes #3103


cmd/internal/flags.go
```diff
@@ -72,4 +73,5 @@ func ServeFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 	flags.StringVar(&opts.Cfg.McpPrmFile, "mcp-prm-file", "", "Path to a manual Protected Resource Metadata (PRM) JSON file. If provided, overrides auto-generation.")
 	flags.StringSliceVar(&opts.Cfg.AllowedOrigins, "allowed-origins", []string{"*"}, "Specifies a list of origins permitted to access this server. Defaults to '*'.")
 	flags.StringSliceVar(&opts.Cfg.AllowedHosts, "allowed-hosts", []string{"*"}, "Specifies a list of hosts permitted to access this server. Defaults to '*'.")
+	flags.Int64Var(&opts.Cfg.HttpMaxRequestBytes, "http-max-request-bytes", server.DefaultHTTPMaxRequestBytes, "Maximum MCP HTTP request body size in bytes.")
```
averikitsch, line 76: Thank you. Can you add this to docs/en/reference/cli.md?

internal/server/mcp.go
```diff
@@ -492,12 +492,20 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
 	// Read body first so we can extract trace context
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
 		// The id cannot be determined from an unreadable body. Per JSON-RPC 2.0,
 		// the response id MUST be null in that case.
 		// See https://www.jsonrpc.org/specification#response_object
+		var maxErr *http.MaxBytesError
+		if errors.As(err, &maxErr) {
+			err = fmt.Errorf("request body exceeds %d bytes", limit)
+			w.WriteHeader(http.StatusRequestEntityTooLarge)
```
averikitsch: nit: writeheader is not needed

averikitsch (changes requested):
We should also update /api endpoint https://github.com/googleapis/mcp-toolbox/blob/eede644599fd7c88775f40025db53d30eb08d381/internal/server/api.go#L203

averikitsch (approved):

PR #3245, opened 2026-05-18
feat(server): add support for toolset filtering in prebuilt CLI flag

duwenxin99:
Toolset Filtering for Prebuilt Configs: Enabled loading specific toolsets from a prebuilt configuration by appending a / suffix (e.g., --prebuilt sqlite/sqlite_database_tools), allowing users to load only a subset of tools rather than the entire configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ You can use the Toolbox in any MCP-compatible IDE or client (e.g., Gemini CLI, G
 
 2. Set the appropriate environment variables to connect, see the [Prebuilt Tools Reference](https://mcp-toolbox.dev/documentation/configuration/prebuilt-configs/).
 
-When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. You can also specify a specific toolset using the `--prebuilt=<database>/<toolset>` syntax (e.g., `--prebuilt=postgres/sql` to only load SQL tools). 
+When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. You can also specify a specific toolset using the `--prebuilt=<database>/<toolset>` syntax (e.g., `--prebuilt=postgres/data` to only load SQL tools). 
 
 Supported databases currently include:
 - **Google Cloud:** AlloyDB, BigQuery, Cloud SQL (PostgreSQL, MySQL, SQL Server), Spanner, Firestore, Knowledge Catalog (formerly known as Dataplex).
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ You can use the Toolbox in any MCP-compatible IDE or client (e.g., Gemini CLI, G
 
 2. Set the appropriate environment variables to connect, see the [Prebuilt Tools Reference](https://mcp-toolbox.dev/documentation/configuration/prebuilt-configs/).
 
-When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. 
+When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. You can also specify a specific toolset using the `--prebuilt=<database>/<toolset>` syntax (e.g., `--prebuilt=postgres/sql` to only load SQL tools). 
 
 Supported databases currently include:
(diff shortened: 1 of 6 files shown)


docs/en/reference/cli.md
averikitsch, line 1: Can we add a updates to the README too (or another high visible place)?

cmd/internal/options.go
averikitsch, line 287: We  should sanitize the name for the user agent. For example we  will auto add "+prebuilt.sqlite/sqlite_database_tools" to the useragent string but we should use "+prebuilt.sqlite.sqlite_database_tools"

cmd/internal/options.go
```diff
@@ -215,6 +223,30 @@ func (opts *ToolboxOptions) LoadConfig(ctx context.Context, parser *ConfigParser
 				logger.ErrorContext(ctx, errMsg.Error())
 				return isCustomConfigured, errMsg
 			}
+
+			if toolsetName != "" {
+				targetToolset, exists := parsed.Toolsets[toolsetName]
+				if !exists {
+					errMsg := fmt.Errorf("toolset '%s' not found in prebuilt configuration '%s'", toolsetName, sourceName)
```
averikitsch: Great job with this error message. nit: we could add which toolsets are available.

cmd/internal/options.go
```diff
@@ -202,7 +202,15 @@ func (opts *ToolboxOptions) LoadConfig(ctx context.Context, parser *ConfigParser
 		logger.InfoContext(ctx, logMsg)
 
 		for _, configName := range opts.PrebuiltConfigs {
-			buf, err := prebuiltconfigs.Get(configName)
+			sourceName := configName
+			toolsetName := ""
+			if strings.Contains(configName, "/") {
```
averikitsch: nit: if we get issues on confusion we could add a regex to check for extra characters periods, etc to tell the users to use "/" for toolsets

cmd/internal/options.go
```diff
@@ -215,6 +223,30 @@ func (opts *ToolboxOptions) LoadConfig(ctx context.Context, parser *ConfigParser
...
+					logger.ErrorContext(ctx, errMsg.Error())
+					return isCustomConfigured, errMsg
+				}
+
+				// Filter tools to only include those in the target toolset
+				filteredTools := make(server.ToolConfigs)
+				for _, tName := range targetToolset.ToolNames {
+					if tCfg, tExists := parsed.Tools[tName]; tExists {
+						filteredTools[tName] = tCfg
+					}
```
averikitsch, line 256: nit: we shouldn't need to log missing tools but there might be an edge case

averikitsch (changes requested):

README.md
```diff
@@ -95,7 +95,7 @@ You can use the Toolbox in any MCP-compatible IDE or client (e.g., Gemini CLI, G
 
 2. Set the appropriate environment variables to connect, see the [Prebuilt Tools Reference](https://mcp-toolbox.dev/documentation/configuration/prebuilt-configs/).
 
-When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. 
+When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. You can also specify a specific toolset using the `--prebuilt=<database>/<toolset>` syntax (e.g., `--prebuilt=postgres/sql` to only load SQL tools).
```
averikitsch:
```suggestion
When you run Toolbox with a `--prebuilt=<database>` flag, you instantly get access to standard tools to interact with that database. You can also specify a specific toolset using the `--prebuilt=<database>/<toolset>` syntax (e.g., `--prebuilt=postgres/data` to only load SQL tools). 
```

docs/en/documentation/configuration/prebuilt-configs/_index.md
```diff
@@ -19,6 +19,9 @@ You can now use `--prebuilt` along `--config`, `--configs`, or
 
 You can also combine multiple prebuilt configs.
 
+**Filtering Toolsets:**
+You can load a specific toolset from a prebuilt configuration by appending a `/` and the toolset name, for example: `--prebuilt=postgres/sql` to only load the SQL tools.
```
averikitsch:
```suggestion
You can load a specific toolset from a prebuilt configuration by appending a `/` and the toolset name, for example: `--prebuilt=postgres/data` to only load the SQL tools.
```

averikitsch (approved):

PR #3257, opened 2026-05-19
feat: respond 404 to client if something is not found by tool

drstrangelooker:


internal/util/errors.go
```diff
@@ -135,6 +135,15 @@ func ProcessGeneralError(err error) ToolboxError {
 		)
 	}
 
+	// Check for NotFound
+	if strings.Contains(errStr, "Error 404") || strings.Contains(errStr, "status 404") || strings.Contains(errStr, "status=404") {
+		return NewClientServerError(
```
duwenxin99, line 140: If the tool tries a access a resource that is not found, it could potentially be an agent error because the agent may have passed in a wrong resource identifier. Categorizing all 404 error as client server error could interrupt the client-side agent retry logic.

duwenxin99 (changes requested):

internal/util/errors.go
drstrangelooker: OK I'll abandon this then.

PR #3264, opened 2026-05-20
refactor(tools): introduce BaseTool with default Tool interface methods

twishabansal:
Add a BaseTool struct in internal/tools/tools.go that provides default implementations for the metadata, parameter, and authorization methods on the Tool interface. Concrete tools can embed BaseTool to drop their boilerplate and override only methods that need custom behavior.

BaseTool does not implement Invoke or ToConfig; concrete tools must provide those. The default EmbedParams passes a nil vector formatter, so tools writing into pgvector or similar must override EmbedParams.

This is the foundational PR for decoupling tools from sources at startup. Subsequent PRs will refactor existing tools to embed BaseTool; this PR introduces it without yet using it, so behavior is unchanged.

## PR Checklist

- [ ] Make sure you reviewed


internal/tools/tools.go
```diff
@@ -169,3 +169,70 @@ func GetCompatibleSource[T any](resourceMgr SourceProvider, sourceName, toolName
...
+// destructive) depends on the tool.
+type BaseTool struct {
+	Name             string
+	Description      string
+	Metadata         Manifest
+	StaticParameters parameters.Parameters
+	AuthRequired     []string
+	ScopesRequired   []string
+	Annotations      *ToolAnnotations
+}
```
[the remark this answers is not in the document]
twishabansal: Done.

internal/tools/tools.go
```diff
@@ -169,3 +169,70 @@ func GetCompatibleSource[T any](resourceMgr SourceProvider, sourceName, toolName
...
+	return b.Name
+}
+
+func (b BaseTool) GetDescription() string {
+	return b.Description
+}
+
+func (b BaseTool) GetAuthRequired() []string {
+	return b.AuthRequired
+}
```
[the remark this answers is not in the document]
twishabansal: Done.

internal/tools/tools.go
```diff
@@ -169,3 +169,69 @@ func GetCompatibleSource[T any](resourceMgr SourceProvider, sourceName, toolName
...
+
+func (b BaseTool) GetDescription() string {
+	return b.Description
+}
+
+func (b BaseTool) GetAuthRequired() []string {
+	return b.Metadata.AuthRequired
+}
+
+func (b BaseTool) GetAnnotations() *ToolAnnotations {
```
Yuan325: each tool's GetAnnotations() might be different. The annotations retrieval was moved from initialize() to GetAnnotations().

twishabansal: I am thinking of updating the annotations directly in the baseTool method in init. Something like [this](https://github.com/googleapis/mcp-toolbox/pull/3265/changes#diff-1d55b914a6cd9088d66f93ed6f1a2e52cd0c427e5e0221c1cbbb858547180377R111).

This ensures that the `GetAnnotations()` method returns the original response and the user does not ened to add a new `GetAnnotations` method. They can just updated the base tool struct during init.

Do you think it makes sense?

internal/tools/tools.go
```diff
@@ -169,3 +169,69 @@ func GetCompatibleSource[T any](resourceMgr SourceProvider, sourceName, toolName
...
+// write embeddings into pgvector (or similar) must override EmbedParams to
+// supply the right formatter.
+//
+// Annotations should be pre-resolved by the concrete tool's Initialize using
+// GetAnnotationsOrDefault, since the appropriate default (read-only vs
+// destructive) depends on the tool.
+type BaseTool struct {
+	Name             string
+	Description      string
+	Metadata         Manifest
```
Yuan325: Should this be named Manifest instead of Metadata to prevent confusion?

internal/tools/tools.go
```diff
@@ -169,3 +169,69 @@ func GetCompatibleSource[T any](resourceMgr SourceProvider, sourceName, toolName
...
+// those. The default EmbedParams passes a nil vector formatter — tools that
+// write embeddings into pgvector (or similar) must override EmbedParams to
+// supply the right formatter.
+//
+// Annotations should be pre-resolved by the concrete tool's Initialize using
+// GetAnnotationsOrDefault, since the appropriate default (read-only vs
+// destructive) depends on the tool.
+type BaseTool struct {
+	Name             string
+	Description      string
```
Yuan325: These values, and potentially the ScopesRequired is a duplicate of Config?

We previously removed these values from the `Tool struct` to prevent having 2 "source of truth". 
For example, in cases where the user update the name of a Tool, we don't want to update in both places (config.Name and BaseTool.Name). We probably want to remove these values here.

internal/tools/tools.go
Yuan325: Got it, make sense~

Yuan325 (approved):
One last comment, otherwise LGTM.

internal/tools/tools.go
twishabansal: Thanks for pointing out. Updated to avoid duplication while achieving the same results.

PR #3323, opened 2026-06-01
fix(source/dataplex): limit search results to pageSize

### LINKED ISSUE #3308: Dataplex source `SearchEntries` method fetches more entries than the page size
### Prerequisites

### Toolbox version

toolbox version 1.3.0+dev.linux.amd64

### Environment

1. OS type and version: `Linux 6.18.14-1rodete3-amd64 #1 SMP PREEMPT_DYNAMIC Deb`
  - Compiled from source: `go build`

### Client

1. Client: `curl` command
2. Version: N/A
3. Exact command:
```bash
curl -X POST http://localhost:5000/mcp \
      -d '{
        "jsonrpc": "2.0",
        "id": "1",
        "method": "tools/call",
        "params": {
          "name": "search_dataplex_entries",
          "arguments": {
            "query": "type:table"
          }
        }
      }'
```

### Expected Behavior

The server responds with at most `pageSize` entries (default 5). This is the expected behavior of the One MCP served Dataplex API.

### Current Behavior

The server makes repeated requests to the Dataplex API until the search is exhausted. If the query is "broad" (like `"type:table"`), this can potentially lead to the server issuing hundreds of sequential requests to the Dataplex API (for example, if there are 500 search result entries and the client is using a page size of 5), making it appear like the server has hung. Even if a sufficiently large page size is used, this is not ideal since it may flood the client LLM's context with hundreds of entries.

### Steps to reproduce?

`tools.yaml`:
```yaml
kind: source
name: my-dataplex-source
type: dataplex
project: <my project>
---
kind: tool
name: search_dataplex_entries
type: dataplex-search-entries
source: my-dataplex-source
description: Search for metadata entries in Dataplex Catalog
```

(Client is as specified previously)

### Additional Details

The methods `SearchAspectTypes`, and `SearchDataQualityScans` have the same behavior. Those may also need to be addressed.
The fix should be as simple as putting the following conditional at the end of the loops within these methods:
```go
		if len(results) >= pageSize {
			break
		}
```

sumedhdk22:
## 1. Description
This PR limits the results collection loops in `SearchEntries`, `SearchAspectTypes`, and `SearchDataQualityScans` to the user-provided `pageSize` value. 
By checking the result count at the top of the iterator loops and breaking early, the server avoids sequential REST/gRPC page fetching round-trips across large collections. This prevents the MCP server from hanging or flooding the client context with thousands of results for broad queries (such as `type=table`).
Additionally, this PR adds a new integration test case to verify the `pageSize` limit check.
## 2. PR Checklist
## 3. Issue Reference
Fixes #3308


harmonisha-wq (approved):

anikasharma03 (approved):

Yuan325 (approved):
Thank you!

PR #3333, opened 2026-06-02
fix(server): return null id for batch request rejection

### LINKED ISSUE #3332: Batch rejection response uses server-generated UUID instead of request id (JSON-RPC 2.0  violation)
### Prerequisites

- [ ] I've updated to the latest version of Toolbox

### Toolbox version

1.3.0+binary.linux.amd64.b001006

### Environment

_No response_

### Client

curl (direct HTTP POST to /mcp endpoint)

### Expected Behavior

Per JSON-RPC 2.0  ("The response MUST contain the same id as the request"), when toolbox rejects a batch request, the error response id must match one of the request ids sent by the client.

Acceptable responses:
  
  Option A — use the first request's id:
  {"jsonrpc":"2.0","id":11,"error":{"code":-32600,"message":"not supporting batch requests"}}

  Option B — use null (when the id cannot be determined):
  {"jsonrpc":"2.0","id":null,"error":{"code":-32600,"message":"not supporting batch requests"}}

  Either is spec-compliant. A server-generated UUID that was never in any request is not.
  

### Current Behavior

  When a batch request is rejected, toolbox returns a response whose id is a freshly generated UUID — not an id from any of the requests in the batch.

  Request sent:
  [
    {"jsonrpc":"2.0","id":11,"method":"tools/list"},
    {"jsonrpc":"2.0","id":12,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1"}}}
  ]
  
 Response received:
  {"jsonrpc":"2.0","id":"a0240994-6225-49be-a8c6-440eccce2942","error":{"code":-32600,"message":"not supporting batch requests"}}

 The id "a0240994-6225-49be-a8c6-440eccce2942" was never in any request. The original ids 11 and 12 receive no response at all.
  
  Impact: any client that tracks pending requests by id will never resolve ids 11 and 12, and will receive an unexpected response for an id it never sent — resulting in a client hang  or memory leak on unbounded pending maps.

### Steps to reproduce?

  1. Start toolbox locally (any prebuilt config works):
     SQLITE_DATABASE="/tmp/test.db" toolbox --prebuilt sqlite --port 5000

  2. Send a batch JSON-RPC request:
     curl -X POST http://127.0.0.1:5000/mcp \
       -d '[
         {"jsonrpc":"2.0","id":11,"method":"tools/list"},
         {"jsonrpc":"2.0","id":12,"method":"initialize",
          "params":{"protocolVersion":"2024-11-05","capabilities":{},
                    "clientInfo":{"name":"test","version":"1"}}}
       ]'
       
  3. Observe the response id — it will be a UUID not present in either request.

### Additional Details

  Found using mcp-scope (https://github.com/SSanju/mcp-scope), a passive JSON-RPC capture tool for MCP. Running mcp-scope check --strict on the capture flagged it as:

    [ERROR] line 25: response for unknown request id=a0240994-6225-49be-a8c6-440eccce2942
    [WARN ] line 23: unmatched request id=11 (no response in capture)
    [WARN ] line 24: unmatched request id=12 (no response in capture)
    FAIL — 1 error(s), 2 warning(s)
    
  Full capture file showing the exact frames: https://github.com/SSanju/mcp-scope/blob/main/examples/mcp-toolbox/capture.jsonl

twishabansal:
## Description

## PR Checklist

- [ ] Make sure you reviewed

🛠️ Fixes #3332


internal/server/mcp.go
```diff
@@ -646,8 +646,10 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
 		var a []any
 		unmarshalErr := json.Unmarshal(body, &a)
 		if unmarshalErr == nil {
+			// A batch has no single request id to echo back. Per JSON-RPC 2.0,
+			// when the id cannot be determined the response id MUST be null.
 			err = fmt.Errorf("not supporting batch requests")
-			return "", jsonrpc.NewError(id, jsonrpc.INVALID_REQUEST, err.Error(), nil), err
+			return "", jsonrpc.NewError(nil, jsonrpc.INVALID_REQUEST, err.Error(), nil), err
 		}
```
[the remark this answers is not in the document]
twishabansal: Will be fixed in a follow up PR

Yuan325 (approved):
Thank you!

internal/server/mcp.go
Yuan325: Will this value be used after this change? Should we remove this id?

internal/server/mcp.go
```diff
@@ -646,8 +646,10 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
 		var a []any
 		unmarshalErr := json.Unmarshal(body, &a)
 		if unmarshalErr == nil {
+			// A batch has no single request id to echo back. Per JSON-RPC 2.0,
```
Yuan325: can we add a reference link here?

internal/server/mcp.go
twishabansal: Good catch. The uuid was still used by the PARSE_ERROR return, but per the same JSON-RPC 2.0 rule (id MUST be null when it can't be determined), echoing back a fabricated uuid for a parse error is also non-compliant. Removed the uuid and now return null for the parse-error case too, and applied the same fix to the sibling unreadable-body block in `httpHandler`.

internal/server/mcp.go
twishabansal: Done — added a link to the JSON-RPC 2.0 spec (https://www.jsonrpc.org/specification#response_object) above each of the relevant returns.

twishabansal: Done.

internal/server/mcp.go
twishabansal: Removed. Thanks!

PR #3335, opened 2026-06-02
fix(server/auth): centralize tool scopes validation

duwenxin99:
All MCP protocol versions will validate auth token scopes to prevent authorization bypass through older protocol versions.

Reported by HE WEI (ギカク) @hewei-gikaku

### DIFF
--- a/internal/server/mcp/util/auth.go
+++ b/internal/server/mcp/util/auth.go
@@ -1,0 +1,72 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


averikitsch (approved):

internal/server/mcp/util/auth.go
```diff
@@ -0,0 +1,72 @@
...
+
+		scopeClaim, _ := claims["scope"].(string)
+		tokenScopes := strings.Fields(scopeClaim)
+
+		// Check if all required scopes are present in the token
+		missing := false
+		for _, ts := range toolScopes {
+			if !slices.Contains(tokenScopes, ts) {
+				missing = true
+				break
```
averikitsch: nit: you could just return here

PR #3338, opened 2026-06-03
fix(tool/spanner-sql,tool/spanner-execute-sql): use read-only annotations when readOnly is set

twishabansal:
## Description

The Spanner SQL tools unconditionally default their annotations to `NewDestructiveAnnotations`, even when the tool is configured with `readOnly: true`.
When `readOnly` is `true`, the tool now defaults to `NewReadOnlyAnnotations`, so clients and LLMs can tell the tool is safe to call. User-supplied `annotations` still take precedence via `GetAnnotationsOrDefault`.


Yuan325 (approved):
Thank you!

Issue #3113, opened 2026-04-22
Default DNS-rebinding protection allows all domains by defaults


JLLeitschuh:
### Expected Behavior

The googleapis/genai-toolbox doesn't, by default, validate the `Origin` header when requests are sent to it, thus violating one of the clauses of both the 2025-06-18 and 2025-11-25 MCP specifications.
 - 2025-06-18: [Base Protocol > Transports: 2.0.1 Security Warning](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#security-warning)
 - 2025-11-25:  [Base Protocol > Transports: 2.0.1 Security Warning](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#security-warning)

> # 2.0.1 Security Warning
> When implementing Streamable HTTP transport:
> 1. Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks
>    - If the Origin header is present and invalid, servers MUST respond with HTTP 403 Forbidden. The HTTP response body MAY comprise a JSON-RPC error response that has no id
> ...
>
> Without these protections, attackers could use DNS rebinding to interact with local MCP servers from remote websites.
> 
> \- 2025-11-25:  Base Protocol > Transports: 2.0.1 Security Warning

The Google MCP Toolbox should not be vulnerable to this attack vector; however, to-date it remains vulnerable.

### Current Behavior

This vulnerability provides an attacker that is able to get a victim to visit a malicious website, or view a malicious ad, full control over the locally running toolbox MCP server.

This allows for database manipulation, data theft, creating new resources, or making HTTP requests from the context of the developers computer, as if they were an MCP connected AI agent. Basically, any functionality this MCP server exposesto an AI agent can be performed by an attacker.

### Steps to reproduce?

https://github.com/user-attachments/assets/be380f35-b4e5-48e0-828b-c974f042be12

### Proof of Concept

1. Install genai-toolbox
2. Create a toolbox.yaml in the current working directory and configure some database connections
3. Launch (the -ui is optional): toolbox -ui
4. Open the following in the browser (tested on Safari & Firefox): http://rebinder.chvibglj.lighthouse-sec.xyz/mcp-inspector.html
5. Set the "Target Host" to "0.0.0.0" for Unix-like platforms (e.g. Linux, macOS) and to "127.0.0.1" (or any other valid local host addresses) for Microsoft Windows.
7. Set the target port to 5000
8. Open the browser console
[...]

PR #3342, opened 2026-06-04
server: default --allowed-hosts to loopback for local (non-remote) deployments

adilburaksen:
## Description

`--allowed-hosts` and `--allowed-origins` default to `["*"]`. The server defaults to the HTTP transport bound to `127.0.0.1`, so out of the box for local development the Host check is a no-op (wildcard). As the existing startup warning already notes, a wildcard `--allowed-hosts` leaves the server "vulnerable to DNS rebinding attacks" — a web page open in the developer's browser can rebind an attacker domain to `127.0.0.1` (requests then become same-origin, so CORS does not apply) and reach the MCP/tool endpoints, which execute against the developer's configured data sources.

This makes that the default rather than something the user must remember to harden.

## Change

`--allowed-hosts` now defaults to a loopback allowlist (`127.0.0.1`, `localhost`, `::1`) **only when** the server is bound to a loopback address and the user did not pass `--allowed-hosts` explicitly. 

- **Local (loopback bind), no flag** → loopback allowlist (DNS-rebinding blocked by default). An INFO line is logged.
- **Non-loopback bind** (`0.0.0.0`, `::`, a specific host — i.e. remote/Cloud Run/containers) → existing `["*"]` behavior is preserved, so remote deployments are unaffected.
- **Explicit `--allowed-hosts`** (including `*`) → always respected unchanged.

Detection of "explicitly set" uses Cobra's `flags.Changed("allowed-hosts")`. The existing wildcard DNS-rebinding warning is preserved and now only fires when the wildcard actually survives.

`--allowed-origins` is intentionally left unchanged (scope kept to the rebinding-relevant Host defense).

## Testing

`go build ./...` and `go test ./cmd/... ./internal/server/...` pass; `gofmt` clean. Added `TestAllowedHostsContextAwareDefault` covering: loopback bind + no flag → `evil.com` 403 / `localhost` 200; non-loopback bind → wildcard preserved (`evil.com` 200); explicit flag respected.

### DIFF
--- a/cmd/internal/flags.go
+++ b/cmd/internal/flags.go
@@ -71,5 +71,5 @@ func ServeFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 	flags.StringVar(&opts.Cfg.ToolboxUrl, "toolbox-url", "", "Specifies the Toolbox URL. Used as the resource field in the MCP PRM file when MCP Auth is enabled. Falls back to TOOLBOX_URL environment variable.")
 	flags.StringVar(&opts.Cfg.McpPrmFile, "mcp-prm-file", "", "Path to a manual Protected Resource Metadata (PRM) JSON file. If provided, overrides auto-generation.")
 	flags.StringSliceVar(&opts.Cfg.AllowedOrigins, "allowed-origins", []string{"*"}, "Specifies a list of origins permitted to access this server. Defaults to '*'.")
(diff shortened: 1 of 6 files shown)


Yuan325:
Hi, please refer to the discussion (#3113) on our decision that we made to keep the default as `*`.
Thank you! :)

PR #3394, opened 2026-06-10
feat(mcp): add Secure Parameters support as Toolbox experimental extension

duwenxin99:
Adds support for secure parameters as part of the Toolbox experimental extension. Tools can now mark parameters as secure (secure: true), and the server validates that sensitive arguments are exclusively passed via secureArguments.

- **Legacy Protocols (< v2026-07-28)**: Secure tools are completely hidden. Attempting to call them results in a non-existent tool (`INVALID_PARAMS`: "tool not found").
- **Latest Protocol (v2026-07-28)**: Calling a secure tool without declaring the `com.google.cloud/toolbox.v1` extension explicitly blocks execution and returns a `400 Bad Request` (`MISSING_REQUIRED_CLIENT_CAPABILITY`).


docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,40 @@ parameters:
     valueType: integer # This enforces the value type for all entries.
 ```
 
+
+### Secure Parameters
+
+Secure parameters are designed to handle sensitive information (such as API keys, passwords, and credentials) that should not be transmitted in plain text through standard client-server arguments or logs.
```
averikitsch: Please update to call out these are parameters that agents should not control along with potentially not wanting to transmit in plain text. We should use example like customer_id versus credentials that should be set in the source.

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,40 @@ parameters:
...
+
+Secure parameters require negotiation of the draft protocol version (`DRAFT-2026-v1`) and the client capability `toolbox/secure-params` to be set to `true`.
```
averikitsch: We need to be more clear on how users set this. Ultimately we may just need to call out the version for both the server and SDKs for this support

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,40 @@ parameters:
...
+
+When a parameter is marked as `secure: true`, the following strict rules are enforced at the transport level:
+
+1. **Capability Requirement**: If a tool has one or more secure parameters, the client **must** negotiate the `"DRAFT-2026-v1"` protocol version during initialization and declare support for `toolbox/secure-params` in its capabilities. Otherwise, invoking the tool will return a `jsonrpc.INVALID_PARAMS` error.
+2. **Mutual Exclusion (Configuration)**: A parameter **cannot** have both `secure: true` and `authServices` specified. An error will be thrown during configuration loading.
+3. **Mutual Exclusion (Arguments)**:
+   - Parameters marked as `secure` **must** be passed in the request's `secureArguments` map, and **must not** be passed in the standard `arguments` map.
+   - Parameters **not** marked as `secure` **must** be passed in the standard `arguments` map, and **must not** be passed in the `secureArguments` map.
+   - Any violation of these rules will result in a `jsonrpc.INVALID_PARAMS` error and execution will be blocked.
+
```
averikitsch: what is actionable here for the customer? We should also note what happens if the extension is not supported.

docs/en/documentation/configuration/tools/_index.md
averikitsch: I thought we agreed on com.google.cloud/secure-params

internal/server/mcp/vdraft/method.go
```diff
@@ -599,3 +605,70 @@ func promptsGetHandler(ctx context.Context, id jsonrpc.RequestId, promptset prom
 		Result:  result,
 	}, nil
 }
+
+func parseSupportsSecureParams(body []byte) bool {
+	var meta struct {
+		Params struct {
+			Meta struct {
```
averikitsch: Capabilities shouldn't be nested in _meta based on your design and spec https://modelcontextprotocol.io/extensions/overview

internal/server/mcp/vdraft/method.go
```diff
@@ -185,8 +187,9 @@ func toolsListHandler(ctx context.Context, id jsonrpc.RequestId, resourceMgr *re
 	}
 
 	urlParams, _ := util.UrlParamsFromContext(ctx)
+	supportsSecureParams := parseSupportsSecureParams(body)
```
averikitsch: We unmarshal the body again here, we've already unmarshalled the request on line 176. Could we refactor so we only have to do this once?

internal/server/mcp/vdraft/method.go
```diff
@@ -246,6 +248,11 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, toolset tools.T
 		return jsonrpc.NewError(id, jsonrpc.INVALID_PARAMS, err.Error(), nil), err
 	}
 
+	toolArguments, err := validateAndMergeSecureParams(resourceMgr.GetSourcesMap(), tool, req, body)
```
averikitsch: Same here, can we unmarshal once and then use that in these helper methods

internal/server/mcp/vdraft/method.go
```diff
@@ -599,3 +605,70 @@ func promptsGetHandler(ctx context.Context, id jsonrpc.RequestId, promptset prom
...
+	}
+	for _, p := range params {
+		if p.GetSecure() {
+			hasSecureParams = true
+			secureParamMap[p.GetName()] = true
+		}
+	}
+
+	if hasSecureParams && !supportsSecureParams {
+		return nil, fmt.Errorf("tool %q requires secure-params extension which is not supported by the client", req.Params.Name)
```
averikitsch: use full extension name in log message

nit: It might be nice to have this check at the beginning of the tool call method

internal/server/mcp/v20260728/method.go
```diff
@@ -599,3 +605,70 @@ func promptsGetHandler(ctx context.Context, id jsonrpc.RequestId, promptset prom
...
+
+	// Validate that secure parameters are only passed in secureArguments
+	for argName := range req.Params.Arguments {
+		if secureParamMap[argName] {
+			return nil, fmt.Errorf("parameter %q is secure and must not be passed in standard arguments", argName)
+		}
+	}
+
+	// Validate that non-secure parameters are not passed in secureArguments
+	for argName := range req.Params.SecureArguments {
```
averikitsch, line 874: How is the SDK setting SecureArguments? 

I think it makes sense to throw the error if its unexpectedly coming via secureparams but based on the current SDK users may bind any param. Would we want to override an agent param?

averikitsch (changes requested):

internal/server/mcp/vdraft/method.go
duwenxin99: Updated.

internal/server/mcp/vdraft/method.go
duwenxin99: Updated.

internal/server/mcp/v20260728/method.go
duwenxin99: Discussing further in doc.

duwenxin99: Update from team discussion, we will keep the current implementation - throwing an error when client passes in an agent param in the secure param section.

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,53 @@ parameters:
...
+    type: string
+    description: Sensitive customer identifier supplied out-of-band by the calling application
+    secure: true
+```
+
+#### Protocol Constraints & Extension Negotiation
+
+Enabling secure parameters requires both the **MCP Toolbox Server** (running protocol version `DRAFT-2026-v1`) and a compatible **client SDK** that supports the `com.google.cloud/secure-params` extension.
+
+Clients declare support by enabling `com.google.cloud/secure-params: true` in their client capabilities (`io.modelcontextprotocol/clientCapabilities.experimental`) or per-request metadata (`_meta`).
```
kurtisvg: Client Capabilities is in _meta in the 07-28-2026 version of the spec right? Why is this OR?

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,53 @@ parameters:
     valueType: integer # This enforces the value type for all entries.
 ```
 
+
+### Secure Parameters
+
+Secure parameters are designed for sensitive runtime context (such as an end-user `customer_id`, tenant identifier, or session token) that **AI agents (LLMs) should not control or see** and that should not be transmitted in plain text through prompt completion requests, model context windows, or standard server logs.
```
kurtisvg, line 235: nit: Is this wording right?

```suggestion
Secure parameters are for values that must not be agent-controlled. They let the calling application supply a parameter's value out-of-band, so the model can neither see it in context nor set or override it — for example, an end user's `customer_id`.
```

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,53 @@ parameters:
...
+parameters:
+  - name: customer_id
+    type: string
+    description: Sensitive customer identifier supplied out-of-band by the calling application
+    secure: true
+```
+
+#### Protocol Constraints & Extension Negotiation
+
+Enabling secure parameters requires both the **MCP Toolbox Server** (running protocol version `DRAFT-2026-v1`) and a compatible **client SDK** that supports the `com.google.cloud/secure-params` extension.
```
kurtisvg: nit: suggest adding a link to the Toolbox SDKs for "client SDK" 
```suggestion
Enabling secure parameters requires:
- **MCP Toolbox Server** (running protocol version `DRAFT-2026-v1`)
- compatible **client SDK** that supports the `com.google.cloud/secure-params` extension.
```

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,53 @@ parameters:
...
+
+When a parameter is marked as `secure: true`, the following strict rules are enforced:
+
+1. **Capability Requirement**: If a tool has one or more secure parameters, the client **must** negotiate protocol version `"DRAFT-2026-v1"` and declare support for the `com.google.cloud/secure-params` capability.
+2. **Mutual Exclusion (Configuration)**: A parameter **cannot** have both `secure: true` and `authServices` specified. An error will be thrown during configuration loading.
+3. **Mutual Exclusion (Arguments)**:
+   - Parameters marked as `secure: true` **must** be passed in the request's `secureArguments` map, and **must not** be passed in the standard `arguments` map.
+   - Parameters marked as `secure: false` **must** be passed in the standard `arguments` map, and **must not** be passed in the `secureArguments` map.
+   - Because `arguments` and `secureArguments` are strictly disjoint, agent-generated arguments cannot override or collide with host-provided secure arguments. Any violation of these rules returns a `jsonrpc.INVALID_PARAMS` error and blocks execution.
```
kurtisvg: This seems like implementation details that a user won't ever see right? 

Can we say something like:

When a parameter is marked as `secure: true`, it will not be presented to the agent as a configurable parameter. Instead, it will rely on the application to set the parameter. If an application fails to set the parameter before the tool is called, it'll return `<some error>`.

Here is how you set a Secure parameter with the Toolbox Python SDK:
`<code snippet>`

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -160,6 +160,53 @@ parameters:
...
+
+#### Client Compatibility & Unsupported Extension Behavior
```
kurtisvg:
```suggestion
#### Unsupported Extension Behavior
```

internal/server/mcp/vdraft/method.go
```diff
@@ -185,8 +186,9 @@ func toolsListHandler(ctx context.Context, id jsonrpc.RequestId, primitiveMgr *p
 	}
 
 	urlParams, _ := util.UrlParamsFromContext(ctx)
+	supportsSecureParams := supportsSecureParams(req.Params.Meta)
```
kurtisvg: nit: This is a var shadow. can we rename the var?

kurtisvg (dismissed):
1. Behavior with URL Params

PR description says:
> server validates that sensitive arguments are exclusively passed via secureArguments rather than standard arguments or URL-bound parameters.

Are we confident that's the case? I'm not sure the logic in the current PR supports this.

2. Should we gate this extension behind a flag? We haven't properly defined it yet, are we confident it won't change before we need to? 

3. DevX on failure case. If I specify "secure: true", should we throw an error unless the extension is enabled?

averikitsch (commented):
@kurtisvg would you like to do another review?

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -230,6 +230,45 @@ parameters:
...
+
+> **Note:** A parameter cannot have both `secure: true` and `authServices` specified.
+
+Here is how you set a secure parameter with the Toolbox Python SDK:
+
+```python
+# Pass secure_args when loading or calling a tool via the Python SDK
+auth_tool = await toolbox.load_tool(
+    "search_secure_data",
+    secure_args={"customer_id": "cust_12345"}
```
anubhav756:
```suggestion
# Pass secure_params when loading or calling a tool via the Python SDK
auth_tool = await toolbox.load_tool(
    "search_secure_data",
    secure_params={"customer_id": "cust_12345"}
```

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -230,6 +230,45 @@ parameters:
...
+statement: |
+  SELECT * FROM sessions WHERE customer_id = $1
+parameters:
+  - name: customer_id
+    type: string
+    description: Sensitive customer identifier supplied out-of-band by the calling application
+    secure: true
+```
+
+When a parameter is marked as `secure: true`, it will not be presented to the agent as a configurable parameter. Instead, it relies on the application to set the parameter. If an application fails to set the parameter before the tool is called, execution is blocked and returns an `invalid_params` error.
```
anubhav756:
```suggestion
When a parameter is marked as `secure: true`, it will not be presented to the agent as a configurable parameter. Instead, it relies on the application to set the parameter. If an application fails to set the parameter before the tool is called, execution returns a tool error indicating that the required parameter was not provided.
```

anubhav756 (commented):
optional: Could we also add a test case for when a required secure param is omitted (e.g. `secureArguments: nil` or `{}`)?

internal/server/mcp/v20260728/method.go
```diff
@@ -840,3 +854,34 @@ func groupsGetHandler(ctx context.Context, id jsonrpc.RequestId, primitiveMgr *p
 		Result:  result,
 	}, nil
 }
+
+// validateAndMergeSecureParams validates and merges standard and secure arguments based on secure parameter definitions.
+func validateAndMergeSecureParams(req CallToolRequest, paramDefs parameters.Parameters) (map[string]any, error) {
```
anubhav756: nit: Should we accept `req` by reference to avoid a copy?

internal/server/mcp/v20260728/method_test.go
```diff
@@ -1344,3 +1362,234 @@ func TestGetResultMetadata(t *testing.T) {
...
+				"params": {
+					"name": "secure_tool",
+					"arguments": {
+						"query": "hello"
+					},
+					"secureArguments": {
+						"api_key": "secret"
+					},
+					"_meta": {
+						"io.modelcontextprotocol/protocolVersion": "2024-11-05",
```
anubhav756: Should this be 2026 version?

tests/mcp_tool.go
```diff
@@ -571,3 +571,162 @@ func GetTemplateParamMCPExpectedTools() []MCPToolManifest {
...
+	for _, o := range options {
+		o(configs)
+	}
+
+	mcpVersion := "2026-07-28"
+
+	// In 2026-07-28 (v20260728), the initialize method is removed in favor of optional server/discover.
+	// Therefore, we only run RunInitialize for older protocol versions.
+	header := map[string]string{}
+	if mcpVersion != "2026-07-28" {
```
anubhav756: What's the purpose of this check? I see we are hardcoding its value in line `584`, so this would be dead code?

anubhav756 (changes requested):

averikitsch (approved):

internal/server/mcp/v20260728/manifests.go
```diff
@@ -45,6 +55,10 @@ func generateToolManifest(name, desc string, authInvoke []string, params paramet
 		ToolInputSchema: inputSchema,
 		Annotations:     toolAnnotations,
 	}
+	if len(secureParams) > 0 {
+		secureInputSchema, _ := generateParamManifest(secureParams, urlParams)
```
anubhav756, line 59: IIUC, `generateParamManifest` has the same schema as regular params, including `required` and `default`.

Per our earlier discussion [here](https://docs.google.com/document/d/1ziHBysFX8RR1IL9_R9z2Xe_135e6YW09h4viNrvaPsw/edit?disco=AAAB9f8p7U0), we discussed about having them non-optional.

So shouldn't we add server-side validation during config parsing to enforce `required: true` and reject any `default` values or `required: false` config for `secure: true` parameters?

anubhav756 (dismissed):

PR #3395, opened 2026-06-10
feat(mcp): add secure parameter support for current MCP versions

duwenxin99:
Add all the stateful MCP protocol integrations (v20241105, v20250326, v20250618, v20251125) for secure parameters.


duwenxin99:
We decided not to support secure param in the current MCP versions.

PR #3415, opened 2026-06-14
fix: support string parameter options

### LINKED ISSUE #3311: Update NewParameter functions
Update all the `NewStringParameterWithDefault`, `NewStringParameterWithRequired` with functional options pattern. This will also allow more flexibility for tools that would like to be not-required and set some allowed values (if value is provided, check it against allowed values).

he-yufeng:
## Summary
- add a small functional-options layer for string parameters
- keep the existing string parameter constructors source-compatible
- allow combinations such as optional string parameters with allowed values

Fixes #3311.

## To verify
- go test ./internal/util/parameters -run 'TestParametersParse|TestParamManifest' -count=1 -v
- go test ./internal/util/parameters -count=1
- git diff --check

### DIFF
--- a/internal/util/parameters/parameters.go
+++ b/internal/util/parameters/parameters.go
@@ -595,8 +595,58 @@ type ParamAuthService struct {
 	Field string `yaml:"field"`
 }
 
+// StringParameterOption configures a StringParameter.
+type StringParameterOption func(*StringParameter)
+
+func applyStringParameterOptions(p *StringParameter, opts ...StringParameterOption) *StringParameter {
+	for _, opt := range opts {
+		opt(p)
+	}
+	return p
+}
+
+func WithStringParameterRequired(required bool) StringParameterOption {
+	return func(p *StringParameter) {
+		p.Required = &required
+	}
+}
+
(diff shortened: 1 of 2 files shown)


he-yufeng:
Rebased onto current main and tightened the string-parameter constructor path so the wrapper constructors initialize the base parameter directly, set their fixed field, then apply any extra options. That keeps the options API while avoiding an extra variadic constructor hop for the built-in option.\n\nValidation:\n- gofmt -l internal/util/parameters/parameters.go internal/util/parameters/parameters_test.go\n- go test ./internal/util/parameters\n- go test ./internal/util/...\n- git diff --check upstream/main..HEAD\n\nI also tried go test ./..., but it did not finish within 5 minutes locally.

Yuan325:
Hey @he-yufeng, thanks so much for putting this together! It seems we already have a similar PR open at [https://github.com/googleapis/mcp-toolbox/pull/3314](https://github.com/googleapis/mcp-toolbox/pull/3314), so I'll go ahead and close this one out to keep things tidy. Thanks again for taking the time to help out!

PR #3416, opened 2026-06-14
docs: document execute_sql least privilege

### LINKED ISSUE #3405: Security: Lack of Input Validation and Restrictions in execute_sql Tools
I have audited several 'execute_sql' tools in the repository (e.g., tidbexecutesql, oceanbaseexecutesql, mindsdbsql) and found that they allow the execution of arbitrary SQL statements without any validation or restriction.

**Vulnerability Detail:**
The tools extract the 'sql' parameter directly from the MCP request and pass it to the database source's 'RunSQL' method. While the tools are marked with 'destructive' annotations (requiring user confirmation), this is insufficient for a professional security boundary.

**Risks:**
1. **Prompt Injection**: An attacker could potentially trick an LLM into executing destructive commands (e.g., 'DROP TABLE', 'TRUNCATE') or exfiltrating sensitive data.
2. **Privilege Escalation**: If the database user used by the MCP server has elevated privileges, this becomes a critical RCE-like vector for the database.

**Recommended Mitigations:**
1. **Read-Only Mode**: Implement a configurable read-only mode for these tools.
2. **Command Blacklisting**: Block dangerous SQL keywords (DROP, ALTER, TRUNCATE, GRANT) unless explicitly enabled.
3. **Parameterization**: Encourage or enforce the use of parameterized queries where possible.
4. **Least Privilege**: Explicitly document the requirement to run the server with a low-privilege database user.

he-yufeng:
Closes #3405.\n\nThis documents the security boundary for prebuilt dynamic execute_sql-style tools: use a dedicated least-privilege database identity, prefer SELECT-only access for exploratory agents, and do not rely on regex keyword blacklists for arbitrary SQL endpoints.\n\nValidation:\n- git diff --check\n- Select-String for execute_sql / SELECT / regex / database-native terms in the updated doc

### DIFF
--- a/docs/en/documentation/configuration/prebuilt-configs/_index.md
+++ b/docs/en/documentation/configuration/prebuilt-configs/_index.md
@@ -25,6 +25,22 @@ You can load a specific toolset from a prebuilt configuration by appending a `/`
 See [Usage Examples](../../../reference/cli.md#usage-examples).
 {{< /notice >}}
 
+## Security for dynamic SQL tools
+
+Some prebuilt configs expose dynamic `execute_sql`-style tools where the agent
+supplies raw SQL. Tool annotations and MCP client confirmations are useful UX
+guardrails, but they are not a database security boundary.
+
+Run these tools with a dedicated database identity that only has the privileges
+the agent should exercise. For exploratory agents, this usually means
+`SELECT`-only access to the specific schemas, tables, or views the agent may
+read. Avoid owner, admin, migration, or application-write accounts.
+
+Prefer custom parameterized tools for fixed workflows. Use dynamic SQL tools for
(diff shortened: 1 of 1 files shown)


he-yufeng:
Rebased onto current main. The old package-lock diff is gone; the PR now only changes the prebuilt-configs documentation.\n\nValidation:\n- git diff --stat upstream/main..HEAD\n- git diff --check upstream/main..HEAD

anubhav756 (approved):
Thank you for this addition @he-yufeng! LGTM

FYI, we are actually in the process of rolling out native, engine-level read-only locks to enforce these boundaries in MCP Toolbox for supported databases. Your changes align to bridge the gap for DBs where we still rely on the user's IAM config.

Thanks again for the PR!

PR #3391, opened 2026-06-10
feat(ci): automate npm publishing via OSS Exit Gate

wangauone:
## Description

Automates external publication of the six `@toolbox-sdk` npm packages (5 platform-specific + 1 wrapper) through the OSS Exit Gate, replacing the manual Phase A/B procedure documented in `DEVELOPER.md`.

On each versioned release:
1. New `publish-npm-to-ar` step packs each platform package with its freshly-built binary already staged in `bin/` (which short-circuits the existing `prepack` GCS-download script), authenticates to the Exit Gate Artifact Registry via the Cloud Build SA's ADC, and `npm publish`es the six tarballs to `us-npm.pkg.dev/oss-exit-gate-prod/mcp-toolbox--npm`.
2. New `trigger-exit-gate` step writes a `publish_all: true` manifest and uploads it to `gs://oss-exit-gate-prod-projects-bucket/mcp-toolbox/npm/manifests/`, which signals Exit Gate to pull the tarballs from AR and publish them to npmjs.org.

The bash logic is extracted into `.ci/publish_npm_to_ar.sh` and `.ci/trigger_exit_gate.sh` for testability and readability. The OSS Exit Gate URLs/bucket are hardcoded `readonly` constants in the scripts (Exit Gate-owned values, not build configuration) so no new substitutions land in the Cloud Build YAML.

The manual procedure in `DEVELOPER.md` is retained as a fallback.

## PR Checklist

- [x] Reviewed CONTRIBUTING.md
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (CI-only change, no source code modified)
- [x] Appropriate docs were updated (DEVELOPER.md)
- [x] No breaking changes

PR #3430, opened 2026-06-15
feat(ci): automate PyPI publishing via OSS Exit Gate

wangauone:
## Description

Follow-up to #3391 (npm automation). Automates external publication of the `toolbox-server` PyPI package through the OSS Exit Gate, mirroring the npm flow.

On each tagged release, the versioned Cloud Build pipeline now:

1. Builds **five platform-specific wheels** (linux x86_64, darwin arm64/x86_64, windows amd64/arm64) — one Linux Cloud Build VM produces all five by varying `TOOLBOX_PLATFORM` per `python -m build` invocation.
2. Uploads them to the Exit Gate AR (`us-python.pkg.dev/oss-exit-gate-prod/mcp-toolbox--pypi`) via `twine --skip-existing` (idempotent).
3. Drops a `publish_all: true` manifest in the Exit Gate's GCS bucket — Exit Gate then publishes to pypi.org via trusted publishing.

A retry-only pipeline (`.ci/pypi_retry.cloudbuild.yaml`) is also included for when the PyPI portion fails after binaries are already in GCS.

Notable side-fixes:
- Rewrites `pypi/setup.py` to drop the runtime GCS download (only worked for linux/amd64); now expects pre-staged binary + `TOOLBOX_PLATFORM` env var
- Bumps `pypi/src/toolbox_server/__init__.py` to `1.4.0` (was drifting at `1.3.0`) and adds the `# x-release-please-version` marker so release-please keeps it synced

OSS Exit Gate prerequisites are already configured: AR repo, builder SA, trusted publisher on pypi.org for `toolbox-server`.

## PR Checklist

- [x] Reviewed CONTRIBUTING.md
- [x] Tests and linter pass
- [x] Code coverage unchanged (CI-only)
- [x] Docs updated (DEVELOPER.md note)
- [x] No breaking changes


.ci/publish_pypi_to_ar.sh
```diff
@@ -0,0 +1,75 @@
...
+  if [[ ! -f "${src}" ]]; then
+    echo "ERROR: ${src} not found in workspace. Earlier build step must run first." >&2
+    exit 1
+  fi
+
+  # Replace bin/ so a previous iteration's binary doesn't leak into this wheel.
+  rm -rf "${BIN_DIR}"
+  mkdir -p "${BIN_DIR}"
+  cp "${src}" "${BIN_DIR}/${dest}"
+  chmod +x "${BIN_DIR}/${dest}"
```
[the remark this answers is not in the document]
wangauone: Done.

pypi/setup.py
```diff
@@ -1,89 +1,77 @@
...
-        raise OSError(f"Unsupported platform: {system}-{machine}")
-    return os_part, arch_part
+class bdist_wheel(_bdist_wheel):
+    def finalize_options(self):
+        super().finalize_options()
+        # Tell bdist_wheel this package contains a platform-specific binary
+        # (not just Python), so the wheel gets a real platform tag instead
+        # of the default "any". Required for our get_tag override to take
+        # effect.
+        self.root_is_pure = False
```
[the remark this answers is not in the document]
wangauone: This will be prevented by the shell script.

.github/workflows/pypi.yml
```diff
@@ -38,9 +43,19 @@ jobs:
       - name: Install build tooling
         run: python -m pip install --upgrade pip build pytest
 
+      - name: Build toolbox binary and stage it for the wheel
+        # setup.py now expects the binary pre-staged in src/toolbox_server/bin/
+        # instead of downloading from GCS at build time.
+        run: |
+          go build -o pypi/src/toolbox_server/bin/toolbox .
+          chmod +x pypi/src/toolbox_server/bin/toolbox
```
helloeve, line 51: why do we need to do this build on the fly? IIUC, this should fetch from the ones uploaded into gs://mcp-toolbox-for-databases

pypi/src/toolbox_server/__init__.py
```diff
@@ -1,2 +1,2 @@
 # Version is managed by release-please
-__version__ = "1.3.0"
+__version__ = "1.4.0"  # x-release-please-version
```
helloeve, line 2: seems like this was not captured by release-please? do we want to update it?

.github/workflows/pypi.yml
wangauone: This is only for PyPI related testing. Not part of the release pipeline. We need to build the package here given that the Go binary might not be available when the workflow is running (e.g. when PR is raised).

pypi/src/toolbox_server/__init__.py
wangauone: Yeah, `# x-release-please-version` is the one to fix this issue. For versions in plain text, release please need this to find the version to do the update.

helloeve (approved):

PR #3314, opened 2026-05-29
refactor(tool/parameters): update NewParameter functions to use functional options pattern

Deeven-Seru:
## Description

This PR refactors all `New*Parameter` constructor functions in `internal/util/parameters/parameters.go` to use the **functional options pattern**, replacing the proliferation of `NewXParameterWithOption(...)` variants.

The old design required a separate constructor for every combination of options (e.g. `NewStringParameterWithDefault`, `NewStringParameterWithRequired`, `NewStringParameterWithAuth`, etc.). This led to a combinatorial explosion of function signatures and made it impossible to combine options without adding yet another constructor.

The new design introduces:
- A `*ParameterOption` functional type per parameter kind
- `With*` option constructors (e.g. `WithStringRequired`, `WithStringDefault`, `WithStringAllowedValues`)
- A single `New*Parameter(name, desc, ...opts)` constructor per type

This makes it trivial to freely combine any options:

```go
// Before — rigid, one constructor per option combo
parameters.NewStringParameterWithRequired("name", "desc", true)

// After — freely composable
parameters.NewStringParameter("name", "desc",
    parameters.WithStringRequired(true),
    parameters.WithStringAllowedValues([]any{"a", "b", "c"}),
)
```

### Changes
- Defined `*ParameterOption` functional types and `With*` functions for all 6 parameter types (`String`, `Int`, `Float`, `Boolean`, `Array`, `Map`)
- Updated all `New*Parameter` constructors to accept variadic `opts ...Option`
- Migrated **600+ call sites** across `internal/` and `tests/` using `gofmt -r` rewrite rules
- Removed all obsolete `NewXParameterWithOption` functions
- Added `WithIntMinValue`, `WithIntMaxValue`, `WithFloatMinValue`, `WithFloatMaxValue` range options

## PR Checklist
- [x] Make sure to open an issue as a bug/issue before writing your code!
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)
- [ ] Make sure to add `!` if this involves a breaking change

## Issue Reference

Fixes #3311 🦕

PR #3445, opened 2026-06-16
fix(util/parameters): report integer type in IntParameter allowed/excluded constructors

he-yufeng:
## Description

`NewIntParameterWithAllowedValues` and `NewIntParameterWithExcludedValues` set their `Type` to `TypeString` instead of `TypeInt`. They were clearly copied from the string constructors and the type field was never updated, so an int parameter created with an allowed/excluded list reports itself as a string everywhere `GetType()` / `Manifest().Type` is consumed (manifests, validation, docs).

Every other int constructor uses `TypeInt`, and the float allowed/excluded constructors correctly use `TypeFloat`, so these two are the odd ones out.

## Fix

Flip both lines from `TypeString` to `TypeInt`. That is the whole change.

## Verification

Added a focused test in `internal/util/parameters`. On current `main` it fails:

```
GetType()="string" want integer
Manifest().Type="string" want integer
```

After the fix `go test ./internal/util/parameters/` passes with no regressions.

### DIFF
--- a/internal/util/parameters/int_parameter_type_test.go
+++ b/internal/util/parameters/int_parameter_type_test.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this — it's obsolete after #3314.

The bug here was that the dedicated `NewIntParameterWithAllowedValues` / `NewIntParameterWithExcludedValues` constructors hardcoded `Type: TypeString`. The functional-options refactor in #3314 removed those constructors entirely: `Type` is now set once to `TypeInt` in `NewIntParameter`, and `WithIntAllowedValues` / `WithIntExcludedValues` only set the value slices without touching `Type`. So there's nothing left to fix on current `main`. Thanks anyway.

Deeven-Seru:
great work @he-yufeng 😊

Issue #3165, opened 2026-05-05
postgres-execute-sql returns null when result set is empty


victorivanov:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox

### Toolbox version

1.1.0

### Environment

1. MacOS 26.3.1 (a) 
2. 1.1.0 Binaries via Homebrew (1.1.0+dev.darwin.arm64)

### Client

Came up when using Codex (0.128.0), which chokes on the null, verified with manual toolbox CLI use

Tool config:
```
---
kind: tool
name: query_readonly_sql
type: postgres-execute-sql
source: u1
description: >
  Execute SQL through the read-only PostgreSQL role only. Intended for SELECT,
  WITH SELECT-only queries, EXPLAIN, SHOW, and schema/data inspectio.
annotations:
  readOnlyHint: true
  destructiveHint: false
  idempotentHint: true
  openWorldHint: false
```

### Expected Behavior

A zero-row SQL result from postgres-execute-sql should be represented as an empty list, which would be consistent with other tools (e.g., postgres-list-tables) and allow the client to distinguish empty results from unknown errors

### Current Behavior

When a postgres-execute-sql tool returns zero rows, Toolbox returns a null tool result payload over MCP.

### Steps to reproduce?

1. Configure a postgres-execute-sql tool
2. Run a `SELECT` query returning no results, e.g. `invoke query_readonly_sql '{"sql": "select 1 where false"}'`
3. Observe null in response (and not `[]`)

### Additional Details

Similar to https://github.com/googleapis/mcp-toolbox/issues/2027

Deeven-Seru:
@duwenxin99  this identical issue affects all 24 relational databases in the project (due to uninitialized var out []any slices in the execution handlers). Is returning null intended behavior here, or should we change it to explicitly initialize these slices to []any{} so they return an empty array?

PR #3446, opened 2026-06-16
fix(sources/cassandra,sources/neo4j): return empty slice instead of nil for empty query results

he-yufeng:
## Description

`#3250` initialized query result slices to an empty array across the SQL sources so that a zero-row result serializes to `[]` instead of `null` (see #3165). The Cassandra and Neo4j sources were not part of that sweep and still declare the result slice with `var out []...`, so they return `nil` for an empty result and the tool response body comes back as `null`.

- `internal/sources/cassandra/cassandra.go`: `RunSQL` returns `out` which stays `nil` when the iterator yields no rows.
- `internal/sources/neo4j/neo4j.go`: `RunQuery` returns `out` which stays `nil` when there are no records. The dry-run path already returns a non-nil slice and `addPlanChildren` only runs when a node has children, so the non-dry-run result slice is the only gap.

## Fix

Initialize both result slices to an empty slice, matching the pattern from #3250. Two lines.

## Verification

`go build ./...` and `go vet ./internal/sources/cassandra/... ./internal/sources/neo4j/...` are clean, and the existing unit tests in both packages still pass. The empty-result behavior itself is exercised by the integration suites (`tests/cassandra`, `tests/neo4j`) that run against live instances, same as the suites #3250 updated. I did not have a live Cassandra/Neo4j instance to run those locally; the change is a direct mirror of the already-merged #3250 for two sources it missed.


he-yufeng:
Friendly ping — small consistency fix here: the Cassandra and Neo4j sources returned `nil` for an empty result set while the other sources return an empty slice, which trips callers that distinguish the two. This aligns them. It's been waiting since mid-June; could a maintainer kick off `/gcbrun` or take a quick look when there's a moment? Happy to rebase.

duwenxin99 (approved):

duwenxin99:
Hi @he-yufeng, can we add integration test cases for these two sources to test empty query results?

he-yufeng:
Added the empty-result coverage in a7141d8084: Cassandra gets a no-match invocation of my-tool-by-id asserting the result serializes as [] (not null), and Neo4j gets a my-empty-cypher-tool that matches no rows with the same assertion. go vet is clean on both suites; the live runs go through the integration CI since there is no local backend here.

he-yufeng:
@duwenxin99 This has your approval and the lanes are green; is anything still blocking the merge? Happy to rebase if main drifted.

he-yufeng:
cassandra.go:106 and neo4j.go:156 still use var out nil. Approved and the tests are in; is this good to merge?

PR #3454, opened 2026-06-16
feat(source/cloud-storage): restrict bucket and local path access

duwenxin99:
- Add allowedBuckets and allowedLocalRoot.
- Defaults to allow all.

Reported by: HE WEI（ギカク)


internal/sources/cloudstorage/cloudstorage.go
```diff
@@ -89,6 +92,53 @@ type Source struct {
 	client *storage.Client
 }
 
+func (s *Source) validateBucket(bucket string) error {
+	if len(s.AllowedBuckets) == 0 {
+		return nil
+	}
+	for _, b := range s.AllowedBuckets {
+		if b == "*" || b == bucket {
```
huangjiahua: Do we need "*" if empty allows every bucket? Same of local path

internal/sources/cloudstorage/cloudstorage_test.go
```diff
@@ -100,3 +124,117 @@ func TestFailParseFromYaml(t *testing.T) {
...
+			allowedRoots: []string{"/workspace"},
+			path:         "/workspace-malicious/file.txt",
+			wantErr:      true,
+		},
+		{
+			desc:         "relative path is cleaned and validated",
+			allowedRoots: []string{"/workspace"},
+			path:         "/workspace/../etc/passwd",
+			wantErr:      true,
+		},
```
huangjiahua, line 215: can we add some tests for windows?

huangjiahua (changes requested):

internal/sources/cloudstorage/cloudstorage.go
duwenxin99: Removed.

internal/sources/cloudstorage/cloudstorage_test.go
duwenxin99: Update test path construction to use filepath.Join(...) to form absolute paths dynamically based on OS.

huangjiahua (approved):
LGTM

averikitsch (approved):

PR #3455, opened 2026-06-16
fix(sources): return empty slice for empty results (cloudloggingadmin, dataplex, cloudsqladmin)

he-yufeng:
## Description

Several source query/list methods build their result slice with `var results []T` and return it directly, so when nothing matches they return `null` instead of `[]`:

- `cloudloggingadmin` list entries: a log filter that matches no entries leaves the slice nil.
- `dataplex/searchcatalog`: a catalog search that matches no entries leaves the slice nil.
- `cloudsqladmin` `ListDatabase` (no empty guard) and `ListInstance` (guarded a nil `Items`, but an empty non-nil `Items` still fell through to a nil result).

All are realistic empty cases. This initializes the slices to an empty slice so the JSON output is a consistent `[]`. It extends #3250 (which fixed the same `null`-vs-`[]` inconsistency across the SQL sources) and #3446 (cassandra/neo4j) to the remaining sources that were missed.

## PR Checklist

- [x] Code coverage does not decrease (no logic change beyond slice initialization)

`go build`, `go vet`, `gofmt` and the existing package unit tests pass for all changed packages. The empty-result paths run against live Cloud Logging / Dataplex / Cloud SQL Admin, so they are not exercised by local unit tests; the change is the same one-line pattern as the merged #3250 (`var results []T` to `results := []T{}`).


he-yufeng:
Gentle nudge on this one — it's been open since mid-June without a review pass. It's a small, self-contained fix returning an empty (non-nil) slice for empty result sets so the JSON response is `[]` instead of `null`, matching the other sources. No conflicts and CI is green. Anything I can do to help move it along, or a reviewer I should tag?

duwenxin99:
Hi @he-yufeng, thanks for the PR! The changes LGTM, but can we add regression integration tests?

he-yufeng:
Regression coverage is in 4640cc0cd8. Cloud SQL Admin has no integration suite (no GCP fixture for it in tests/), so ListDatabase/ListInstance get a unit test with the Admin API stubbed over httptest: an {"items": []} response must marshal as [], and it fails as "null" on the pre-fix code. Cloud Logging gets a no-match filter case in the query-logs integration flow, and Dataplex search-entries gets a well-formed no-match query asserting exactly [] (the existing not-found case tolerates null, this one pins the strict shape). The two integration cases go through CI since there are no live backends here; the unit test passes locally.

NoClueMike (approved):

he-yufeng:
The 4 nil slice cases are still present, and the regression tests went in on 08-15. duwenxin99 could you take another look when you get a chance?

PR #3460, opened 2026-06-17
fix(gda): support mTLS and GOOGLE_API_USE_MTLS_ENDPOINT for GDA client

Genesis929:
## Description

The Gemini Data Analytics (GDA) API calls made via custom REST HTTP client lacked dynamic mTLS endpoint routing, violating AIP-4114 and breaking Context Aware Access (CAA).

This change fixes the issue by updating the GDA client helper to dynamically select the API endpoint based on standard AIP-4114 environment variables (`GOOGLE_API_USE_CLIENT_CERTIFICATE` and `GOOGLE_API_USE_MTLS_ENDPOINT`).

The following tools have been updated to use `NewGDAClient` instead of hardcoded URLs:
- `bigquery-conversational-analytics`
- `conversational-analytics-ask-data-agent`
- `conversational-analytics-get-data-agent-info`
- `conversational-analytics-list-accessible-data-agents`
- `looker-conversational-analytics`

## Description

## PR Checklist

- [ ] Make sure you reviewed


shobsi (approved):

internal/util/gda.go
```diff
@@ -0,0 +1,81 @@
...
+	defaultOpts := []option.ClientOption{
+		option.WithEndpoint(GetGDAEndpoint()),
+		option.WithScopes("https://www.googleapis.com/auth/cloud-platform"),
+	}
+
+	allOpts := append(defaultOpts, opts...)
+
+	client, _, err := htransport.NewClient(ctx, allOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create GDA HTTP client: %w", err)
```
Yuan325, line 63: These error would not be due to the token String or values provided by agent right? Just wanna make sure since we are setting the error as ClientServerError. If this is something that the agent can alter and run it successfully by retrying, we would wanna set it as AgentError~

Yuan325 (approved):
Thank you!

internal/util/gda.go
Genesis929: I asked gemini for the error type here, it says:
This client initialization error only happens locally due to client/credentials config, not token or input values (actual network validation happens later in `client.Do`). Since the agent cannot fix this by altering parameters or retrying, returning `ClientServerError` is correct here.

PR #3472, opened 2026-06-18
fix(tools/http): report header value type with %T, not %t

he-yufeng:
## Description

`getHeaders` in `internal/tools/http/http.go` rejects a non-string header value with:

```go
return nil, fmt.Errorf("header param %s got value of type %t, not string", p.GetName(), headerValue)
```

`%t` is the boolean verb, but `headerValue` is an `any`, so the message comes out garbled — e.g. a header param given the integer `42` produces:

```
header param X-Token got value of type %!t(int=42), not string
```

The intent is to name the value's *type*, which is `%T`. With the fix the message reads `... got value of type int, not string`. `go vet` does not catch this because the argument is an interface value whose dynamic type it can't check at the call site.

## Testing

Added an internal-package test (`http_internal_test.go`) that calls `getHeaders` with a non-string header value and asserts the error names the real type and contains no malformed format token. It fails before the change (the message contains `%!t(int=42)`) and passes after.

```
go test ./internal/tools/http/   # ok
go vet ./internal/tools/http/    # clean
gofmt -l internal/tools/http/    # no diffs
```


internal/tools/http/http_internal_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2026 Google LLC
```
duwenxin99, line 1: Could you move this test to `http_test.go` instead of creating a new file? Thank you!

duwenxin99 (approved):

he-yufeng:
@duwenxin99 Same question here: approved and green, anything left before this can land?

he-yufeng:
http.go:264 still formats with %t. This has been approved by duwenxin99 since 08-04; is anything blocking a merge?

PR #3478, opened 2026-06-18
feat(tool/cloud-storage): support configurable parameters

huangjiahua:
## Description
Adds optional tool-level config fields for the first Cloud Storage tools so configured values are hidden from runtime parameter schemas and used during invocation. Also adds project support to create-bucket and updates docs/tests.

This change has been communicated internally.

## PR Checklist

## Issue Reference
N/A

### DIFF
--- a/docs/en/integrations/cloud-storage/tools/cloud-storage-create-bucket.md
+++ b/docs/en/integrations/cloud-storage/tools/cloud-storage-create-bucket.md
@@ -3,14 +3,20 @@ title: "cloud-storage-create-bucket"
 type: docs
 weight: 2
 description: >
-  A "cloud-storage-create-bucket" tool creates a Cloud Storage bucket in the configured source project.
+  A "cloud-storage-create-bucket" tool creates a Cloud Storage bucket in a configured or runtime-selected project.
 ---
 
 ## About
 
-A `cloud-storage-create-bucket` tool creates a new Cloud Storage bucket in the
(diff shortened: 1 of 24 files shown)


Yuan325 (changes requested):
Thank you for submitting this! Can you please update the parameters with the new format here: https://github.com/googleapis/mcp-toolbox/pull/3314

The integration tests should also be updated to reflect the addition of these new parameters :)

huangjiahua:
Thanks for the review @Yuan325! Both points are addressed:

1. **New parameter format (#3314):** Rebased onto `main` and migrated all parameter construction in the Cloud Storage tools to the functional-options pattern (e.g. `NewStringParameter(key, desc, parameters.WithStringRequired(false))`, `WithStringDefault`, `WithBooleanDefault`, `WithIntDefault`). No `New*ParameterWith*` constructors remain.

2. **Integration tests:** Added configured-variant tools to `tests/cloudstorage/cloud_storage_integration_test.go` that bake in the new parameters (`project`, `location`, `uniform_bucket_level_access`, `bucket`, `prefix`, `delimiter`). Each variant has a manifest assertion confirming the configured params are dropped from the exposed schema, plus an invocation test confirming the configured values are used.

`go build ./...`, `go test -race ./internal/tools/cloudstorage/...`, and `golangci-lint run` (v2) all pass.

Yuan325 (approved):

PR #3490, opened 2026-06-22
fix: scope MySQL list_tables to configured database

### LINKED ISSUE #2212: cloud-sql-mysql prebuilt tools queries to entire instance instead of configured database in that instance
### Prerequisites

### Toolbox version

0.24.0

### Environment

_No response_

### Client

_No response_

### Expected Behavior

When using the cloud-sql-mysql tools, it is expected that queries should be executed against environment variables or configuration.

1. List tables: Should only return tables belonging to the configured/connected database.
2. Data queries: When asking for data (e.g., "Show me customers"), the tool should query the default configured database without requiring me to explicitly name the database in every prompt.

### Current Behavior

The tool currently treats the connection as instance-wide rather than database-specific.

1. List tables: Returns tables from every database existing in that Cloud SQL instance, not just the one intended to connect to.
Image

Image

2. Data querying: General prompts like "Show me customers data" for a particular table related query does not show data from the database but asks clarifying questions which database you are talking about

Image

3. Workaround: The tool only functions correctly when I explicitly append the database name to the prompt (e.g., "What about customers data in db name?").

### Steps to reproduce?

1. Install the Gemini CLI extension for Cloud SQL MySQL.
2. Create more than one database in a single Cloud SQL instance.
3. Export the environment variable to connect to just one of those databases.
4. Give the prompt "List tables" — notice it returns tables from every database in the instance, not just the one in the variable.
5. Give the prompt "Show me customers data" — the tool asks "Which DB are you talking about?" instead of using the connected database.

### Additional Details

_No response_

luohui1:
## Description

Scopes the MySQL `list_tables` query to the configured source database when one is set, so Cloud SQL MySQL prebuilt tools no longer list tables from every database in the instance.

The existing all-user-schemas behavior is preserved for sources without a configured database. This also updates the prebuilt tool descriptions and MCP schema expectation to describe the configured-database default.

## PR Checklist

- [x] Make sure you reviewed

## Testing

- `go test ./internal/tools/mysql/mysqllisttables -count=1`
- `go test ./internal/tools/mysql/mysqllisttables -run 'TestInvokeScopesListTablesToConfiguredDatabase|TestInvokePreservesUnscopedBehaviorWithoutConfiguredDatabase' -count=1`
- `go test ./internal/server/mcp/jsonrpc ./internal/tools/mysql/mysqllisttables -count=1`

AI disclosure: this PR was prepared by Codex (GPT-5) acting as an AI coding agent on behalf of @luohui1.

Fixes #2212


duwenxin99:
Hi @Myst9, could you take a look at this PR change and make sure this behavior is what we want? Thank you!

PR #3511, opened 2026-06-24
refactor: rename ResourceManager to PrimitiveManager 

dishaprakash:
Rename ResourceManager to PrimitiveManager to make the distinction between Toolbox Resources and MCP Resources implementation clear


Yuan325 (approved):
Thank you!!! Checked the main primitives.go and assumes that everything else is updated accordingly since the tests are passing :)

PR #3520, opened 2026-06-24
fix(server): avoid a nil-flusher panic in the SSE handler

he-yufeng:
## What

In `sseHandler`, when the `http.ResponseWriter` does not implement `http.Flusher`, the code renders a 500 error response but then falls through instead of returning:

```go
flusher, ok := w.(http.Flusher)
if !ok {
    err = fmt.Errorf("unable to retrieve flusher for sse")
    _ = render.Render(w, r, newErrResponse(err, http.StatusInternalServerError))
}
session := &sseSession{
    writer:  w,
    flusher: flusher, // nil here
    ...
}
```

`flusher` is the nil interface, so the first `flusher.Flush()` (right after the endpoint event is written) panics with `invalid memory address or nil pointer dereference`. The error response the branch just rendered never takes effect.

## Fix

Return after rendering the error, which is what the branch was already trying to do.

## Why it matters

The `!ok` branch exists specifically to handle a writer without a flusher, but the missing `return` turns that intended 500 into a panic. A wrapped or non-standard `ResponseWriter` that doesn't forward `Flush` (some middleware, custom transports) would hit this. The fix makes the handler return the 500 it already builds instead of crashing the request.

## How to verify

```
go test ./internal/server/ -run TestSseHandlerWriterWithoutFlusher
```

The added test drives `sseHandler` with a `ResponseWriter` that intentionally does not implement `http.Flusher` and asserts a 500 is returned. On `main` it panics at `flusher.Flush()` (mcp.go); with the fix it returns the 500 cleanly. The full `internal/server` package, `gofmt`, and `go vet` all pass.


he-yufeng:
Gentle nudge on this one. It's a small guard for the SSE handler: when the `ResponseWriter` doesn't implement `http.Flusher` the code currently keeps going instead of returning, which can panic on the nil flusher. The change just returns early in that case. It's been green with only the automated review since late June, so if a maintainer has a minute to look I'd appreciate it, and I'm happy to rebase if it has drifted.

duwenxin99 (approved):

duwenxin99:
Hi @he-yufeng, thanks for opening the PR! The changes LGTM. Could you rebase on the latest main to resolve the lint failure? Thank you!

PR #3563, opened 2026-07-06
chore: Updates the version string for latest MCP conformance

anubhav756:
Updates the version string to the latest one updated in the official MCP repo.


Yuan325:
We can use this to check for conformance, but it might be better to keep as is until the official MCP release? To prevent confusing users since our docs all uses the `DRAFT-2026-1` version string.

anubhav756:
Closing this one. We will update the string post official MCP dated version release.

PR #3566, opened 2026-07-06
chore: add more draft specs updates

Yuan325:
This PR add the following updates to the draft specs: 
* MCP updated error code for `HeaderMismatch`, `MissingRequiredClientCapability`, `UnsupportedProtocolVersion` (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2907)
* Added the default `ttlMs` and `cacheScope` value in server to fulfill conformance tests: https://modelcontextprotocol.io/seps/2549-TTL-for-list-results#cache-scope
    * Defaulted TtlMs to 300000 (in milliseconds; equivalent to 5 minutes). Client SHOULD refresh every 5 minutes
    * Defaulted cacheScope to public
* Send `INVALID_PARAMS` error with `http.StatusBadRequest` when it's due to _meta


anubhav756:
Conformance test runner is crashing with a `_meta validation probe failed` error because it couldn't parse the headerless `body`.

Seems like calling `w.WriteHeader` pushes the HTTP headers out before `render.JSON` can attach the `Content-Type: application/json` header?

internal/server/mcp/vdraft/types.go
anubhav756, line 103: The spec requires clients to declare their capabilities (like `sampling`) inside the `_meta` field of every request. If a client tries to trigger a feature it didn't declare in that request's `_meta` field, the server must block it. The conformance tests checks this by sending a trap request (`test_missing_capability`) with an empty capabilities list to see if we catch it.

However, right now, our server code doesn't seem to be able to tell the difference between a client saying whether it supports the feature vs the client passing an empty object. So seems like it accidentally lets the trap request through.

I think changing one variable type to a pointer, the server can see when a client hasn't included the feature in their `_meta` payload, allowing it to correctly block the request and pass the test.

anubhav756:
FYI, to pass the capability enforcement tests, we need to intercept the diagnostic tool `test_missing_capability` in `vdraft/method.go`. If the client calls it but `capabilities.Sampling == nil`, the spec requires us to return a `-32021` JSON-RPC error.

anubhav756 (dismissed):

internal/server/mcp/vdraft/types.go
Yuan325: I think the missing capabilities is only triggered when the server require certain capability. However, Toolbox currently are not supporting any of those capabilities so we won't be checking for those.

averikitsch (approved):

internal/server/mcp/vdraft/types.go
averikitsch: This is fine with me. With the current set up do we not pass the test?

dishaprakash:
@Yuan325 just to confirm, we're going with a default TTL and CacheScope for conformance for now right?

internal/server/mcp/vdraft/manifests_test.go
```diff
@@ -254,6 +254,10 @@ func TestGenerateListToolsResult(t *testing.T) {
 		t.Fatalf("unable to generate list tools result: %s", err)
 	}
 	want := ListToolsResult{
+		CacheableResult: CacheableResult{
+			TtlMs:      0,
```
averikitsch: This seems like a fine default for now. What is the polling frequency for loading new tools?

Yuan325:
@dishaprakash Yes, seems like those values are required. But it should still be updated when we eventually implement Groups~

averikitsch (approved):

PR #3567, opened 2026-07-06
fix(source/bigquery): recover default-ADC clients after on-disk re-auth

### LINKED ISSUE #3573: BigQuery default-ADC source keeps using stale credentials after gcloud re-auth (requires process restart)
## What happens

A BigQuery source using Application Default Credentials (the default, non-`useClientOAuth` path) fetches credentials once at `Initialize` via `google.FindDefaultCredentials(...)` in `internal/sources/bigquery/bigquery.go` and holds the resulting in-memory `oauth2.TokenSource` (and the `Client`/`RestService` built from it) for the lifetime of the process.

That in-memory token source transparently refreshes *access* tokens from its cached refresh material, so ordinary expiry is handled. The gap: it never re-reads ADC from disk. If a user re-runs

```
gcloud auth application-default login
```

a new refresh token is written to disk and the old one is revoked. Once the cached access token expires, the held source can no longer mint tokens (its refresh token is dead), and every BigQuery tool invocation on that long-running process fails with an authentication error. The only way to recover today is to restart the whole MCP Toolbox process so `Initialize` runs again.

## Expected

After re-authenticating on disk with `gcloud auth application-default login`, a long-running session should recover on its own once its cached access token refreshes — no process restart.

## Repro

1. Run the toolbox with a BigQuery source using default ADC (no `useClientOAuth`).
2. Invoke a BigQuery tool (e.g. `bigquery-execute-sql`); confirm it works.
3. In another terminal, run `gcloud auth application-default login` and complete the flow (a new refresh token is written and the old one revoked).
4. After the cached access token expires, invoke the tool again on the still-running process.

**Observed:** invocations keep failing with an authentication error until the process is restarted.
**Expected:** the invocation picks up the new on-disk credentials and succeeds.

## Note

This is the same class of problem as #2173 (which moved token acquisition per-invoke for the conversational-analytics tool); the default-ADC source path — shared by the BigQuery tools and the source's Dataplex client used by `bigquery-search-catalog` — wasn't covered.

Fix proposed in #3567.

hiteshgupta88:
## Description

BigQuery sources using Application Default Credentials build their token source once at `Initialize` (via `google.FindDefaultCredentials`) and hold it for the process lifetime. That source refreshes access tokens from its cached refresh material, but it never re-reads ADC from disk - so after a user re-runs `gcloud auth application-default login` (which writes a new refresh token and revokes the old one), the cached source can no longer mint tokens once its current access token expires, and every BigQuery tool invocation fails until the whole process is restarted.

This wraps the credentials in a self-reloading token source, now shared for reuse as `sources.NewReloadingDefaultCredentials` in `internal/sources/util.go`:

- **Reload only when it can help.** On a mint failure the ADC file is re-read (a small local read) and the reload proceeds only if the on-disk JSON actually changed since the source was built. Rate limits, server errors, timeouts, and network drops are returned untouched. Keying on the on-disk change rather than the error's shape also keeps recovery working for ADC credential types whose token sources do not preserve the oauth2 error chain (`impersonated_service_account` and `external_account` stringify the inner error with `%v`), and it means permanently-broken credentials never trigger repeated reloads.
- **Impersonation recovers too.** `impersonate.CredentialsTokenSource` mints tokens by calling the IAM Credentials API with base credentials discovered from ADC, so it went stale the same way. Both `initBigQueryConnection` and `initDataplexConnection` now discover the base credentials explicitly, wrapped in the reloading source, and pass them via `option.WithCredentials` (honored verbatim by both auth paths; the base scope matches what the impersonate package self-discovers, so no scope change).
- **Guards on adoption, with actionable errors.** Reloaded credentials may differ from the originals only in fields a routine re-auth or key rotation touches (`refresh_token`, `private_key`, `private_key_id`, `expiry`); a change to any other field (`type`, `client_email`, `token_uri`, `credential_source`, `universe_domain`, ...) is refused naming the field, since those decide which principal the server runs as and where credential material is sent. Separately, `X-Goog-User-Project` is baked into a client at construction, so when clients derive it from the ADC file, a change to a non-empty baked quota project is refused; a quota project pinned via the source's `quotaProject` field or `GOOGLE_CLOUD_QUOTA_PROJECT` takes precedence over the file in the client libraries, making drift harmless, so it is tolerated. Impersonation base credentials skip only the quota half of the guard (their quota project only labels the internal IAM mint call). Every adoption and refusal is logged through the server's configured logger with the source name.

Known residual, documented in code: an `authorized_user` re-auth to a different Google account under gcloud's constant client ID is indistinguishable on disk from a routine re-auth, so it is adopted (and logged). Note also that `BigQueryTokenSourceWithScope` re-discovers ADC per invocation (see #2173) and is unchanged; the guards here apply to the shared long-lived sources, not to that per-call path.

This heals the same stale-cached-credentials problem that #2173 addressed for the conversational-analytics tool; here the shared default-ADC source is made self-refreshing instead.

### Tests

`internal/sources/util_test.go`: happy path (no reload), reload on disk change (regardless of error shape), no reload when disk unchanged (transient errors returned untouched; metadata-server JSON-less credentials), discovery-failure propagation, reloaded-source-still-failing, concurrent reload (race-safe, exactly one discovery), field/quota drift guard tables, terminal refusal, quota-project pin precedence (config over `GOOGLE_CLOUD_QUOTA_PROJECT` over file), and a disk-integration test through a real `GOOGLE_APPLICATION_CREDENTIALS` file with no network. `go build`, `go vet`, `golangci-lint run`, and `go test -race ./internal/sources/... ./internal/sources/bigquery/...` all pass.

Verified end to end against live BigQuery: a server started on revoked ADC recovers without restart once valid credentials are written to disk; an unchanged broken file surfaces the plain `invalid_grant` with no reload; a re-auth that changes the quota project or an identity-bearing field is refused with the actionable error. The impersonation wiring is verified against the client library sources; it was not exercised against a live service account.

## PR Checklist

- [x] Code coverage does not decrease (adds unit tests; no source paths lose coverage)
- [ ] Appropriate docs were updated (if necessary) - N/A (internal behavior change)
- [ ] Make sure to add `!` if this involves a breaking change - not a breaking change

## Issue Reference

Fixes #3573 🦕


internal/sources/bigquery/bigquery.go
```diff
@@ -684,6 +684,65 @@ func NormalizeValue(v any) any {
...
+	}
+	next, err := r.reload()
+	if err != nil {
+		return nil, fmt.Errorf("failed to reload default credentials: %w", err)
+	}
+	r.mu.Lock()
+	r.cur = next
+	r.mu.Unlock()
+	return next.Token()
+}
```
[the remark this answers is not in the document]
hiteshgupta88: Incorporated.

hiteshgupta88:
Hi! 👋 Gentle nudge on this one. I've just rebased it onto current main (it was ~31 commits behind — clean rebase, unit tests pass with -race). cla/google and the lint/header checks are green; the only red is integration-test-pr, which I think needs a maintainer to run it against the test project for a fork PR — could someone confirm that and kick it off, and take a look at the change when you have a chance? It heals the same stale-ADC class as #2173, for the default-ADC BigQuery/Dataplex path. Happy to adjust anything — thanks!

internal/sources/bigquery/bigquery.go
```diff
@@ -735,11 +804,21 @@ func initBigQueryConnection(
...
+		// call until the process is restarted.
 		cred, err := google.FindDefaultCredentials(ctx, credScopes...)
 		if err != nil {
 			return nil, nil, nil, fmt.Errorf("failed to find default Google Cloud credentials with scopes %v: %w", credScopes, err)
 		}
+		// Replace only the token source with a self-refreshing one and keep the
+		// rest of the credential via WithCredentials, so ADC metadata such as
+		// the quota project (X-Goog-User-Project, read from cred.JSON) is still
+		// applied to requests.
+		cred.TokenSource = newReloadingDefaultTokenSource(credScopes, cred.TokenSource)
```
duwenxin99: What if the user re-auth and:
1. changes the quota project
2. switches to a different Google account

Requests authenticated with the newly reloaded access token will continue sending the stale quota project header.

internal/sources/bigquery/bigquery.go
duwenxin99, line 722: The impersonation token source also uses service ADC under the hood. Can we make sure that this also recovers when the ADC rotates? Same applies to `initDataplexConnection()`.

internal/sources/bigquery/bigquery.go
```diff
@@ -689,6 +689,75 @@ func NormalizeValue(v any) any {
 	return v
 }
 
+// reloadingTokenSource wraps the Application Default Credentials token source so
+// that, when the cached source can no longer mint a token -- for example after
+// a user re-runs `gcloud auth application-default login`, which writes a new
+// refresh token and revokes the old one -- it reloads ADC from disk and
+// recovers, instead of failing until the process is restarted. It is safe for
+// concurrent use.
+type reloadingTokenSource struct {
```
duwenxin99: This implementation could benefit every other GCP source that uses ADC. Can we abstract it to `internal/sources/util.go` so we can reuse it in the future across other sources as well?

internal/sources/bigquery/bigquery.go
```diff
@@ -689,6 +689,75 @@ func NormalizeValue(v any) any {
...
+// mint one, it reloads ADC from disk once and retries. Ordinary access-token
+// expiry is handled transparently by the underlying source's own refresh, so
+// reaching the reload path typically means the on-disk credentials were rotated
+// (e.g. by `gcloud auth application-default login`); a transient mint error also
+// lands here and simply resolves on the retry.
+func (r *reloadingTokenSource) Token() (*oauth2.Token, error) {
+	r.mu.Lock()
+	cur := r.cur
+	r.mu.Unlock()
+	if cur != nil {
```
duwenxin99: Reloading it for every request may lead to meaningless I/O cost when the error is 5xx or not auth-related.
We should inspect the error and do not reload in cases of rate limits, server errors, timeouts, and network drops etc.

duwenxin99 (changes requested):
Hi @hiteshgupta88, thank you for opening the PR! The changes should be a very helpful addition to our GCP sources. I left some comments for optimization. Let me know if you have any questions.

internal/sources/bigquery/bigquery.go
hiteshgupta88: Done - moved it to internal/sources/util.go behind a single entry point, sources.NewReloadingDefaultCredentials(ctx, params), which does the discovery + wrapping in one call so other GCP sources can adopt it with a two-line change. Happy to wire up the remaining ADC-using sources (dataplex, cloudloggingadmin, datalineage, cloudmonitoring, ...) in a follow-up PR - kept this one scoped to the BigQuery source.

internal/sources/bigquery/bigquery.go
hiteshgupta88: Agreed on avoiding the pointless I/O. We ended up with a predicate that is a bit stronger than inspecting the error: on a mint failure we re-read the ADC file (a small local read, no network) and reload only if the bytes actually changed since the source was built; if the file is unchanged the original error is returned untouched. That covers every transient case (rate limits, 5xx, timeouts, network drops) without classifying error shapes - which turned out to matter, because two ADC credential types (impersonated_service_account, external_account) stringify the inner oauth2 error with %v, so an errors.As-based invalid_grant check would silently disable recovery for them. It also means permanently-broken credentials never reload repeatedly: the file has not changed, so the reload path is never re-entered.

internal/sources/bigquery/bigquery.go
hiteshgupta88: Done - both initBigQueryConnection and initDataplexConnection now discover the base credentials explicitly, wrap them in the same reloading source, and hand them to impersonate.CredentialsTokenSource via option.WithCredentials, which both auth paths honor verbatim, so the IAM generateAccessToken calls mint through the self-healing source. The base scope stays cloud-platform, matching what the impersonate package self-discovers. One honest caveat: the impersonation wiring is verified against the client library sources; the live end-to-end verification ran on the default-ADC path.

internal/sources/bigquery/bigquery.go
hiteshgupta88: Great question - this shaped most of the revision. X-Goog-User-Project is resolved once at client construction, so a reload cannot retro-fix it; instead of silently serving with a stale header, the reload is refused with an actionable error naming both projects when an unpinned, non-empty baked quota project changes on disk. When the quota project is pinned (the source's quotaProject field, or GOOGLE_CLOUD_QUOTA_PROJECT, which the client libraries give precedence over the file), drift never reaches request headers, so the reload proceeds. For the account-switch half we went further: reloaded credentials may differ from the originals only in the fields a routine re-auth or key rotation touches (refresh_token, private_key, private_key_id, expiry) - a change to any other field (type, client_email, client_id, token_uri, credential_source, universe_domain, ...) is refused naming the field, since those decide which principal the server runs as and where credential material is sent. The one undetectable case is an authorized_user re-auth to a different Google account under gcloud's constant client ID - indistinguishable on disk from the routine re-auth this feature exists to survive - so it is adopted, and every adoption and refusal is logged with the source name. Two things we deliberately left out of this PR, flagging for your call: (1) the Dataplex catalog client does not currently receive the source's quotaProject on any path; making it consistent with the BigQuery clients would change live quota attribution for existing configs, so we would rather do it as a separate PR if you want it. (2) An alternative design is to rebuild the clients when credentials change (keyed like the existing user-OAuth client cache), which would make the drift guards unnecessary - bigger surgery, but happy to go that way if you prefer it.

hiteshgupta88:
Hi @duwenxin99! Gentle nudge on this one - all four of your comments are addressed in 9168d04 (reloading source moved to internal/sources as NewReloadingDefaultCredentials, reload now keyed on the ADC file actually changing on disk rather than error shape, impersonation base credentials covered in both init paths, and identity/quota-project drift guards with actionable errors). The replies on your threads have the details, and two of them end with small questions for you (whether you want the Dataplex quotaProject consistency change and the other ADC-using sources wired up here or as follow-up PRs). integration-test-pr also needs a maintainer to authorize the run for a fork PR when you get a chance. Happy to adjust anything - thanks!

PR #3569, opened 2026-07-07
refactor(server): make group the source of truth in ResourceManager

twishabansal:
## Description

Makes `internal/group` the source of truth in `ResourceManager` and derives the `tools.Toolset` / `prompts.Promptset` views from it.

Stacked on https://github.com/googleapis/mcp-toolbox/pull/3488


NoClueMike (approved):
Truth Works...

internal/server/mcp/v20241105/method_test.go
```diff
@@ -181,7 +181,7 @@ func TestToolsListHandler(t *testing.T) {
 		{
 			name:        "invalid json body",
 			rawBody:     []byte(`{invalid json}`),
-			toolset:     toolsets[""],
+			toolset:     groups[""].ToToolset(),
```
Yuan325: we can also update this to `group group.Group`. And update the `toolsListHandler to take the `group.Group`, as well as the `GenerateListToolsResult()`. The `GenerateListToolResult()` will eventually grab the `group.ToolNames`~

twishabansal: I moved the materialization to the ResourceManager boundary (GetToolset builds the toolset from the group's names), so the group stays a clean primitive. The full group.Group thread-through is coming in #3576.

cmd/internal/skills/command.go
```diff
@@ -251,6 +252,16 @@ func (c *skillsCmd) collectTools(ctx context.Context, opts *internal.ToolboxOpti
 		return toolsetTools
 	}
```
Yuan325:
```suggestion
```

Seems like this could be removed?

internal/server/resources/resources.go
Yuan325, line 1: I like that we are keeping the GetToolset and GetPromptset in `resources.go` for now~ Since the non-mcp API might be using it. For the mcp endpoint, we might want to migrate all `GetToolset` and `GetPromptset` to `GetGroup` at some point (can do it in later PRs though).

internal/server/resources/resources_test.go
```diff
@@ -73,19 +63,25 @@ func TestUpdateServer(t *testing.T) {
...
+	if gotToolset.Name != "example-toolset" || !gotToolset.ContainsTool("example-tool") {
+		t.Errorf("error updating server, toolset = %+v, want name %q containing tool %q", gotToolset, "example-toolset", "example-tool")
 	}
 
-	gotPrompt, _ := resMgr.GetPrompt("example-prompt")
-	if diff := cmp.Diff(gotPrompt, newPrompts["example-prompt"]); diff != "" {
-		t.Errorf("error updating server, prompts (-want +got):\n%s", diff)
+	gotPrompt, ok := resMgr.GetPrompt("example-prompt")
+	if !ok || gotPrompt == nil {
+		t.Errorf("error updating server, prompt %q not found", "example-prompt")
```
Yuan325: might not want to update this~

internal/server/resources/resources_test.go
```diff
@@ -73,19 +63,25 @@ func TestUpdateServer(t *testing.T) {
...
-	gotPromptset, _ := resMgr.GetPromptset("example-promptset")
-	if diff := cmp.Diff(gotPromptset, newPromptsets["example-promptset"], cmp.AllowUnexported(prompts.Promptset{})); diff != "" {
-		t.Errorf("error updating server, promptset (-want +got):\n%s", diff)
+	gotPromptset, ok := resMgr.GetPromptset("example-promptset")
+	if !ok {
+		t.Fatal("expected promptset \"example-promptset\" to exist")
+	}
+	if gotPromptset.Name != "example-promptset" || !gotPromptset.ContainsPrompt("example-prompt") {
+		t.Errorf("error updating server, promptset = %+v, want name %q containing prompt %q", gotPromptset, "example-promptset", "example-prompt")
 	}
```
Yuan325, line 109: can we add a test case for `GetGroup()`? and compare the struct with `cmp.Diff`~

internal/server/resources/resources_test.go
```diff
@@ -39,24 +41,12 @@ func TestUpdateServer(t *testing.T) {
...
 	newEmbeddingModels := map[string]embeddingmodels.EmbeddingModel{"example-model": nil}
 	newTools := map[string]tools.Tool{"example-tool": nil}
-	newToolsets := map[string]tools.Toolset{
-		"example-toolset": {
-			ToolsetConfig: tools.ToolsetConfig{
-				Name: "example-toolset",
-			},
-			Tools: []*tools.Tool{},
-		},
+	newPrompts := map[string]prompts.Prompt{"example-prompt": testutils.NewMockPrompt("example-prompt", "", prompts.Arguments{})}
```
Yuan325, line 45: Might want to add a `wantToolset` and `wantPromptset` and compare the full struct directly with `cmp.Diff` to make sure there's no extra field that is returned~

internal/server/api.go
```diff
@@ -80,6 +80,7 @@ func toolsetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 		_ = render.Render(w, r, newErrResponse(err, http.StatusInternalServerError))
 		return
 	}
+	manifest.ServerVersion = s.version
```
Yuan325, line 83: Do we need this here? 



also, we might need to add `toolsetManifest()` and `promptsetManifest()` function to `Group struct`.

note that this endpoint will not be getting the groups capability (being able to retrieve group that consist of both toolset and promptset etc). We are just migrating the toolset and promptset src code over to Groups for the api endpoint.

We can take care of this in a separate PR though~ :)

internal/server/config.go
```diff
@@ -59,8 +59,6 @@ type ServerConfig struct {
 	ToolsetConfigs ToolsetConfigs
```
Yuan325: Might want to remove this too?

internal/server/config.go
```diff
@@ -161,7 +159,6 @@ type EmbeddingModelConfigs map[string]embeddingmodels.EmbeddingModelConfig
 type ToolConfigs map[string]tools.ToolConfig
 type ToolsetConfigs map[string]tools.ToolsetConfig
```
Yuan325: Might want to remove this too?

internal/server/server.go
```diff
@@ -327,63 +296,72 @@ func initializeTools(ctx context.Context, cfg ServerConfig, instrumentation *tel
...
+	allPromptNames := make([]string, 0, len(promptsMap))
+	for name := range promptsMap {
+		allPromptNames = append(allPromptNames, name)
 	}
-	cfg.ToolsetConfigs[""] = tools.ToolsetConfig{Name: "", ToolNames: allToolNames}
 
-	toolsetsMap := make(map[string]tools.Toolset)
+	// Convert each legacy toolset into a tools-only group config, then seed the
+	// default nameless group with all tools and all prompts.
+	groupConfigs := make(map[string]group.GroupConfig, len(cfg.ToolsetConfigs)+1)
```
```suggestion
	groupConfigs := make(map[string]group.GroupConfig, len(cfg.GroupConfigs)+1)
```

We probably want to minimize keeping toolset/promptset stuff as soon as possible (to keep the close cleaner and ease of removal in the future) so converting them to `GroupConfigs` in the unmarshaling stage might be the best. With that, we can just use the `GroupConfigs` here~

internal/server/config.go
```diff
@@ -161,7 +159,6 @@ type EmbeddingModelConfigs map[string]embeddingmodels.EmbeddingModelConfig
...
 type PromptConfigs map[string]prompts.PromptConfig
-type PromptsetConfigs map[string]prompts.PromptsetConfig
 
 func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, AuthServiceConfigs, EmbeddingModelConfigs, ToolConfigs, ToolsetConfigs, PromptConfigs, error) {
```
Yuan325: In this function, we can remove the `toolsetConfigs` and `promptConfigs`. Those will be converted into `groupConfigs` and add to that instead~

internal/server/config.go
Yuan325, line 529: we can keep all the code above, but before returning, we can convert this into `groups.GroupConfig`.

internal/server/server_test.go
```diff
@@ -288,19 +277,25 @@ func TestUpdateServer(t *testing.T) {
...
 	}
 
-	gotToolset, _ := s.ResourceMgr.GetToolset("example-toolset")
-	if diff := cmp.Diff(gotToolset, newToolsets["example-toolset"], cmp.AllowUnexported(tools.Toolset{})); diff != "" {
-		t.Errorf("error updating server, toolset (-want +got):\n%s", diff)
+	gotToolset, ok := s.ResourceMgr.GetToolset("example-toolset")
+	if !ok {
+		t.Fatal("expected toolset \"example-toolset\" to exist")
+	}
+	if gotToolset.Name != "example-toolset" || !gotToolset.ContainsTool("example-tool") {
```
Yuan325: Similar with `resources_test.go`, we might want to check the full struct :)

Yuan325 (changes requested):
Thank you! LMK if you have more questions~ 

headsup, there's a PR (https://github.com/googleapis/mcp-toolbox/pull/3511) that update all `internal/server/resources/*` to `internal/server/primitives/*`. That change might cause alot of merge conflict with this PR~

cmd/internal/skills/command.go
twishabansal: Thanks for catching that. Done!

internal/server/resources/resources_test.go
twishabansal: Reverted back to original

internal/server/resources/resources_test.go
twishabansal: Updated

internal/server/resources/resources_test.go
twishabansal: Added. Thanks for pointing out!

internal/server/server_test.go
twishabansal: Done. Thanks for pointing out :)

internal/server/resources/resources.go
twishabansal: For the MCP endpoint, we are updating the toolset and promptset methods to use groups in                                https://github.com/googleapis/mcp-toolbox/pull/3576. This is a follow up PR for the groups primitive.

twishabansal:
> Thank you! LMK if you have more questions~
> 
> headsup, there's a PR (#3511) that update all `internal/server/resources/*` to `internal/server/primitives/*`. That change might cause alot of merge conflict with this PR~

Thank you for pointing this out. Currently I'm merging all the groups feat PRs into https://github.com/googleapis/mcp-toolbox/pull/3605. It will help us resolve any merge conflicts while also landing the complete feature in main at once. I will take care of any merge conflicts before merging the master PR: https://github.com/googleapis/mcp-toolbox/pull/3605

internal/server/config.go
twishabansal: Updated https://github.com/googleapis/mcp-toolbox/pull/3575 to remove this!

internal/server/config.go
twishabansal: Updated https://github.com/googleapis/mcp-toolbox/pull/3575 to remove this!

internal/server/config.go
twishabansal: Updated in #3575

internal/server/config.go
twishabansal: Done in #3575

internal/server/server.go
twishabansal: Done in #3575

internal/server/api.go
twishabansal: Created another PR for this: https://github.com/googleapis/mcp-toolbox/pull/3608

 Before #3569, the toolset was built once at startup with the real version baked into Toolset.Manifest.ServerVersion, and stored. The handler just read it out but the version came along for free.

#3569 stops storing toolsets. Now GetToolset rebuilds the toolset fresh on every request via Initialize("", ...) and it passes an empty version, because ResourceManager doesn't have the version (it only holds the resource graph, not s.version).

So without a fix, /api/toolset would return "serverVersion": "".

This line  puts the version back, using s.version from the Server. It's needed simply because the version is no longer carried by the derived toolset, so the handler has to supply it.
  
  (#3608 then removes this line by building the manifest with the version passed in up front, so there's nothing left to patch.)

Yuan325 (approved):

PR #3575, opened 2026-07-08
feat(group): parse kind: group config and wire into initialization

twishabansal:
## Description

Adds parsing for `kind: group` configuration documents and wires the parsed groups through initialization. Groups are the source of truth from which toolset and promptset views are derived; this PR lets users declare them directly in config.

- `UnmarshalResourceConfig` now parses both flat (`kind: group`) and nested (`groups:`) formats, returning a new `GroupConfigs` value.
- Validation: rejects duplicate default (nameless) or named groups, and rejects a default group that declares `tools` or `prompts` (it always contains all configured tools and prompts).
- `GroupConfigs` flows through config merge, dynamic reload, and `initializeGroups`; a `kind: group` definition overrides a same-named toolset (with a warning).
- Mechanically updated all `UnmarshalResourceConfig` call sites for the new return arity.

Stacked on #3569


cmd/internal/config.go
```diff
@@ -150,7 +151,7 @@ func (p *ConfigParser) ParseConfig(ctx context.Context, raw []byte) (Config, err
 	}
 
 	// Parse contents
-	config.Sources, config.AuthServices, config.EmbeddingModels, config.Tools, config.Toolsets, config.Prompts, err = server.UnmarshalResourceConfig(ctx, raw)
+	config.Sources, config.AuthServices, config.EmbeddingModels, config.Tools, config.Toolsets, config.Prompts, config.Groups, err = server.UnmarshalResourceConfig(ctx, raw)
```
Yuan325: With change from the #3569, we can probably remove `config.Toolsets` here.

cmd/internal/config.go
```diff
@@ -376,11 +380,24 @@ func mergeConfigs(files ...Config) (Config, error) {
...
 		}
+
+		// Check for conflicts and merge groups
+		for name, grp := range file.Groups {
+			if _, exists := merged.Groups[name]; exists {
+				if name == "" {
+					conflicts = append(conflicts, fmt.Sprintf("default nameless group (file #%d)", fileIndex+1))
+				} else {
+					conflicts = append(conflicts, fmt.Sprintf("group '%s' (file #%d)", name, fileIndex+1))
+				}
```
Yuan325: do we need to check for empty name here? iirc, we should one group with the default name?

cmd/internal/config.go
```diff
@@ -376,11 +380,24 @@ func mergeConfigs(files ...Config) (Config, error) {
 				merged.Prompts[name] = prompt
 			}
 		}
+
+		// Check for conflicts and merge groups
+		for name, grp := range file.Groups {
+			if _, exists := merged.Groups[name]; exists {
```
Yuan325, line 375: We might also want to check for conflict with toolset name.

internal/server/config.go
```diff
@@ -159,20 +162,22 @@ type EmbeddingModelConfigs map[string]embeddingmodels.EmbeddingModelConfig
...
+type GroupConfigs map[string]group.GroupConfig
 
-func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, AuthServiceConfigs, EmbeddingModelConfigs, ToolConfigs, ToolsetConfigs, PromptConfigs, error) {
+func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, AuthServiceConfigs, EmbeddingModelConfigs, ToolConfigs, ToolsetConfigs, PromptConfigs, GroupConfigs, error) {
 	// prepare configs map
 	var sourceConfigs SourceConfigs
 	var authServiceConfigs AuthServiceConfigs
 	var embeddingModelConfigs EmbeddingModelConfigs
 	var toolConfigs ToolConfigs
 	var toolsetConfigs ToolsetConfigs
```
Yuan325: Can probably remove this here as well.

internal/server/config.go
```diff
@@ -184,27 +189,34 @@ func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, Au
...
-				return nil, nil, nil, nil, nil, nil, fmt.Errorf("%s missing 'kind' field or it is not a string", formatDocLocation(docIndex, keyToken(doc.Body, "kind"), doc.Body))
+				return nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("%s missing 'kind' field or it is not a string", formatDocLocation(docIndex, keyToken(doc.Body, "kind"), doc.Body))
 			}
-			return nil, nil, nil, nil, nil, nil, fmt.Errorf("missing 'kind' field or it is not a string: %v", resource)
+			return nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("missing 'kind' field or it is not a string: %v", resource)
 		}
 		if name, ok = resource["name"].(string); !ok {
+			// A `kind: group` may omit `name` (or leave it empty) to target the
+			// default nameless group; every other resource requires a name.
+			if rawName, present := resource["name"]; kind == "group" && (!present || rawName == nil) {
```
Yuan325:
```suggestion
			if kind == "group" {
```

Should be good with checking the kind?

internal/server/config.go
```diff
@@ -277,22 +289,40 @@ func UnmarshalResourceConfig(ctx context.Context, raw []byte) (SourceConfigs, Au
...
+			}
+			if groupConfigs == nil {
+				groupConfigs = make(GroupConfigs)
+			}
+			if _, exists := groupConfigs[name]; exists {
+				if name == "" {
+					return nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("more than one default (nameless) group declared; only one is allowed")
+				}
+				return nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("group %q declared more than once", name)
+			}
```
Yuan325, line 314: good catch on this, we might wanna add this check to other primitives as well. Seems like we are not checking for duplicated source/tool/prompt/etc. right now? 


```suggestion
			if _, exists := groupConfigs[name]; exists {
				return nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("group %q declared more than once", name)
			}
```
Other than that, is there any reason we explicitly check for `name == ""`? I think the rule is the same as other groups~ only once allowed.

internal/server/config.go
```diff
@@ -483,6 +513,33 @@ func UnmarshalYAMLToolsetConfig(ctx context.Context, name string, r map[string]a
...
+func UnmarshalYAMLGroupConfig(ctx context.Context, name string, r map[string]any) (group.GroupConfig, error) {
+	// `name` is resolved by the caller; strip it so the strict decoder only sees
+	// the group body fields (description, tools, prompts).
+	body := make(map[string]any, len(r))
+	for k, v := range r {
+		if k == "name" {
+			continue
+		}
+		body[k] = v
+	}
```
Yuan325: wondering why do we strip the name off? iirc, the decoder should be able to decode the full thing even if name is in it~

internal/server/config.go
```diff
@@ -483,6 +513,33 @@ func UnmarshalYAMLToolsetConfig(ctx context.Context, name string, r map[string]a
...
+	dec, err := util.NewStrictDecoder(body)
+	if err != nil {
+		return group.GroupConfig{}, fmt.Errorf("error creating decoder: %s", err)
+	}
+	var gc group.GroupConfig
```
Yuan325:
```suggestion
	gc := group.GroupConfig{Name: name}
```

internal/server/config.go
```diff
@@ -483,6 +513,33 @@ func UnmarshalYAMLToolsetConfig(ctx context.Context, name string, r map[string]a
...
+	if err := dec.DecodeContext(ctx, &gc); err != nil {
+		return group.GroupConfig{}, fmt.Errorf("unable to unmarshal group: %s", err)
+	}
+	gc.Name = name
```
Yuan325:
```suggestion
```

internal/server/config.go
```diff
@@ -483,6 +513,33 @@ func UnmarshalYAMLToolsetConfig(ctx context.Context, name string, r map[string]a
...
+		}
+		body[k] = v
+	}
+	dec, err := util.NewStrictDecoder(body)
+	if err != nil {
+		return group.GroupConfig{}, fmt.Errorf("error creating decoder: %s", err)
+	}
+	var gc group.GroupConfig
+	if err := dec.DecodeContext(ctx, &gc); err != nil {
+		return group.GroupConfig{}, fmt.Errorf("unable to unmarshal group: %s", err)
```
Yuan325, line 539:
```suggestion
		return nil, fmt.Errorf("unable to unmarshal group: %s", err)
```

internal/server/server_test.go
```diff
@@ -1324,14 +1324,166 @@ scopesRequired:
...
-			_, _, _, _, _, _, err := server.UnmarshalResourceConfig(ctx, []byte(tc.yaml))
+			_, _, _, _, _, _, _, err := server.UnmarshalResourceConfig(ctx, []byte(tc.yaml))
 			if (err != nil) != tc.wantError {
 				t.Fatalf("UnmarshalResourceConfig() returned error: %v, wantError: %v", err, tc.wantError)
 			}
 		})
 	}
 }
 
+func TestGroupConfigParsing(t *testing.T) {
```
Yuan325, line 1359: Thanks for adding this!

We can probably combine this 2 unit test into a single one by adding `want group.GroupConfig`. if `wantError == false`, then we can compare the return GroupConfig to make sure its equal. If you can add one test case with `kind: toolset` and make sure it returns the right `GroupConfig` as well that will be great!

Yuan325 (changes requested):

cmd/internal/config.go
twishabansal: Done.

internal/server/config.go
twishabansal: Done.

internal/server/config.go
twishabansal: Thanks for pointing out. Fixed.

internal/server/config.go
twishabansal: Done.

internal/server/config.go
twishabansal: fixed

internal/server/server_test.go
twishabansal: Done. Thanks for the suggestion!

cmd/internal/config.go
twishabansal: toolsets no longer have a separate namespace — a kind: toolset becomes a GroupConfig at unmarshal. So a toolset/group name collision is already caught. There's no separate toolset map left to check against.

cmd/internal/config.go
twishabansal: Updated.

internal/server/config.go
twishabansal: Fixed.

Deeven-Seru:
Hi @Yuan325 and @twishabansal,

I think this simplification unintentionally swallows invalid name types (like `name: 123` or booleans); since they fail the `.(string)` assertion, they now silently default to `""` instead of throwing a validation error. 

I think it's better to probably reinstate:
```go
if rawName, present := resource["name"]; kind == "group" && (!present || rawName == nil)

internal/server/config.go
twishabansal: Added another PR for duplicate validation across the different primitives: https://github.com/googleapis/mcp-toolbox/pull/3612.

Removed the `name==""` check

PR #3576, opened 2026-07-08
fix(server): scope MCP prompts to the connected group

twishabansal:
## Description

The MCP route is `/{toolsetName}`, so `promptsetName` was always empty and `GetPromptset("")` returned every prompt regardless of which group the client connected to. This fixes `prompts/list` (and `prompts/get`) to scope prompts to the connected group by resolving the group from the URL name and serving its derived promptset. `tools/list` is unchanged — tools were already scoped by the toolset name.

### Old vs new response

Given two groups `docs` (prompt `summarize`) and `i18n` (prompt `translate`), connected to `/mcp/docs`:

**`prompts/list`** request:

```json
{ "jsonrpc": "2.0", "id": 1, "method": "prompts/list" }
```

Old (bug — returns every prompt on the server):

```json
{ "jsonrpc": "2.0", "id": 1, "result": { "prompts": [
  { "name": "summarize", "description": "Summarize a document" },
  { "name": "translate", "description": "Translate text" }
] } }
```

New (scoped to the `docs` group):

```json
{ "jsonrpc": "2.0", "id": 1, "result": { "prompts": [
  { "name": "summarize", "description": "Summarize a document" }
] } }
```

**`prompts/get`** for a prompt in another group (`translate`, which belongs to `i18n`):

Old (bug — succeeds, since the default all-prompts set was used):

```json
{ "jsonrpc": "2.0", "id": 2, "result": { "description": "Translate text", "messages": [ ... ] } }
```

New (correctly rejected):

```json
{ "jsonrpc": "2.0", "id": 2, "error": {
  "code": -32602,
  "message": "invalid prompt name: prompt with name \"translate\" does not exist"
} }
```

Stacked on #3575


internal/server/mcp.go
```diff
@@ -580,7 +579,6 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 	}
 
 	toolsetName := chi.URLParam(r, "toolsetName")
```
Yuan325: Might wanna update this to `groupName := chi.URLParam(r, "toolsetName")` eventually to keep it clean~

internal/server/mcp_test.go
```diff
@@ -1856,3 +1859,65 @@ func TestExtractMeta(t *testing.T) {
 		})
 	}
 }
+
+func TestMcpPromptScopingByGroup(t *testing.T) {
```
Yuan325, line 1867: can we have a comment that explains more on what this unit test is? Is this similar to the test we already have in `manifests_test.go`?

Yuan325 (changes requested):
Some questions but overall LGTM, will review again once resolve the 2 little comment :) Thank you!

internal/server/mcp.go
twishabansal: Updated!

internal/server/mcp_test.go
twishabansal: Added a comment to explain this test.

This test exercises the full routing + processMcpMessage path, unlike TestGenerateListPromptsResult in mcp/v20251125/manifests_test.go, which unit tests the GenerateListPromptsResult builder for a single group in isolation.

Let me know if you feel this should be modified or removed.

Yuan325 (approved):
Thank youuu!!!

PR #3578, opened 2026-07-09
docs(source/looker): correct default timeout to 600s

anxkhn:
The Looker source Reference table documents the `timeout` field default as 120s
("By default, 120s is applied"), but the code initializes it to 600s. In
`newConfig` (`internal/sources/looker/looker.go`) the `Config` struct is seeded
with `Timeout: "600s"` before the user config is decoded, so when `timeout` is
left unset the effective default is 600s. That value is parsed via
`time.ParseDuration` and passed to the Looker SDK `rtl.ApiSettings.Timeout`, so
the documented 120s is a 5x understatement of the real behavior.

This updates the single Reference row to read "By default, 600s is applied" so
the docs match the code default. Docs-only change; no source code or tests are
affected.

### PR Checklist

Notes for the checklist:
- No issue was opened first: this is a one-line documentation correction with no
  behavior change. Happy to file an issue if the maintainers prefer.
- No source code changed, so unit tests, linter, and coverage are unaffected
  (this is a Markdown-only edit).

### Issue Reference

No open issue tracks this documentation mismatch. (If an issue is required, I can
open one describing the 120s/600s discrepancy.)


drstrangelooker (approved):

drstrangelooker:
Thanks for the  correction!

PR #3579, opened 2026-07-09
feat(server): set MCP server instructions from group description

twishabansal:
## Description

Makes MCP server `instructions` reflect the connected group's `description`. Previously the `instructions` field existed in every protocol version's result type but was never set.

Stacked on #3576 (base `feat/groups-pr2-prompt-scoping`).


twishabansal:
Closing: design decision — the MCP `initialize`/`serverDiscover` handshake describes the server and must be group-agnostic, so server-level `instructions` is dropped entirely. A group's description will be surfaced only through the groups introspection methods (`groups/list`), not the handshake.

PR #3580, opened 2026-07-09
docs(tool/postgres-list-schemas): correct default limit to 10

anxkhn:
The `postgres-list-schemas` tool documentation states that the `limit` parameter
defaults to `50`, but the tool registers `limit` with `WithIntDefault(10)`
(`internal/tools/postgres/postgreslistschemas/postgreslistschemas.go:121`). That
default flows into `LIMIT COALESCE($3::int, NULL)`, so omitting `limit` actually
returns 10 rows, not 50.

This aligns the doc to the code by changing the documented default from `50` to
`10`. It is a docs-only change: the code default is intentional (the sibling
`postgres-list-database-stats` tool also uses a code default of `10` and its doc
already documents `10`), so the doc is the side that was inconsistent. Changing
the code default instead would be a behavior change, so the surgical fix is to
correct the documentation.

## PR Checklist

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/mcp-toolbox/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea


Yuan325 (approved):
Thank you!!

PR #3584, opened 2026-07-09
feat(skills): make description optional during skills gen

twishabansal:
## Description

Once description becomes optional, skills generation can take in description directly from `Groups`.

Note: group description takes precedence when present, `--description` is used otherwise, and both may be empty.

Based on [#3605](https://github.com/googleapis/mcp-toolbox/pull/3605)


cmd/internal/skills/command_test.go
```diff
@@ -366,6 +365,70 @@ func TestGenerateSkill_MissingArguments(t *testing.T) {
...
+	c := &skillsCmd{name: "my-skill", description: "flag desc", toolset: "my-toolset"}
+	contents, err := c.buildSkillContents(nil, groupsMap)
+	if err != nil {
+		t.Fatalf("buildSkillContents failed: %v", err)
+	}
+
+	if got := contents["my-skill"].description; got != "flag desc" {
+		t.Errorf("toolset mode: got %q, want %q", got, "flag desc")
+	}
+}
```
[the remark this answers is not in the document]
twishabansal: Toolset mode intentionally uses only the `--description` flag, not the group's description (group-description precedence applies only in group mode). This is already covered by `TestBuildSkillContents_ToolsetModeUsesFlagDescription`; the proposed test would fail against the current code.

cmd/internal/skills/command.go
```diff
@@ -257,14 +270,14 @@ func (c *skillsCmd) collectTools(ctx context.Context, opts *internal.ToolboxOpti
...
-		return skillsToTools, nil
+		skillsToContents[c.name] = skillContent{tools: getToolsFromGroup(g), description: c.description}
+		return skillsToContents, nil
 	}
 
 	if len(groupsMap) <= 1 {
 		// Default to all tools if no named group found
-		skillsToTools[c.name] = toolsMap
-		return skillsToTools, nil
+		skillsToContents[c.name] = skillContent{tools: toolsMap, description: c.description}
```
Yuan325: should we check if the group have an existing description?

cmd/internal/skills/command_test.go
Yuan325, line 1: We can probably combine the 3 Test functions into one with the table-struct testing format? Seems to be checking the same thing with different cases. Might want to compare the content struct directly (also to make sure that no additional items was added)

Yuan325 (changes requested):

cmd/internal/skills/command_test.go
twishabansal: Thanks for pointing out. Updated!

cmd/internal/skills/command.go
twishabansal: Updated to check existing description here too

cmd/internal/skills/command.go
```diff
@@ -229,17 +235,24 @@ func run(cmd *skillsCmd, opts *internal.ToolboxOptions) error {
 	return nil
 }
 
-func (c *skillsCmd) collectTools(ctx context.Context, opts *internal.ToolboxOptions) (map[string]map[string]tools.Tool, error) {
+func (c *skillsCmd) collectTools(ctx context.Context, opts *internal.ToolboxOptions) (map[string]skillContent, error) {
```
helloeve: nit: rename this to collectContents?

helloeve (approved):

Yuan325 (approved):

cmd/internal/skills/command.go
twishabansal: Done.

PR #3585, opened 2026-07-09
feat(skills): add --group flag to generate a skill from one group

twishabansal:
## Description

Adds a `--group <name>` flag to `skills-generate` that selects a single named group and generates one skill (named exactly `--name`) from that group's tools.

The description follows the same precedence as PR #3584: the group's own `description` is used when present, `--description` is the fallback. `--group` is mutually exclusive with `--toolset`.

This complements the existing behaviors:
- no flag + multiple groups → one skill per group (`--name-<group>`)
- `--toolset X` → single skill from toolset X (legacy; uses `--description`)
- `--group X` → single skill from group X (uses group description, falls back to `--description`)

Based on https://github.com/googleapis/mcp-toolbox/pull/3584


cmd/internal/skills/command.go
Yuan325, line 295: We can probably dedup these~

cmd/internal/skills/command_test.go
```diff
@@ -391,6 +481,48 @@ func TestBuildSkillContents_GroupDescriptionPrecedence(t *testing.T) {
 	}
 }
 
+func TestBuildSkillContents_GroupMode(t *testing.T) {
```
Yuan325: This could probably be added to the #3584 PR's test under one of the testcases~?

cmd/internal/skills/command_test.go
```diff
@@ -315,6 +315,96 @@ toolsets:
...
+    source: my-sqlite
+    description: "bye tool"
+    statement: "SELECT 'bye' as greeting"
+toolsets:
+  greeting:
+    tools:
+      - hello-sqlite
+  farewell:
+    tools:
+      - bye-sqlite
```
Yuan325: can we update this to use flat format for examples? Thank youuuu

Yuan325 (changes requested):
Overall implementation LGTM :) Some feedbacks to keep the code cleaner :) Tyyy!

cmd/internal/skills/command.go
twishabansal: Updated!

cmd/internal/skills/command_test.go
twishabansal: Done.

cmd/internal/skills/command_test.go
twishabansal: Updated!

helloeve (approved):

Yuan325 (approved):

PR #3587, opened 2026-07-09
docs(skills): document skills-generate --group, --name, and description precedence

twishabansal:
Documents the `skills-generate` group/name/description flags.

Note: Currently this PR contains groups docs: https://github.com/googleapis/mcp-toolbox/pull/3629 which will be removed when that PR is merged.


Yuan325 (approved):
Wondering have we thought about using the group-name/toolset-name as the name if `--name` is not provided`? In that case, the only time we require name is when user are generating skills from the default group.

Otherwise LGTM

twishabansal:
> Wondering have we thought about using the group-name/toolset-name as the name if `--name` is not provided`? In that case, the only time we require name is when user are generating skills from the default group.

This makes sense. I have updated https://github.com/googleapis/mcp-toolbox/pull/3586 to use this as well as this PR to include the updated description. Thanks for the suggestion!

twishabansal:
Superseded: the docs in this PR have been redistributed to the corresponding feature PRs so each ships with its own documentation.

- skills-generate `--description` optional → #3584
- skills-generate `--group` flag + "Generating from a Group" example → #3585
- skills-generate `--name` defaulting + prebuilt example → #3586
- Groups feature docs (`groups/_index.md`, the Groups section in `configuration/_index.md`, prompts/toolsets notices) already live in #3605.

Closing as no longer needed.

PR #3590, opened 2026-07-09
fix(server/mcp): return method-not-found errors with HTTP 200, not 404

### LINKED ISSUE #3588: Streamable HTTP: JSON-RPC "method not found" is returned as HTTP 404, which spec-compliant clients must treat as session expiry
# Streamable HTTP: JSON-RPC "method not found" is returned as HTTP 404, which spec-compliant clients must treat as session expiry

> [!NOTE]
> This investigation was performed and this issue drafted by an AI assistant (Claude Code), working from a real production incident; it was reviewed by a human before submission. The code reference, spec citations, and observed behavior were verified against the v1.6.0 release, current `main`, and live traffic — but the curl reproduction below is written from that analysis rather than freshly executed, so please forgive (and correct) any minor inaccuracies.

## Summary

When Toolbox receives a JSON-RPC request for a method it does not implement (e.g. `resources/list` — Toolbox serves only tools and prompts), it correctly builds a JSON-RPC `-32601 Method not found` error, but returns it with **HTTP status 404** instead of 200:

```go
// internal/server/mcp.go
case jsonrpc.METHOD_NOT_FOUND:
    render.Status(r, http.StatusNotFound)
```

Under the MCP Streamable HTTP transport, HTTP 404 has a reserved, load-bearing meaning that is unrelated to JSON-RPC method errors: it signals that the **session has been terminated**, and the client is *required* to drop everything and re-initialize.

From the spec ([Transports § Session Management](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management)):

> 3. The server **MAY** terminate the session at any time, after which it **MUST** respond to requests containing that session ID with HTTP 404 Not Found.
> 4. When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached.

A JSON-RPC *error response* is still a response, and the spec expects it delivered as a normal HTTP exchange ([§ Sending Messages to the Server](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server), point 5): for a POSTed JSON-RPC *request* the server returns either an SSE stream or a single JSON object — protocol-level errors like `-32601` belong in the JSON-RPC `error` member on HTTP 200, not in the HTTP status.

## Impact

Any spec-compliant client that probes an optional method gets its session killed:

- Clients commonly probe `resources/list` / `prompts/list` regardless of advertised capabilities. Against Toolbox, each probe returns 404 → the client obeys the spec, discards the session, and re-initializes → connect/reconnect churn.
- The failure compounds behind MCP gateways/aggregators that multiplex several servers under one endpoint: the gateway fans `resources/list` out, Toolbox answers 404, and the gateway propagates that status to the end client. We hit this in production with Toolbox behind [agentgateway](https://github.com/agentgateway/agentgateway) with claude.ai as the client — the connector was stuck in a permanent initialize → 404 → re-initialize loop (~24 initializes/hour) and intermittently failed to list tools until we reconfigured the gateway to fail open on the 404.
- As a lesser side effect, 4xx on POST also collides with the [backwards-compatibility detection](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#backwards-compatibility) flow, where clients interpret 4xx as "this is an old HTTP+SSE server".

Note that Toolbox's `initialize` response correctly advertises only `tools` and `prompts` capabilities — the problem is solely the HTTP status used when an unadvertised method is called anyway.

## Reproduction

Any prebuilt config reproduces it (v1.6.0, also present on current `main`):

```sh
toolbox --prebuilt mysql --address 127.0.0.1 --port 5000
```

```sh
# initialize — fine, HTTP 200
curl -si http://127.0.0.1:5000/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"repro","version":"0"}}}'

# resources/list — HTTP/1.1 404 Not Found, body contains the -32601 JSON-RPC error
curl -si http://127.0.0.1:5000/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'Mcp-Session-Id: <id from initialize>' \
  -d '{"jsonrpc":"2.0","id":2,"method":"resources/list","params":{}}'
```

## Suggested fix

Return the `-32601` JSON-RPC error with **HTTP 200** (drop the `METHOD_NOT_FOUND → http.StatusNotFound` mapping in `internal/server/mcp.go`), reserving HTTP 404 for its spec-defined meaning: a request carrying an `Mcp-Session-Id` the server no longer recognizes.

mostafa-hz:
Fixes #3588

## Problem

Toolbox maps JSON-RPC `-32601` (method not found) to HTTP 404. Under the MCP streamable HTTP transport, 404 has a reserved meaning unrelated to method errors ([Transports § Session Management](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management)):

> 3. The server **MAY** terminate the session at any time, after which it **MUST** respond to requests containing that session ID with HTTP 404 Not Found.
> 4. When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached.

So any spec-compliant client that probes an optional method (e.g. `resources/list`, which Toolbox doesn't implement) has its session killed and re-initializes — a permanent connect/reconnect loop. Details, real-world impact (Toolbox behind an MCP multiplexing gateway with claude.ai as the client), and repro in #3588.

## Change

- `internal/server/mcp.go`: drop the `METHOD_NOT_FOUND → http.StatusNotFound` mapping so the `-32601` error is returned in the JSON-RPC `error` member on HTTP 200 — consistent with how other request-level errors (e.g. invalid toolset, `-32600`) are already returned. A comment marks the omission as intentional so the mapping doesn't get reintroduced.
- `internal/server/mcp_test.go`: the two existing cases covering this path (`missing method`, `invalid method`) now expect `http.StatusOK`; their JSON-RPC error body assertions are unchanged.

`go test ./internal/server/...`, `go vet`, and `gofmt` are clean.

> [!NOTE]
> This fix was investigated and drafted by an AI assistant (Claude Code) working from a real production incident, and reviewed by a human before submission.

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -650,8 +650,10 @@ func httpHandler(s *Server, w http.ResponseWriter, r *http.Request) {
 					render.Status(r, http.StatusUnauthorized)
 				}
 			}
-		case jsonrpc.METHOD_NOT_FOUND:
-			render.Status(r, http.StatusNotFound)
+		// METHOD_NOT_FOUND (-32601) is intentionally not mapped to an HTTP
+		// error: under MCP streamable HTTP, 404 means the session expired and
+		// requires clients to discard it and re-initialize, so protocol-level
+		// errors must ride on HTTP 200.
 		case jsonrpc.HEADER_MISMATCH, jsonrpc.UNSUPPORTED_PROTOCOL_VERSION:
 			render.Status(r, http.StatusBadRequest)
 		case jsonrpc.INVALID_PARAMS:
--- a/internal/server/mcp_test.go
+++ b/internal/server/mcp_test.go
@@ -921,7 +921,7 @@ func TestMcpEndpoint(t *testing.T) {
 						Request: jsonrpc.Request{},
 					},
 					methodName:     "",
-					wantStatusCode: http.StatusNotFound,
+					wantStatusCode: http.StatusOK,
 					want: map[string]any{
 						"jsonrpc": "2.0",
 						"id":      "missing-method",
@@ -943,7 +943,7 @@ func TestMcpEndpoint(t *testing.T) {
 						},
 					},
 					methodName:     "foo",
-					wantStatusCode: http.StatusNotFound,
+					wantStatusCode: http.StatusOK,
 					want: map[string]any{
 						"jsonrpc": "2.0",
 						"id":      "invalid-method",


Yuan325:
Please refer to the discussion in the issue. Closing this PR as not fixed. Thank you!! :)

PR #3605, opened 2026-07-13
feat: add groups support

twishabansal:
## Description
 
Master PR for adding groups support to MCP Toolbox. Groups currently contain tools and prompts.


Yuan325 (approved):

Yuan325:
Approving assuming that these are all previously approved PRs :)

PR #3608, opened 2026-07-13
refactor(server): build /api toolset manifest from group

twishabansal:
Stacked on #3575


internal/server/api.go
```diff
@@ -66,15 +66,27 @@ func toolsetHandler(s *Server, w http.ResponseWriter, r *http.Request) {
...
+		tool, ok := s.ResourceMgr.GetTool(name)
+		if !ok {
+			err = fmt.Errorf("tool %q does not exist", name)
+			s.logger.DebugContext(ctx, err.Error())
+			_ = render.Render(w, r, newErrResponse(err, http.StatusInternalServerError))
+			return
+		}
+		toolsMap[name] = tool
+	}
+
```
Yuan325: We might not need to build this here~ can we use `s.resourceMgr.GetToolsMap()`? (or `s.primitiveMgr` after the merge).

Yuan325 (approved):
please update the one little change, otherwise LGTM :) Thank you!

internal/server/api.go
twishabansal: Updated.

PR #3612, opened 2026-07-13
fix(server): reject duplicate names across all primitives

### LINKED ISSUE #3648: Check duplicated primitive during unmarshal process
**Problem:**
Currently, duplicated primitive names are only validated in two specific scenarios:
* During ParseConfig() when using a nested format configuration file.
* During configuration merges when using the `--configs` or `--config-folder` flags.

However, if a user provides a single flat-format configuration file, the server boots without warning, and the duplicated primitive is silently overwritten.

**Proposed Solution:**
We want to enforce strict validation and fail fast. The server should check for duplicate primitive names during the UnmarshalPrimitiveConfig() stage and throw an error to prevent the server from starting with a misconfigured state.

twishabansal:
## Description

Follow-up to #3575

Previously only `kind: group` rejected duplicate names. Every other resource kind — `source`, `authService`, `tool`, `toolset`, `embeddingModel`, and `prompt` silently kept the *last* declaration when a name was reused within a config, so a duplicated resource would be dropped with no error.

This makes duplicate detection uniform: `UnmarshalResourceConfig` now returns `"<kind> %q declared more than once"` for any repeated name, matching the behavior already in place for groups.

Stacked on top of #3575 

Fixed #3648


Yuan325 (approved):
Thank you!!!

PR #3615, opened 2026-07-14
feat: core agent-level suppression for read-only tools

anubhav756:
## Description

This PR introduces core framework support for agent-level tool suppression and dynamic annotations in read-only mode.

When a data source or admin service is configured as `readonly: true`, write-capable and destructive tools bound to that source are automatically suppressed from tool registration and pruned from tool groups. This saves valuable LLM agent context window space, eliminates agent confusion by preventing tool hallucination/misuse on read-only resources, and acts as a defense-in-depth layer in conjunction with database-level session locking.

### Design

1. **`sources.Source` Interface Extension**:
   - Added `IsReadOnly() bool` to the `sources.Source` interface.
   - Implemented `IsReadOnly()` across all existing data source drivers (50+ sources).
   - Added `readonly` configuration support to `alloydb-admin` and `cloud-sql-admin` sources.

2. **Tool Suppression Protocol**:
   - Added `ShouldSuppress(sources.Source) bool` to the `tools.Tool` interface.
   - In `BaseTool`, default suppression logic checks if the bound source is in read-only mode and suppresses any tool whose `ReadOnlyHint` annotation is `false`.
   - Read-only tools (`ReadOnlyHint == true`) and unannotated tools are preserved (with a warning logged for unannotated tools to encourage explicit annotation).
   - Tools like `postgresexecutesql` and `mysqlexecutesql` override `ShouldSuppress` to remain enabled because they enforce read-only semantics at the database connection/session level.

3. **Server-Level Tool Filtering & Group Pruning**:
   - During tool initialization (`initializeTools`), write-capable tools bound to read-only sources are suppressed and omitted from the server's registered tools map with an informative log.
   - During group initialization (`initializeGroups`), suppressed tools are cleanly pruned from configured groups while keeping the group intact and preserving standard validation errors for genuinely missing tools.

4. **Dynamic Tool Annotations in MCP Manifests**:
   - Updated MCP version manifests (`v20241105`, `v20250326`, `v20250618`, `v20251125`, `v20260728`) to query `tool.Manifest(src)` dynamically.

5. **Cloud SQL PG Vector Assist Annotations**:
   - Updated vector assist tools (`vectorassistapplyspec`, `vectorassistdefinespec`, `vectorassistgeneratequery`, `vectorassistgetspec`, `vectorassistimprovequeryrecall`, `vectorassistlistspecs`, `vectorassistmodifyspec`) with annotations (`ReadOnlyAnnotations` vs `DestructiveAnnotations`) and support for custom YAML annotations.

## PR Checklist

### DIFF
--- a/docs/en/integrations/alloydb-admin/source.md
+++ b/docs/en/integrations/alloydb-admin/source.md
@@ -46,3 +46,4 @@ useClientOAuth: true
 | type           |  string  |     true     | Must be "alloydb-admin".                                                                                                                       |
 | defaultProject |  string  |     false    | The Google Cloud project ID to use for AlloyDB infrastructure tools.                                                                           |
 | useClientOAuth |  boolean |     false    | If true, the source will use client-side OAuth for authorization. Otherwise, it will use Application Default Credentials. Defaults to `false`. |
+| readOnly       |  boolean |     false    | When set to `true`, suppresses write-capable admin tools. Default: `false`.                                                                    |
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -274,6 +274,7 @@ func InitializeOfflineConfigs(ctx context.Context, cfg ServerConfig) (
 func initializeTools(ctx context.Context, cfg ServerConfig, sourcesMap map[string]sources.Source, instrumentation *telemetry.Instrumentation, l log.Logger) (map[string]tools.Tool, error) {
 	toolsMap := make(map[string]tools.Tool)
 	for name, tc := range cfg.ToolConfigs {
+		var src sources.Source
 		t, err := func() (tools.Tool, error) {
 			_, span := instrumentation.Tracer.Start(
 				ctx,
@@ -286,16 +287,16 @@ func initializeTools(ctx context.Context, cfg ServerConfig, sourcesMap map[strin
 			if err != nil {
 				return nil, fmt.Errorf("unable to initialize tool %q: %w", name, err)
 			}
-			if !cfg.SkipSourceValidation {
-				srcName := t.GetSourceName()
-				var src sources.Source
+
+			if srcName := t.GetSourceName(); srcName != "" && sourcesMap != nil {
 				var ok bool
-				if srcName != "" {
-					src, ok = sourcesMap[srcName]
-					if !ok {
-						return nil, fmt.Errorf("unable to retrieve source %s for tool %s", srcName, name)
-					}
+				src, ok = sourcesMap[srcName]
+				if !ok && !cfg.SkipSourceValidation {
+					return nil, fmt.Errorf("unable to retrieve source %q for tool %q", srcName, name)
 				}
+			}
+
+			if !cfg.SkipSourceValidation {
 				err = t.ValidateSource(src)
 				if err != nil {
 					return nil, err
@@ -306,6 +307,11 @@ func initializeTools(ctx context.Context, cfg ServerConfig, sourcesMap map[strin
 		if err != nil {
 			return nil, err
 		}
+
+		if t.ShouldSuppress(ctx, src) {
+			continue
+		}
+
 		toolsMap[name] = t
 	}
 	toolNames := make([]string, 0, len(toolsMap))
@@ -351,17 +357,20 @@ func initializeGroups(ctx context.Context, cfg ServerConfig, toolsMap map[string
 
 	groupsMap := make(map[string]group.Group)
 	for name, gc := range groupConfigs {
-		if cfg.IgnoreUnknownTools {
-			filteredToolNames := make([]string, 0, len(gc.ToolNames))
-			for _, tn := range gc.ToolNames {
-				if _, ok := toolsMap[tn]; ok {
-					filteredToolNames = append(filteredToolNames, tn)
-				} else {
-					l.WarnContext(ctx, fmt.Sprintf("Skipping missing tool %q in group %q", tn, name))
-				}
+		filteredToolNames := make([]string, 0, len(gc.ToolNames))
+		for _, tn := range gc.ToolNames {
+			if _, ok := toolsMap[tn]; ok {
+				filteredToolNames = append(filteredToolNames, tn)
+			} else if _, isTool := cfg.ToolConfigs[tn]; isTool {
+				l.InfoContext(ctx, fmt.Sprintf("Removing suppressed tool %q from group %q", tn, name))
+			} else if cfg.IgnoreUnknownTools {
+				l.WarnContext(ctx, fmt.Sprintf("Skipping missing tool %q in group %q", tn, name))
+			} else {
+				// Keep it so that Initialize returns the expected error
+				filteredToolNames = append(filteredToolNames, tn)
 			}
-			gc.ToolNames = filteredToolNames
 		}
+		gc.ToolNames = filteredToolNames
 
 		g, err := func() (group.Group, error) {
 			_, span := instrumentation.Tracer.Start(
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -50,10 +50,14 @@ import (
 	v20260728 "github.com/googleapis/mcp-toolbox/internal/server/mcp/v20260728"
 	"github.com/googleapis/mcp-toolbox/internal/sources"
 	"github.com/googleapis/mcp-toolbox/internal/sources/alloydbpg"
+	_ "github.com/googleapis/mcp-toolbox/internal/sources/postgres"
+	_ "github.com/googleapis/mcp-toolbox/internal/sources/sqlite"
 	"github.com/googleapis/mcp-toolbox/internal/telemetry"
 	"github.com/googleapis/mcp-toolbox/internal/testutils"
 	"github.com/googleapis/mcp-toolbox/internal/tools"
 	_ "github.com/googleapis/mcp-toolbox/internal/tools/http"
+	"github.com/googleapis/mcp-toolbox/internal/tools/mysql/mysqlexecutesql"
+	"github.com/googleapis/mcp-toolbox/internal/tools/postgres/postgresexecutesql"
 	"github.com/googleapis/mcp-toolbox/internal/util"
 )
 
@@ -2096,4 +2100,269 @@ func TestNewServer_Extensions(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestShouldSuppressTool(t *testing.T) {
+	ctx, err := testutils.ContextWithNewLogger()
+	if err != nil {
+		t.Fatalf("error setting up logger: %s", err)
+	}
+
+	readOnlySource := &testutils.MockSource{MockSourceConfig: testutils.MockSourceConfig{Name: "readonly-db", ReadOnly: true}}
+	readWriteSource := &testutils.MockSource{MockSourceConfig: testutils.MockSourceConfig{Name: "readwrite-db", ReadOnly: false}}
+
+	boolPtr := func(b bool) *bool { return &b }
+
+	initTool := func(cfg testutils.MockToolConfig) tools.Tool {
+		t, err := cfg.Initialize(ctx)
+		if err != nil {
+			panic(err)
+		}
+		return t
+	}
+
+	testCases := []struct {
+		desc   string
+		source sources.Source
+		tool   tools.Tool
+		want   bool
+	}{
+		{
+			desc:   "nil source -> not suppressed",
+			source: nil,
+			tool:   initTool(testutils.MockToolConfig{ConfigBase: tools.ConfigBase{Name: "some-tool"}, Source: "readonly-db"}),
+			want:   false,
+		},
+		{
+			desc:   "nil tool -> not suppressed",
+			source: readOnlySource,
+			tool:   nil,
+			want:   false,
+		},
+		{
+			desc:   "write tool on read-write source (readOnlyHint: false) -> not suppressed",
+			source: readWriteSource,
+			tool:   initTool(testutils.MockToolConfig{ConfigBase: tools.ConfigBase{Name: "write-tool"}, Source: "readwrite-db", Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(false)}}),
+			want:   false,
+		},
+		{
+			desc:   "write tool on read-only source (readOnlyHint: false) -> suppressed",
+			source: readOnlySource,
+			tool:   initTool(testutils.MockToolConfig{ConfigBase: tools.ConfigBase{Name: "write-tool"}, Source: "readonly-db", Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(false)}}),
+			want:   true,
+		},
+		{
+			desc:   "read-only tool on read-only source (readOnlyHint: true) -> not suppressed",
+			source: readOnlySource,
+			tool:   initTool(testutils.MockToolConfig{ConfigBase: tools.ConfigBase{Name: "readonly-tool"}, Source: "readonly-db", Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(true)}}),
+			want:   false,
+		},
+		{
+			desc:   "unannotated tool on read-only source -> not suppressed",
+			source: readOnlySource,
+			tool:   initTool(testutils.MockToolConfig{ConfigBase: tools.ConfigBase{Name: "unannotated-tool"}, Source: "readonly-db", Annotations: nil}),
+			want:   false,
+		},
+		{
+			desc:   "tool with non-nil annotations but nil readOnlyHint on read-only source -> not suppressed",
+			source: readOnlySource,
+			tool:   initTool(testutils.MockToolConfig{ConfigBase: tools.ConfigBase{Name: "nil-hint-tool"}, Source: "readonly-db", Annotations: &tools.ToolAnnotations{ReadOnlyHint: nil}}),
+			want:   false,
+		},
+		{
+			desc:   "mysql-execute-sql on read-only source -> not suppressed (session locked at DB layer)",
+			source: readOnlySource,
+			tool: func() tools.Tool {
+				t, err := mysqlexecutesql.Config{
+					ConfigBase: tools.ConfigBase{Name: "mysql-execute-sql", Description: "execute sql query"},
+					Type:       "mysql-execute-sql",
+					Source:     "readonly-db",
+				}.Initialize(ctx)
+				if err != nil {
+					panic(err)
+				}
+				return t
+			}(),
+			want: false,
+		},
+		{
+			desc:   "postgres-execute-sql on read-only source -> not suppressed (session locked at DB layer)",
+			source: readOnlySource,
+			tool: func() tools.Tool {
+				t, err := postgresexecutesql.Config{
+					ConfigBase: tools.ConfigBase{Name: "postgres-execute-sql", Description: "execute sql query"},
+					Type:       "postgres-execute-sql",
+					Source:     "readonly-db",
+				}.Initialize(ctx)
+				if err != nil {
+					panic(err)
+				}
+				return t
+			}(),
+			want: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			if tc.tool == nil {
+				return
+			}
+			got := tc.tool.ShouldSuppress(ctx, tc.source)
+			if got != tc.want {
+				t.Errorf("Tool.ShouldSuppress(ctx) = %v, want %v", got, tc.want)
+			}
+			gotNilLogger := tc.tool.ShouldSuppress(context.Background(), tc.source)
+			if gotNilLogger != tc.want {
+				t.Errorf("Tool.ShouldSuppress(nil logger) = %v, want %v", gotNilLogger, tc.want)
+			}
+		})
+	}
+}
+
+func TestInitializeGroups(t *testing.T) {
+	ctx, err := testutils.ContextWithNewLogger()
+	if err != nil {
+		t.Fatalf("error setting up logger: %s", err)
+	}
+	instrumentation, err := telemetry.CreateTelemetryInstrumentation("0.0.0")
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	ctx = util.WithInstrumentation(ctx, instrumentation)
+
+	boolPtr := func(b bool) *bool { return &b }
+
+	t.Run("suppressed tool is pruned from group", func(t *testing.T) {
+		cfg := server.ServerConfig{
+			SourceConfigs: server.SourceConfigs{
+				"readonly-db": &testutils.MockSourceConfig{Name: "readonly-db", ReadOnly: true},
+			},
+			ToolConfigs: server.ToolConfigs{
+				"allowed_read_tool": &testutils.MockToolConfig{
+					ConfigBase:  tools.ConfigBase{Name: "allowed_read_tool"},
+					Source:      "readonly-db",
+					Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(true)},
+				},
+				"suppressed_write_tool": &testutils.MockToolConfig{
+					ConfigBase:  tools.ConfigBase{Name: "suppressed_write_tool"},
+					Source:      "readonly-db",
+					Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(false)},
+				},
+			},
+			GroupConfigs: server.GroupConfigs{
+				"my-custom-group": group.GroupConfig{
+					Name:      "my-custom-group",
+					ToolNames: []string{"allowed_read_tool", "suppressed_write_tool"},
+				},
+			},
+		}
+
+		s, err := server.NewServer(ctx, cfg)
+		if err != nil {
+			t.Fatalf("failed to create server: %v", err)
+		}
+
+		grp, ok := s.PrimitiveMgr.GetGroup("my-custom-group")
+		if !ok {
+			t.Fatalf("expected group 'my-custom-group' to be registered, but not found")
+		}
+
+		// Verify allowed tool is present in group
+		if !grp.ContainsTool("allowed_read_tool") {
+			t.Errorf("expected group to contain 'allowed_read_tool'")
+		}
+
+		// Verify suppressed tool was pruned from group
+		if grp.ContainsTool("suppressed_write_tool") {
+			t.Errorf("expected group to NOT contain suppressed tool 'suppressed_write_tool'")
+		}
+
+		// Verify group ToolNames slice only contains allowed_read_tool
+		wantTools := []string{"allowed_read_tool"}
+		if diff := cmp.Diff(wantTools, grp.ToolNames); diff != "" {
+			t.Errorf("group ToolNames mismatch (-want +got):\n%s", diff)
+		}
+	})
+
+	t.Run("group with all tools suppressed remains registered with empty tools", func(t *testing.T) {
+		cfg := server.ServerConfig{
+			SourceConfigs: server.SourceConfigs{
+				"readonly-db": &testutils.MockSourceConfig{Name: "readonly-db", ReadOnly: true},
+			},
+			ToolConfigs: server.ToolConfigs{
+				"write_tool_1": &testutils.MockToolConfig{
+					ConfigBase:  tools.ConfigBase{Name: "write_tool_1"},
+					Source:      "readonly-db",
+					Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(false)},
+				},
+				"write_tool_2": &testutils.MockToolConfig{
+					ConfigBase:  tools.ConfigBase{Name: "write_tool_2"},
+					Source:      "readonly-db",
+					Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(false)},
+				},
+			},
+			GroupConfigs: server.GroupConfigs{
+				"admin-group": group.GroupConfig{
+					Name:      "admin-group",
+					ToolNames: []string{"write_tool_1", "write_tool_2"},
+				},
+			},
+		}
+
+		s, err := server.NewServer(ctx, cfg)
+		if err != nil {
+			t.Fatalf("failed to create server: %v", err)
+		}
+
+		grp, ok := s.PrimitiveMgr.GetGroup("admin-group")
+		if !ok {
+			t.Fatalf("expected group 'admin-group' to be registered even when empty, but not found")
+		}
+
+		if len(grp.ToolNames) != 0 {
+			t.Errorf("expected group ToolNames to be empty, got: %v", grp.ToolNames)
+		}
+	})
+
+	t.Run("group with no tools suppressed initializes normally", func(t *testing.T) {
+		cfg := server.ServerConfig{
+			SourceConfigs: server.SourceConfigs{
+				"readwrite-db": &testutils.MockSourceConfig{Name: "readwrite-db", ReadOnly: false},
+			},
+			ToolConfigs: server.ToolConfigs{
+				"tool_1": &testutils.MockToolConfig{
+					ConfigBase:  tools.ConfigBase{Name: "tool_1"},
+					Source:      "readwrite-db",
+					Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(false)},
+				},
+				"tool_2": &testutils.MockToolConfig{
+					ConfigBase:  tools.ConfigBase{Name: "tool_2"},
+					Source:      "readwrite-db",
+					Annotations: &tools.ToolAnnotations{ReadOnlyHint: boolPtr(true)},
+				},
+			},
+			GroupConfigs: server.GroupConfigs{
+				"all-tools-group": group.GroupConfig{
+					Name:      "all-tools-group",
+					ToolNames: []string{"tool_1", "tool_2"},
+				},
+			},
+		}
+
+		s, err := server.NewServer(ctx, cfg)
+		if err != nil {
+			t.Fatalf("failed to create server: %v", err)
+		}
+
+		grp, ok := s.PrimitiveMgr.GetGroup("all-tools-group")
+		if !ok {
+			t.Fatalf("expected group 'all-tools-group' to be registered, but not found")
+		}
+
+		wantTools := []string{"tool_1", "tool_2"}
+		if diff := cmp.Diff(wantTools, grp.ToolNames); diff != "" {
+			t.Errorf("group ToolNames mismatch (-want +got):\n%s", diff)
+		}
+	})
 }
--- a/internal/sources/bigquery/bigquery.go
+++ b/internal/sources/bigquery/bigquery.go
@@ -313,6 +313,10 @@ type Session struct {
 	LastUsed     time.Time
 }
 
+func (s *Source) IsReadOnly() bool {
+	return false
+}
+
 func (s *Source) SourceType() string {
 	// Returns BigQuery Google SQL source type
 	return SourceType
--- a/internal/sources/mysql/mysql.go
+++ b/internal/sources/mysql/mysql.go
@@ -91,6 +91,10 @@ type Source struct {
 	Pool *sql.DB
 }
 
+func (s *Source) IsReadOnly() bool {
+	return false
+}
+
 func (s *Source) SourceType() string {
 	return SourceType
 }
--- a/internal/sources/sources.go
+++ b/internal/sources/sources.go
@@ -66,6 +66,7 @@ type SourceConfig interface {
 type Source interface {
 	SourceType() string
 	ToConfig() SourceConfig
+	IsReadOnly() bool
 }
 
 // InitConnectionSpan adds a span for database pool connection initialization
--- a/internal/testutils/mocks.go
+++ b/internal/testutils/mocks.go
@@ -28,13 +28,17 @@ import (
 
 // MockSourceConfig is used to mock source config in tests
 type MockSourceConfig struct {
-	Name string `yaml:"name"`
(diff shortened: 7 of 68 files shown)


internal/server/readonly_tools_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/googleapis/mcp-toolbox/internal/log"
+	"github.com/googleapis/mcp-toolbox/internal/sources"
+	"github.com/googleapis/mcp-toolbox/internal/telemetry"
+	"github.com/googleapis/mcp-toolbox/internal/tools"
+	"github.com/googleapis/mcp-toolbox/internal/util"
+	"github.com/googleapis/mcp-toolbox/internal/util/parameters"
+)
+
+type mockSource struct {
```
Yuan325: Can we have this in `internal/testutils/mocks.go` instead? We can then import it here~ thank you.

internal/server/readonly_tools_test.go
```diff
@@ -0,0 +1,162 @@
...
+func (m *mockSource) SourceType() string {
+	return "mock-db"
+}
+func (m *mockSource) ToConfig() sources.SourceConfig {
+	return nil
+}
+
+var _ sources.ReadOnlySource = (*mockSource)(nil)
+
+type mockToolConfig struct {
```
Yuan325: I believe we have existing mockTool at `testutils/mocks.go`

internal/server/readonly_tools_test.go
Yuan325, line 1: why do we need a separate test file for readonly_tools? Can we move these tests into server_test.go instead since these are testing changes within server.go

internal/server/server.go
```diff
@@ -183,11 +184,28 @@ func InitializeConfigs(ctx context.Context, cfg ServerConfig) (
...
+				} else if _, isTool := cfg.ToolConfigs[tn]; isTool {
+					l.InfoContext(ctx, fmt.Sprintf("Removing suppressed tool %q from toolset %q", tn, tsName))
+				} else {
+					filtered = append(filtered, tn)
+				}
+			}
+			tsCfg.ToolNames = filtered
+			cfg.ToolsetConfigs[tsName] = tsCfg
+		}
+	}
```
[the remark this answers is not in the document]
Yuan325: please fix. Also we might want to add this change within the toolset initialize function. There might be ways to simplify this (e.g. we might not need to add the whole if statement. There's duplicated code within the initializeToolset functions).

internal/server/server.go
```diff
@@ -317,6 +335,33 @@ func initializeTools(ctx context.Context, cfg ServerConfig, instrumentation *tel
 		if err != nil {
 			return nil, err
 		}
+
+		if sourcesMap != nil {
```
Yuan325: can we simplify this chunk? it's doing alot of reflect checks that seems irrelevant?

Yuan325 (changes requested):

internal/server/server.go
```diff
@@ -317,6 +335,33 @@ func initializeTools(ctx context.Context, cfg ServerConfig, instrumentation *tel
...
+						if src, ok := sourcesMap[sourceName]; ok {
+							if rs, ok := src.(sources.ReadOnlySource); ok && rs.IsReadOnlyMode() {
+								if t.GetAnnotations() != nil && t.GetAnnotations().ReadOnlyHint != nil {
+									if !*t.GetAnnotations().ReadOnlyHint {
+										l.InfoContext(ctx, fmt.Sprintf("Suppressing write-capable tool %q bound to read-only source %q", name, sourceName))
+										continue
+									}
+								} else {
+									l.WarnContext(ctx, fmt.Sprintf("Tool %q bound to read-only source %q lacks ReadOnlyHint annotation; executing this tool may fail if it attempts write operations. If this tool is meant to be read-only, please add 'readOnlyHint: true' to its annotations. Otherwise, add 'readOnlyHint: false' to suppress it in read-only mode and save agent context window.", name, sourceName))
+								}
```
Yuan325: Please simplify these 7-level nested if-statements.

internal/server/server.go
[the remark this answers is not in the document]
anubhav756: Updated this. We now copy the config map (`cfg.ToolsetConfigs`) to a local map `toolsetConfigs` during initialization to fix imutability. I kept the filtering logic inside server package's `initializeToolsets` function rather than moving it down into `tools.ToolsetConfig.Initialize()` so that the suppression filtering relies on server-level options like the logger (`log.Logger`), the full list of original `cfg.ToolConfigs`, and the `IgnoreUnknownTools` config flag. If we moved the filter logic down to `ToolsetConfig.Initialize`, we would have to pass the server logger and the server settings down into the tools package making the tools package dependent on server-specific details.

internal/server/server.go
```diff
@@ -183,7 +184,7 @@ func InitializeConfigs(ctx context.Context, cfg ServerConfig) (
 	}
 	l.InfoContext(ctx, fmt.Sprintf("Initialized %d embeddingModels: %s", len(embeddingModelsMap), strings.Join(embeddingModelNames, ", ")))
 
-	toolsMap, err := initializeTools(ctx, cfg, instrumentation, l)
+	toolsMap, err := initializeTools(ctx, cfg, sourcesMap, instrumentation, l)
 	if err != nil {
 		return nil, nil, nil, nil, nil, nil, nil, err
 	}
```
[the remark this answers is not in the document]
anubhav756: Added

internal/server/readonly_tools_test.go
anubhav756: Makes sense. Moved.

internal/server/readonly_tools_test.go
anubhav756: Thanks for pointing that out! Updated and reused that mock.

internal/server/server.go
anubhav756: Thanks for the suggestion. Simplified this with `reflect.Indirect()`.

PS: We would still need `reflect.Indirect()` since `tools.ToolConfig` is a general interface and does not enforce a `GetSource()` method. Since different tool configs expose their bound data source via `Source`, we use reflection to dynamically discover this.

internal/server/server.go
anubhav756: Thanks for the suggestion! Created `shouldSuppressTool()` with a flat early returns to replace this `if` pyramid.

internal/server/readonly_tools_test.go
anubhav756: Thanks for the suggestion! Added `MockReadOnlySource` and `MockSourceConfig` to `mocks.go`.

internal/server/server.go
```diff
@@ -327,6 +333,49 @@ func initializeTools(ctx context.Context, cfg ServerConfig, instrumentation *tel
...
+	if v.Kind() != reflect.Struct {
+		return false
+	}
+
+	sourceField := v.FieldByName("Source")
+	if !sourceField.IsValid() || sourceField.Kind() != reflect.String {
+		return false
+	}
+
+	sourceName := sourceField.String()
```
Yuan325: Instead of this, let's implement `GetSource()` within the `type Tool interface{}`. For tools that does not have a source, we can just return `""`.

This will be type safe, faster, easier to maintain and also cleaner

internal/sources/sources.go
```diff
@@ -68,6 +68,12 @@ type Source interface {
 	ToConfig() SourceConfig
 }
 
+// ReadOnlySource is the interface for a source that supports read-only mode.
+type ReadOnlySource interface {
+	Source
+	IsReadOnlyMode() bool
+}
+
```
Yuan325: I'd recommend adding `IsReadOnlyMode() bool` directly to the Source interface instead of using an optional interface check. Since read-only status seems like a property that we will eventually expand to most of our sources, forcing all sources to explicitly implement it (even if they just return false) gives us compile-time safety and prevents bugs where we forget to add the method to a supported source.

internal/server/server.go
```diff
@@ -327,6 +333,49 @@ func initializeTools(ctx context.Context, cfg ServerConfig, instrumentation *tel
...
+	if sourceName == "" {
+		return false
+	}
+
+	src, ok := sourcesMap[sourceName]
+	if !ok {
+		return false
+	}
+
+	rs, ok := src.(sources.ReadOnlySource)
```
Yuan325: Probably wouldn't need to convert this. Every source should have the `IsReadOnlyMode()` function? In that case we should be able to just check `src.IsReadOnlyMode()` directly.

internal/server/server.go
```diff
@@ -327,6 +333,49 @@ func initializeTools(ctx context.Context, cfg ServerConfig, instrumentation *tel
 	return toolsMap, nil
 }
 
+func shouldSuppressTool(ctx context.Context, l log.Logger, tc tools.ToolConfig, t tools.Tool, toolName string, sourcesMap map[string]sources.Source) bool {
```
Yuan325: Seems like we will only need ctx, logger, t, sourcesMap

internal/server/server.go
```diff
@@ -335,34 +384,39 @@ func initializeToolsets(ctx context.Context, cfg ServerConfig, toolsMap map[stri
...
+				l.WarnContext(ctx, fmt.Sprintf("Skipping missing tool %q in toolset %q", tn, name))
+			} else {
+				// Keep it so that Initialize returns the expected error
+				filteredToolNames = append(filteredToolNames, tn)
 			}
-			tc.ToolNames = filteredToolNames
-			cfg.ToolsetConfigs[name] = tc
 		}
 
+		tcCopy := tc
```
Yuan325: Do we need a copy here? Can we just replace it with `tc.ToolNames = filteredToolNames`?

cmd/root_test.go
Yuan325, line 1: Why did this change? I don't think we made any updates here?

internal/server/server_test.go
Yuan325, line 1: The new test function should be `TestShouldSuppressTool(t *testing.T)`. And only one test function should be sufficient by covering all the cases (readonly-tool is present, write-tool is suppressed, if source name is empty, the possible values of annotations -- basically making sure we check all the return statements of that new function).

Yuan325 (changes requested):
Thank you! Please let me know if you have any questions on the comments~

internal/sources/sources.go
Yuan325: Another nit: update the name to `ReadOnly() bool` or if we can't set it as `ReadOnly()`, the next best choice is `IsReadOnly() bool`. (closer to Go's preference of concise, clean namings). Thankyouuuuu

cmd/root_test.go
anubhav756: Good question!

**tldr:** due to changes asked by [this comment](https://github.com/googleapis/mcp-toolbox/pull/3615#discussion_r3584462896).

Added detailed reasoning in a note in the PR description.

Yuan325:
Update: checks for tools should be part of checking tool's compatibility on source. I'll submit a PR to make some updates. Please hold this off for a little while.

internal/sources/sources.go
anubhav756: Thanks for this suggestion Yuan! Really helpful. Implemented it and updated all the sources with it :)

internal/server/server.go
anubhav756: Updated with the `IsReadOnly` suggested below. Thanks!

internal/server/server.go
anubhav756: Thanks for pointing that out! Done.

internal/server/server.go
anubhav756: Ah thanks for pointing that out! Thanks 😄

internal/server/server_test.go
anubhav756: Done. Thanks for the suggestion!

internal/server/server.go
```diff
@@ -283,16 +284,16 @@ func initializeTools(ctx context.Context, cfg ServerConfig, sourcesMap map[strin
...
+
+			if srcName := t.GetSourceName(); srcName != "" && sourcesMap != nil {
 				var ok bool
-				if srcName != "" {
-					src, ok = sourcesMap[srcName]
-					if !ok {
-						return nil, fmt.Errorf("unable to retrieve source %s for tool %s", srcName, name)
-					}
+				src, ok = sourcesMap[srcName]
+				if !ok && !cfg.SkipSourceValidation {
```
Yuan325, line 294: Do we need to check for `!cfg.SkipSourceValidation` here? seems irrelevant~

internal/server/server.go
```diff
@@ -313,6 +319,33 @@ func initializeTools(ctx context.Context, cfg ServerConfig, sourcesMap map[strin
...
+	}
+
+	annotations := t.GetAnnotations()
+	if annotations == nil || annotations.ReadOnlyHint == nil {
+		if l != nil {
+			l.WarnContext(ctx, fmt.Sprintf("Tool %q bound to read-only source %q lacks ReadOnlyHint annotation; executing this tool may fail if it attempts write operations. If this tool is meant to be read-only, please add 'readOnlyHint: true' to its annotations. Otherwise, add 'readOnlyHint: false' to suppress it in read-only mode and save agent context window.", toolName, sourceName))
+		}
+	}
+
+	return false
```
Yuan325: Can we move all these part of the logic into `t.ShouldSuppress()`? We can send `ctx` into the function and grab `logger` from context.

here we can just `return t.ShouldSuppress(ctx, src)`

internal/testutils/mocks.go
```diff
@@ -122,21 +168,16 @@ func (t MockTool) Invoke(ctx context.Context, s sources.Source, params parameter
...
 	// default to true
 	return !t.unauthorized
 }
 
 func (t MockTool) ValidateSource(src sources.Source) error {
-	if src == nil || src.SourceType() == "mock-source" {
+	if t.ErrValidateSource != nil {
+		return t.ErrValidateSource
+	}
+	if src == nil || src.SourceType() == "" || src.SourceType() == "mock-source" || src.SourceType() == "mock-db" || src.SourceType() == "cloud-sql-postgres" {
```
Yuan325: Why do we need to update this?

internal/testutils/mocks.go
```diff
@@ -81,6 +96,11 @@ var _ tools.ToolConfig = MockToolConfig{}
 // MockTool is used to mock tools in tests
 type MockTool struct {
 	tools.BaseTool[MockToolConfig]
+	Name                       string
+	Source                     string
+	Type                       string
+	Annotations                *tools.ToolAnnotations
+	ErrValidateSource          error
```
Yuan325: Do we need this?

internal/testutils/mocks.go
```diff
@@ -107,6 +127,32 @@ func NewMockTool(name, desc, source string, params []parameters.Parameter, unaut
 	return mt
 }
 
+func (t MockTool) GetName() string {
+	if t.Name != "" {
+		return t.Name
+	}
+	return t.BaseTool.GetName()
+}
+
```
Yuan325: This should already be implemented with BaseTool?

internal/testutils/mocks.go
```diff
@@ -81,6 +96,11 @@ var _ tools.ToolConfig = MockToolConfig{}
 // MockTool is used to mock tools in tests
 type MockTool struct {
 	tools.BaseTool[MockToolConfig]
+	Name                       string
```
Yuan325: shouldn't need this since it is already in MockToolConfig

internal/testutils/mocks.go
```diff
@@ -148,7 +189,24 @@ func (t MockTool) ParseParams(data map[string]any, claimsMap map[string]map[stri
...
+	return t.BaseTool.GetAnnotations()
+}
+
+func (t MockTool) ShouldSuppress(src sources.Source) bool {
+	if src == nil || !src.IsReadOnly() {
+		return false
+	}
+	if t.Type == "postgres-execute-sql" || t.Type == "mysql-execute-sql" {
+		return false
+	}
```
Yuan325: The type for this should be `mock-tool`.

We probably don't wanna test our existing `type` supression with these mocks. If we want to test `postgres-execute-sql` or `mysql-execute-sql` specific, we should test those in the integration test instead. Please remove~

internal/tools/cloudsqlpg/vectorassistdefinespec/vectorassistdefinespec.go
```diff
@@ -102,7 +103,7 @@ func (cfg Config) Initialize(context.Context) (tools.Tool, error) {
 	return Tool{
 		BaseTool: tools.NewBaseTool(
 			cfg,
-			nil,
+			tools.GetAnnotationsOrDefault(cfg.Annotations, tools.NewDestructiveAnnotations),
```
Yuan325: Please double check if this tool is destructive~ seems like this is defining specs and not applying anything.

internal/tools/cloudsqlpg/vectorassistgeneratequery/vectorassistgeneratequery.go
```diff
@@ -64,8 +64,9 @@ type compatibleSource interface {
 
 type Config struct {
 	tools.ConfigBase `yaml:",inline"`
-	Type             string `yaml:"type" validate:"required"`
-	Source           string `yaml:"source" validate:"required"`
+	Type             string                 `yaml:"type" validate:"required"`
+	Source           string                 `yaml:"source" validate:"required"`
+	Annotations      *tools.ToolAnnotations `yaml:"annotations,omitempty"`
```
Yuan325, line 69: makes me think if we should move Annotations to `tools.ConfigBase` since it seems like we are requiring it at all tools now. (just a note for now, please keep this in mind for after readonly is fully implemented, we can add a PR to move this last~)


_feel free to resolve this after you've seen this_

internal/tools/tools.go
```diff
@@ -154,6 +155,7 @@ type Manifest struct {
 	Description  string                         `json:"description"`
 	Parameters   []parameters.ParameterManifest `json:"parameters"`
 	AuthRequired []string                       `json:"authRequired"`
+	Annotations  *ToolAnnotations               `json:"annotations,omitempty"`
```
Yuan325: This might be irrelevant. this `Manifest struct` is used for the native api which is not mcp. I don't think we are adding this feature to the api right?

internal/tools/tools_test.go
```diff
@@ -142,3 +143,63 @@ func TestBaseToolEmbedParamsPassthrough(t *testing.T) {
...
 	}
 }
+
+type fakeSource struct {
+	readOnly bool
+}
+
+func (f fakeSource) SourceType() string             { return "fake" }
+func (f fakeSource) IsReadOnly() bool               { return f.readOnly }
+func (f fakeSource) ToConfig() sources.SourceConfig { return nil }
```
Yuan325: Please remove~ we can probably use MockSource for this.

internal/server/mcp/v20241105/manifests.go
```diff
@@ -118,7 +118,15 @@ func GenerateListToolsResult(pMgr *primitives.PrimitiveManager, g group.Group, u
...
-		toolManifest := generateToolManifest(toolName, tool.GetDescription(), tool.GetAuthRequired(), params, tool.GetAnnotations(), urlParams)
+		manifest, err := tool.Manifest(src)
+		if err != nil {
+			return ListToolsResult{}, fmt.Errorf("error getting manifest for tool %q: %w", toolName, err)
+		}
+		ann := manifest.Annotations
+		if ann == nil {
+			ann = tool.GetAnnotations()
+		}
+		toolManifest := generateToolManifest(toolName, tool.GetDescription(), tool.GetAuthRequired(), params, ann, urlParams)
```
Yuan325: These changes seems irrelevant. Instead you might want to update the code within `generateToolManifest()`. That's where we set the annotations. 

We might want to update the test too.

internal/tools/mysql/mysqlexecutesql/mysqlexecutesql.go
```diff
@@ -126,3 +126,23 @@ func (t Tool) ValidateSource(source sources.Source) error {
...
+func (t Tool) Manifest(source sources.Source) (tools.Manifest, error) {
+	m, err := t.BaseTool.Manifest(source)
+	if err != nil {
+		return tools.Manifest{}, err
+	}
+	if source != nil && source.IsReadOnly() {
+		m.Annotations = tools.NewReadOnlyAnnotations()
+	}
+	return m, nil
+}
```
Yuan325: This might be irrelevant. This is used for the `/api` native endpoint's manifest. Not the MCP manifest. The MCP's manifest is generated within `mcp/<version>/manifest.go`

Yuan325 (changes requested):
Thank you! Please lmk if you have any questions on the comments

internal/server/server.go
anubhav756: I think it's required for offline mode for `skills-generate`?

IIUC, in offline mode we intentionally skip initializing any sources and pass an empty `sourcesMap`:

Because `sourcesMap` is empty in this mode, `ok` would be `false`.

If we remove `&& !cfg.SkipSourceValidation` here, then offline tool init would start failing with `"unable to retrieve source..."` for any tool configured with a source. So that is why this check is important.

Am I missing something?

internal/server/server.go
anubhav756: I want to keep `t.ShouldSuppress(src)` as a pure boolean predicate without side effects. I think the logging in `server.ShouldSuppressTool()` maintains the separation of concerns?

The warning for missing `ReadOnlyHint` annotations and the suppression info log are server-init concerns rather than properties of the tool itself.

We also have some tools overriding `ShouldSuppress` (like `mysqlexecutesql` and `postgresexecutesql`). If we move everything inside base's `t.ShouldSuppress()`, then we would bypass this warning logic or need to replicate it in all these tools.

Because of this the server layer feels cleaner and more idiomatic.

Thoughts?

internal/testutils/mocks.go
anubhav756: Now I made all of them consistent to use the default value `mock-source`, and now it's also consistent with how we have a default value of `mock-tool`.

Earlier I had added mock sources with different names so they had to be allowlisted here.

Thanks for the suggestion!

internal/testutils/mocks.go
anubhav756: I had implemented bare literal mock tools in test. Now I updated it to use the MockToolConfig like the others and was able to cleanup and reuse. Thanks a lot for the suggestion! (PS: This resolves some other suggestions below so resolving them as well).

internal/tools/cloudsqlpg/vectorassistdefinespec/vectorassistdefinespec.go
anubhav756: Thanks for catching that! Yes it's just additive so it must be write not destructive. Updated both `definespec` and `modifyspec` tools. Thanks!

internal/tools/cloudsqlpg/vectorassistgeneratequery/vectorassistgeneratequery.go
anubhav756: Agreed! I've created a backlog issue to track this: https://github.com/googleapis/mcp-toolbox/issues/3809

Thanks for bringing this up!

internal/tools/tools.go
anubhav756: Makes sense! Removed `Annotations` from `tools.Manifest`.

internal/tools/tools_test.go
anubhav756: Done, replaced with `testutils.MockSource`. Thanks!

internal/server/mcp/v20241105/manifests.go
anubhav756: Thanks for the suggestion! Reverted to passing `tool.GetAnnotations()` into `generateToolManifest()` and cleaned up the test.

internal/tools/mysql/mysqlexecutesql/mysqlexecutesql.go
anubhav756: This part is to implement the dynamic flipping of MCP annotations for these multi-mode tools. But I see your point that it shouldn't affect `/api`. Let me take the alternative route of doing that in `GetAnnotations` function only. That way, it will also enable simplifying the tools by enabling removal of the custom overrides of `ShouldSuppress()` logic in those tools, but would require changing in a bunch of places since it would be a signature change. Will add that in a follow up PR. In this PR I'll just remove for now.

internal/server/server.go
Yuan325: I'd say moving those into the tool's level is still a cleaner. These lines seems to be just adding logs rather than doing any logic parsing. That way we don't have to add these additional checks for tools's annotations that are not necessary here~

For tools that are overriding ShouldSuppress, I don't think it affects these logs that are added here?

internal/testutils/mocks.go
```diff
@@ -122,20 +146,15 @@ func (t MockTool) Invoke(ctx context.Context, s sources.Source, params parameter
...
-
 func (t MockTool) Authorized(verifiedAuthServices []string) bool {
 	// default to true
 	return !t.unauthorized
 }
 
 func (t MockTool) ValidateSource(src sources.Source) error {
+	if t.Cfg.ErrValidateSource != nil {
+		return t.Cfg.ErrValidateSource
+	}
```
Yuan325: Doesn't seem like we need these?

internal/testutils/mocks.go
```diff
@@ -56,20 +67,25 @@ func (s MockSource) ToConfig() sources.SourceConfig {
...
-	tools.ConfigBase `yaml:",inline"`
-	Source           string                `yaml:"source"`
-	Parameters       parameters.Parameters `yaml:"parameters"`
-	Type             string                `yaml:"type"`
+	tools.ConfigBase  `yaml:",inline"`
+	Source            string                 `yaml:"source"`
+	Parameters        parameters.Parameters  `yaml:"parameters"`
+	Type              string                 `yaml:"type"`
+	Annotations       *tools.ToolAnnotations `yaml:"annotations,omitempty"`
+	ErrValidateSource error
```
Yuan325: Doesn't seem like we need this field? please remove

internal/server/server_test.go
```diff
@@ -2050,3 +2052,422 @@ func TestNewServer_Extensions(t *testing.T) {
...
+				gotDirect := tc.tool.ShouldSuppress(tc.source)
+				if gotDirect != tc.want {
+					t.Errorf("Tool.ShouldSuppress() = %v, want %v", gotDirect, tc.want)
+				}
+			}
+		})
+	}
+}
+
+func TestInitializeTools_ValidateSourceFailure(t *testing.T) {
```
Yuan325: This seems irrelevant to the changes here.

internal/server/server_test.go
```diff
@@ -2050,3 +2052,422 @@ func TestNewServer_Extensions(t *testing.T) {
...
+	_, _, _, _, _, _, err = server.InitializeConfigs(ctx, cfg)
+	if err == nil {
+		t.Fatalf("expected error from ValidateSource failure, got nil")
+	}
+	if !strings.Contains(err.Error(), "incompatible source type") {
+		t.Errorf("unexpected error message: %v", err)
+	}
+}
+
+func TestNegativeUnsupportedSourceRejection(t *testing.T) {
```
Yuan325: this test seems irrelevant, please remove

internal/server/server_test.go
```diff
@@ -2050,3 +2052,422 @@ func TestNewServer_Extensions(t *testing.T) {
...
+				t.Fatalf("expected unmarshaling error when configuring readonly: true on unsupported source, got nil")
+			}
+			if !strings.Contains(err.Error(), "unknown field") && !strings.Contains(err.Error(), "readonly") {
+				t.Errorf("expected unknown field or readonly error, got: %v", err)
+			}
+		})
+	}
+}
+
+func TestNegativeDirectToolCallBypass(t *testing.T) {
```
Yuan325: this test seems irrelevant, please remove

internal/server/server_test.go
```diff
@@ -2050,3 +2052,422 @@ func TestNewServer_Extensions(t *testing.T) {
...
+		t.Fatalf("failed to create server: %v", err)
+	}
+
+	_, ok := s.PrimitiveMgr.GetTool("suppressed_write_tool")
+	if ok {
+		t.Fatalf("expected suppressed tool 'suppressed_write_tool' to be omitted from registered tools, but it was found")
+	}
+}
+
+func TestNegativeUnannotatedToolWarning(t *testing.T) {
```
Yuan325: this test seems irrelevant, please remove

internal/server/server_test.go
```diff
@@ -2050,3 +2052,422 @@ func TestNewServer_Extensions(t *testing.T) {
...
+	if !ok {
+		t.Fatalf("expected group 'admin-group' to be registered even when empty, but not found")
+	}
+
+	if len(grp.ToolNames) != 0 {
+		t.Errorf("expected group ToolNames to be empty, got: %v", grp.ToolNames)
+	}
+}
+
+func TestAdminSourceReadOnlySuppression(t *testing.T) {
```
Yuan325: this test seems irrelevant, please remove

internal/server/server_test.go
```diff
@@ -2050,3 +2052,422 @@ func TestNewServer_Extensions(t *testing.T) {
...
+		t.Errorf("expected unannotated tool to not be suppressed, but it was")
+	}
+
+	logs := buf.String()
+	if !strings.Contains(logs, "lacks ReadOnlyHint annotation") {
+		t.Errorf("expected warning about missing ReadOnlyHint annotation in logs, got:\n%s", logs)
+	}
+}
+
+func TestInitializeGroups_SuppressedToolPruned(t *testing.T) {
```
Yuan325: can we combine this and the next test to `TestInitializeGroups()` and we can have test cases with tools that are suppressed and tools that are not suppressed, and check the initialization.

internal/tools/mysql/mysqlexecutesql/mysqlexecutesql_test.go
Yuan325, line 1: The tests here seems irrelevant since we hard coded return false.

internal/tools/postgres/postgresexecutesql/postgresexecutesql_test.go
Yuan325, line 1: similarly, the tests here seems irrelevant since we are returning false.

Yuan325 (changes requested):

docs/en/integrations/alloydb-admin/source.md
```diff
@@ -46,3 +46,4 @@ useClientOAuth: true
 | type           |  string  |     true     | Must be "alloydb-admin".                                                                                                                       |
 | defaultProject |  string  |     false    | The Google Cloud project ID to use for AlloyDB infrastructure tools.                                                                           |
 | useClientOAuth |  boolean |     false    | If true, the source will use client-side OAuth for authorization. Otherwise, it will use Application Default Credentials. Defaults to `false`. |
+| readonly       |  boolean |     false    | When set to `true`, suppresses write-capable admin tools. Default: `false`.                                                                    |
```
Yuan325:
```suggestion
| readOnly       |  boolean |     false    | When set to `true`, suppresses write-capable admin tools. Default: `false`.                                                                    |
```

Please update for others as well.

Yuan325:
Are we still planning to update this: Enables multi-mode tools (like SQL execution tools) to dynamically report ReadOnlyHint: true and DestructiveHint: false in MCP tool manifests when connected to a read-only source.

internal/server/server.go
anubhav756: I see your point. I think I am mostly worried about tools that might override `ShouldSuppress()` and may forget about adding these logs (in addition to introducing side-effects in `ShouldSuppress()`). Since currently it's just these two tools, and they don't need this logic for now, adding this side-effect to `ShouldSuppress()` for now.

internal/testutils/mocks.go
anubhav756: Removed. Thanks!

internal/testutils/mocks.go
anubhav756: Removed this error field, and relying back on the "invalid-type" way for injecting errors. Thanks for the suggestion!

internal/server/server_test.go
anubhav756: I see.. thanks for catching!

internal/server/server_test.go
anubhav756: Got it. Removed. Thanks!

internal/server/server_test.go
anubhav756: Removed. Thanks!

internal/server/server_test.go
anubhav756: Removed. Thanks for the suggestion!

internal/server/server_test.go
anubhav756: Got it. Removed. Thanks!

internal/server/server_test.go
anubhav756: Good point! Added and combined :)

internal/tools/mysql/mysqlexecutesql/mysqlexecutesql_test.go
anubhav756: Although we hard code return false, these tests in `mysqlexecutesql_test.go` and `postgresexecutesql.go` ensures that these tools are not accidentally hidden, otherwise there would be no way to run any SQL in read-only mode for these sources. If someone removes this override method, for instance, or if the logic in the surrounding code changes such that even if we return `false`, it somehow ends up being hidden?

I think it makes sense to cover this? For now I removed these from here and added these two cases to the server_test.go's `TestShouldSuppressTool` so we can reuse its logic instead of creating new boilerplate in these test files.

internal/tools/postgres/postgresexecutesql/postgresexecutesql_test.go
anubhav756: Consolidating this discussion in the comment [above](https://github.com/googleapis/mcp-toolbox/pull/3615/#discussion_r3768746692).

docs/en/integrations/alloydb-admin/source.md
anubhav756: Sure! Updated across all PRs. Thanks for the suggestion!

anubhav756:
> Are we still planning to update this: Enables multi-mode tools (like SQL execution tools) to dynamically report ReadOnlyHint: true and DestructiveHint: false in MCP tool manifests when connected to a read-only source.

As mentioned in [this comment](https://github.com/googleapis/mcp-toolbox/pull/3615#discussion_r3763581521), the multi-mode tool dynamic annotation was implemented in `Manifest()` but following the suggestion, I removed it from there, and have implemented in a follow up PR here: https://github.com/googleapis/mcp-toolbox/pull/3816

Yuan325 (approved):

internal/sources/bigquery/bigquery.go
```diff
@@ -321,6 +321,10 @@ type Session struct {
 	LastUsed     time.Time
 }
 
+func (s *Source) IsReadOnly() bool {
+	return false
```
shobsi, line 317: shouldn't it be consistent with `BigQueryWriteMode`?

anubhav756: Thanks for the suggestion @shobsi! I created a [doc](https://docs.google.com/document/d/1Bhm9D4zoQpexaBO9kAyHaf-NwwXuDN6P326qrTH3Y-w/edit?resourcekey=0-YSVPQ_x6TaArPtsdJtHx0w&pli=1&tab=t.btvxffp6ychn) for proposing BigQuery integration that includes this proposed change and makes it consistent as per your suggestion.

Will come back with any updates on this.

anubhav756: Hi @shobsi

Following up on this: we've aligned the BigQuery design in the [design doc](https://docs.google.com/document/d/1Bhm9D4zoQpexaBO9kAyHaf-NwwXuDN6P326qrTH3Y-w/edit?resourcekey=0-YSVPQ_x6TaArPtsdJtHx0w&pli=1&tab=t.btvxffp6ychn#heading=h.35zcfv1t3h0k) and implemented it with `writeMode` in a follow-up PR.

Could you please take a look at https://github.com/googleapis/mcp-toolbox/pull/3851 when you get a chance?

Let me know if you have any feedback or if this looks good for approval.

Thanks!

shobsi: Thanks for putting this together, taking a look

anubhav756: Thanks Shobhit! Closing this thread here to continue discussion in https://github.com/googleapis/mcp-toolbox/pull/3851.

internal/sources/mysql/mysql.go
```diff
@@ -91,6 +91,10 @@ type Source struct {
 	Pool *sql.DB
 }
 
+func (s *Source) IsReadOnly() bool {
+	return false
```
shuzhou-gc, line 95: Does it mean that the customer configuration is ***silently*** ignored?

shuzhou-gc (approved):

internal/sources/mysql/mysql.go
anubhav756: No, it won't be silently ignored. `readOnly` will be treated as any other unknown field for community MySQL and will fail with an `unknown field "readOnly"` error.

### Context
As discussed [here](https://github.com/googleapis/mcp-toolbox/pull/3618#discussion_r3856385601), in community MySQL, there is no session-level lock (unlike Cloud SQL MySQL's `cloudsql_session_read_only=locked`).

Therefore, we don't expose a `readOnly` config flag for it, and `IsReadOnly()` returns `false` so Toolbox does not claim a 100% read-only guarantee or suppress write-capable tools at the source layer.

shuzhou-gc: Got it, thanks.

PR #3618, opened 2026-07-14
feat(cloudsqlmysql): add native read-only session locking

anubhav756:
## Description

This PR implements native, protocol-level Read-Only session enforcement for Cloud SQL for MySQL.

## Changes
- Appends `&session_variables=cloudsql_session_read_only%3Dlocked` to the connection DSN when `readOnly: true` is configured on the source.
- Added `ReadOnly bool` to `cloudsqlmysql.Config` and implemented `IsReadOnly() bool` on `cloudsqlmysql.Source`.
- Updated prebuilt configuration templates (`cloud-sql-mysql.yaml` and `cloud-sql-mysql-admin.yaml`) to dynamically resolve read-only mode via the `CLOUDSQL_MYSQL_READONLY` environment variable (`readonly: ${CLOUDSQL_MYSQL_READONLY:false}`).
- Enforces non-bypassable session locking directly within the MySQL database engine upon connection handshake, thwarting prompt injection mutations and multi-statement breakout attempts.
- Added unit tests, integration vulnerability tests, and updated integration READMEs.

### DIFF
--- a/docs/en/integrations/cloud-sql-mysql/prebuilt-configs/cloud-sql-for-mysql.md
+++ b/docs/en/integrations/cloud-sql-mysql/prebuilt-configs/cloud-sql-for-mysql.md
@@ -14,8 +14,8 @@ description: "Details of the Cloud SQL for MySQL prebuilt configuration."
     *   `CLOUD_SQL_MYSQL_DATABASE`: The name of the database to connect to.
     *   `CLOUD_SQL_MYSQL_USER`: The database username.
     *   `CLOUD_SQL_MYSQL_PASSWORD`: The password for the database user.
-    *   `CLOUD_SQL_MYSQL_IP_TYPE`: The IP type i.e. "Public
-     or "Private" (Default: Public).
+    *   `CLOUD_SQL_MYSQL_IP_TYPE`: The IP type i.e. "Public" or "Private" (Default: Public).
+    *   `CLOUD_SQL_MYSQL_READONLY`: (Optional) When set to `true`, enforces read-only execution on the database connection (`read_only_connection:true`) and suppresses write-capable tools. Default: `false`.
 *   **Permissions:**
     *   **Cloud SQL Client** (`roles/cloudsql.client`) to connect to the
         instance.
--- a/docs/en/integrations/cloud-sql-mysql/source.md
+++ b/docs/en/integrations/cloud-sql-mysql/source.md
@@ -140,3 +140,4 @@ instead of hardcoding your secrets into the configuration file.
 | password  |  string  |    false     | Password of the MySQL user (e.g. "my-password"). Defaults to attempting IAM authentication if unspecified.              |
 | ipType    |  string  |    false     | IP Type of the Cloud SQL instance, must be either `public`,  `private`, or `psc`. Default: `public`.                    |
 | sqlCommenter | boolean |  false     | Overrides the global `--sql-commenter` flag for this source. When set, it takes priority; when omitted, the global flag applies. |
+| readOnly  | boolean  |    false     | When set to `true`, enforces read-only execution on the database connection (`read_only_connection:true`) and suppresses write-capable tools. Default: `false`. |
--- a/internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
+++ b/internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
@@ -22,11 +22,13 @@ database: ${CLOUD_SQL_MYSQL_DATABASE}
 user: ${CLOUD_SQL_MYSQL_USER:}
 password: ${CLOUD_SQL_MYSQL_PASSWORD:}
 ipType: ${CLOUD_SQL_MYSQL_IP_TYPE:PUBLIC}
+readOnly: ${CLOUD_SQL_MYSQL_READONLY:false}
 ---
 kind: source
 name: cloud-sql-admin-source
 type: cloud-sql-admin
 defaultProject: ${CLOUD_SQL_MYSQL_PROJECT:}
+readOnly: ${CLOUD_SQL_MYSQL_READONLY:false}
 ---
 kind: source
 name: cloud-monitoring-source
--- a/internal/sources/cloudsqlmysql/cloud_sql_mysql.go
+++ b/internal/sources/cloudsqlmysql/cloud_sql_mysql.go
@@ -60,6 +60,7 @@ type Config struct {
 	User         string         `yaml:"user"`
 	Password     string         `yaml:"password"`
 	Database     string         `yaml:"database"`
+	ReadOnly     bool           `yaml:"readOnly"`
 	SQLCommenter *bool          `yaml:"sqlCommenter"`
 }
 
@@ -68,7 +69,7 @@ func (r Config) SourceConfigType() string {
 }
 
 func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.Source, error) {
-	pool, err := initCloudSQLMySQLConnectionPool(ctx, tracer, r.Name, r.Project, r.Region, r.Instance, r.IPType.String(), r.User, r.Password, r.Database)
+	pool, err := initCloudSQLMySQLConnectionPool(ctx, tracer, r.Name, r.Project, r.Region, r.Instance, r.IPType.String(), r.User, r.Password, r.Database, r.ReadOnly)
 	if err != nil {
 		return nil, fmt.Errorf("unable to create pool: %w", err)
 	}
@@ -93,7 +94,7 @@ type Source struct {
 }
 
 func (s *Source) IsReadOnly() bool {
-	return false
+	return s.ReadOnly
 }
 
 func (s *Source) SourceType() string {
@@ -209,7 +210,36 @@ func getConnectionConfig(ctx context.Context, user, pass string) (string, string
 	return user, pass, useIAM, nil
 }
 
-func initCloudSQLMySQLConnectionPool(ctx context.Context, tracer trace.Tracer, name, project, region, instance, ipType, user, pass, dbname string) (*sql.DB, error) {
+func buildDSN(user, pass, driverName, project, region, instance, dbname, userAgent string, useIAM, readOnly bool) string {
+	connAttrs := fmt.Sprintf("program_name:%s", url.QueryEscape(userAgent))
+	if readOnly {
+		connAttrs += ",read_only_connection:true"
+	}
+
+	if useIAM {
+		return fmt.Sprintf("%s@%s(%s:%s:%s)/%s?connectionAttributes=%s",
+			user,
+			driverName,
+			project,
+			region,
+			instance,
+			dbname,
+			connAttrs,
+		)
+	}
+	return fmt.Sprintf("%s:%s@%s(%s:%s:%s)/%s?connectionAttributes=%s",
+		user,
+		pass,
+		driverName,
+		project,
+		region,
+		instance,
+		dbname,
+		connAttrs,
+	)
+}
+
+func initCloudSQLMySQLConnectionPool(ctx context.Context, tracer trace.Tracer, name, project, region, instance, ipType, user, pass, dbname string, readOnly bool) (*sql.DB, error) {
 	//nolint:all // Reassigned ctx
 	ctx, span := sources.InitConnectionSpan(ctx, tracer, SourceType, name)
 	defer span.End()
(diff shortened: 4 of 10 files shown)


internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
```diff
@@ -22,11 +22,13 @@ database: ${CLOUD_SQL_MYSQL_DATABASE}
 user: ${CLOUD_SQL_MYSQL_USER:}
 password: ${CLOUD_SQL_MYSQL_PASSWORD:}
 ipType: ${CLOUD_SQL_MYSQL_IP_TYPE:PUBLIC}
+readOnly: ${CLOUD_SQL_MYSQL_READONLY:false}
```
shuzhou-gc, line 25: How do we plan to catch the potential conflict between this configuration option and the actions from individual tools?

averikitsch: Hi Shu, the full design doc can be found [go/toolbox-read-only](http://goto.google.com/toolbox-read-only). This works in 2 ways 1) locks the source to readonly transactions so if tools try to write it will be blocked at the engine level and 2) write tools (using the annotations) will be removed from the tool list so the agent doesn't try to use them.

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -209,7 +210,38 @@ func getConnectionConfig(ctx context.Context, user, pass string) (string, string
...
+			project,
+			region,
+			instance,
+			dbname,
+			url.QueryEscape(userAgent),
+		)
+	}
+
+	if readOnly {
+		dsn += "&session_variables=cloudsql_session_read_only%3Dlocked"
```
shuzhou-gc: I guess this will eventually result in a read_only ExecuteSql session, right?

What is the current underlying implementation?

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
shuzhou-gc: Thanks Averi, this is very helpful. There are a few details that I'm still not clear:

1. What do you mean by "readonly transactions" here? Is it Cloud SQL implementation of read_only `ExecuteSQL` interface, or native `readonly transaction`?
1. This PR only updates Cloud SQL for MySQL port, why not covering vanilla MySQL as well?
1. W.R.T the 2nd point in your earlier comment, I don't see the implementation in this PR. I guess it is planned in some follow-up work?

anubhav756: Thanks for the questions @shuzhou-gc!

> 1. What do you mean by "readonly transactions" here? Is it Cloud SQL implementation of read_only ExecuteSQL interface, or native readonly transaction?

We are not doing this through software checks in the tool code, and we are not just wrapping queries in a normal read-only transaction (which can be [bypassed by SQL injection](https://docs.google.com/document/d/10fYXXryqdumZ6x05tF6gEmUiVtGln5NYRGlWWqhAT4I/edit?pli=1&resourcekey=0-EkVFpz7tXMCWYSJgMnVV9g&tab=t.iyygyk9dnr7k#heading=h.ojcn9022i3fm) like `COMMIT; INSERT ...`).
   
Instead, when connecting to Cloud SQL MySQL, we tell the database server directly `cloudsql_session_read_only = locked`. This tells Cloud SQL to physically lock the connection itself. If any query attempts to insert, update, or drop data, the database server immediately rejects it with an error.

> 2. This PR only updates Cloud SQL for MySQL port, why not covering vanilla MySQL as well?

We are aiming for a 100% guarantee that writes are impossible at the database level (see https://docs.google.com/document/d/10fYXXryqdumZ6x05tF6gEmUiVtGln5NYRGlWWqhAT4I/edit?resourcekey=0-EkVFpz7tXMCWYSJgMnVV9g&tab=t.0#heading=h.28r91vesizt2). Cloud SQL MySQL supports this native `locked` setting, whereas regular self-hosted MySQL does not have this built-in lock and would rely on softer checks that can be bypassed.
   
To avoid giving users a false sense of security, our current MVP only covers database engines that have native, unbreakable locks (Cloud SQL MySQL, Cloud SQL Postgres, AlloyDB, and Spanner). We plan to work with other database engines (like vanilla MySQL) [Post-MVP](https://docs.google.com/document/d/1Bhm9D4zoQpexaBO9kAyHaf-NwwXuDN6P326qrTH3Y-w/edit?resourcekey=0-YSVPQ_x6TaArPtsdJtHx0w&tab=t.0#heading=h.oxnz8ns2iilt).

> 3. W.R.T the 2nd point in your earlier comment, I don't see the implementation in this PR. I guess it is planned in some follow-up work?

The code that implements hiding write tools from the LLM lives in our shared core PR (https://github.com/googleapis/mcp-toolbox/pull/3615 and https://github.com/googleapis/mcp-toolbox/pull/3816).

In this Cloud SQL MySQL PR, all we had to do was implement `IsReadOnly() bool { return s.ReadOnly }`. Once this returns `true`, the shred core code automatically hides the write tools from the LLM without needing any extra code in this PR.


Let me know if this helps clear things up! 🙂

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
anubhav756: Absolutely! Every query run against this source ends up in a locked read-only session.

Here is how it works:
1. At connection time, when Toolbox opens the database connection pool, it adds `&session_variables=cloudsql_session_read_only%3Dlocked` to the connection string.
2. At the database level, Cloud SQL MySQL applies this setting to every connection in the pool. The `locked` keyword ensures that no query can change or unlock the session later.
3. During query execution, tools, like `mysql-execute-sql` or custom SQL tools, simply run queries normally against this pool.
   * `SELECT` queries work as usual and return results.
   * Any modifying query (`INSERT`, `UPDATE`, `DELETE`, `CREATE`, `DROP`) is blocked and returned as an error by the Cloud SQL database server itself.

So instead of having a special tool-side mode or software check, the database connection itself does all the heavy lifting to ensure 100% read-only safety.

anubhav756:
Hi @shuzhou-gc

Checking in to see if the explanations above addressed your questions, or if you have any further thoughts on the approach. If everything looks good, could you please give this an LGTM?

Thanks! 🙂

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
shuzhou-gc: Thanks for the info. What would be the behavior if a customer configures "read only" in for a community mysql?

shuzhou-gc (approved):

Yuan325 (approved):

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
anubhav756: If a customer adds `readOnly: true` to a community MySQL source, Toolbox will fail at startup (`unknown field "readOnly"`). This is the same behavior as community Postgres as well.

This is because, as documented [here](https://docs.google.com/document/d/10fYXXryqdumZ6x05tF6gEmUiVtGln5NYRGlWWqhAT4I/edit?resourcekey=0-EkVFpz7tXMCWYSJgMnVV9g&tab=t.0#heading=h.t4nftr4ej5nu), Cloud SQL (MySQL & PostgreSQL) and AlloyDB provide kernel-level locks that prevent LLM prompt injections or transaction breakout attacks. However, community MySQL & PostgreSQL lack session lockout APIs (client queries can easily override session variables via `SET SESSION TRANSACTION READ WRITE`).

We intentionally avoid best-effort / non-guaranteed toggles that give users a false sense of security.

For community engines, the only true 100% security boundary is configuring database-level permissions (`SELECT`-only user credentials). We will add guidance for it post this MVP as documented [here](https://docs.google.com/document/d/10fYXXryqdumZ6x05tF6gEmUiVtGln5NYRGlWWqhAT4I/edit?resourcekey=0-EkVFpz7tXMCWYSJgMnVV9g&tab=t.iyygyk9dnr7k#heading=h.mlq6fas8cqs).

shuzhou-gc: I see, thanks.

docs/en/integrations/cloud-sql-mysql/source.md
```diff
@@ -140,3 +140,4 @@ instead of hardcoding your secrets into the configuration file.
 | password  |  string  |    false     | Password of the MySQL user (e.g. "my-password"). Defaults to attempting IAM authentication if unspecified.              |
 | ipType    |  string  |    false     | IP Type of the Cloud SQL instance, must be either `public`,  `private`, or `psc`. Default: `public`.                    |
 | sqlCommenter | boolean |  false     | Overrides the global `--sql-commenter` flag for this source. When set, it takes priority; when omitted, the global flag applies. |
+| readOnly  | boolean  |    false     | When set to `true`, enforces read-only execution at the database session level (`cloudsql_session_read_only=locked`) and suppresses write-capable tools. Default: `false`. |
```
rajavi-vivek: Isn't this specific to Postgres?

docs/en/integrations/cloud-sql-mysql/prebuilt-configs/cloud-sql-for-mysql.md
```diff
@@ -14,8 +14,8 @@ description: "Details of the Cloud SQL for MySQL prebuilt configuration."
     *   `CLOUD_SQL_MYSQL_DATABASE`: The name of the database to connect to.
     *   `CLOUD_SQL_MYSQL_USER`: The database username.
     *   `CLOUD_SQL_MYSQL_PASSWORD`: The password for the database user.
-    *   `CLOUD_SQL_MYSQL_IP_TYPE`: The IP type i.e. "Public
-     or "Private" (Default: Public).
+    *   `CLOUD_SQL_MYSQL_IP_TYPE`: The IP type i.e. "Public" or "Private" (Default: Public).
+    *   `CLOUD_SQL_MYSQL_READONLY`: (Optional) When set to `true`, enforces read-only execution at the database session level (`cloudsql_session_read_only=locked`) and suppresses write-capable tools. Default: `false`.
```
rajavi-vivek: Here again cloudsql_session_read_only=locked ins specific to Postgres.

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
rajavi-vivek: Cloud SQL MySQL only provides session-level  "readonly-ness". I presume this is sufficient?

rajavi-vivek (commented):
Please see comments about clousql_session_readonly

internal/prebuiltconfigs/tools/cloud-sql-mysql.yaml
anubhav756: Yes, because each source has its own connection pool. For read-only sources, all of the connections in that pool are created with `cloudsql_session_read_only=locked`, and the agent cannot spawn new connections or switch pools, session-level locking is sufficient.

docs/en/integrations/cloud-sql-mysql/source.md
anubhav756: Could you elaborate a bit? 🙂 

Should Cloud SQL MySQL be using DSN  as `?connectionAttributes=program_name:...,cloudsql_read_only_connection:true` instead of `?options='-c cloudsql_session_read_only=locked'`?

docs/en/integrations/cloud-sql-mysql/prebuilt-configs/cloud-sql-for-mysql.md
anubhav756: Interesting. Let's continue discussion in the [comment thread above](https://github.com/googleapis/mcp-toolbox/pull/3618#discussion_r3863109761)?

docs/en/integrations/cloud-sql-mysql/source.md
rajavi-vivek: "connectionAttributes=program_name:...,read_only_connection:true" is the string to use for MySQL

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
```diff
@@ -209,7 +210,38 @@ func getConnectionConfig(ctx context.Context, user, pass string) (string, string
 	return user, pass, useIAM, nil
 }
 
-func initCloudSQLMySQLConnectionPool(ctx context.Context, tracer trace.Tracer, name, project, region, instance, ipType, user, pass, dbname string) (*sql.DB, error) {
+func buildDSN(user, pass, driverName, project, region, instance, dbname, userAgent string, useIAM, readOnly bool) string {
+	var dsn string
```
rajavi-vivek: There is no need to set the session variable for MySQL. Instead you can do the following.

```
var connAttrs string

connAttrs = fmt.Sprintf("program_name:%s", url.QueryEscape(userAgent))
if (readOnly) {
  connAttrs  += ",read_only_connection=true"
}

if useIAM {
		dsn = fmt.Sprintf("%s@%s(%s:%s:%s)/%s?connectionAttributes=%s",  ...., connAttrs)
} else {
 		dsn = fmt.Sprintf("%s@%s(%s:%s:%s)/%s?connectionAttributes=%s",  ...., connAttrs)      
}
```

rajavi-vivek (commented):
Please see the suggested changes.

internal/sources/cloudsqlmysql/cloud_sql_mysql.go
anubhav756: Done. Thanks for the suggestion! 🙂

docs/en/integrations/cloud-sql-mysql/source.md
anubhav756: Got it. Thanks for catching that. Fixed!

rajavi-vivek (approved):
Thanks for pulling in the recommended changes.

PR #3619, opened 2026-07-14
feat(cloudsqlpg): add native read-only session locking

anubhav756:
## Description

This PR implements native, protocol-level Read-Only session enforcement for Cloud SQL for PostgreSQL.

## Changes
- Appends `options='-c cloudsql_session_read_only=locked'` to the connection DSN when `readOnly: true` is configured on the source.
- Added `ReadOnly bool` to `cloudsqlpg.Config` and implemented `IsReadOnly() bool` on `cloudsqlpg.Source`.
- Updated prebuilt configuration templates (`cloud-sql-postgres.yaml` and `cloud-sql-postgres-admin.yaml`) to dynamically resolve read-only mode via the `CLOUDSQL_PG_READONLY` environment variable (`readonly: ${CLOUDSQL_PG_READONLY:false}`).
- Enforces non-bypassable read-only mode at the PostgreSQL kernel level, preventing physical Transaction ID (XID) allocation for DML/DDL mutations, transaction breakout attacks (e.g. `COMMIT; DROP TABLE;`), and side-effecting stored procedures.
- Added unit tests in `cloud_sql_pg_test.go`, integration vulnerability tests and updated integration READMEs.

### DIFF
--- a/docs/CLOUDSQLPG_README.md
+++ b/docs/CLOUDSQLPG_README.md
@@ -94,6 +94,7 @@ export CLOUD_SQL_POSTGRES_DATABASE="<your-database-name>"
 export CLOUD_SQL_POSTGRES_USER="<your-database-user>"  # Optional
 export CLOUD_SQL_POSTGRES_PASSWORD="<your-database-password>"  # Optional
 export CLOUD_SQL_POSTGRES_IP_TYPE="PUBLIC"  # Optional: `PUBLIC`, `PRIVATE`, `PSC`. Defaults to `PUBLIC`.
+export CLOUD_SQL_POSTGRES_READONLY="true"  # Optional: Restricts tools and enforces read-only session locking on the database connection. Defaults to `false`.
 ```
 
 Add the following configuration to your MCP client (e.g., `settings.json` for Gemini CLI, `mcp_config.json` for Antigravity):
--- a/internal/prebuiltconfigs/tools/cloud-sql-postgres-admin.yaml
+++ b/internal/prebuiltconfigs/tools/cloud-sql-postgres-admin.yaml
@@ -16,6 +16,7 @@ kind: source
 name: cloud-sql-admin-source
 type: cloud-sql-admin
 defaultProject: ${CLOUD_SQL_POSTGRES_PROJECT:}
+readOnly: ${CLOUD_SQL_POSTGRES_READONLY:false}
 ---
 kind: tool
 name: create_instance
--- a/internal/sources/cloudsqlpg/cloud_sql_pg.go
+++ b/internal/sources/cloudsqlpg/cloud_sql_pg.go
@@ -58,6 +58,7 @@ type Config struct {
 	Database     string         `yaml:"database" validate:"required"`
 	User         string         `yaml:"user"`
 	Password     string         `yaml:"password"`
+	ReadOnly     bool           `yaml:"readOnly"`
 	SQLCommenter *bool          `yaml:"sqlCommenter"`
 }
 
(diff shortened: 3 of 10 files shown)


docs/CLOUDSQLPG_README.md
```diff
@@ -94,6 +95,7 @@ export CLOUD_SQL_POSTGRES_DATABASE="<your-database-name>"
 export CLOUD_SQL_POSTGRES_USER="<your-database-user>"  # Optional
 export CLOUD_SQL_POSTGRES_PASSWORD="<your-database-password>"  # Optional
 export CLOUD_SQL_POSTGRES_IP_TYPE="PUBLIC"  # Optional: `PUBLIC`, `PRIVATE`, `PSC`. Defaults to `PUBLIC`.
+export CLOUDSQL_PG_READONLY="true" # Optional: Restricts tools and executes queries on read-only endpoints.
```
divyang-dalal: to be consistent with other variables we should use CLOUD_SQL_POSTGRES_READONLY

internal/prebuiltconfigs/tools/cloud-sql-postgres-admin.yaml
averikitsch, line 1: Remove changes to the Cloud Run Admin source

averikitsch (changes requested):

docs/CLOUDSQLPG_README.md
anubhav756: Made consistent. Thanks for the suggestion!

internal/prebuiltconfigs/tools/cloud-sql-postgres-admin.yaml
anubhav756: Context: The main reason I added this was so users can run the admin server safely in read-only mode.

It defaults to `false`, so nothing changes for regular users unless they turn it on with the environment variable.

I included it here because all Cloud SQL PG tools (both DB engine and admin) are in this PR.

Could you share your thoughts on why we should remove it from the admin source? I'm happy to take it out if you feel it doesn't belong here, just wanted to understand your perspective! :)

averikitsch: Have we ensured that the tools are properly annotated and can we update the documentation for the admin source?

anubhav756: Yes, these tools are properly annotated. I think I had updated the documentation for the readonly field and explain the suppression behavior. Let me know if you have any questions :)

averikitsch (approved):

averikitsch:
nit: please update the admin source docs

anubhav756:
> nit: please update the admin source docs

Updated docs. Thanks for bringing that up!

internal/sources/cloudsqlpg/cloud_sql_pg.go
```diff
@@ -169,20 +168,26 @@ func getConnectionConfig(ctx context.Context, user, pass, dbname string) (string
...
 		}
 		user = email
+		dsn = fmt.Sprintf("user=%s dbname=%s sslmode=disable application_name=%s", user, dbname, userAgent)
+	} else {
+		// Construct IAM connection string with username
+		dsn = fmt.Sprintf("user=%s dbname=%s sslmode=disable application_name=%s", user, dbname, userAgent)
+	}
+
+	if readOnly {
+		dsn += " options='-c cloudsql.session_read_only=locked'"
```
postsql: This has to be

	dsn += " options='-c cloudsql_session_read_only=locked'"

notice '.' --> '_'

anubhav756: Fixed! Thanks for the info.

postsql (approved):

PR #3620, opened 2026-07-14
feat(source/postgres): add optional connectTimeout

### LINKED ISSUE #2227: Add timeout for postgres source
### Prerequisites

Adding timeout for generic postgresql [source](https://github.com/googleapis/genai-toolbox/blob/main/internal/sources/postgres/postgres.go).

### Suggested Solution(s)

We can add it via db conn pool.ConnConfig.ConnectTimeout = defaultConnectTimeout (ref: https://medium.com/@neelkanthsingh.jr/understanding-database-connection-pools-and-the-pgx-library-in-go-3087f3c5a0c)

### Alternatives Considered

_No response_

### Additional Details

_No response_

vismaytiwari:
## Description

Adds an opt-in `connectTimeout` field to the postgres source (requested in #2227). Today the source sets no connect timeout, so a server that accepts the TCP connection but never completes the handshake can block indefinitely.

When `connectTimeout` is set (a Go duration string, e.g. `"5s"`), it is applied to pgx's `ConnConfig.ConnectTimeout`. When omitted, nothing is applied and connection behavior is unchanged — per the discussion on #2227 this is intentionally opt-in with **no default**, so existing users see no change.

## PR Checklist

- [x] Make sure to open an issue as a bug/issue before writing your code (#2227)
- [x] Code coverage does not decrease (added `TestApplyConnectTimeout` + a YAML-parse case)
- [x] Appropriate docs were updated (postgres source reference)
- [ ] Make sure to add `!` if this involves a breaking change — n/a, opt-in and non-breaking

**Verification**

```
go test ./internal/sources/postgres/          # ok (incl. new TestApplyConnectTimeout: set → 5s, unset → 0/no timeout, invalid → error)
golangci-lint run ./internal/sources/postgres/...   # 0 issues
```

🛠️ Fixes #2227


internal/sources/postgres/postgres.go
```diff
@@ -131,7 +136,22 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any) (an
...
+func applyConnectTimeout(config *pgxpool.Config, connectTimeout string) error {
+	if connectTimeout == "" {
+		return nil
+	}
+	duration, err := time.ParseDuration(connectTimeout)
+	if err != nil {
+		return fmt.Errorf("unable to parse connectTimeout %q as time.Duration: %w", connectTimeout, err)
+	}
+	config.ConnConfig.ConnectTimeout = duration
+	return nil
```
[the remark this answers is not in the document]
vismaytiwari: Good catch — added the `duration <= 0` guard (rejects `0s` and negatives with a clear error), plus test cases for both. Thanks!

internal/sources/postgres/postgres.go
```diff
@@ -60,14 +61,18 @@ type Config struct {
 	QueryParams   map[string]string `yaml:"queryParams"`
 	QueryExecMode string            `yaml:"queryExecMode" validate:"omitempty,oneof=cache_statement cache_describe describe_exec exec simple_protocol"`
 	SQLCommenter  *bool             `yaml:"sqlCommenter"`
+	// ConnectTimeout optionally bounds how long a single connection attempt may
+	// take, parsed as a Go duration (e.g. "5s"). When unset, no timeout is
+	// applied and connection behavior is unchanged.
+	ConnectTimeout string `yaml:"connectTimeout"`
```
Yuan325: can we make this as a integer type instead of string? With that, we can also use the yaml validator. Something like this:

```
ConnectTimeout *int `yaml:"connectTimeout" validate:"omitEmpty,gte=1"`
```

internal/sources/postgres/postgres.go
```diff
@@ -131,7 +136,25 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any) (an
...
+	if err != nil {
+		return fmt.Errorf("unable to parse connectTimeout %q as time.Duration: %w", connectTimeout, err)
+	}
+	if duration <= 0 {
+		return fmt.Errorf("connectTimeout %q must be a positive duration", connectTimeout)
+	}
+	config.ConnConfig.ConnectTimeout = duration
+	return nil
+}
+
```
Yuan325: We might not need this extra function here since we are handling the validation during the unmarshaling process.

internal/sources/postgres/postgres.go
```diff
@@ -158,6 +181,10 @@ func initPostgresConnectionPool(ctx context.Context, tracer trace.Tracer, name,
 	}
 	config.ConnConfig.DefaultQueryExecMode = execMode
 
+	if err := applyConnectTimeout(config, connectTimeout); err != nil {
```
Yuan325: we can check if connectTimeout is not nil, we can set the duration here.

internal/sources/postgres/postgres_internal_test.go
Yuan325, line 1: Wouldn't need this test file :)

internal/sources/postgres/postgres_test.go
Yuan325, line 157: Let's add a test case here to make sure the validator catch on <1 value for `connectTimeout`.

Yuan325 (changes requested):
Hi @vismaytiwari Thank you for your contribution! Please let me know if you have any questions on the feedback :) Will review again once updated

vismaytiwari:
Thanks @Yuan325, all done in fd5dc2e:

- `connectTimeout` is now `*int` with `validate:"omitempty,gte=1"`, so it's validated at unmarshal time and rejects `< 1`.
- Dropped `applyConnectTimeout` and set it inline in `initPostgresConnectionPool` when it's non-nil.
- Removed the internal test file, and added a `connectTimeout: 0` case to `TestFailParseFromYaml` to cover the validator.
- Also updated the `source.md` param table since the field is seconds now, not a duration string.

Ready for another look.

Yuan325 (approved):
Awesome @vismaytiwari , thank you again for your contribution!! I'll go ahead and merge this. It will be included in our release tomorrow :)

PR #3622, opened 2026-07-14
fix(source/cockroachdb): detect existing LIMIT under any whitespace

### LINKED ISSUE #3553: CockroachDB ApplyQueryLimits appends a second LIMIT for multiline/tab-formatted queries
### Prerequisites

### Toolbox version

main (built from source; the same code is in v1.5.0)

### Environment

The bug is in source-code logic and is independent of OS/runtime; it is reproducible from a unit test of the CockroachDB source.

### Expected Behavior

`Source.ApplyQueryLimits` in the CockroachDB source should not append a `LIMIT` clause when the query already has one, regardless of how the `LIMIT` keyword is surrounded by whitespace (a single space, a newline, or a tab).

### Current Behavior

The existing-`LIMIT` check is a space-anchored, upper-cased substring match:

```go
normalized := strings.ToUpper(sql)
if !strings.Contains(normalized, " LIMIT ") {
    // ... append "LIMIT <MaxRowLimit>"
}
```

Because it looks for the literal `<space>LIMIT<space>`, a query whose `LIMIT` is separated by other whitespace (for example a formatted multiline query) is treated as having no limit, so a second `LIMIT` is appended, producing invalid SQL:

```
SELECT *
FROM users
LIMIT 50

-> SELECT *\nFROM users\nLIMIT 50 LIMIT 100   (syntax error)
```

This affects newline- and tab-formatted queries (and the trailing-append branch already trims arbitrary whitespace, so the mismatch is purely in this detection).

Source: `internal/sources/cockroachdb/cockroachdb.go` (the `ApplyQueryLimits` SELECT branch).

### Steps to reproduce?

With `MaxRowLimit > 0`, call `ApplyQueryLimits` (or run a SELECT through the CockroachDB source) on a query that already contains a `LIMIT` separated by a newline or tab:

1. Input: `"SELECT *\nFROM users\nLIMIT 50"`
2. Observed output: `"SELECT *\nFROM users\nLIMIT 50 LIMIT 100"` (a second `LIMIT` is appended)
3. Expected output: the query is returned unchanged (it already has a `LIMIT`)

The same happens for a lowercase multiline `limit` and for a tab-separated `\tLIMIT`.

### Additional Details

Happy to send a small PR: replace the space-anchored substring check with a precompiled case-insensitive word-boundary regexp (`(?i)\bLIMIT\b`), matching the existing `regexp.MustCompile` style already used in this file, and add multiline/lowercase/tab cases to `TestApplyQueryLimits`.

he-yufeng:
## Description

`Source.ApplyQueryLimits` for the CockroachDB source detected an existing `LIMIT` clause with `strings.Contains(strings.ToUpper(sql), " LIMIT ")`, which only matches a space-delimited keyword. A query whose `LIMIT` is preceded or followed by a newline or tab — for example a formatted multiline query — was treated as having no limit, so a second `LIMIT` was appended, producing invalid SQL like `... ORDER BY id\nLIMIT 50 LIMIT 1000`.

This matches `LIMIT` as a standalone keyword with a case-insensitive `\bLIMIT\b` regexp (compiled once at package scope), so any surrounding whitespace is recognized. A column named `limit_col` is still correctly treated as not being a `LIMIT` clause. The existing `TestApplyQueryLimits` is extended with newline-, tab-, multiline-, and newline-before-value cases, which fail against the old check and pass with this fix.

## PR Checklist

## Issue Reference

Fixes #3553 🦕


duwenxin99:
Thanks for opening the PR! This is a duplicate of https://github.com/googleapis/mcp-toolbox/pull/3554#pullrequestreview-4695795340 so closing it for now.

PR #3623, opened 2026-07-14
feat: Implement Text type MCP Resource

dishaprakash:
### Overview
This PR implements the concrete `text` resource type. The text resource allows admins to define static, read-only textual content directly within their configuration files without relying on external file systems

### Key Changes
* **`text` Resource Implementation (`internal/resources/text`)**:
  * Implemented the `Config` mapping which inherits `BaseConfig` and enforces a mandatory `Text` string payload.
  * Ensures `MimeType` gracefully defaults to `text/plain` if omitted.
* **BaseConfig Changes**:
  * Handles missing `Annotations` and safely defaults the resource `priority` to `1.0`
* **Dynamic Sizing**: 
  * Evaluates and sets the `BaseConfig.Size` dynamically at parse-time by calculating the byte-length of the text payload (`len([]byte(c.Text))`)

### Extensive Testing
* Added highly comprehensive test coverage in `text_test.go` verifying:
  * **Happy Paths & Missing Fields:** Validating fallback behaviors and ensuring empty payloads cleanly trigger config errors.
  * **Multi-byte Unicode Sizing:** Verifying that strings containing emojis (e.g., `Hello 🌍`) correctly calculate `Size` using the total byte-length rather than character count, guaranteeing protocol transport accuracy.
  * **Whitespace Safety:** Verifying that strings containing pure whitespace/newlines (`"   \n  "`) are appropriately accepted as valid payloads.
  * **Strict Decoder Validation:** Testing raw YAML deserialization and ensuring the `yaml.Strict()` parser accurately flags and rejects typo'd config fields .


internal/resources/text/text_test.go
```diff
@@ -0,0 +1,226 @@
...
+			res, err := tc.config.Initialize(ctx)
+			if tc.wantError {
+				if err == nil {
+					t.Fatalf("Initialize() expected error, got nil")
+				}
+				if tc.errContains != "" {
+					expectedErr := "missing required 'text' field for text resource \"test3\""
+					if err.Error() != expectedErr {
+						t.Errorf("Initialize() err = %v, want %v", err, expectedErr)
+					}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/resources_test.go
```diff
@@ -213,7 +213,7 @@ annotations:
 	if cfg.MimeType != "text/plain" {
 		t.Errorf("Expected MimeType 'text/plain', got %q", cfg.MimeType)
 	}
-	if val, ok := cfg.Annotations["key1"].(string); !ok || val != "value1" {
+	if cfg.Annotations == nil || cfg.Annotations.Priority == nil || *cfg.Annotations.Priority != 0.5 {
 		t.Errorf("Expected annotation key1=value1, got %v", cfg.Annotations["key1"])
 	}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/text/text_test.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+
+			if cfg.Annotations.Priority == nil || *cfg.Annotations.Priority != tc.wantPrior {
+				t.Errorf("Annotations.Priority = %v, want %v", cfg.Annotations.Priority, tc.wantPrior)
+			}
+
+			if cfg.Size == nil {
+				t.Fatalf("Size is nil, expected dynamic calculation")
+			}
+			expectedSize := int64(len([]byte(tc.config.Text)))
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+	if c.Annotations == nil {
+		c.Annotations = &resources.ResourceAnnotations{}
+	}
+	if c.Annotations.Priority == nil {
+		p := 1.0
+		c.Annotations.Priority = &p
+	}
+
+	// Invert at boot by calculating the length of the string array as per design doc
+	size := int64(len([]byte(c.Text)))
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (c *Config) Initialize(ctx context.Context) (resources.Resource, error) {
+	if c.Text == "" {
+		return nil, fmt.Errorf("missing required 'text' field for text resource %q", c.Name)
+	}
+
+	// Default MimeType if unset
+	if c.MimeType == "" {
+		c.MimeType = "text/plain"
+	}
```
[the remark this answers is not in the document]
dishaprakash: Doing this already in config.go

internal/resources/text/text_test.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+			expectedSize := int64(len(tc.config.Text))
+			if *cfg.Size != expectedSize {
+				t.Errorf("Size = %d, want %d", *cfg.Size, expectedSize)
+			}
+		})
+	}
+}
+
+func TestTextResourceYAMLUnmarshaling(t *testing.T) {
```
Yuan325, line 152: Can we keep this test similar to our other tests? where we have 2 function one for unmarshaling and another for fail test cases and in table-struct formatting? Tyyyy~

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+	cfg := &Config{BaseConfig: resources.BaseConfig{Name: name, Type: resourceType}}
+	if err := decoder.DecodeContext(ctx, cfg); err != nil {
+		return nil, err
+	}
+	return cfg, nil
+}
+
+// Config represents the uninitialized textual resource configuration from YAML.
+type Config struct {
+	resources.BaseConfig `yaml:",inline"`
```
Yuan325, line 50: nit: do you mind submitting another PR, changing `BaseConfig` to `ConfigBase` to be consistent with tools' naming? (doesn't have to be this PR)

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+	Text                 string `yaml:"text"`
+}
+
+var _ resources.ResourceConfig = (*Config)(nil)
```
Yuan325:
```suggestion
var _ resources.ResourceConfig = &Config{}
```

Can we just do this?

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (c *Config) ResourceConfigType() string {
+	return resourceType
+}
+
+func (c *Config) Initialize(ctx context.Context) (resources.Resource, error) {
+	if c.Text == "" {
```
Yuan325: Is this required? If yes, we can add a validate:required to check during unmarshaling

Please add a test case for this after adding the validator~ to make sure that the validator will flag it if it was empty. (I'm not sure if it checks for empty string, or just the definition of field (e.g. will validator flag `text: ` as pass or fail?)

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+const resourceType = "text"
+
+func init() {
+	if !resources.Register(resourceType, newConfig) {
+		panic(fmt.Sprintf("resource type %q already registered", resourceType))
+	}
+}
+
+func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (resources.ResourceConfig, error) {
+	cfg := &Config{BaseConfig: resources.BaseConfig{Name: name, Type: resourceType}}
```
Yuan325: we can set default values here (e.g. https://github.com/googleapis/mcp-toolbox/blob/f3e7ca9a8f49ce79f5b6fbef23c45a643c4e9d44/internal/sources/alloydbpg/alloydb_pg.go#L45)

internal/resources/text/text.go
Yuan325: We can set default value in the `newConfig()`

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	// Default annotations.priority to 1.0 (critical) if unset
+	if c.Annotations == nil {
+		c.Annotations = &resources.ResourceAnnotations{}
+	}
+	if c.Annotations.Priority == nil {
+		p := 1.0
+		c.Annotations.Priority = &p
+	}
```
Yuan325: Similarly we can set the default at `newConfig()`.

internal/resources/text/text.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return &Resource{config: *c}, nil
+}
+
+// Resource represents the initialized textual resource that returns plain text payloads.
+type Resource struct {
+	config Config
+}
+
+var _ resources.Resource = (*Resource)(nil)
```
Yuan325:
```suggestion
var _ resources.Resource = &Resource{}
```

Yuan325 (changes requested):
Thank you!! 

qq, I don't see the `mimeType` and `size` field in the design for `type: text`. If that field is not required for all types, should we remove it from the ConfigBase?

dishaprakash:
> I don't see the `mimeType` and `size` field in the design for `type: text`. If that field is not required for all types, should we remove it from the ConfigBase?

These are not required fields but the user can add them regardless of the resource type. If not set, it has defaults which we are setting in the `newConfig` and `Initialize` now

internal/resources/text/text_test.go
dishaprakash: Done

internal/resources/text/text.go
dishaprakash: Sure will do as a part of the next PRs

internal/resources/text/text.go
dishaprakash: Done

internal/resources/text/text.go
dishaprakash: Done

internal/resources/text/text.go
dishaprakash: Moved this to resources.go for common defaults

internal/resources/text/text.go
dishaprakash: Done

internal/resources/text/text.go
dishaprakash: Added mimetype default this way

internal/resources/text/text.go
dishaprakash: Added validate check and extra test cases

internal/resources/text/text.go
```diff
@@ -0,0 +1,86 @@
...
+		panic(fmt.Sprintf("resource type %q already registered", resourceType))
+	}
+}
+
+func newConfig(ctx context.Context, name string, decoder *yaml.Decoder) (resources.ResourceConfig, error) {
+	cfg := &Config{
+		BaseConfig: resources.BaseConfig{
+			Name:     name,
+			Type:     resourceType,
+			MimeType: "text/plain",
```
Yuan325, line 38: could also add the default uri here (from the previous PR)

internal/resources/text/text.go
```diff
@@ -0,0 +1,86 @@
...
+	Text                 string `yaml:"text" validate:"required"`
+}
+
+var _ resources.ResourceConfig = &Config{}
+
+func (c *Config) ResourceConfigType() string {
+	return resourceType
+}
+
+func (c *Config) Validate() error {
```
Yuan325: we can check the URI compatibility here~ Iirc, this will also replace the BaseConfig's `Validate()` function? Was that intended? pls correct me if I'm wrong~

internal/resources/text/text.go
Yuan325: Qq, did you verify that the yaml validator do not check for empty string? If the validator is checking for empty string, we can just remove it from here.

internal/resources/text/text.go
```diff
@@ -0,0 +1,86 @@
...
+func (c *Config) Initialize(ctx context.Context) (resources.Resource, error) {
+	size := int64(len(c.Text))
+	c.Size = &size
+
+	return &Resource{config: *c}, nil
+}
+
+// Resource represents the initialized textual resource that returns plain text payloads.
+type Resource struct {
+	config Config
```
Yuan325:
```suggestion
	Config
```

internal/resources/text/text.go
```diff
@@ -0,0 +1,86 @@
...
+func (c *Config) Validate() error {
+	if c.Text == "" {
+		return fmt.Errorf("Field validation for 'Text' failed: missing required field")
+	}
+	return c.BaseConfig.Validate()
+}
+
+func (c *Config) Initialize(ctx context.Context) (resources.Resource, error) {
+	size := int64(len(c.Text))
+	c.Size = &size
```
Yuan325: Since this is not part of the config file, let's remove this from ConfigBase. We should add this to the `type Resource struct` instead. The `Config` should be solely items that are declared within the config file.

Yuan325 (commented):
Wondering if we should add integration test to resources as well? 

e.g. write a temp file, and then try invoking the resource.

internal/resources/text/text.go
dishaprakash: Yes added a test for it in server_test with a malformed config. I had left it as an extra check but I have removed it now.

internal/resources/text/text.go
dishaprakash: Ah since we removed the extra validation, we can rely on the base Validate() method here now
Done

internal/resources/text/text.go
dishaprakash: Done

internal/resources/text/text.go
dishaprakash: Done

internal/resources/text/text.go
dishaprakash: Done

Yuan325 (approved):

PR #3626, opened 2026-07-15
docs(source/snowflake): remove unsupported timeout field from reference

anxkhn:
The Snowflake source Reference table documents a `timeout` field
(`integer`, "The connection timeout in seconds. Defaults to 60."), but the
Snowflake source has no such option:

- The Snowflake `Config` struct in `internal/sources/snowflake/snowflake.go`
  declares only `name`, `type`, `account`, `user`, `password`, `database`,
  `schema`, `warehouse`, and `role`. There is no `timeout` field, and
  `initSnowflakeConnection` never applies a connection timeout.
- Source configs are parsed with a strict YAML decoder (`yaml.Strict()` in
  `internal/util/util.go`), so an unknown key is a hard error, not a silently
  ignored no-op.

The net effect is that a user who follows the docs and adds `timeout: 60` to a
Snowflake source config does not get a 60-second timeout: their configuration
fails to load with an `unknown field "timeout"` error. The documented option
actively breaks config loading.

This PR removes the fabricated `timeout` row from the reference table. The
`warehouse` and `role` rows are left untouched: those defaults ("COMPUTE_WH" /
"ACCOUNTADMIN") really are applied in `initSnowflakeConnection`, so the docs are
correct there.

It also adds a table-driven parse-failure case to `TestFailParseFromYaml` in
`internal/sources/snowflake/snowflake_test.go` asserting that a config with
`timeout: 60` is rejected. This ties the doc change to observable behavior and
guards against the field being re-documented without being implemented.

Scope note: adding an actual connection timeout would be a `feat`; this PR is
the surgical `docs` fix that makes the reference match current behavior.

### PR Checklist

- [x] Make sure you reviewed CONTRIBUTING.md


internal/sources/snowflake/snowflake_test.go
Yuan325, line 1: I'll remove this test cases~ since we have a test cases that checks for extra field.Removing it from docs should be sufficient. Thank youuuu

Yuan325 (approved):

PR #3630, opened 2026-07-15
fix(server): reject non-string group names

twishabansal:
## Description

Reinstates the `(!present || rawName == nil)` guard in `UnmarshalResourceConfig` that was accidentally dropped during the merge of the groups stack into `feat/groups`.

Without the guard, a config like:

```yaml
kind: group
name: 123
```

would silently default to the empty (default) group instead of returning a parse error. Deeven Seru flagged this in the review of #3575.

The fix ensures non-string name values (`name: 123`, `name: true`) are rejected with `"missing 'name' field or it is not a string"`, while legitimately absent or null names are still accepted as the default group.

Adds regression tests to `TestParseConfigGroupNameValidation` covering all four cases.

### DIFF
--- a/cmd/internal/config_test.go
+++ b/cmd/internal/config_test.go
@@ -2477,13 +2477,14 @@ description: the default group
 			wantErr: false,
 		},
 		{
-			description: "group with null name is accepted as default group",
+			description: "group with null name is rejected",
 			in: `
 kind: group
 name: ~
 description: the default group
 `,
-			wantErr: false,
+			wantErr:     true,
+			errContains: "missing 'name' field or it is not a string",
 		},
 		{
 			description: "non-group resource with integer name is rejected",
--- a/cmd/internal/config_test.go
+++ b/cmd/internal/config_test.go
@@ -2486,6 +2486,14 @@ description: the default group
 			wantErr: false,
 		},
 		{
+			description: "group with quoted numeric name is accepted",
+			in: `
+kind: group
+name: "123"
+`,
+			wantErr: false,
+		},
+		{
 			description: "non-group resource with integer name is rejected",
 			in: `
 kind: tool
--- a/cmd/internal/config_test.go
+++ b/cmd/internal/config_test.go
@@ -2438,3 +2438,83 @@ tools:
 		})
 	}
 }
+
+func TestParseConfigGroupNameValidation(t *testing.T) {
+	ctx, err := testutils.ContextWithNewLogger()
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	tcs := []struct {
+		description string
+		in          string
+		wantErr     bool
+		errContains string
+	}{
+		{
+			description: "group with integer name is rejected",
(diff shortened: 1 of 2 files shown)


Yuan325:
Good catch on name validation! :) 

Is there a reason to completely reject non-string names? For MCP specification, the tool name have these rules (https://modelcontextprotocol.io/specification/draft/server/tools#tool-names) : 

- Length: 1–128 characters
- Case-sensitive
- Allowed characters: A-Z, a-z, 0-9, _, -, .
- No spaces, commas, or special characters
- Unique within a server_

https://github.com/googleapis/mcp-toolbox/blob/1b416edb3bb9fe1c277f83eae9e10a4a5fa5ec7b/internal/server/config.go#L520

Seems like numbers and certain characters are allowed~ should we follow these? prompt names does not have any rules.

Deeven-Seru:
Thanks a lot  @Yuan325,

The key distinction here is between the **allowed characters** inside a name and the underlying **data type** parsed from the YAML.

While the MCP spec allows digits (`0-9`) as *characters*, the protocol uses JSON-RPC, meaning the name identifier must fundamentally be typed as a `string`. 

If a user writes `name: 123` in their config without quotes, the Go YAML parser strictly evaluates it as an `int`. This correctly causes the `.(string)` type assertion to fail because an integer is not a string.

If we were to accept raw integers and secretly convert them to strings under the hood, we'd be bypassing strict YAML typing, which is innaccurate (e.g., a boolean `name: true` turning into `"true"`). 

Users are still completely free to use numbers in their names! They just need to explicitly quote them (e.g., `name: "123"`). The YAML parser will then treat it as a `string` which passes the type assertion, and our `NameValidation` regex perfectly handles validating the characters against the spec.

Yuan325:
Hi @Deeven-Seru, in the config file, if a user defines name as an integer (like name: 123) and our server rejects it, that is completely acceptable and expected. Because we unmarshal into a generic map[string]any first, the parser evaluates 123 as an int, meaning the .(string) type assertion will fail and safely return an error stating that a string is required.

I also noticed that @twishabansal has already added test cases specifically covering these scenarios (like name: 123), which should address your concerns and ensure our validation behaves correctly.

LMK if you have any other thoughts! :)

Yuan325:
I also just realized that the NameValidation was removed previously during migration to flat-format config file, I'll submit a PR to add that back to Tool's unmarshaling~ Thanks for point that out @Deeven-Seru :)

Yuan325 (approved):

cmd/internal/config_test.go
```diff
@@ -2438,3 +2438,83 @@ tools:
 		})
 	}
 }
+
+func TestParseConfigGroupNameValidation(t *testing.T) {
```
Yuan325, line 2442: let's add test cases where `name: "123"` is parsed successfully :)

twishabansal: Added. Thanks for pointing out!

PR #3641, opened 2026-07-16
refactor(prompts): remove orphaned promptset layer

twishabansal:
## Description

Stacked on top of #3605.

The prompts coverage check in #3605 fails at **69.2%** (needs **80%**). Investigating why turned up the real cause: the `prompts.Promptset` layer is **dead code**.

Groups are the source of truth for prompt scoping. `prompts/list` and `prompts/get` resolve entirely from `group.Group` (`GenerateListPromptsResult` and `group.ContainsPrompt`). `PrimitiveManager.GetPromptset` — the only caller of `prompts.Promptset.Initialize` — has no production caller itself, so `Promptset` / `PromptsetConfig` / `PromptsetManifest` and `GetPromptset` are never used.

Rather than add tests for unreachable code to satisfy the gate, this PR removes the orphaned layer:
- `internal/prompts/promptsets.go` (+ its unit test)
- `PrimitiveManager.GetPromptset`
- the promptset scaffolding in `testutils.SetUpResources` and the two server tests

Dropping the uncovered promptset statements brings `./internal/prompts/...` coverage to **84.0%**, clearing the gate with no new tests.

**Groups still contain prompts** — this only removes the unused parallel view; `group.Group.PromptNames` / `ContainsPrompt` and the MCP prompt handlers are unchanged.


internal/server/primitives/primitives_test.go
```diff
@@ -44,8 +44,7 @@ func TestUpdateServer(t *testing.T) {
 	newTools := map[string]tools.Tool{"example-tool": nil}
 	newPrompts := map[string]prompts.Prompt{"example-prompt": testutils.NewMockPrompt("example-prompt", "", prompts.Arguments{})}
 	newGroups := map[string]group.Group{
-		"example-toolset":   group.NewGroup(group.GroupConfig{Name: "example-toolset", ToolNames: []string{"example-tool"}}),
-		"example-promptset": group.NewGroup(group.GroupConfig{Name: "example-promptset", PromptNames: []string{"example-prompt"}}),
```
Yuan325, line 48: Do we have a test case somewhere that checks for the group with both tools and prompts?

internal/testutils/testutils.go
```diff
@@ -197,30 +197,23 @@ func SetUpResources(t *testing.T, mockTools []MockTool, mockPrompts []MockPrompt
...
+	// Build the authoritative groups map. Toolsets contribute their tool names;
+	// all prompts belong to the default (nameless) group, matching the legacy
+	// default-toolset behavior.
 	groupNames := make(map[string]struct{})
 	for name := range toolsets {
 		groupNames[name] = struct{}{}
 	}
-	for name := range promptsets {
-		groupNames[name] = struct{}{}
+	if len(allPrompts) > 0 {
```
Yuan325, line 207: should we also update the toolsets above as well? Since groupsNames[""] should consist of both allTools and allPrompts.

internal/server/primitives/primitives_test.go
twishabansal: Present in the base PR: https://github.com/googleapis/mcp-toolbox/pull/3605/changes#diff-dc9092dbcf96f1fdf0898cebd810b7385193a362c601ca694dff3f419f4eb253R53

internal/testutils/testutils.go
twishabansal: toolset cleanup is done in follow up PR: https://github.com/googleapis/mcp-toolbox/pull/3643

Yuan325 (approved):

PR #3643, opened 2026-07-16
refactor(tools): remove unused GetToolset

twishabansal:
## Description

Stacked on top of #3641.

Companion cleanup to the promptset removal. `PrimitiveManager.GetToolset` is the toolset twin of the `GetPromptset` dead code removed in #3641 — and it is equally orphaned.

Groups are the source of truth for named collections. The REST `/toolset/{name}` path renders directly from `group.Group.ToolsetManifest`, and MCP tool listing resolves from `group.Group.ToolNames` / `ContainsTool`. Nothing in production calls `GetToolset`, so the "materialize a `tools.Toolset` view from the group on demand" helper is never reached.

This PR removes the orphaned method and its now-dangling test assertions.


Yuan325 (approved):
Thanks for cleaning this up!!! :yay

PR #3650, opened 2026-07-17
fix(config): reject duplicate flat primitive names

### LINKED ISSUE #3648: Check duplicated primitive during unmarshal process

loulanyue:
## Description

Flat-format configuration files currently allow a later primitive with the same kind and name to silently overwrite the earlier definition. This change tracks successfully unmarshaled primitive names per kind and fails at the second definition with its YAML document, line, and column.

The validation remains kind-scoped, so resources of different kinds can continue to share a name. Unknown tools skipped through `ignoreUnknownTools` are not registered and retain the existing behavior.

## Validation

- Unit: `TestUnmarshalPrimitiveConfigRejectsDuplicateNames`
- Compatibility: `TestUnmarshalPrimitiveConfigAllowsNameAcrossKinds`
- Integration: `TestParseConfigRejectsDuplicatePrimitiveNames`
- `go test -race ./internal/server ./cmd/internal`
- `go test -cover ./internal/server ./cmd/internal` (62.0% statements in both packages)
- `go vet ./internal/server ./cmd/internal`
- `go mod tidy && git diff --exit-code -- go.mod go.sum`

`golangci-lint` was not available locally and is left to CI.

## PR Checklist

- [x] Reviewed CONTRIBUTING.md
- [x] Existing issue describes and scopes the change
- [x] Relevant tests and available local static checks pass
- [x] No documentation update is necessary
- [x] This is not a breaking change

## AI assistance

Codex assisted with issue investigation, implementation, and test execution. The resulting patch and validation output were reviewed before submission.

Fixes #3648

### DIFF
--- a/cmd/internal/config_test.go
+++ b/cmd/internal/config_test.go
@@ -853,6 +853,33 @@ func TestParseConfig(t *testing.T) {
 	}
 }
 
+func TestParseConfigRejectsDuplicatePrimitiveNames(t *testing.T) {
+	ctx, err := testutils.ContextWithNewLogger()
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	raw := []byte(`kind: authService
+name: duplicate
+type: google
+clientId: first
+---
+kind: authService
+name: duplicate
+type: google
+clientId: second
+`)
+
+	parser := ConfigParser{}
+	_, err = parser.ParseConfig(ctx, raw)
+	if err == nil {
+		t.Fatal("ParseConfig() accepted duplicate authService names")
+	}
+	want := `document 2 (line 7, column 1): duplicate authService name "duplicate"`
+	if err.Error() != want {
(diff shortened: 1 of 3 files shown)


Yuan325:
Hi @loulanyue Thank you for your contributions, however, another core maintainer had already submitted a PR on this fix. I'll go ahead and close this issue for now. Thank you againnnn :)

PR #3652, opened 2026-07-17
feat: Add Resource Template structs and buildtime validation

dishaprakash:
### Description
This PR introduces support for `resourceTemplates` within MCP Toolbox. It expands the configuration parser to accept `kind: resourceTemplate` definitions (only targeting `type: file` as a resourceTemplate) and handles URI template registration and conflict resolution alongside regular `resources`. 

### Key Changes
- **Configuration Parser Updates (`cmd/internal` & `internal/server`):**
  - Updated `UnmarshalPrimitiveConfig` to fully support flat YAML mapping for `resourceTemplates`.
  - Added robust validation in `UnmarshalYAMLResourceTemplateConfig` to enforce that `type` exclusively defaults to and must be `"file"`.
  - Added strict RFC 6570 validation for `uriTemplate` definitions to ensure that only the `{path}` variable is utilized and that the base URI complies with the `file://` scheme.
  - Implemented merging and conflict detection logic to explicitly prevent name and URI collisions between `resources` and `resourceTemplates`.

### Notes for Reviewers
The scope of this PR is strictly contained to local file resource evaluation and native `{path}` variable handling for templates. No JSON-RPC handler modifications were included in this change.

Review is necessary only for code changes in resources/ directory. Rest of the code contains downflowing changes from changes in function signature


internal/server/primitives/primitives.go
```diff
@@ -191,3 +197,42 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
+	for _, rt := range r.resourceTemplates {
+		tmpl := rt.ToConfig().GetURITemplate()
+		if strings.HasSuffix(tmpl, "{path}") {
+			prefix := strings.TrimSuffix(tmpl, "{path}")
+			if strings.HasPrefix(uri, prefix) {
+				pathParam := strings.TrimPrefix(uri, prefix)
+				return nil, rt, map[string]any{"path": pathParam}, nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives_test.go
```diff
@@ -97,9 +111,61 @@ func TestUpdateServer(t *testing.T) {
...
+	}
+	if res != nil {
+		t.Errorf("Expected resource to be nil for template match, got %v", res)
+	}
+	if tmpl == nil {
+		t.Errorf("Expected to find template, got nil")
+	}
+	if params == nil || params["path"] != "/dynamic/dir/file.log" {
+		t.Errorf("Expected params to contain path=/dynamic/dir/file.log, got %v", params)
+	}
```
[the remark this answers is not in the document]
dishaprakash: Done

cmd/internal/config_test.go
```diff
@@ -2324,6 +2421,18 @@ func TestMergeConfigs(t *testing.T) {
...
+			name:      "merge with template name conflicts",
+			files:     []Config{file1, fileWithTemplateNameConflict},
+			wantErr:   true,
+			errString: "resourceTemplate 'tmpl1' (file #2)",
+		},
+		{
+			name:      "merge with template URI conflicts",
+			files:     []Config{file1, fileWithTemplateURIConflict},
+			wantErr:   true,
+			errString: "resourceTemplate 'tmpl3' (file #2) URI \"mock://{path}\" collides with resource 'tmpl1'",
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/file/file.go
```diff
@@ -290,3 +322,179 @@ func (r *FileResource) ToConfig() resources.ResourceConfig {
...
+		}
+		if !isAllowed {
+			return nil, fmt.Errorf("security violation: path %q is not within any allowedPaths", resolvedPath)
+		}
+	} else {
+		// Even if no sandbox, block hidden files for templates
+		if containsHidden(resolvedPath) {
+			return nil, fmt.Errorf("security violation: path %q contains hidden file components", resolvedPath)
+		}
+	}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -191,3 +198,47 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
 }
+
+// GetResourceTemplatesMap returns a copy of the resource templates map.
+func (r *PrimitiveManager) GetResourceTemplatesMap() map[string]resources.ResourceTemplate {
+	copied := make(map[string]resources.ResourceTemplate)
+	for name, rt := range r.resourceTemplates {
+		copied[name] = rt
+	}
+	return copied
+}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -191,3 +198,47 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
+
+// GetResourceTemplate returns a specific resource template by name.
+func (r *PrimitiveManager) GetResourceTemplate(name string) (resources.ResourceTemplate, bool) {
+	rt, exists := r.resourceTemplates[name]
+	return rt, exists
+}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -191,3 +198,47 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
+				continue
+			}
+			matches := re.FindStringSubmatch(uri)
+			if len(matches) == 2 {
+				return nil, rt, map[string]any{"path": matches[1]}, nil
+			}
+		}
+	}
+	return nil, nil, nil, fmt.Errorf("no resource or template found for URI: %s", uri)
+}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -27,15 +30,16 @@ import (
...
+	sources           map[string]sources.Source
+	authServices      map[string]auth.AuthService
+	embeddingModels   map[string]embeddingmodels.EmbeddingModel
+	tools             map[string]tools.Tool
+	toolsets          map[string]tools.Toolset
+	prompts           map[string]prompts.Prompt
+	promptsets        map[string]prompts.Promptset
+	resources         map[string]resources.Resource
+	resourceTemplates map[string]resources.ResourceTemplate
 }
```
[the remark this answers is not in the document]
dishaprakash: Done

cmd/internal/config.go
```diff
@@ -33,13 +33,14 @@ import (
...
-	Prompts         server.PromptConfigs         `yaml:"prompts"`
-	Resources       server.ResourceConfigs       `yaml:"resources"`
+	Sources           server.SourceConfigs           `yaml:"sources"`
+	AuthServices      server.AuthServiceConfigs      `yaml:"authServices"`
+	EmbeddingModels   server.EmbeddingModelConfigs   `yaml:"embeddingModels"`
+	Tools             server.ToolConfigs             `yaml:"tools"`
+	Toolsets          server.ToolsetConfigs          `yaml:"toolsets"`
+	Prompts           server.PromptConfigs           `yaml:"prompts"`
+	Resources         server.ResourceConfigs         `yaml:"resources"`
+	ResourceTemplates server.ResourceTemplateConfigs `yaml:"resource_templates"`
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/file/file.go
```diff
@@ -72,6 +82,28 @@ func validateExtension(path string) error {
...
+// containsHidden checks if any component of the path starts with a dot
+func containsHidden(path string) bool {
+	parts := strings.Split(path, string(filepath.Separator))
+	for _, part := range parts {
+		if strings.HasPrefix(part, ".") && part != "." && part != ".." {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/file/file.go
```diff
@@ -290,3 +313,192 @@ func (r *FileResource) ToConfig() resources.ResourceConfig {
...
+func (c *TemplateConfig) Initialize(ctx context.Context) (resources.ResourceTemplate, error) {
+	// Validate and resolve allowed paths if specified
+	var unresolvedAllowedPaths []string
+	var resolvedAllowedPaths []string
+	baseDir := resources.GetBaseDirFromContext(ctx)
+	if baseDir == "" {
+		baseDir = "."
+	}
+
+	for _, p := range c.AllowedPaths {
```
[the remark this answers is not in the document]
dishaprakash: This design is unsecure by default and will be communicated through documentation. Although hidden files will not available in an unsandboxed environment

internal/server/primitives/primitives.go
```diff
@@ -191,3 +199,61 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
+					continue
+				}
+				r.templateRegexCache.Store(tmpl, re)
+			}
+			matches := re.FindStringSubmatch(uri)
+			if len(matches) == 2 {
+				return nil, rt, map[string]any{"path": matches[1]}, nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/file/file.go
```diff
@@ -290,3 +313,192 @@ func (r *FileResource) ToConfig() resources.ResourceConfig {
...
+
+	// Explicitly block backward traversal in the raw input
+	if containsTraversal(pathStr) {
+		return nil, fmt.Errorf("security violation: path %q contains backward traversal components (..)", pathStr)
+	}
+
+	// If the client provides an absolute path like "/logs/server.txt",
+	// filepath.Abs will evaluate it as the OS root (not relative to the sandbox).
+	// This will cause the subsequent filepath.Rel sandbox check to fail if it's not actually within allowedPaths.
+	absPath, err := filepath.Abs(filepath.Clean(pathStr))
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/config.go
```diff
@@ -651,3 +677,67 @@ func keyToken(body ast.Node, key string) *token.Token {
...
+			for _, varName := range tmpl.Varnames() {
+				if varName != "path" {
+					return nil, fmt.Errorf("invalid uriTemplate %q: only the 'path' variable is supported (found %q)", name, varName)
+				}
+			}
+
+			// Strip all {variables} to validate the base URI structure natively
+			re := regexp.MustCompile(`\{[^}]+\}`)
+			parseableURI := re.ReplaceAllString(uriStr, "dummy")
+			parsed, err := url.ParseRequestURI(parseableURI)
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/config.go
```diff
@@ -555,7 +581,7 @@ func UnmarshalYAMLResourceConfig(ctx context.Context, name string, r map[string]
 		if uriStr, isString := uriVal.(string); !isString {
 			return nil, fmt.Errorf("invalid 'uri' field for resource %q (must be a string)", name)
 		} else {
-			parsed, err := url.Parse(uriStr)
+			parsed, err := url.ParseRequestURI(uriStr)
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/file/file.go
```diff
@@ -290,3 +313,226 @@ func (r *FileResource) ToConfig() resources.ResourceConfig {
...
+func (c *TemplateConfig) Initialize(ctx context.Context) (resources.ResourceTemplate, error) {
+	// Validate and resolve allowed paths if specified
+	var unresolvedAllowedPaths []string
+	var resolvedAllowedPaths []string
+	baseDir := resources.GetBaseDirFromContext(ctx)
+	if baseDir == "" {
+		baseDir = "."
+	}
+
+	for _, p := range c.AllowedPaths {
```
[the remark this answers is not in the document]
dishaprakash: This design is unsecure by default and this will be captured in the documentation

internal/resources/file/file.go
```diff
@@ -290,3 +313,226 @@ func (r *FileResource) ToConfig() resources.ResourceConfig {
...
+			if idx := strings.Index(base, "#"); idx != -1 {
+				base = base[:idx]
+			}
+
+			// Handle Windows drive letter quirks (e.g., /C:/...)
+			if len(base) > 2 && base[0] == '/' && base[2] == ':' {
+				base = base[1:]
+			}
+
+			pathStr = filepath.FromSlash(base)
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -191,3 +199,69 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
+					continue
+				}
+				r.templateRegexCache.Store(tmpl, re)
+			}
+			matches := re.FindStringSubmatch(matchURI)
+			if len(matches) == 2 {
+				return nil, rt, map[string]any{"path": matches[1]}, nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -15,6 +15,9 @@
 package primitives
 
 import (
+	"fmt"
+	"regexp"
+	"strings"
 	"sync"
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/server/primitives/primitives.go
```diff
@@ -191,3 +200,78 @@ func (r *PrimitiveManager) GetPromptsMap() map[string]prompts.Prompt {
...
+	for _, rt := range r.resourceTemplates {
+		tmpl := rt.ToConfig().GetURITemplate()
+		if strings.Contains(tmpl, "{path}") {
+			// Parse the template URI to isolate the path component
+			// Replace {path} with a dummy placeholder to make it a valid parseable URI
+			dummyTmpl := strings.ReplaceAll(tmpl, "{path}", "dummy_path_placeholder")
+			parsedTmpl, err := url.Parse(dummyTmpl)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
dishaprakash: Done

internal/resources/file/file.go
```diff
@@ -290,3 +312,221 @@ func (r *FileResource) ToConfig() resources.ResourceConfig {
...
+	if containsTraversal(pathStr) {
+		return nil, fmt.Errorf("security violation: path %q contains backward traversal components (..)", pathStr)
+	}
+
+	// If the path is relative, reconstruct the full path from the URI template
+	if !filepath.IsAbs(pathStr) {
+		uriTemplate := r.config.URITemplate
+		if strings.Contains(uriTemplate, "{path}") {
+			// Interpolate {path} back into the template to capture prefix AND suffix
+			fullURI := strings.Replace(uriTemplate, "{path}", pathStr, 1)
```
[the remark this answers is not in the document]
dishaprakash: Done

cmd/internal/config.go
```diff
@@ -404,6 +410,22 @@ func mergeConfigs(files ...Config) (Config, error) {
...
 
+		// Check for conflicts and merge resource templates
+		for name, rt := range file.ResourceTemplates {
+			if _, exists := merged.ResourceTemplates[name]; exists {
+				conflicts = append(conflicts, fmt.Sprintf("resourceTemplate '%s' (file #%d)", name, fileIndex+1))
+			} else {
+				// Check for URI collisions with static resources
+				uriStr := rt.GetURITemplate()
+				if existingName, exists := seenResourceURIs[uriStr]; exists {
+					conflicts = append(conflicts, fmt.Sprintf("resourceTemplate '%s' (file #%d) URI %q collides with resource '%s'", name, fileIndex+1, uriStr, existingName))
```
Yuan325: Might wanna have consistency with the resource's conflict message. `conflicts = append(conflicts, fmt.Sprintf("resourceTemplate URI '%s' used by '%s' and '%s' (file #%d)", resource.GetURI(), existingName, name, fileIndex+1))`

cmd/internal/config_test.go
```diff
@@ -621,9 +624,21 @@ tools:
...
+	resources.RegisterTemplate("mock", func(ctx context.Context, name string, decoder *yaml.Decoder) (resources.ResourceTemplateConfig, error) {
+		var cfg testutils.MockResourceTemplateConfig
+		cfg.Name = name
+		cfg.Type = "mock"
+		if err := decoder.DecodeContext(ctx, &cfg); err != nil {
+			return nil, err
+		}
+		return &cfg, nil
+	})
+
```
Yuan325: Do we need this?

cmd/internal/config_test.go
```diff
@@ -838,13 +855,12 @@ func TestParseConfig(t *testing.T) {
...
-							URI:  "mock://test",
-							Annotations: &resources.ResourceAnnotations{
-								Priority: func() *float64 { p := 1.0; return &p }(),
-							},
+						ResourceConfigBase: resources.ResourceConfigBase{
+							ConfigBase: resources.ConfigBase{
+								Name: "my-resource",
+								Type: "mock",
+								Annotations: &resources.ResourceAnnotations{Priority: float64Ptr(1.0)},},
+							URI: "mock://test",
```
Yuan325, line 1001: Why do we update this? Can we just update the testutils's Mock to include a new URI field?

internal/resources/resources.go
```diff
@@ -68,25 +68,30 @@ type ResourceAnnotations struct {
...
+func (c ConfigBase) GetName() string                      { return c.Name }
+func (c ConfigBase) GetTitle() string                     { return c.Title }
+func (c ConfigBase) GetDescription() string               { return c.Description }
+func (c ConfigBase) GetMimeType() string                  { return c.MimeType }
+func (c ConfigBase) GetAnnotations() *ResourceAnnotations { return c.Annotations }
+
+// ResourceConfigBase contains the fields for a specific resource configuration.
+type ResourceConfigBase struct {
+	ConfigBase `yaml:",inline"`
+	URI        string `yaml:"uri,omitempty"`
```
Yuan325:
```suggestion
	URI        string `yaml:"uri" validate:"required"`
```

Seems like we are setting default values for these in newConfig, so I'm assuming these are technically required~

internal/resources/file/file.go
```diff
@@ -102,6 +122,17 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+// containsTraversal checks if any component of the path is a backward traversal
+func containsTraversal(path string) bool {
+	parts := strings.Split(filepath.ToSlash(path), "/")
```
Yuan325: will this work on unix system? (paths are separated by `\`)


Might also want to make sure that we check for `%2e%2e` or `%2f..%2f`

Let's make sure to test all the edge cases~

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
...
+
+// TemplateConfig represents the configuration for a file resource template.
+type TemplateConfig struct {
+	resources.ResourceTemplateConfigBase `yaml:",inline"`
+	AllowedPaths                         []string `yaml:"allowedPaths,omitempty"`
+}
+
+// ResourceTemplateConfigType returns the resource template type identifier.
+func (c *TemplateConfig) ResourceTemplateConfigType() string {
+	return "file"
```
Yuan325:
```suggestion
	return resourceType
```

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
...
+func (c *TemplateConfig) Validate() error {
+	if err := c.ResourceTemplateConfigBase.Validate(); err != nil {
+		return err
+	}
+	parsed, _ := url.Parse(strings.ReplaceAll(c.URITemplate, "{path}", "path"))
+	if parsed.Scheme != "file" {
+		return fmt.Errorf("invalid scheme for file resource template %q: must be 'file'", c.Name)
+	}
+	return nil
+}
```
Yuan325, line 459: This seems to run a very similar checks as the `ResourceTemplateConfigBase.Validate()`. Do we still need this method?

If each type is expected to have its own custom validation, we might want to move the default implementation from ResourceTemplateConfigBase here (seems like the main check is making sure that the URITemplate is not ""). That way, the compiler ensures every type implements its own validation, preventing subclasses from accidentally falling back to the base validator.

internal/resources/file/file.go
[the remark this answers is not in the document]
Yuan325: wondering is this a design of the MCP specs?

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
...
+	return &FileTemplate{
+		config:                 c,
+		unresolvedAllowedPaths: unresolvedAllowedPaths,
+		resolvedAllowedPaths:   resolvedAllowedPaths,
+	}, nil
+}
+
+// FileTemplate handles reading content from a file template URI.
+type FileTemplate struct {
+	config                 *TemplateConfig
```
Yuan325:
```suggestion
	TemplateConfig
```

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
 	}
 	return size, nil
 }
+
+// TemplateConfig represents the configuration for a file resource template.
+type TemplateConfig struct {
+	resources.ResourceTemplateConfigBase `yaml:",inline"`
+	AllowedPaths                         []string `yaml:"allowedPaths,omitempty"`
+}
```
Yuan325, line 431:
```suggestion
}

var _ resources.ResourceTemplateConfig = &TemplateConfig{}
```

Let's make sure to check validation against the interface

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
...
+		resolvedAllowedPaths:   resolvedAllowedPaths,
+	}, nil
+}
+
+// FileTemplate handles reading content from a file template URI.
+type FileTemplate struct {
+	config                 *TemplateConfig
+	unresolvedAllowedPaths []string
+	resolvedAllowedPaths   []string
+}
```
Yuan325, line 517:
```suggestion
}

var _ resources.ResourceTemplate = &FileTemplate{}
```

internal/resources/file/template_test.go
Yuan325, line 1: can we convert these tests to table-driven struct? Might also be able to combine some of the Test

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
...
+		return nil, fmt.Errorf("failed to evaluate symlinks for %q: %w", absPath, err)
+	}
+
+	// Second security check on the resolved path to prevent symlink escape attacks
+	if err := checkSandbox(resolvedPath, r.resolvedAllowedPaths); err != nil {
+		return nil, err
+	}
+
+	// Security check for extension
+	if err := validateExtension(resolvedPath); err != nil {
```
Yuan325, line 623: Wanted to check, this should cover files that are symlinked? Do we need to also validate the user-requested path?

e.g. If resolvedPath points to a file without an extension or with a different extension (e.g. through hardlinks or symlinked targets like /tmp/xyz123), the check evaluates the target rather than the requested name. 

if pathStr is secret.env symlinked to readme.txt, it passes extension validation if only resolvedPath is checked.

internal/resources/file/file.go
```diff
@@ -357,3 +388,237 @@ func (r *FileResource) GetCurrentSize() (int64, error) {
...
+
+	if int64(len(content)) > limit {
+		truncated := content[:limit]
+		for len(truncated) > 0 {
+			runeVal, size := utf8.DecodeLastRune(truncated)
+			if runeVal == utf8.RuneError && size == 1 {
+				truncated = truncated[:len(truncated)-1]
+			} else {
+				break
+			}
```
Yuan325, line 670: If a multi-byte UTF-8 character is cut mid-sequence at limit, you peel off 1 byte.

However, on the next iteration, the remaining prefix of that cut multi-byte sequence will still be invalid UTF-8, returning RuneError && size == 1 again until the entire truncated rune is stripped. While this eventually terminates, standard Go provides utf8.Valid(truncated) or utf8.DecodeRune:

```
// Clean alternative:
for len(truncated) > 0 && !utf8.Valid(truncated) {
    truncated = truncated[:len(truncated)-1]
}
```

^Please double check this.

internal/resources/file/file_test.go
```diff
@@ -633,3 +628,18 @@ func TestFileResource_ToConfigNonRegularFile(t *testing.T) {
...
+	if cfg.ResourceConfigType() != "file" {
+		t.Errorf("expected ResourceConfigType to be 'file', got %q", cfg.ResourceConfigType())
+	}
+
+	res := file.FileResource{Config: cfg}
+	if res.ToConfig() == nil {
+		t.Errorf("expected ToConfig to return a valid config")
+	}
+}
+
```
Yuan325: probably wouldn't need this~

internal/resources/file/file_test.go
```diff
@@ -33,6 +33,7 @@ import (
 
 const defaultMaxFileSize = 5 * 1024 * 1024
 
+
```
Yuan325: extra?

internal/resources/file/file_test.go
```diff
@@ -69,12 +70,6 @@ func TestFileResource_Validation(t *testing.T) {
 		wantErrMsg string
 		failDecode bool
 	}{
-		{
-			name:       "missing path",
-			yamlStr:    "type: file",
-			wantErrMsg: "required' tag",
-			failDecode: true,
-		},
```
Yuan325, line 77: Should we keep this?

internal/resources/resources.go
```diff
@@ -203,3 +217,105 @@ func DecodeConfig(ctx context.Context, resourceType, name string, decoder *yaml.
...
+	GetURITemplate() string
+	Read(ctx context.Context, params map[string]any) (any, error)
+	ToConfig() ResourceTemplateConfig
+}
+
+// ResourceTemplateConfigBase contains the specific fields for resource template configurations.
+type ResourceTemplateConfigBase struct {
+	ConfigBase  `yaml:",inline"`
+	URITemplate string `yaml:"uriTemplate" validate:"required"`
+}
```
Yuan325, line 248: Is `file` the only type that will be implementing the `ResourceTemplate`? If yes, it might be worth to not have a `ResourceTemplateConfigBase`, instead just move this directly into `file.go`

internal/server/primitives/primitives.go
```diff
@@ -126,9 +131,20 @@ func (r *PrimitiveManager) SetPrimitives(sourcesMap map[string]sources.Source, a
...
+func (r *PrimitiveManager) GetSourcesMap() map[string]sources.Source {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	copiedMap := make(map[string]sources.Source, len(r.sources))
+	for k, v := range r.sources {
+		copiedMap[k] = v
+	}
+	return copiedMap
+}
+
```
Yuan325: We removed this function~

internal/server/primitives/primitives.go
```diff
@@ -158,3 +174,53 @@ func (r *PrimitiveManager) GroupsList() []group.Group {
...
+// GetResourceTemplatesMap returns a copy of the resource templates map.
+func (r *PrimitiveManager) GetResourceTemplatesMap() map[string]resources.ResourceTemplate {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	copied := make(map[string]resources.ResourceTemplate)
+	for name, rt := range r.resourceTemplates {
+		copied[name] = rt
+	}
+	return copied
+}
```
Yuan325: Remove if not needed. We are trying not to implement map~ Always use the primitive manager and `.Get<Primitive>` function whenever possible.

Yuan325 (changes requested):

internal/server/primitives/primitives.go
```diff
@@ -158,3 +174,53 @@ func (r *PrimitiveManager) GroupsList() []group.Group {
...
+				continue
+			}
+			matches := re.FindStringSubmatch(uri)
+			if len(matches) == 2 {
+				return nil, rt, map[string]any{"path": matches[1]}, nil
+			}
+		}
+	}
+	return nil, nil, nil, fmt.Errorf("no resource or template found for URI: %s", uri)
+}
```
Yuan325: Is this function needed? seems like we are not querying it anywhere. Can we add this when it is needed~?

internal/server/config.go
```diff
@@ -707,3 +734,59 @@ func keyToken(body ast.Node, key string) *token.Token {
...
+	var resourceType string
+	if typeVal, ok := r["type"]; ok {
+		var isString bool
+		resourceType, isString = typeVal.(string)
+		if !isString {
+			return nil, fmt.Errorf("invalid 'type' field for resourceTemplate %q (must be a string)", name)
+		}
+	} else {
+		return nil, fmt.Errorf("missing required 'type' field for resourceTemplate %q", name)
+	}
```
Yuan325: We can do:
```
resourceType, ok := r["type"].(string)
	if !ok {
		return nil, fmt.Errorf("missing 'type' field or it is not a string")
	}
```
similar to the other primitives

internal/server/config.go
```diff
@@ -707,3 +734,59 @@ func keyToken(body ast.Node, key string) *token.Token {
...
+
+	if uriVal, ok := r["uriTemplate"]; ok {
+		if uriStr, isString := uriVal.(string); !isString {
+			return nil, fmt.Errorf("invalid 'uriTemplate' field for resourceTemplate %q (must be a string)", name)
```
Yuan325: do we need to do the validation here? This should be caught during the config unmarshaling?

internal/server/config.go
```diff
@@ -707,3 +734,59 @@ func keyToken(body ast.Node, key string) *token.Token {
...
+			// Strip all {variables} to validate the base URI structure natively
+			re := regexp.MustCompile(`\{[^}]+\}`)
+			parseableURI := re.ReplaceAllString(uriStr, "dummy")
+			parsed, err := url.Parse(parseableURI)
+			if err != nil || parsed.Scheme == "" {
+				return nil, fmt.Errorf("invalid 'uriTemplate' field for resourceTemplate %q: must be a valid RFC-compliant absolute URI with a scheme", name)
+			}
+
+			// Update the map with the normalized URI
+			r["uriTemplate"] = uriStr
```
Yuan325: can we do these validation at the primitive side?

Would be good if we can keep this section consistent with other primitives.

internal/server/mcp_test.go
Yuan325, line 1: here it seems like we are adding the mock but it doesn't seem like we are checking those resources or template resources anywhere? Can we add those method checks?

internal/server/server_test.go
```diff
@@ -2250,7 +2271,130 @@ annotations: {}`,
...
+			} else {
+				if err != nil {
+					t.Errorf("unexpected error for %q: %v", tt.name, err)
+				}
+			}
+		})
+	}
+}
+
+func TestResourceTemplateConfigValidation(t *testing.T) {
```
Yuan325: Can we build on top of existing test by adding new test cases rather than starting a new validation test function.

Yuan325 (changes requested):

internal/resources/resources.go
dishaprakash: Currently yes, but for total conformance we need to add a text type resource template as well. 
I abstracted this struct here so it would be easier for whenever we support more types.

internal/server/server_test.go
dishaprakash: Makes sense. Added both static and resource template config validation into one test function.

internal/server/mcp_test.go
dishaprakash: Ah this is because the rpc wiring is done in the next PR. I'll make sure to add these checks in that

internal/server/config.go
dishaprakash: Makes sense, moved to resource.go

internal/server/config.go
dishaprakash: I've added this and also added validate required tag to both `name` and `type` fields as well.

internal/server/primitives/primitives.go
dishaprakash: This is required for the JSON RPC Endpoints, I will add it with the next PR in that case

internal/server/primitives/primitives.go
dishaprakash: I had re-introduced it since the groups implementation was initially going to be after implementing the rpc endpoints. Since we are now adding groups before the wiring, I'll remove this.

internal/server/primitives/primitives.go
dishaprakash: Removed!

internal/resources/file/file_test.go
dishaprakash: Ah removed this by mistake. Added back in

internal/resources/file/file_test.go
dishaprakash: Removed

internal/resources/file/file_test.go
dishaprakash: Removed

internal/resources/file/file.go
dishaprakash: You're right! Updated the code to validate decoded url and the extension once resolved as well

internal/resources/file/file.go
dishaprakash: Done

internal/resources/file/file.go
dishaprakash: Done

internal/resources/file/file.go
dishaprakash: The func within `ResourceTemplateConfigBase` follows generic validation for the URI Template and it's RFC compliancy and checks if it structured correctly or not as well. 
When we are implementing new types of resource templates, I think it would be easier to have this validation de-duplicated.

What do you think?

internal/resources/file/file.go
dishaprakash: Done

internal/resources/file/file.go
dishaprakash: I've updated the code to convert all backslashes into forward slashes and then check.

Also updated testcases for these specific encoded urls as well

internal/resources/resources.go
dishaprakash: Ah this was removed in the refactor, added back!

cmd/internal/config_test.go
dishaprakash: Removed

cmd/internal/config.go
dishaprakash: Updated!

internal/server/config.go
dishaprakash: Ah you're right! Removed the redundant check.

internal/resources/file/file.go
dishaprakash: On a deeper look, this implementation is cleaner but `utf.Valid` function reads the entire content of string (resource) per loop which would increase time complexity. It also flags not just the end of the string which is truncated but tries to remove any invalid byte in the entire string.

Our older implementation, even though verbose, provides better time complexity and does not alter any other data in the string.

internal/resources/file/file.go
dishaprakash: Done

internal/resources/file/file.go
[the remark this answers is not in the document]
dishaprakash: The MCP Spec does not specify the parameters of the security boundaries here (only for sensitive resources, for which we will add authentication as a future feature). The unsecure by default design was something we came up with during the design OH / bikeshedding.

Do you have a different approach that you'd like to see here?

internal/resources/file/template_test.go
dishaprakash: Done

cmd/internal/config_test.go
dishaprakash: Can I get more information about this?

internal/resources/file/file.go
Yuan325: Sounds good

Yuan325 (approved):
Thank you! Please resolve lint failure before merging :)

PR #3660, opened 2026-07-20
perf(sources,util): optimize row serialization and avoid speculative json parsing

### LINKED ISSUE #3659: Performance: Optimize Row Serialization and Avoid Speculative JSON Parsing
### Description
The current implementation of row serialization in the MCP Toolbox has significant performance bottlenecks when processing large database payloads:
1. `orderedmap.Row.MarshalJSON` creates a large number of intermediate JSON allocations.
2. In `sources/sqlite`, speculative JSON parsing (`json.Unmarshal`) is performed on every string column, leading to heavy overhead and memory allocation for standard text columns.

### Expected Behavior
Row serialization should stream directly to a buffer to minimize allocations, and speculative JSON parsing should employ a fast-path heuristic filter (`IsProbablyJSON`) to rapidly reject non-JSON strings.

### Steps to Reproduce
Run a query returning thousands of rows with standard text columns against SQLite and observe CPU/memory profiling.

Deeven-Seru:
### Description
Optimizes the core performance bottlenecks in the MCP toolbox during row serialization.

1. Eliminates speculative JSON parsing in SQLite by introducing a fast-path heuristic filter (`util.IsProbablyJSON`), reducing OOM risks and CPU overhead on large text datasets.
2. Refactors `orderedmap.Row.MarshalJSON` to stream output linearly into a single buffered `json.Encoder`, resulting in a ~2.5x speedup during row serialization.

### PR Checklist

### Issue Reference
Fixes #3659 🦕


internal/util/util.go
Yuan325, line 1: Please add unit tests for this :) Thank youuu

internal/sources/sqlite/sqlite.go
```diff
@@ -138,7 +139,7 @@ func (s *Source) RunSQL(ctx context.Context, statement string, params []any) (an
 			// Handle JSON data
 			if jsonString, ok := val.(string); ok {
 				var unmarshaledData any
-				if json.Unmarshal([]byte(jsonString), &unmarshaledData) == nil {
+				if util.IsProbablyJSON(jsonString) && json.Unmarshal([]byte(jsonString), &unmarshaledData) == nil {
```
Yuan325, line 142: Wondering if you've tried using `json.Valid()` instead of creating our own verification function?

Deeven-Seru:
Hi @Yuan325, thanks for the review!

1. **Unit Tests**: I have just pushed comprehensive unit tests for `IsProbablyJSON` to the PR.

2. **json.Valid()**: I did test `json.Valid()`, but unfortunately it defeats the purpose of this specific optimization. 
`json.Valid()` runs in `O(N)` time because it scans every single byte of the entire string to ensure full JSON correctness. If a normal text column contains a 10,000-character article, `json.Valid()` will still scan the entire 10,000 characters just to return `false` at the end. That is almost as CPU-intensive as `json.Unmarshal()`.
`IsProbablyJSON` is designed to run in `O(1)` time. It strictly checks the boundaries (first and last characters) to verify if the string *even has a chance* of being a JSON object (`{}`), array (`[]`), string (`""`), or matches exact primitive values. It immediately rejects 99.9% of normal text strings in under a nanosecond without ever scanning the middle of the string!

Issue #483, opened 2025-04-24
Add data source support for Iceberg (catalogs)


brandon-powers:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

### What are you trying to do that currently feels hard or impossible?

Here are the currently supported data sources: https://googleapis.github.io/genai-toolbox/resources/sources/. 

The idea is to add support for a source that can interact with Iceberg catalogs; whether it's REST-based, AWS Glue, or eventually BigQuery Metastore. The tools would be related to namespaces, tables, metadata exploration (e.g. `get_average_file_size_for_table`), etc. 

I'd imagine this being used for mostly read-only use cases, as it wouldn't rely on a dedicated compute engine, but an **[iceberg-go](https://github.com/apache/iceberg-go)** implementation from Apache.

Similar to the addition of Entra ID support, I think the toolbox provides useful out-of-the-box functionality & performance that can be leveraged, instead of creating a separate MCP server.  

### Suggested Solution(s)

- `iceberg` addition to the Tools and Sources directories for implementation using `iceberg-go`
- Config update [here](https://github.com/googleapis/genai-toolbox/blob/main/internal/server/config.go#L163)
- Reference examples from [Neo4j](https://github.com/googleapis/genai-toolbox/pull/189) & [Couchbase](https://github.com/googleapis/genai-toolbox/pull/307)

### Alternatives Considered

For the REST-based catalog, we may be able to utilize the HTTP source, though a direct integration would be useful + can branch by catalog type. 

### Additional Details

I'd be interested in implementing this, though wanted to raise it for feedback before doing so.

kurtisvg:
Hey @brandon-powers -- definitely open to accepting contributions on this. 

I'm not familiar with iceberg at all, so will rely on your input on what makes the most sense. Could you provide sketchs for what a tools config might look like for those 3 options you outlined?

debabsah:
Hi @kurtisvg, this one has been quiet for a while, so if @brandon-powers is not actively working on it I would be glad to take it on. I work on data tooling (I have the mssql tools in flight in this repo) and Iceberg catalogs are squarely in my wheelhouse.

To answer your question about what a tools config might look like, here is a sketch that follows the current source/tool shape. I would sequence the REST catalog first (most broadly useful and cleanest to land), then add Glue and BigQuery Metastore as follow-ups, all read-only via [apache/iceberg-go](https://github.com/apache/iceberg-go):

[...]

PR #3667, opened 2026-07-21
feat(iceberg): add iceberg source and catalog exploration tools

debabsah:
## Description
Adds an `iceberg` source plus three read-only catalog exploration tools (`iceberg-list-namespaces`, `iceberg-list-tables`, `iceberg-get-table-info`), the first milestone agreed in #483.

The source connects to an Iceberg REST catalog through [apache/iceberg-go](https://github.com/apache/iceberg-go) (pinned at v0.6.0), with an optional warehouse and an optional static bearer token (`accessToken`, following the trino source's field naming). It verifies connectivity at initialization and exposes only catalog read methods, and each tool carries the read-only annotation. `iceberg-get-table-info` returns the table's native catalog metadata: schema, partition spec, sort order, properties, location, and the current snapshot summary when the table has one, all from the load-table response, so nothing reads data files from the warehouse.

The PR adds the source (`internal/sources/iceberg/`), the three tools (`internal/tools/iceberg/`), their unit tests, the registrations in `cmd/internal/imports.go`, docs pages under `docs/en/integrations/iceberg/`, an integration test (`tests/iceberg/`), and an `iceberg` shard in `.ci/integration.cloudbuild.yaml`. The integration test self-provisions `apache/iceberg-rest-fixture:1.10.1` with testcontainers (the image iceberg-go itself tests against; it needs no credentials or object store), seeds a namespace and a table through the iceberg-go client, and covers the manifest endpoint, the three tools, and the missing-namespace and missing-table error paths.

A few notes for review:
1. This adds `github.com/apache/iceberg-go` v0.6.0 (Apache-2.0) as the one new direct dependency. Its storage layer pulls a sizable indirect surface (arrow-go, AWS and Azure SDKs); the go.mod and go.sum deltas are exactly what `go get github.com/apache/iceberg-go@v0.6.0` plus `go mod tidy` produce (re-running tidy is a zero diff), and only the catalog client is used here. Keeping this PR to the source plus the three catalog tools is meant to isolate that review.
2. The generic invoke test suites assume statement-based tools, so the integration test asserts each catalog tool directly instead, the way `tests/arcadedb` does.
3. Only `catalog: rest` is accepted for now; the flavor field exists so glue or hive backends can be added later without a config break. The follow-ups agreed on #483 (snapshot listing, table and column statistics) will come as separate small PRs.

## PR Checklist
- [x] Make sure to open an issue before writing the code (scope discussed and agreed in #483)
- [x] Ensure the tests and linter pass (gofmt, go vet, go build ./cmd/..., the unit tests, the integration test against the live REST catalog fixture, and both docs linters all pass)
- [x] Code coverage does not decrease (adds unit tests for the source and each tool, plus the integration test)
- [x] Appropriate docs were updated (source page and one page per tool under docs/en/integrations/iceberg/)
- [ ] Make sure to add `!` if this involve a breaking change (it does not)

Part of #483

## Update 2026-08-26

Main is merged in, and a follow-up commit ports the three tools to the current tool API: `Invoke` takes `sources.Source` and type-asserts it, and `GetSourceName` plus `ValidateSource` are defined per tool because `BaseTool` no longer supplies them. No behavior change.

On dependencies: `apache/iceberg-go` pulls in 22 new modules and moves several shared ones forward, including `arrow-go` 18.4.0 to 18.6.0, `aws-sdk-go-v2/config` 1.31.8 to 1.32.17, and `feature/s3/manager` 1.16.15 to 1.20.12. `go mod tidy` produces no diff, `go mod verify` reports all modules verified, and every version difference against main is an upgrade rather than a downgrade. Happy to split the dependency move out if you would rather review it separately.

Re-verified against current main: gofmt clean, `go vet`, `go build ./cmd/...`, unit tests for the source and all three tools, and `TestIcebergToolEndpoints` against a live `apache/iceberg-rest-fixture:1.10.1` catalog (all six subtests pass).


debabsah:
@Yuan325 this is up to date with main and ported to the current tool API. No behavior change, and the integration test passes against a live `apache/iceberg-rest-fixture:1.10.1` catalog. Let me know if you would like the dependency move split into its own PR.

debabsah:
Merged main again (the `Source` interface gained `IsReadOnly` in #3872, added). No behavior change; unit tests and the live-catalog integration test pass.

PR #3670, opened 2026-07-21
chore(server): make groups/list and groups/get unreachable until the gated extension release

twishabansal:
## Description

Makes `groups/list` / `groups/get` unreachable in `feat/groups` so groups ships as Part A only (scoping). These methods must be gated behind the `com.google.cloud/toolbox` extension; shipping them reachable now and gating later would be a breaking change.

Removes only the `case GROUPS_LIST` / `case GROUPS_GET` dispatch arms in `ProcessMethod` across all 5 MCP versions (methods now return `METHOD_NOT_FOUND`). All other Part B code — types, handlers, manifests, telemetry, tests — is retained and still unit-tested. Re-enabling later = re-add the arms alongside the gate. Also trims the docs describing these methods.


Yuan325 (approved):
Thank you!

PR #3673, opened 2026-07-21
Improve Spanner tool descriptions for graph queries

### LINKED ISSUE #3519: docs/feat(sources/spanner): Improve LLM context and prompting for Spanner Graph and Vector query syntax
### Problem Statement
Spanner recently introduced support for Graph database queries (using PGQ / Graph query syntax) and Vector Search (using vector search functions and indexes). However, LLMs/agents struggle to utilize these features correctly because:
- There is a lack of structured context or prompting informing the LLM of the specific Spanner Graph and Vector syntax.
- The default tool schemas (`list_tables`, `execute_sql`) do not provide enough semantic hinting for models to craft graph/vector-specific queries.
- Graph/vector query syntax is variable and highly dependent on model-specific knowledge.

### Proposed Solution
Close the capability gap for Graph and Vector workloads by:
1. **Improving Context/Prompting:** Injecting specific guidelines or system prompts for the LLM on how to construct Spanner Graph and Vector queries.
2. **Metadata & Annotations:** Adding descriptions/examples to the Spanner prebuilt tools (such as `list_graphs` and `execute_sql`) that explain how they relate to graph and vector search.
3. **Graph-Optimized/Vector-Optimized Tools:** Introducing helper tools or custom prompt templates specifically designed to aid the LLM in structuring Spanner Graph queries and vector distances.

kamran-nizamani:
## Summary

This PR improves the descriptions of the prebuilt Spanner SQL tools to provide clearer guidance for LLM agents without changing any tool behavior.

### Changes

* Expanded `execute_sql` description with clearer guidance for DML operations.
* Expanded `execute_sql_dql` description with clearer guidance for read-only queries.
* Added guidance to use `list_graphs` before writing graph queries.
* Added guidance to use Spanner Graph query syntax for graph traversal.

### Testing

* ✅ `go test ./internal/prebuiltconfigs/...`
* ✅ Project tests completed successfully.

### Notes

This change only improves tool metadata and prompting guidance. No functional behavior has been modified.

Fixes #3519


cmd/internal/skills/generator.go
```diff
@@ -42,6 +42,16 @@ All scripts can be executed using Node.js. Replace ` + "`" + `<param_name>` + "`
 {{.AdditionalNotes}}
 {{end}}
 
+
+## Security Guidelines
```
averikitsch, line 46: Is this change associate with the Spanner tool changes? We should separate the different improvements

internal/prebuiltconfigs/tools/spanner.yaml
averikitsch, line 1: Thanks for putting this together and taking on this issue!

Looking at the changes, I'm a bit hesitant about trying to force tool usage patterns by adding this much explicit guidance directly into the tool descriptions. Ideally, agents should be smart enough to handle these workflows naturally. If they do need explicit guidance, that orchestration might be better suited for a 'skill' rather than being hardcoded into the base tool definitions.

Could we try reducing the scope of these description updates for now and test them with some evals first? I'd love to see how a more minimal change impacts the agent's behavior and measure the results before we commit to this approach. Let me know what you think!

averikitsch (changes requested):

averikitsch:
@TanmayVartak any thoughts here?

PR #3676, opened 2026-07-21
fix(auth/google): require clientId when mcpEnabled is false

CorsenAI:
## Summary

In non-MCP mode, `GetClaimsFromHeader` validates ID tokens via `idtoken.Validate(ctx, token, a.ClientID)`. When `clientId` is empty, `idtoken.Validate` receives an empty audience and **skips the audience check entirely**, so the service silently accepts **any Google-signed ID token**, regardless of which OAuth client it was issued for. `Initialize()` guards this for the MCP path (`audience` or `clientId` required when `mcpEnabled: true`) but the non-MCP branch has no equivalent guard.

The reference docs already mark `clientId` as "Required for validating ID tokens in non-MCP web apps (GetClaimsFromHeader)", but `Initialize()` never enforced it. Still reproduces on `main`.

## Change

`Initialize()` now fails closed when `mcpEnabled: false` and `clientId` is empty, mirroring the existing guard used when `mcpEnabled: true`. The validation call sites keep the original `idtoken.Validate` usage.

## Tests

- `Initialize()` rejects empty `clientId` in non-MCP mode.
- `GetClaimsFromHeader` with no token returns no claims.

`go test ./internal/auth/google/` and `go vet` pass.

## Note

Re-opens the fix from #3480 (labelled `priority: p1`, assigned to @duwenxin99), which auto-closed when the original fork was deleted; the CLA is now signed (check green). Happy to align with any in-flight internal fix.

### DIFF
--- a/internal/auth/google/google.go
+++ b/internal/auth/google/google.go
@@ -59,6 +59,9 @@ func (cfg Config) Initialize() (auth.AuthService, error) {
 			return nil, fmt.Errorf("`audience` or `clientId` is required when `mcpEnabled` is true")
 		}
 	} else {
+		if cfg.ClientID == "" {
+			return nil, fmt.Errorf("`clientId` is required when `mcpEnabled` is false")
+		}
 		if cfg.Audience != "" {
 			return nil, fmt.Errorf("`audience` is not allowed when `mcpEnabled` is false")
 		}
--- a/internal/auth/google/google.go
+++ b/internal/auth/google/google.go
@@ -30,6 +30,12 @@ import (
 
 const AuthServiceType string = "google"
 
+// validateIDToken validates a Google-issued ID token against the expected
+// audience. It is a package-level indirection over idtoken.Validate so tests
+// can substitute a fake validator and exercise the audience-binding logic
+// without performing live network calls to Google's certificate endpoint.
+var validateIDToken = idtoken.Validate
+
 // validate interface
 var _ auth.AuthServiceConfig = Config{}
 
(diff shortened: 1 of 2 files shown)


internal/auth/google/google.go
```diff
@@ -120,10 +137,21 @@ func (a AuthService) GetAuthorizationServer() string {
...
+}
+
 // Verifies Google ID token and return claims
 func (a AuthService) GetClaimsFromHeader(ctx context.Context, h http.Header) (map[string]any, error) {
 	if token := h.Get(a.Name + "_token"); token != "" {
-		payload, err := idtoken.Validate(ctx, token, a.ClientID)
+		payload, err := a.validateIDToken(ctx, token, a.ClientID)
 		if err != nil {
 			return nil, fmt.Errorf("google ID token verification failure: %w", err)
 		}
```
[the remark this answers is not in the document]
CorsenAI: Good catch - addressed in 9214bb3. GetClaimsFromHeader now resolves the audience as clientId with an audience fallback and fails closed when neither is set, so idtoken.Validate is never called with an empty audience. Mirrors ValidateMCPAuth. Added test coverage (clientId used / audience fallback / fail-closed).

internal/auth/google/google.go
```diff
@@ -30,6 +30,12 @@ import (
 
 const AuthServiceType string = "google"
 
+// validateIDToken validates a Google-issued ID token against the expected
+// audience. It is a package-level indirection over idtoken.Validate so tests
+// can substitute a fake validator and exercise the audience-binding logic
+// without performing live network calls to Google's certificate endpoint.
+var validateIDToken = idtoken.Validate
```
duwenxin99: This variable doesn't seem necessary. Let's keep the original usage of `idtoken.Validate`

internal/auth/google/google.go
```diff
@@ -59,6 +65,16 @@ func (cfg Config) Initialize() (auth.AuthService, error) {
 			return nil, fmt.Errorf("`audience` or `clientId` is required when `mcpEnabled` is true")
 		}
 	} else {
+		// In non-MCP mode the only audience binding available for ID token
```
duwenxin99: Can we clean up the comments and keep them concise?

duwenxin99 (changes requested):

CorsenAI:
Addressed both review comments in 779f298: reverted to `idtoken.Validate` directly (no indirection), removed the validator-dependent test, and trimmed the comments. The fail-closed Initialize() guard for non-MCP mode remains. CLA signature is being arranged on our side.

PR #3680, opened 2026-07-22
feat(config): support non-empty environment expansion

### LINKED ISSUE #3635: Env-var interpolation treats set-but-empty variables as valid values; support Compose-style ${VAR:-default} and ${VAR:?} semantics
### Prerequisites

Closest matches found: #3543 (structural config validation) and #3634 (SQL-statement validation) cover *validation timing*; #2086 is about missing env vars in the Gemini CLI extension packaging. None address interpolation semantics for **set-but-empty** variables, which is a distinct gap.

Express "this optional field should only exist when its env var has a non-empty value" in `tools.yaml` — and more generally, get fail-fast behavior when a variable is *set but empty*.

Current behavior in `ConfigParser.parseEnv` (`cmd/internal/config.go`):

| State of `VAR` | `${VAR}` | `${VAR:default}` |
|---|---|---|
| unset | ✅ startup error ("environment variable not found") | uses default |
| **set to `""`** | **silently interpolates `""`** | **silently interpolates `""`** |
| set to value | interpolates value | interpolates value |

The set-but-empty row is the trap, because in containerized deployments (Cloud Run, Kubernetes, Compose) a declared env var is **always set** — commonly to the empty string when the underlying setting is absent (`--set-env-vars="FOO=${FOO:-}"` patterns, K8s `value: ""`, secret-optional wiring). The "unset → error" safety net never fires in exactly the environments Toolbox is most deployed to.

Concrete failure from our deployment: a `cloud-gemini-data-analytics-query` tool referenced an optional context set:

```yaml
agentContextReference:
  contextSetId: ${WEATHERBOT_GDA_CONTEXT_SET_ID}
```

Before the context set was created, the Cloud Run service had the variable declared-but-empty. Interpolation produced `contextSetId: ""` — structurally valid YAML, passes config parsing, server starts cleanly — and the failure only surfaces when the Gemini Data Analytics API rejects the empty resource name **at query time**, wrapped in an MCP error, through an agent, to an end user. Config-time mistake, conversation-time detection.

### Suggested Solution(s)

Adopt the Docker Compose / bash parameter-expansion conventions in `parseEnv`, which are already muscle memory for the target audience:

1. **`${VAR:-default}`** — substitute default when `VAR` is **unset or empty** (this is the big one; Compose semantics).
2. **`${VAR:?}` / `${VAR:?error message}`** — **startup error when unset or empty**. This is the correct tool for fields like `contextSetId` where an empty value is never valid: the operator gets the failure at deploy, not at query time.
3. Keep existing behavior for back-compat: `${VAR}` (error only when unset) and `${VAR:default}` (default only when unset — note this collides visually with bash's `:-`; worth documenting explicitly, or accepting both `:` and `-` forms).
4. **Optional, field-level complement:** the `cloud-gemini-data-analytics-query` tool kind could treat an empty-string `agentContextReference.contextSetId` as "no context reference" instead of forwarding `""` to the API. Empty is never a meaningful resource name, so dropping the reference is strictly better than a guaranteed runtime rejection. (Analogous per-field handling may make sense for other optional resource-name fields.)

Items 1–3 are a contained change to the regex + branch logic in `parseEnv`, plus docs. Happy to contribute the implementation with unit tests per CONTRIBUTING.md once maintainers agree on the shape — flagging especially whether you'd prefer to (a) add `:-`/`:?` alongside the existing `:` syntax or (b) also revise `:` to Compose semantics in a major release.

### Alternatives Considered

- **Our current workaround:** sentinel comments around the optional block in `tools.yaml` (`# ⟪AGENT_CONTEXT_BEGIN⟫` / `# ⟪AGENT_CONTEXT_END⟫`) plus an `awk` filter in the deploy script that strips the block when the variable is empty, producing a second, ephemeral config file. It works, but every consumer of the config now depends on shell preprocessing, local `run` and CI `deploy` scripts must duplicate the awk logic, and the file on disk no longer matches what the server actually loaded.
- **Templating engines** (envsubst, gomplate, Helm-style) in front of Toolbox: heavier version of the same workaround, same drift problem.
- **Always requiring the value:** not viable for genuinely optional features — the whole point of an optional context set is that the tool must work before it exists.

### Additional Details

- Implementation pointer: `ConfigParser.parseEnv` in `cmd/internal/config.go` — regex `\$\{(\w+)(:([^}]*))?\}`, substitution via `os.LookupEnv` where `found == true` for empty strings.
- Precedent: [Docker Compose variable interpolation](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/) supports exactly this matrix (`${VAR:-default}`, `${VAR-default}`, `${VAR:?err}`, `${VAR?err}`), so the semantics are already familiar and well-specified.
- Related in spirit to #3543 / #3634 (fail-at-config-time rather than query-time), but orthogonal in mechanism: those are validation passes, this is interpolation semantics.

jstar0:
## Description

Container deployments commonly define optional environment variables as empty
strings. Because `os.LookupEnv` reports those variables as present, Toolbox's
existing interpolation accepted the empty value and could defer an invalid
configuration failure until an MCP tool call reached a downstream API.

This change adds two explicit expansion forms:

- `${VAR:-default}` uses `default` when `VAR` is unset or empty.
- `${VAR:?message}` fails configuration parsing when `VAR` is unset or empty;
  the message is optional and is included in the error when provided.

The documented `${VAR}` and `${VAR:default}` behavior is unchanged. The new
operator spellings were previously interpreted as legacy default strings that
started with `-` or `?`; adopting the requested operators intentionally
redefines that narrow edge. The configuration documentation now includes the
full unset, empty, and non-empty behavior matrix.

Tests cover the compatibility matrix, required/optional tracking, the offline
lenient parser path, source locations and custom errors, unsupported no-colon
forms, and expansion through `ParseConfig`.

## Verification

- `go test -race ./cmd/internal -run '^(TestParseEnv|TestParseConfigEnvExpansion)$' -count=1`
- `go test -race ./cmd/... ./internal/... -count=1`
- `golangci-lint run --timeout 10m`
- `go mod tidy && git diff --exit-code -- go.mod go.sum`

## PR Checklist

- [x] Reviewed `CONTRIBUTING.md`
- [x] Opened and discussed an issue before implementation
- [x] Tests and linter pass
- [x] New source branches are covered by regression tests
- [x] Configuration documentation is updated
- [x] Reviewed the compatibility boundary for the new operators

Fixes #3635


Yuan325:
Thank you @jstar0 for submitting this. I'll take a look asap :)

jstar0:
Merged current `main` on exact head `6b8416fab02`.

The conflict was in `parseEnv`: this change keeps `${VAR:-default}` / `${VAR:?message}` and the structured missing-variable errors, and now uses main's comment-aware YAML lexer tokenization so placeholders inside comments stay unexpanded. Comment-skip tests now expect `requiredEnvVars` only for the active placeholders.

Verification: `go test ./cmd/internal -run 'TestParseEnv|TestParseConfigEnvExpansion'` passed. Google CLA is green on this head.

PR #3684, opened 2026-07-22
feat(server): list available tools and nearest match in tool-not-found errors

### LINKED ISSUE #3636: Include available tool names (or nearest match) in the tool-not-found error so LLM agents can self-correct
### Prerequisites

No existing issue found for enriching the tool-not-found error. Related in spirit: #3634 (validation) catches config-time mistakes; this one addresses a *runtime* mistake made by the calling LLM, not by the config author.

When an MCP client calls `tools/call` with a tool name that doesn't exist, Toolbox returns:

```
invalid tool name: tool with name "lookup_sensor" does not exist
```

For a human that's fine. For an LLM agent it's a dead end: the model knows the name was wrong but has no signal about what's right, so the common failure mode is retrying the same wrong name (possibly forever) or apologizing to the end user. We hit this in production when we renamed a tool (`lookup_sensor` → `list_sensors`): the agent had the stale name in its context and looped on it.

Our workaround is client-side: the dispatcher pattern-matches the error, appends the cached tool list, and returns a synthetic error like *"Tool \"lookup_sensor\" does not exist. The available tools right now are: list_sensors, latest_observation, … Pick the closest match by intent — do NOT retry \"lookup_sensor\"."* With that, the model self-corrects in the same turn, every time. But every MCP client has to reinvent this.

### Suggested Solution(s)

Enrich the server-side error for unknown tool names:

1. **List the available tool names** in the error message (for typical toolsets of 5–50 tools this is a few hundred bytes).
2. Optionally include a **nearest-match suggestion** (edit distance or prefix match): `did you mean "list_sensors"?`.
3. For very large toolsets, cap the list (e.g. 25 names + "and N more") to bound the payload.

Error messages returned through MCP are consumed by LLMs as prompt text — making them actionable is one of the highest leverage-per-byte ergonomic improvements available. The same enrichment could apply to unknown toolset names on `tools/list`.

### Alternatives Considered

- **Client-side self-correction** (what we do): works but must be rebuilt in every client, requires the client to have a cached tool list, and drifts if tools change mid-session.
- **Prompting the agent to only use listed tools:** helps but does not survive stale context after renames; models fall back to remembered names under pressure.
- **Doing nothing:** the retry-loop failure mode burns end-user conversations, which is the most expensive place to discover a rename.

### Additional Details

Happy to contribute the implementation (error construction + tests) if maintainers agree on the shape — flagging especially whether the nearest-match suggestion is wanted in v1 or the plain name list is preferred.

AndyFooBlah:
## Description

Implements #3636, as invited by @duwenxin99 there.

When a `tools/call` (any MCP protocol version) or HTTP API tool lookup fails, the error now appends the available tool names and — when one is close enough by case-insensitive edit distance — a nearest-match suggestion:

```
invalid tool name: tool with name "lookup_sensor" does not exist. Did you mean "list_sensors"? Available tools: latest_observation, list_sensors
```

MCP errors are consumed by LLM agents as prompt text, so listing the valid names lets an agent self-correct in the same turn instead of retry-looping on a stale or misspelled name (the failure mode described in the issue).

Design points, per the issue discussion:
- **Backward compatible prefix** — the message still starts with `invalid tool name: tool with name %q does not exist`, so clients pattern-matching the old message keep working.
- **Scoped names** — MCP handlers list the current toolset's tools; the HTTP API handlers (no toolset scope) list the global tool map's names.
- **Bounded payload** — names are sorted for determinism and capped at 25, with `(and N more)` for larger toolsets.
- **Conservative suggestion** — nearest match is suggested only when the edit distance is at most half the longer name's length; otherwise only the list is shown.

The helper lives in `internal/tools` (`UnknownToolError`) and is unit-tested (message shapes, cap, tie-breaking, Levenshtein). The shared integration-test expectations were trimmed to the stable message prefix (they are substring assertions).

Happy to extend the same treatment to unknown toolset/prompt names in a follow-up if wanted.

Fixes #3636

## PR Checklist
- [x] Appropriate docs were updated (n/a — error-message ergonomics only)


internal/server/primitives/primitives.go
```diff
@@ -144,3 +144,14 @@ func (r *PrimitiveManager) GroupsList() []group.Group {
 
 	return groupsList
 }
+
+// ToolNames returns the names of all registered tools.
+func (r *PrimitiveManager) ToolNames() []string {
```
duwenxin99: We should not return all tool names. Some tools may need to be omitted due to certain configurations on the server side, which is why we don't have a generic `ToolsList` method. This naive tool listing may be working for now, but soon won't work with our upcoming server extension support that will selectively expose tools based on client extensions declaration.

internal/util/util.go
```diff
@@ -357,6 +357,24 @@ func IgnoreUnknownToolsFromContext(ctx context.Context) bool {
 	return false
 }
 
+// toolSuggestionsKey is the key used to store the tool-suggestions mode within
+// context. The value is the raw flag string; the tools package owns its
+// interpretation (this package cannot import tools without a cycle).
+const toolSuggestionsKey contextKey = "toolSuggestions"
```
duwenxin99: Not sure whether it is necessary to store it in the context as this is not a request-scoped config. We should probably store it somewhere within `primitiveMgr` or pass it directly in function signatures.

internal/tools/suggest.go
```diff
@@ -0,0 +1,205 @@
...
+	if !ok {
+		return SuggestionsFull
+	}
+	return SuggestionMode(mode).resolve()
+}
+
+// UnknownToolError returns the error for a tool name that could not be
+// resolved. Beyond the base message, and depending on mode, it lists the
+// available tool names (capped at maxListedNames) and, when one is close
+// enough, a nearest-match suggestion. MCP errors are consumed by LLM agents as
```
duwenxin99: Can we clean up the comments to be more concise?

internal/server/mcp/v20250618/method.go
```diff
@@ -167,13 +167,13 @@ func toolsCallHandler(ctx context.Context, id jsonrpc.RequestId, g group.Group,
 
 	// Verify tool belongs to the current group before resolving globally.
 	if !g.ContainsTool(toolName) {
-		err = fmt.Errorf("invalid tool name: tool with name %q does not exist", toolName)
+		err = tools.UnknownToolError(toolName, g.ToolNames, tools.SuggestionModeFromContext(ctx))
 		return jsonrpc.NewError(id, jsonrpc.INVALID_PARAMS, err.Error(), nil), err
```
duwenxin99, line 171: Are we sure most clients provide protocol errors to the LLM?

internal/tools/tool_suggest.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return fmt.Errorf("%s", b.String())
+}
+
+// nearestName returns the candidate most similar to name, if any candidate is
+// similar enough to be a plausible rename or typo. Similarity is
+// case-insensitive Levenshtein distance; a candidate qualifies when the
+// distance is at most half the longer name's length. Candidates must be
+// sorted so ties resolve deterministically.
+func nearestName(name string, candidates []string) (string, bool) {
```
duwenxin99, line 137: We should rename the file to something like `tool_suggest.go`. Also, we should probably also implement some sort of short-circuiting to prevent calculating distance for all tool names.

duwenxin99 (changes requested):

AndyFooBlah:
Thanks @duwenxin99 — all five addressed in e9de2ec, except the fourth, which I think is a real design question rather than something I should just patch. Taking them in order.

### `PrimitiveManager.ToolNames()` — removed

You're right, and the problem was worse than the listing. The nearest-match suggestion was computed from that same unscoped slice, so even the `/api` path capped at `nearest` could still surface the name of a tool the server meant to withhold — the cap bounded *how many* names leaked, not *which*.

So rather than filter it, I deleted the method. Its only callers were the two `/api` routes, and those carry no group scope, so there is no set of names there known to be disclosable. They now return the bare unknown-tool message for every mode, and `TestApiUnknownToolNeverDisclosesNames` pins that across all three (the test name it looks up is one edit from a real tool, so it fails loudly if anything ever consults the registry there again).

Suggestions are now an MCP-only affordance, where `g.ToolNames` already bounds them to the connected group — which should also mean extension-based filtering is inherited for free, since it will land on the group rather than on the registry. Let me know if that assumption is wrong.

### Context storage — moved

Agreed, it isn't request-scoped. It now lives on `PrimitiveManager` with a `ToolSuggestions()` getter, set once during server construction; `util.WithToolSuggestions` and `ToolSuggestionsFromContext` are gone. Every handler that needed it already had `primitiveMgr` in scope, so no signatures changed.

Same feedback applies to my #3711, which was carrying server-wide result caps the same way — I've fixed that there too rather than leave the pattern in flight.

### Rename and short-circuiting — done

`suggest.go` → `tool_suggest.go`. On short-circuiting: only the closest candidate is ever used, so the scan now avoids computing any distance that couldn't change the answer:

- a candidate whose **length** alone can't beat the incumbent is skipped without entering the matrix (length difference lower-bounds edit distance);
- the matrix walk **aborts** once an entire row reaches the cutoff, since distances never decrease as rows advance;
- an **exact match** ends the scan.

The risk with this kind of pruning is that it quietly changes *which* candidate wins, so I added `TestNearestNameMatchesNaive`, which sweeps name/candidate-list combinations and asserts the optimized scan agrees with an unoptimized reference on every input. I checked it actually bites: with a tempting-but-wrong prune (skipping candidates that can't clear the threshold on their own) it fails with `nearestName("ab", [TOOL_A a]) = ("", false), naive = ("a", true)`.

### Comments — tightened

Trimmed throughout, and dropped `AtMost` and the rank map, which the `/api` change made dead.

### "Are we sure most clients provide protocol errors to the LLM?" — no, we're not

This is the one I don't want to paper over, because I think you've found the weak premise in the original issue.

The error *type* is right. The spec puts "Unknown tools" under **Protocol Errors** — JSON-RPC, with `-32602` in its worked example, which is what we return. So this isn't a case where we should be using `isError: true` instead.

But spec-correct placement is not the same as model visibility, and the spec doesn't promise the latter: its message-flow diagram routes `Tool result` → `Client` → `Process result` → LLM, and the client guidance is to "validate tool results before passing to LLM." Protocol errors aren't described as reaching the model at all. Whether the text lands in front of an LLM is entirely a property of the host: agent loops that catch a failed tool call and feed the message back as an observation will show it; clients that raise to application code will surface it to a developer instead.

So I'd revise the claim to: this helps a developer or an agent framework that echoes tool-call failures back to the model, and it is not guaranteed to reach the model on every client. That's a narrower benefit than the issue implied, and I'm happy to soften the PR description accordingly.

If you'd rather have the guarantee, the alternative is returning unknown-tool as a tool execution error (`isError: true`) so it always reaches the model — at the cost of deviating from the spec's stated mapping, and of `tools/call` no longer producing a protocol error for a name that isn't there. I'd lean toward staying spec-correct, but it's your call and I'll implement whichever you prefer.

Rebased on current main; `go build ./...` and `go test ./internal/... ./cmd/...` are green locally.

AndyFooBlah:
@duwenxin99 friendly ping on this one — no rush if you're heads-down, but it's blocked on a call only you can make.

All five review comments are addressed in e9de2ec (pushed 08-20, CI green): `PrimitiveManager.ToolNames()` is gone and `/api` no longer discloses names in any mode, the mode moved off the context onto `PrimitiveManager`, the file is renamed with the nearest-match scan short-circuited, and the comments are tightened.

The one thing I deliberately didn't act on is your question about whether clients surface protocol errors to the LLM — because I think you were right and it changes the feature's shape:

- **Keep it a protocol error** (what the PR does). Spec-correct — the MCP spec lists "Unknown tools" under Protocol Errors with `-32602` — but the spec never promises those reach the model, so the benefit is client-dependent.
- **Return `isError: true`** instead. Guarantees the model sees the suggestion, at the cost of deviating from the spec's stated mapping.

I lean toward staying spec-correct and narrowing the claim in the PR description, but I'll implement whichever you prefer. Happy to close this if you'd rather not carry the feature at all.

AndyFooBlah:
Rebased onto current main and resolved conflicts; build/vet/tests green locally. Ready for another look.

AndyFooBlah:
Status note so this doesn't look abandoned: the branch has gone stale against `main` again and I'm going to leave it that way for now rather than rebase on a timer.

All five review comments were addressed in e9de2ec back on 08-20, and the remaining open item is the design question above — whether unknown-tool should stay a JSON-RPC protocol error (spec-correct, but not guaranteed to reach the model) or become an `isError: true` tool result (guaranteed, but off-spec). I'd rather not rewrite the feature in either direction until someone weighs in.

**I'll rebase and get it green within a day of anyone picking it up for review** — just comment here and I'll refresh it. No urgency implied; the queue is clearly deep.

PR #3704, opened 2026-07-24
feat(migrate): convert toolset to group kind during migration

twishabansal:
## Description

Parsing a toolset as a group is stricter than the old toolset path, so three behaviors change:

- A `description` on a toolset is **dropped** with a warning: a toolset has no description of its own, so promoting one would publish a description the collection never declared. Use `kind: group` to keep it.
- Unknown fields are **rejected** at startup rather than silently ignored.
- A `kind: toolset` and `kind: group` sharing a name is a **duplicate-name error**; it used to warn and let the group win.

Docs under `configuration/groups/` and `configuration/toolsets/` said toolsets need "no migration", and now list these three changes.

Stacked on #3764, which converts the two cloud-storage toolsets carrying descriptions so those descriptions apply rather than being dropped by the rule above.


cmd/internal/config.go
```diff
@@ -168,8 +168,11 @@ func (p *ConfigParser) ParseConfig(ctx context.Context, raw []byte) (Config, err
 	return config, nil
 }
 
-// ConvertConfig converts configuration file to flat format.
-func ConvertConfig(raw []byte) ([]byte, error) {
+// ConvertConfig converts configuration file to flat format. When migrateToolset
+// is true, toolsets are additionally rewritten to the group kind (both nested
+// and already-flat inputs); the runtime parse path passes false to leave
+// toolsets untouched.
+func ConvertConfig(raw []byte, migrateToolset bool) ([]byte, error) {
```
Yuan325: is there any concerns to migrate toolset for all?

cmd/internal/config.go
```diff
@@ -259,16 +270,32 @@ func hasKindField(input yaml.MapSlice) bool {
 	return false
 }
 
-// transformDocs transforms the configuration file from nested to flat format
-// yaml.MapSlice will preserve the order in a map
-func transformDocs(kind string, input yaml.MapSlice) ([]yaml.MapSlice, error) {
+// migrateToolsetKind rewrites a top-level `kind: toolset` to `kind: group` in a
+// flat-format doc, preserving field order. Docs of any other kind are left
+// unchanged.
+func migrateToolsetKind(input yaml.MapSlice) {
```
Yuan325: This function might not be necessary~ are we able to just update the key in the `case "toolsets"` at line 227 (existing main branch line) | 233 (curent line)?

Yuan325 (changes requested):
Thank you! left a couple of comments~ lmk what you think about them, overall LGTM :)

cmd/internal/config.go
twishabansal: Good call. Dropped the flag, ConvertConfig now always rewrites toolsets to groups.

This causes 2 small behaviour changes:

1. Unknown fields are now caught at startup. UnmarshalYAMLToolsetConfig rebuilt the map as justTools and dropped everything else like description field. Raised a PR to fix an affected config file: https://github.com/googleapis/mcp-toolbox/pull/3764.
2. Name clashes between a kind: toolset and a kind: group are now a duplicate-name error instead of a warning.

twishabansal:
Hi @Yuan325 

The previous version rewrote the toolset kind in two places: once for nested input, once for already-flat docs. I realised it makes more sense to have two sequential passes instead of one forked pass: transformDocs flattens nested input exactly as before and still emits kind: toolset, then migrateToolsetKind rewrites the kind on every doc once it's flat. Both shapes converge on a single function.

This seems to have made the code changes simpler and more robust. Let me know what you think!

cmd/internal/config.go
```diff
@@ -259,6 +264,69 @@ func hasKindField(input yaml.MapSlice) bool {
...
+// warnDescriptionDropped reports a discarded toolset description. Warning is
+// best effort: a caller without a logger in context still gets the conversion.
+func warnDescriptionDropped(ctx context.Context, name string) {
+	logger, err := util.LoggerFromContext(ctx)
+	if err != nil {
+		return
+	}
+	logger.WarnContext(ctx, fmt.Sprintf("toolset %q: dropping description, which a toolset does not support; declare the collection as `kind: group` to keep it", name))
+}
+
```
Yuan325: We can probably move this inline~ seems like it's only used once in the `migrateToolsetKind()`

cmd/internal/config.go
```diff
@@ -259,6 +264,69 @@ func hasKindField(input yaml.MapSlice) bool {
...
+	for i, item := range input {
+		if i == descIndex {
+			continue
+		}
+		if i == kindIndex {
+			item.Value = "group"
+		}
+		migrated = append(migrated, item)
+	}
+
```
Yuan325: Do we need this? Since we already have the kindIndex, can we just replace the value directly with group?

cmd/internal/config.go
```diff
@@ -259,6 +264,69 @@ func hasKindField(input yaml.MapSlice) bool {
...
+func mapSliceString(items yaml.MapSlice, key string) string {
+	for _, item := range items {
+		if k, ok := item.Key.(string); ok && k == key {
+			if v, ok := item.Value.(string); ok {
+				return v
+			}
+		}
+	}
+	return ""
+}
```
Yuan325: Wondering if we can just get a nameIndex during the first for loop of `migrateToolsetKind()`, with that we won't need this extra function~

Yuan325 (changes requested):

Yuan325 (approved):
LGTM, feel free to merge after resolving the comments :)

cmd/internal/config.go
twishabansal: Done.

cmd/internal/config.go
twishabansal: Updated!

cmd/internal/config.go
twishabansal: Done!

PR #3709, opened 2026-07-24
chore: remove groups list/get handlers from non-draft versions

twishabansal:
## Description

The `groups/list` and `groups/get` handlers were defined in the `v20241105`, `v20250326`, `v20250618`, and `v20251125` MCP protocol versions but were **never routed** by their `ProcessMethod` switch statements : only `vdraft` actually exposes the groups feature. This PR removes that dead code from the four non-draft versions.

The groups feature is untouched in `vdraft`.


Yuan325 (approved):
Thank you!

PR #3711, opened 2026-07-25
feat(server): declarative per-tool result caps with structured truncation notice

### LINKED ISSUE #3637: Declarative per-tool result caps (maxRows / max response bytes) with a structured truncation notice
### Prerequisites

No existing issue found for result-size limits ("row limit", "truncate response", "max rows" all came up empty). #3405 (closed) covered execute_sql *input* restrictions; this is about *output* size. #753 (BigQuery selective fields) is adjacent but per-tool-kind.

Keep tool responses within the transport and context-window budgets of LLM agent clients — declaratively, per tool.

Agent transports have hard limits: Gemini Live's WebSocket empirically closes the connection (code 1007) somewhere around ~1 MB per frame; every LLM has a context budget where a 500-row JSON dump is pure waste. In our deployment, a time-series tool (`postgres-sql` kind) asked for two weeks of 5-minute sensor readings returned ~4,000 rows ≈ 800 KB, which killed the Live session mid-conversation.

We currently defend in two hand-rolled layers, both of which feel like they belong in Toolbox:

1. **Inside the SQL statement:** `LIMIT LEAST(GREATEST(COALESCE($6, 5000), 1), 5000)` plus a `row_limit` parameter the agent may pass — boilerplate repeated in every time-series tool, easy to forget on the next tool.
2. **In the client:** a 256 KB cap that replaces oversized results with a structured hint ("re-call with a tighter filter — add location and measurement_type…"), which reliably steers the model to a narrower query on the next turn.

### Suggested Solution(s)

Per-tool (and/or server-level default) response caps in `tools.yaml`:

```yaml
observations_in_range:
  kind: postgres-sql
  source: my-pg
  maxRows: 5000            # truncate result sets beyond this
  maxResponseBytes: 262144  # and/or a byte ceiling
  statement: |
    SELECT ...
```

Behavior on hitting a cap: return the truncated result **plus a structured truncation notice** (e.g. an `isTruncated: true` flag and a text part like "Result truncated to 5,000 of 12,481 rows — narrow the query filters to see the rest"). The notice text matters: it's consumed by an LLM, and in our experience an explicit "here's how to narrow it" instruction converts a failure into a correct follow-up call almost every time.

Enforcement at the Toolbox layer is engine-agnostic (count rows / bytes while marshaling) — no per-source SQL rewriting required, though a future optimization could push `maxRows` down as a LIMIT for SQL sources.

### Alternatives Considered

- **LIMIT clauses in every statement** (what we do): works per-tool but is boilerplate, silently missing on new tools, and can't express a byte ceiling (row width varies).
- **Client-side caps** (what we also do): must be rebuilt per client; the server has already paid the query + serialization cost for bytes that get thrown away.
- **Prompting the agent to always pass narrow filters:** unreliable under paraphrase; guardrails shouldn't depend on model compliance.

### Additional Details

The two failure modes this prevents — transport disconnects and context-window flooding — are both invisible at config-authoring time and only appear under realistic data volumes, which makes a declarative guard rail especially valuable. Happy to contribute the implementation if maintainers agree on the shape (per-tool field names, notice format, whether byte caps belong in v1).

AndyFooBlah:
## Description

Implements #3637, as invited by @anubhav756 there. This follows the [design proposal posted on the issue](https://github.com/googleapis/mcp-toolbox/issues/3637#issuecomment-5077131751); the one open design question (truncation-notice placement) is implemented as an **extra text content item** for MCP — happy to switch to `_meta` (or mirror into both) if preferred.

Two optional fields on the shared tool config, inherited by every tool kind with no per-kind changes:

```yaml
tools:
  search_flights:
    kind: postgres-sql
    source: my-pg
    statement: SELECT * FROM flights WHERE origin = @origin
    maxRows: 200             # cap list-shaped results by element count
    maxResponseBytes: 65536  # cap serialized result size
```

When a cap fires, the caller gets a structured notice so an LLM can distinguish a capped result from a complete one:

```json
{"truncation": {"truncated": true, "appliedLimit": "maxRows", "returnedRows": 200, "totalRows": 1434}}
```

Design points:
- **Single enforcement point** — a `tools.CapResult` helper applied at the seven invocation sites (six MCP protocol versions' `toolsCallHandler` + the HTTP API's `toolInvokeHandler`), so all ~100 tool kinds are covered by one code path.
- **Never a broken row** — for list-shaped results, `maxResponseBytes` drops whole trailing elements; byte accounting matches what the MCP surface actually renders (one text item per element). Non-list results are cut at a UTF-8 boundary.
- **Non-breaking** — both surfaces only add to the response when truncation occurs: MCP appends a final content item (existing items unchanged), the HTTP API adds a `truncation` sibling field with `omitempty`. Tools without caps (the default) take the existing code path with zero overhead.
- **Config validation** — `validate:"gte=0"` on both fields.
- **Non-goals (per the issue)** — no per-request overrides, no pagination semantics, no change to streaming/binary tools.

Testing: unit tests for `CapResult` (row cap, byte cap, both, boundary cases, UTF-8 safety, first-row-over-budget) plus a handler-level test asserting the notice appears in MCP content. Docs: new "Result Caps" section in the tools configuration reference.

## PR Checklist
- [x] Appropriate docs were updated

Fixes #3637 🦕

### DIFF
--- a/cmd/internal/flags.go
+++ b/cmd/internal/flags.go
@@ -77,4 +77,6 @@ func ServeFlags(flags *pflag.FlagSet, opts *ToolboxOptions) {
 	flags.Int64Var(&opts.Cfg.HttpMaxRequestBytes, "http-max-request-bytes", server.DefaultHTTPMaxRequestBytes, "Maximum MCP HTTP request body size in bytes.")
 	flags.BoolVar(&opts.Cfg.EnableDraftSpecs, "enable-draft-specs", false, "Opt-in and test upcoming draft MCP specifications.")
 	flags.StringSliceVar(&opts.Cfg.DisableExt, "disable-ext", []string{}, "Specifies MCP extension URIs disabled on this server.")
+	flags.IntVar(&opts.Cfg.MaxRows, "max-rows", 0, "Default cap on the number of rows a tool result may contain. 0 means uncapped. Overridden per tool by the maxRows field.")
+	flags.IntVar(&opts.Cfg.MaxResponseBytes, "max-response-bytes", 0, "Default cap on the serialized size of a tool result, in bytes. 0 means uncapped. Overridden per tool by the maxResponseBytes field.")
 }
--- a/docs/en/documentation/configuration/tools/_index.md
+++ b/docs/en/documentation/configuration/tools/_index.md
@@ -457,6 +457,48 @@ You can bind specific arguments to tools at the transport level using URL query 
 
 For a comprehensive guide, see the [URL Parameter Binding](./url_parameter_binding.md) documentation.
 
+## Result Caps
+
+You can bound the size of any tool's results by declaring caps in its
+configuration. This protects both the server and the calling LLM's context
+window from unexpectedly large result sets (for example, a query without a
+`LIMIT` clause matching millions of rows).
+
+| **field**        | **type** | **default** | **description**                                                       |
+|------------------|:--------:|:-----------:|-----------------------------------------------------------------------|
+| maxRows          |   int    | 0 (uncapped)| Maximum number of elements returned for list-shaped results.          |
+| maxResponseBytes |   int    | 0 (uncapped)| Maximum serialized size of the returned result, in bytes.             |
+
+```yaml
+kind: tool
+name: search_flights
+type: postgres-sql
+source: my-pg-instance
+statement: |
+  SELECT * FROM flights WHERE origin = @origin
+maxRows: 200
+maxResponseBytes: 65536
+```
+
+Both caps apply to every invocation surface (MCP `tools/call` and the HTTP
+API). For list-shaped results, `maxResponseBytes` drops whole trailing rows
+rather than cutting a row mid-way. A result that is not list-shaped and
+exceeds the byte cap is returned as a `truncatedResult` string holding the
+leading fragment, since a document cut mid-way is no longer valid JSON.
+
+When a cap fires, the response includes a structured truncation notice so the
+calling model can tell a capped result from a complete one — as an additional
+text content item in MCP responses, or a `truncation` field alongside `result`
+in HTTP API responses:
+
+```json
+{"truncation": {"truncated": true, "appliedLimit": "maxRows", "returnedRows": 200, "totalRows": 1434}}
+```
+
+Result caps are a guardrail, not a paging mechanism: agents that need to walk
+a large result set should page explicitly (for example, with `LIMIT` and
+`OFFSET` parameters in the statement).
+
 ## Using tools with MCP Toolbox Client SDKs
 
 Once your tools are defined in your configuration, you can retrieve them directly from your application code.
--- a/docs/en/documentation/configuration/tools/_index.md
+++ b/docs/en/documentation/configuration/tools/_index.md
@@ -464,10 +464,10 @@ configuration. This protects both the server and the calling LLM's context
 window from unexpectedly large result sets (for example, a query without a
 `LIMIT` clause matching millions of rows).
 
-| **field**        | **type** | **default** | **description**                                                       |
-|------------------|:--------:|:-----------:|-----------------------------------------------------------------------|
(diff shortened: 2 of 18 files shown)


AndyFooBlah:
Thanks — the diagnosis here is correct and worth fixing, but the specific suggestion would break this path, so I've fixed it a different way.

**Why `json.RawMessage` doesn't work here:** the bytes being returned are a *deliberately truncated* serialization, i.e. not valid JSON (`{"blob":"xxxxxxx` with no closing quote or brace). `json.Marshal` validates the output of `MarshalJSON`, so wrapping an invalid fragment in `json.RawMessage` fails outright rather than passing the bytes through:

```
json.Marshal(json.RawMessage(truncated))
  → err = json: error calling MarshalJSON for type json.RawMessage: unexpected end of JSON input
```

That would turn a cosmetic double-escaping wart into a 500 on every byte-capped non-list result.

**What the raw `string` actually produced** (confirming the underlying finding — there is a real extra serialization layer):

```
{"result":"\"{\\\"blob\\\":\\\"xxxxxxxxxxxxxxx\""}   // truncated, double-escaped
{"result":"{\"blob\":\"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\"}"}  // untruncated, for comparison
```

**Fix applied:** the truncated fragment is now carried as a string value inside a valid JSON object rather than as a bare string:

```go
return map[string]string{"truncatedResult": string(kept)}, &Truncation{...}
```

This marshals exactly once, is valid JSON on the wire, and names the fragment for the calling model so a partial document doesn't read as a malformed one. Added `TestCapResultNonListRemainsMarshalable`, which round-trips the capped result through `json.Marshal`/`Unmarshal` — it fails against both the old bare-string version and the suggested `json.RawMessage` version, so this stays fixed.

Docs updated to describe the `truncatedResult` shape. Also rebased onto current main, which picked up the `group.Group` refactor and the new `v20260728` protocol version (now wired for caps too). Worth noting `v20260728` has a native `_meta` on its result — happy to move the truncation notice there for that version if maintainers prefer, though I've kept the extra content item everywhere for now since `content` is what reliably reaches the model.

AndyFooBlah:
Rebased on current main and pushed a follow-up commit (50f2084).

@duwenxin99 made a point on #3684 that applied here too: a server-wide setting was being carried on the request context, when it's server configuration rather than per-request state. This PR was doing the same thing with the result caps, so I've moved them rather than leave the pattern in flight:

- the caps now live on `PrimitiveManager` (`ResultCaps()` / `SetResultCaps()`, set once at server construction);
- `util.WithResultCaps` / `ResultCapsFromContext` are gone;
- `tools.CapResultForTool` takes the defaults as arguments instead of reaching into the context for them.

Precedence is unchanged — a tool's declared cap wins, including an explicit `0`, and an omitted field inherits the server default — and the existing precedence tests still cover all four combinations, now seeded through `SetResultCaps` instead of the context.

Still open from my earlier comment, whenever you get to it: the naming question (`maxRows` vs aligning with BigQuery's `maxQueryResultRows`), and whether you want the follow-up that makes BigQuery's existing source-level cap emit the same truncation notice — today it drops rows silently, so a model reads a capped result as a complete one.

AndyFooBlah:
Rebased onto current main and resolved conflicts; build/vet/tests green locally. Ready for another look.

AndyFooBlah:
Status note so this doesn't look abandoned: the branch has gone stale against `main` again and I'm going to leave it that way for now rather than rebase on a timer.

The PR is feature-complete and was green as of the 08-29 rebase: per-tool `maxRows`/`maxResponseBytes`, server-wide `--max-rows`/`--max-response-bytes` defaults with per-tool override, and the structured truncation notice. Two questions from #3637 are still open whenever someone gets to it — whether to align naming with BigQuery's `maxQueryResultRows`, and whether you want the follow-up making BigQuery's existing source-level cap emit the same truncation notice (today it drops rows silently, so a model reads a capped result as complete).

**I'll rebase and get it green within a day of anyone picking it up for review** — just comment here and I'll refresh it.

PR #3723, opened 2026-07-27
feat(server/mcp): introduce generic client extension registry

duwenxin99:
1. Description

- Added mcp.InitializeProtocols to orchestrate startup setup in internal/server/mcp, decoupling server.go from protocol-specific packages.
- Add generic extension infra to the 20260728 mcp version.

2. PR Checklist

3. Issue Reference
Fixes #0 🦕

### DIFF
--- a/internal/server/mcp/v20260728/extensions.go
+++ b/internal/server/mcp/v20260728/extensions.go
@@ -17,7 +17,7 @@ package v20260728
 import "slices"
 
 // SupportedExtensions lists all MCP extension URIs supported by Toolbox by default.
-var SupportedExtensions = []string{}
+var SupportedExtensions = map[string]any{}
 
 // ServerExtensions is the map of extension URIs enabled on this server.
 var ServerExtensions map[string]any
@@ -29,29 +29,21 @@ func Initialize(disabledExts []string) {
 
 // InitializeExtensions configures the active ServerExtensions map based on disabled extensions.
 func InitializeExtensions(disabledExts []string) {
-	var enabled []string
-	for _, ext := range SupportedExtensions {
+	ServerExtensions = make(map[string]any)
+	for ext, extConfig := range SupportedExtensions {
 		if ext != "" && !slices.Contains(disabledExts, ext) {
-			enabled = append(enabled, ext)
+			ServerExtensions[ext] = extConfig
 		}
-	}
-
-	ServerExtensions = make(map[string]any, len(enabled))
-	for _, uri := range enabled {
-		ServerExtensions[uri] = map[string]any{}
 	}
 }
 
 // ParseSupportedExtensions returns a map of extension URIs that are supported by both the client and the server.
 func ParseSupportedExtensions(clientExtensions map[string]any) map[string]any {
 	supported := make(map[string]any)
-	if len(clientExtensions) == 0 {
+	if len(clientExtensions) == 0 || len(ServerExtensions) == 0 {
 		return supported
 	}
 	for uri, clientExtVal := range clientExtensions {
-		if clientExtVal == nil {
-			continue
-		}
 		if _, ok := ServerExtensions[uri]; ok {
 			supported[uri] = clientExtVal
 		}
--- a/internal/server/mcp/v20260728/extensions.go
+++ b/internal/server/mcp/v20260728/extensions.go
@@ -24,11 +24,6 @@ var ServerExtensions map[string]any
 
 // Initialize performs version-specific protocol setup for v20260728.
 func Initialize(disabledExts []string) {
-	InitializeExtensions(disabledExts)
-}
-
-// InitializeExtensions configures the active ServerExtensions map based on disabled extensions.
-func InitializeExtensions(disabledExts []string) {
 	ServerExtensions = make(map[string]any)
 	for ext, extConfig := range SupportedExtensions {
 		if ext != "" && !slices.Contains(disabledExts, ext) {
@@ -49,10 +44,4 @@ func ParseSupportedExtensions(clientExtensions map[string]any) map[string]any {
 		}
 	}
 	return supported
-}
-
-// SupportsExtension returns true if both the client and server support the given extension URI.
-func SupportsExtension(supportedExts map[string]any, uri string) bool {
-	_, ok := supportedExts[uri]
-	return ok
 }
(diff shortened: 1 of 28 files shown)


internal/server/mcp/v20260728/extensions.go
Yuan325, line 1: Might want to add this in `/2026-07-28` instead~ To be version-controlled schema.

internal/server/mcp/v20260728/extensions.go
```diff
@@ -0,0 +1,133 @@
...
+func WithClientExtensions(ctx context.Context, exts ClientExtensions) context.Context {
+	return context.WithValue(ctx, extensionsKey{}, exts)
+}
+
+// ClientExtensionsFromContext retrieves the client's extension map from the context.
+func ClientExtensionsFromContext(ctx context.Context) ClientExtensions {
+	exts, _ := ctx.Value(extensionsKey{}).(ClientExtensions)
+	return exts
+}
+
```
Yuan325: We might be able to just extract it during the `validateMetadata()` function within `method.go`. Probably wouldn't need to keep track in the context? 

Otherwise if we have to, I'd suggest adding it along with other `xxFromContext()` in `util/util.go`~

internal/server/mcp/v20260728/manifests.go
Yuan325, line 1: Doesn't seem like we need to make any update here~?  Might be a merge conflict

internal/server/mcp/v20260728/method.go
```diff
@@ -96,6 +96,14 @@ func validateMetadata(id jsonrpc.RequestId, params RequestParams, stdio bool) (a
...
 }
 
+// withClientExtensions attaches client extensions from request metadata to the context.
+func withClientExtensions(ctx context.Context, meta *RequestMetaObject) context.Context {
+	if meta == nil || meta.MetaClientCapabilities == nil {
+		return ctx
+	}
+	return util.WithClientExtensions(ctx, util.ExtractClientExtensions(meta.MetaClientCapabilities.GetExtensions()))
+}
+
```
Yuan325: Should we add this in `util/util.go` instead? Since all of our `context.Withxyz` is there

internal/server/mcp/v20260728/method.go
Yuan325, line 1: Do we need to add all the `ctx = withClientExtensions(ctx, req.Params.Meta)`? Can we add it as needed (e.g. when adding secure param to toolbox extension, we will check for the client extension in toolsList and toolsCalls handler).


Another thought: we can modify the `validateMetadata()` function to return the clientCapabilities, since we are checking the metadata field there. Also, do we really need to save those information to context? We only need to retrieve the information once during toolsList -> to decide which tools to remove, and once during toolsCall -> to decide if user can access the tools. Might not need to save the whole map into context (generally also a bad practice to save map into Go context)

internal/server/mcp/v20260728/types.go
```diff
@@ -119,12 +119,22 @@ type ResultMetaObject struct {
...
 }
 
+// GetExtensions returns standard extensions advertised by the client.
+func (c *ClientCapabilities) GetExtensions() map[string]any {
+	if c == nil {
+		return nil
+	}
+	return c.Extensions
+}
+
```
Yuan325: This file is kept for the schemas, let's not add any function here~

internal/server/server_test.go
```diff
@@ -1980,3 +1980,66 @@ func TestDefaultToolsetIsAlphabeticallySorted(t *testing.T) {
...
+			name:       "default enables toolbox extension",
+			enableExt:  nil,
+			disableExt: nil,
+			want:       []string{"com.google.cloud/toolbox.v1"},
+		},
+		{
+			name:       "custom enable and disable",
+			enableExt:  []string{"com.google.cloud/toolbox.v1", "io.modelcontextprotocol/tasks"},
+			disableExt: []string{"io.modelcontextprotocol/tasks"},
+			want:       []string{"com.google.cloud/toolbox.v1"},
```
Yuan325, line 2018: seems like we might want to throw an error? since there's conflict between enableExt and disableExt

internal/server/server_test.go
```diff
@@ -1980,3 +1980,66 @@ func TestDefaultToolsetIsAlphabeticallySorted(t *testing.T) {
...
+		},
+		{
+			name:       "disable default extension",
+			enableExt:  nil,
+			disableExt: []string{"com.google.cloud/toolbox.v1"},
```
Yuan325: Here too, we might want to throw error since there's conflict on the default enableExt and disableExt.

internal/util/util.go
```diff
@@ -393,3 +394,65 @@ func EnableDraftSpecsFromContext(ctx context.Context) (bool, bool) {
...
+// SupportsExtension returns true if both the client and server support the given extension URI.
+func SupportsExtension(ctx context.Context, uri string) bool {
+	exts := ClientExtensionsFromContext(ctx)
+	if exts == nil || !exts[uri] {
+		return false
+	}
+	if serverExts, ok := ctx.Value(serverExtensionsKey).([]string); ok {
+		return slices.Contains(serverExts, uri)
+	}
+	return true
```
Yuan325: should this be false?

Yuan325 (changes requested):

internal/server/mcp/v20260728/manifests.go
duwenxin99: This is so that we can generate tool manifest based on the extensions we support. For example, the secure params require certain tools to be removed from the manifest if the extension is not supported etc.

internal/server/mcp/v20260728/extensions.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+	return supported
+}
+
+// SupportsExtension returns true if both the client and server support the given extension URI.
+func SupportsExtension(supportedExts map[string]any, uri string) bool {
+	_, ok := supportedExts[uri]
+	return ok
+}
```
Yuan325: Since this is just one liner function we might not need it~

internal/server/mcp/v20260728/extensions.go
```diff
@@ -0,0 +1,58 @@
...
+
+// InitializeExtensions configures the active ServerExtensions map based on disabled extensions.
+func InitializeExtensions(disabledExts []string) {
+	ServerExtensions = make(map[string]any)
+	for ext, extConfig := range SupportedExtensions {
+		if ext != "" && !slices.Contains(disabledExts, ext) {
+			ServerExtensions[ext] = extConfig
+		}
+	}
+}
```
Yuan325, line 33: can we combine this 2 into a single function?

Yuan325 (approved):
LGTM with some nit

PR #3730, opened 2026-07-28
fix(util): convert exponent-form JSON numbers in ConvertNumbers

saivedant169:
## Summary

`util.ConvertNumbers` returns an error for JSON numbers written in exponent notation (for example `1e5`, `1e-07`, `1e+21`), even though these are valid JSON numbers that Go's own `encoding/json` produces.

## Mechanism

`ConvertNumbers` decides between int and float based on whether the number string contains a `.`:

```go
case json.Number:
    if strings.Contains(v.String(), ".") {
        return v.Float64()
    }
    return v.Int64()
```

Exponent-form numbers contain no `.`, so they fall to `json.Number.Int64()`, which calls `strconv.ParseInt` and rejects exponent syntax:

```
strconv.ParseInt: parsing "1e5": invalid syntax
```

This is reachable from normal input. `DecodeJSON` parses numbers with `UseNumber()`, and `json.Marshal` emits exponent form for large or small floats:

```
json.Marshal(1e-07) = 1e-7
json.Marshal(1e+21) = 1e+21
```

A generic map parameter (`MapParameter.Parse` with an empty `ValueType`) runs its values through `ConvertNumbers`, so a tool call whose map value is a scientific-notation number fails with `failed to parse integer or float values in map`.

## Fix

Try `Int64()` first and fall back to `Float64()`. This preserves integers, and routes decimals, exponent notation, and values outside the int64 range to float:

```go
case json.Number:
    if i, err := v.Int64(); err == nil {
        return i, nil
    }
    return v.Float64()
```

Plain integers still decode to `int64` and decimals to `float64`, so existing behavior is unchanged apart from the previously erroring cases.

## Testing

Added `TestConvertNumbers` covering integers, decimals, exponent forms, an integer beyond the int64 range, passthrough of non-numbers, and nested maps and slices.

```
$ go test ./internal/util/...
ok  	github.com/googleapis/mcp-toolbox/internal/util
ok  	github.com/googleapis/mcp-toolbox/internal/util/parameters
```

`gofmt` and `go vet ./internal/util/` are clean.


internal/util/convertnumbers_test.go
Yuan325, line 1: Can we move this to util_test.go?

Yuan325 (changes requested):
Hi @saivedant169 Thank you for your contribution! Would like to merge this in after the feedback is resolved! :)

saivedant169:
Done, moved `TestConvertNumbers` into `util_test.go` and removed the separate file. Thanks!

saivedant169:
@Yuan325 bump on this one, the test move is done. CI hasn't run yet, still waiting on approval. Happy to rebase if it's gone stale.

Yuan325 (approved):

Yuan325:
Thank you @saivedant169 ! This LGTM :) Will include this update in our next release

saivedant169:
@Yuan325 macOS failed in TestSingleEdit in cmd, a file-watcher timeout unrelated to this change. Could you re-run the job? Thanks again!

PR #3734, opened 2026-07-28
fix(tools/bigquery): run the allowedDatasets check before the dry run

### LINKED ISSUE #3717: bigquery-execute-sql: run allowedDatasets check before dry-run (non-existent foreign datasets return BQ 404)
## Summary

With `allowedDatasets` set, `bigquery-execute-sql` documents that queries accessing tables outside the list are rejected. In practice the allowlist check runs **only after a successful BigQuery dry-run**, so a query against a **non-existent** foreign dataset returns a BQ **404 Not Found** (via `ProcessGcpError` → AgentError) instead of the clear AgentError:

`query accesses dataset 'PROJECT.DATASET', which is not in the allowed list`

Security is usually still fail-safe (no data returned), but the policy signal is wrong and agents/evals cannot distinguish “table missing” from “dataset not allowlisted”.

## Code path (`main`)

`internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go` `Invoke`:

1. `DryRunQuery(...)` — on error, **return immediately** via `ProcessGcpError`
2. Only then: writeMode checks + `allowedDatasets` loop (dry-run `ReferencedTables` + `TableParser`)

So the SQL `TableParser` path never runs when dry-run fails for `dataset/table not found`.

## Repro

Source config with e.g. `allowedDatasets: [omnifox_admin, …]` (any non-empty list).

```sql
SELECT 1 FROM `my-project.definitely_not_allowlisted.no_such_table` LIMIT 1
```

**Actual:** readable googleapi Error 404 Not found: Table …  
**Expected:** AgentError mentioning `definitely_not_allowlisted` is not in the allowed list (independent of whether the table exists).

Contrast: when dry-run **succeeds** against a real table outside the list, the allowlist message appears as designed.

## Ask

Run `TableParser` (or equivalent static dataset extraction) **before** dry-run when `allowedDatasets` is non-empty, and reject with the existing “not in the allowed list” AgentError. Keep dry-run for statement-type / referenced-table enrichment afterward.

## Context

Follow-up to https://github.com/googleapis/mcp-toolbox/issues/3716 (403 → opaque 500). Same deploy; eval task scored PARTIAL because rejection was safe but not allowlist-shaped.

Checked against Toolbox **1.6.0** behavior and current `main` source (2026-07-27).

he-yufeng:
## What

Fixes #3717. With `allowedDatasets` configured, `bigquery-execute-sql` only checked the allowlist after a successful dry run, so a query against a non-existent foreign dataset failed the dry run first and surfaced a BigQuery 404 via `ProcessGcpError`. Callers could not tell "table missing" apart from "dataset not allowlisted".

## How

The parser-based allowlist check now runs before the first BigQuery call, so the policy error fires whenever it is determinable. The post-dry-run block keeps its `ReferencedTables` check, which catches the shapes the parser cannot see (views expanded server-side, DDL target/destination tables), and the `statementType` gates and INFORMATION_SCHEMA/EXTERNAL_QUERY rejections are unchanged.

## Verification

New regression test with a mock server that fails every dry run with a 404: a query against `forbidden_dataset.no_such_table` surfaces the allowlist error ahead of the 404, and a query against `allowed_dataset.no_such_table` still sees the 404 pass through (the `ProcessGcpError` path is preserved). The existing `TestInvokeDatasetRestrictions` matrix (12 cases) passes unchanged, plus `go vet` clean and the whole `internal/tools/bigquery/...` tree green.


he-yufeng:
Good catch, fixed in 62ba2e5.

One nuance on the trigger path: a domain-scoped project written directly in the query can't reach this check, since the table parser caps identifiers at three parts and skips longer ones. The reachable path is a domain-scoped DEFAULT project on the client: an unqualified `dataset.table` reference then expands to `example.com:project.dataset.table`, and the strict three-part split skipped the check for every unqualified table. The fix splits from the right so the project keeps its dots, applied to both the pre-dry-run parser pass and the dry-run stats pass, plus a test with a domain-scoped default project.

internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
```diff
@@ -149,6 +160,24 @@ func (t Tool) Invoke(ctx context.Context, s sources.Source, params parameters.Pa
...
+	// instead of the allowlist error, which callers cannot tell apart from
+	// "table missing".
+	if len(source.BigQueryAllowedDatasets()) > 0 {
+		parsedTables, parseErr := bqutil.TableParser(sql, bqClient.Project())
+		if parseErr != nil {
+			return nil, util.NewAgentError("could not parse tables from query to validate against allowed datasets", parseErr)
+		}
+		for _, tableID := range parsedTables {
+			projectID, datasetID, ok := splitTableID(tableID)
+			if ok && !source.IsDatasetAllowed(projectID, datasetID) {
```
Genesis929: Security Issue: Fail-Open Vulnerability. If splitTableID fails to parse properly and returns ok == false, the && condition short-circuits. Instead of blocking the request, it silently allows it through. Permission checks should always fail-closed. Please change this to if !ok || !source.IsDatasetAllowed(...).

he-yufeng: Fixed in 01b6c18048, fail-closed as requested. An unattributable table reference now stops the query with its own error (`could not attribute table '%s' to a dataset ...`) instead of short-circuiting past the gate.

One honest note on reachability after tracing the parser: with today's `TableParser`, every emitted ID goes through `formatTableID`, which only emits `project.dataset.table` or default-project-qualified `dataset.table`, so each ID carries at least three segments and `!ok` is not reachable via SQL right now. So this is defense in depth at a permission gate (exactly the right default there) rather than a currently exploitable bypass, and there is no SQL input I can add as a regression test that would exercise the new branch. Existing dataset-restriction tests (before-dry-run, domain-scoped project, INFORMATION_SCHEMA forms) all still pass locally with `go test`.

he-yufeng:
Merge conflict with main resolved: main's new TestGetAnnotations landed at the same anchor as this PR's two new tests, so both are kept now. The bigquery tool tree tests pass locally (`go test ./internal/tools/bigquery/...`), branch is mergeable again.

he-yufeng:
Verified this still stands apart from the approved #3979: dry-run still runs before the allowedDatasets check, and the 3-segment split still bypasses the domain-scoped project. Worth a look alongside that one.

PR #3738, opened 2026-07-28
fix(tools/bigquery): keep the provider error classification in bigquery-execute-sql

### LINKED ISSUE #3716: BigQuery 403 from impersonated SA should be AgentError, not HTTP 500 / Cloudflare 502
## Summary

When `bigquery-execute-sql` runs as an **impersonated** service account that lacks `dataViewer` on a referenced table, BigQuery returns **403 Access Denied**. Toolbox currently surfaces that as an HTTP **500** / JSON-RPC `INTERNAL_ERROR`. LLM connectors (e.g. Claude via Cloudflare) then show an opaque **502 origin_bad_gateway**, so the model cannot recover and operators think the MCP is down.

Invalid SQL already maps correctly to a readable AgentError. Access Denied for “SA cannot read this table” should get the same treatment — distinct from true credential misconfiguration (token refresh failure).

## Environment

- Toolbox **1.6.0** (also checked behavior notes against current `main` as of 2026-07-27)
- Source: BigQuery with `impersonateServiceAccount`, `writeMode: blocked`, `allowedDatasets` including a platform dataset + underlying datasets needed for authorized UNION view expansion
- Profile SA: `dataViewer` on the platform dataset only; `metadataViewer` (not `dataViewer`) on tenant core datasets so authorized views can dry-run-expand

## Observed behavior

| Query | Result |
|-------|--------|
| `SELECT … FROM platform_dataset.view` (authorized UNION view) | OK |
| `SELECT … FROM tenant_core.table` (SA has metadataViewer only) | HTTP 500 (~0.5s) → Cloudflare 502 |
| `COUNT(*) AS rows` (invalid alias) | Readable BQ 400 tool / AgentError |

## Code path (current `main`)

1. `internal/tools/bigquery/bigqueryexecutesql`: dry-run failure → `util.ProcessGcpError(err)`
2. `internal/util/errors.go` `ProcessGcpError`: googleapi **403** → `NewClientServerError(..., StatusForbidden, …)` (`CategoryServer`)
3. On the non-`clientAuth` / impersonation path, that server-category error is remapped to HTTP **500** / JSON-RPC INTERNAL_ERROR
4. Edge connectors (Cloudflare in front of Cloud Run) present that as **502 origin_bad_gateway**

Related prior work: invalid SQL → AgentError (#3055 / #3056). Same class of “agent can recover from a readable tool error” applies to SA Access Denied on a specific table.

## Ask

1. When `bigquery-execute-sql` (and related BigQuery tools) receive googleapi **403** from an impersonated / ADC identity for a **table/dataset access** denial, return a normal tool / AgentError result with `isError: true` and the BQ message visible — **not** HTTP 500 — unless the failure is truly a misconfigured credential (e.g. token refresh / impersonation setup broken).
2. Optionally: support a `metadataDatasets` list (or per-tool metadata scoping) separate from `allowedDatasets`, so cores can remain listed for UNION dry-run expansion / `datasets.get` without advertising them as SELECT-able in metadata tools.

## Workaround we ship today

Tool-description + house-rule steering away from direct core SELECTs; curated golden-query tools; authorized views on a platform dataset for all SELECT traffic.

## Related observation (separate but adjacent)

`allowedDatasets` validation runs **only after a successful dry-run** (`bigqueryexecutesql.go`: dry-run error returns before the “not in the allowed list” check / SQL table parser). Querying a **non-existent** foreign dataset therefore yields a readable BQ **404**, not the allowlist AgentError — even though the dataset is clearly outside `allowedDatasets`. Happy to open a follow-up issue if useful: run the SQL table parser allowlist check **before** (or on failure of) dry-run.

Thanks!

he-yufeng:
## What

Fixes #3716. On the actual query run, `bigquery-execute-sql` wrapped every provider failure in a blanket `NewClientServerError("error running sql", 500, err)`, so a BigQuery 403 from an impersonated service account missing `dataViewer` surfaced as HTTP 500 and connectors showed an opaque bad-gateway instead of anything recoverable. The dry-run path in the same tool already routed through `util.ProcessGcpError`, which made the two stages inconsistent with each other.

The actual-run error path now goes through `util.ProcessGcpError` too, the same classification the other GCP tools use (bigtable, firestore, datalineage, and this tool's dry run): 401/403 keep their status with the provider cause attached, and everything else becomes a readable `AgentError` so the model sees the real message (invalid SQL, missing table) instead of a generic 500.

## How

One-line swap at the `RunSQL` error site, plus `TestInvokeRunSqlErrorClassification`: a 403 from the provider must come back as `ClientServerError` with code 403 and the Access Denied cause visible, and a 400 must come back as an `AgentError` with the cause visible. Existing tests (`TestInvokeDatasetRestrictions` and the rest of the package) still pass.

## Verification

- `go test ./internal/tools/bigquery/bigqueryexecutesql/` - all pass, including the two new classification subtests
- `go vet ./internal/tools/bigquery/bigqueryexecutesql/` - clean

### DIFF
--- a/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
+++ b/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql.go
@@ -237,7 +237,7 @@ func (t Tool) Invoke(ctx context.Context, s sources.Source, params parameters.Pa
 	logger.DebugContext(ctx, fmt.Sprintf("executing `%s` tool query: %s", resourceType, sql))
 	resp, err := source.RunSQL(ctx, bqClient, sql, statementType, nil, connProps, map[string]string{"mcp-toolbox-tool": resourceType})
 	if err != nil {
-		return nil, util.NewClientServerError("error running sql", http.StatusInternalServerError, err)
+		return nil, util.ProcessGcpError(err)
 	}
 	return resp, nil
 }
--- a/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql_test.go
+++ b/internal/tools/bigquery/bigqueryexecutesql/bigqueryexecutesql_test.go
@@ -16,6 +16,7 @@ package bigqueryexecutesql_test
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"strings"
@@ -28,8 +29,10 @@ import (
 	"github.com/googleapis/mcp-toolbox/internal/tools"
 	bqutil "github.com/googleapis/mcp-toolbox/internal/tools/bigquery/bigquerycommon"
 	"github.com/googleapis/mcp-toolbox/internal/tools/bigquery/bigqueryexecutesql"
+	"github.com/googleapis/mcp-toolbox/internal/util"
 	"github.com/googleapis/mcp-toolbox/internal/util/parameters"
 	bigqueryrestapi "google.golang.org/api/bigquery/v2"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 )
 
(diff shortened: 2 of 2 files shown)


Yuan325 (approved):

Yuan325:
Thank you for your contributions! @he-yufeng :)

PR #3758, opened 2026-07-31
feat(tool/looker): add create_merge_query tool

### LINKED ISSUE #3756: Add create_merge_query to Looker MCP
### Prerequisites

I want to merge different explores, this could be done through: [create_merge_query](https://docs.cloud.google.com/looker/docs/reference/looker-api/4.0/26.6/methods/Query/create_merge_query) 

### Suggested Solution(s)

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

LacombeLouis:
## Description

Adds a new Looker tool, `looker-create-merge-query` (registered as
`create_merge_query`), exposing Looker's
[Merged Results](https://cloud.google.com/looker/docs/merged-results) through the
API (`POST /merge_queries`).

**Impact:** a `query` call can only target one explore, so today an agent that
needs fields from two explores has to run both queries and stitch the rows
together itself — losing Looker's join semantics and burning extra tool calls.
This tool performs the merge server-side, including across explores backed by
different databases.

**Solution:** the tool takes a `source_queries` array, one entry per explore.
Order is significant — the first entry is the primary query, and the results of
every later entry are joined onto it using that entry's `merge_fields`, matching
Looker's left-outer-join semantics. Each source query is created with
`sdk.CreateQuery` (idempotent, so an identical query is reused rather than
duplicated) and the resulting query ids are handed to `sdk.CreateMergeQuery`.
Merge-level `pivots`, `sorts`, `limit`, `column_limit`, `total`,
`dynamic_fields` and `vis_config` are supported.

### Important API limitation

Looker has **no endpoint that runs a merge query** — only create and get. So the
tool returns the merge query `id`, its `result_maker_id`, the created source
queries (`name`, `model`, `explore`, `query_id`, `slug`) and a `url` that opens
the merged results in the Looker UI, rather than rows of data. This is stated
explicitly in the tool description and the docs page so the planner does not
expect data back and instead falls back to `query` per explore when it needs the
rows in-conversation.

### Notable details

- **Strict source-query keys.** An unknown key inside a source query is
  rejected rather than silently dropped, so a typo like `explores` surfaces as an
  error instead of a query quietly missing a filter.
- **Unquoted-parameter filters.** Reuses
  `lookercommon.EscapeUnquotedParameterFilters` per source query, so LookML
  `type: unquoted` parameters behave exactly as they do in `looker-query`.
- **Discoverability.** Adds a note to the prebuilt `query` tool description
  pointing at `create_merge_query`. The shared parameter descriptions in
  `lookercommon` are deliberately left untouched — changing them forces edits
  across five separate integration-test expectation blocks.
- **Annotations.** Uses `NewWriteAnnotations` (non-destructive write), since the
  call does create objects in Looker but nothing user-visible or destructive.

### Testing

- Unit tests: YAML config parse (success + failure), `processSourceQuery` (happy
  path, every optional key, 10 error cases), `processSourceQueries` (arity guard),
  `asRowLimit` and `mergeURL`.
- `go test -race ./cmd/... ./internal/...` passes.
- `go vet` and `gofmt` clean, including `tests/looker`.
- `.ci/lint-docs-tool-page.sh` passes (320 pages).
- Integration tests: tool config plus a `RunToolGetTestByName` manifest
  assertion (verified byte-for-byte against the generated manifest) and a
  `RunToolInvokeParametersTest` invoke case.

### Reviewer notes

- **The integration invoke test is unverified against a live instance.** It
  merges two `system__activity` `look` queries joined on `look.id` and assumes
  `look.id` / `look.title` exist there. I had no Looker instance available, so
  this is the most likely thing to need adjusting.
- `golangci-lint` is not installed in my environment and was not run; CI will be
  the first real check.
- `.ci/lint-docs-sample-filters.sh` fails locally, but it reports every filter in
  every untouched file as invalid — a pre-existing local tooling problem, not
  something this branch causes.

## PR Checklist

- [x] Make sure you reviewed
- [x] Make sure to add `!` if this involve a breaking change (not a breaking change)

🛠️ Fixes #3756 🦕


duwenxin99:
Hi @drstrangelooker, this contributor is adding a new Looker tool. Could you take a look and see if this is what you want? Thank you!

PR #3774, opened 2026-08-04
docs(cli): document the migrate sub command

twishabansal:
## Description

`toolbox migrate` is missing from the CLI reference, which documents `invoke` and `skills-generate`. This adds a `migrate` entry describing its syntax, flags, and the in-place rewrite / `.bak` backup behavior.

Behavior documented was verified by running the command against a nested `tools.yaml`.

## PR Checklist


docs/en/reference/cli.md
```diff
@@ -61,6 +61,28 @@ For more detailed instructions, see [Invoke Tools via CLI](../documentation/conf
...
+```
+
+**Flags:**
+
+- `--config`: (Optional) Path to a configuration file to migrate. Defaults to `tools.yaml` when no other config flag is set.
+- `--configs`: (Optional) Comma-separated list of configuration files to migrate.
+- `--config-folder`: (Optional) Path to a folder whose `.yaml` and `.yml` files should be migrated.
+- `--dry-run`: (Optional) Print the migrated configuration to stdout instead of writing it.
+
+--config, --configs, and --config-folder are mutually exclusive. Each file is rewritten in place, with the original kept alongside it with .bak appended (e.g., tools.yaml.bak); a file that needs no changes is left as is. Comments other than top-level ones are not preserved, so review the result before deleting the backup.
```
```suggestion
`--config`, `--configs`, and `--config-folder` are mutually exclusive. Each file is rewritten in place, with the original kept alongside it with `.bak` appended (e.g., `tools.yaml.bak`); a file that needs no changes is left as is. Comments other than top-level ones are not preserved, so review the result before deleting the backup.
```

docs/en/reference/cli.md
```diff
@@ -61,6 +61,28 @@ For more detailed instructions, see [Invoke Tools via CLI](../documentation/conf
 
 </details>
 
+<details>
+<summary><code>migrate</code></summary>
+
+Rewrites configuration files from the legacy nested format (top-level `sources:`, `tools:`, `toolsets:` maps) into the flat format, where each resource is its own YAML document with a `kind` field.
```
Yuan325, line 67:
```suggestion
Rewrites configuration files from the legacy nested format (top-level `sources:`, `tools:`, `toolsets:` maps) into the flat format, where each resource is its own YAML document with a `kind` field.

This subcommand also converts `toolset` primitive into `group` primitive.
```

Feel free to reword this as you see fit :)

Yuan325 (approved):
Thank you for adding thiss!! :)

PR #3775, opened 2026-08-04
fix(docs): redirect integration sections to source

### LINKED ISSUE #3752: Dead links at `/integrations/*`
### Prerequisites

Navigating to the base URL of an integration, such as https://mcp-toolbox.dev/integrations/alloydb/, currently results in a dead end. Users clicking on the top-level integration name (e.g., AlloyDB PostgreSQL) expect to see an overview or the main configuration page, but instead hit a page with no content.

Expected behavior
The base path for any integration (e.g., /integrations/<integration_name>/) should automatically redirect users to its corresponding /source endpoint (e.g., /integrations/<integration_name>/source/).

### Suggested Solution(s)

Implement an automatic redirect rule in the documentation site's routing configuration so that whenever a user navigates to an integration's root directory, they are seamlessly redirected to the source documentation page.
For example:

https://mcp-toolbox.dev/integrations/alloydb/ ➡️ https://mcp-toolbox.dev/integrations/alloydb/source/
This should be applied globally across all existing integrations to prevent similar dead ends.

### Alternatives Considered

This could be another page with an overview.

### Additional Details

_No response_

lyydsheep:
Fixes #3752.

Integration sections that contain a `source.md` page now redirect their root URL to that source page. This applies globally through the shared Hugo section layout, so new integrations inherit the behavior automatically.

Validation:
- Verified the Hugo template references the child `source` page and emits both a meta refresh and JavaScript fallback.
- `git diff --check` passes.

### DIFF
--- a/.hugo/layouts/docs/section.html
+++ b/.hugo/layouts/docs/section.html
@@ -1,4 +1,10 @@
 {{ define "main" }}
+{{ $sourcePages := where .Pages "File.BaseFileName" "source" }}
+{{ if and (hasPrefix .RelPermalink "/integrations/") (gt (len $sourcePages) 0) }}
+<meta http-equiv="refresh" content="0; url={{ (index $sourcePages 0).RelPermalink }}">
+<script>window.location.replace({{ (index $sourcePages 0).RelPermalink | jsonify }});</script>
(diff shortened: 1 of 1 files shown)


Yuan325:
Hi @lyydsheep thank you for your contribution. I was testing the `/integrations/alloydb/` path but seems like it's still showing a blank page.

PR #3791, opened 2026-08-08
feat(server): add --defer-source-connect and --defer-env-var-parsing

### LINKED ISSUE #2341: Skip connectivity checks on start up

akangsha7:
Adds the two flags that turn on the deferred path the sources gained in #3907.

**Stacked on #3907** — based on `feat/lazy-source-init-all-sources`, so this diff shows only the flag work. It retargets to `main` automatically when #3907 merges.

* `--defer-source-connect` — each source connects on the first tool call that needs it. Toolbox starts with nothing reachable, `tools/list` returns the full catalog, and a connect failure comes back as a tool error the agent can read while the server stays up. Only successful connections are kept, so a source that comes up later works without a restart.
* `--defer-env-var-parsing` — an unset `${VAR}` resolves to its own name instead of failing the parse, with one warning naming every variable replaced. Requires `--defer-source-connect`; alone it would substitute placeholders and then immediately connect with them, so that combination is rejected at startup rather than left to fail confusingly.
* Both default to false. The eager path is unchanged.

Substitution already runs over the whole config document (`ParseConfig`, `cmd/internal/config.go`), so **tool** environment variables are covered along with source ones — the Sep 1 sync had this down as open work and it is not.

## This replaces the SourceResolver design

The previous revision of this PR put laziness in a `SourceResolver` and added 93 lines to `internal/server/primitives/primitives.go`. #3907 moved laziness into the sources instead, so the resolver is deleted and `primitives.go` is untouched — which was the gate from the Aug 19 design review.

Three things the old revision needed turn out to be unnecessary under the new design, each checked rather than assumed:

* **Static manifest fallback.** It existed because the resolver could return *no* source. A source is now always present. All 41 tools overriding `Manifest`/`GetParameters` read config-only accessors (`BigQueryProject()`, `BigQueryAllowedDatasets()`, `GetDefaultProject()`, `AllowedFHIRStores()`) — plain struct-field reads, none reaching `ConnectOnce`. Full schemas render with nothing connected.
* **`isError` handling.** Already correct. `method.go` maps a `util.CategoryAgent` error to `CallToolResult{IsError: true}`, and a connect failure becomes an `AgentError` through `util.ProcessGeneralError`.
* **A separate failure metric.** The existing recorder in each `method.go` already fires on an `Invoke` error with `error.type` attached. The old revision needed its own because resolution failed *before* `Invoke`; now the connect failure happens inside it.

Reload needed no ordering fix either — `handleDynamicReload` swaps everything in one `SetPrimitives` call, and there is no second registry to keep in step.

## Testing

`go build`, `go test -race ./cmd/... ./internal/...`, `golangci-lint run` all pass. Verified end to end against a config naming an unreachable database:

| case | result |
|---|---|
| neither flag | startup fails on the connect, as today |
| `--defer-env-var-parsing` alone | `Error: --defer-env-var-parsing requires --defer-source-connect` |
| `--defer-source-connect` | starts, logs the deferred sources, `tools/list` returns the catalog, `tools/call` returns `isError: true` with the connect error, server stays up |
| both, `${VAR}` unset | starts, one warning naming the substituted variable |
| database brought up mid-run | next call succeeds with no restart |

Towards #3791 🦕

## PR Checklist

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/mcp-toolbox/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

### DIFF
--- a/cmd/internal/config.go
+++ b/cmd/internal/config.go
@@ -51,12 +51,18 @@ type ConfigParser struct {
 	requiredEnvVars []string
 
 	// AllowMissingEnvVars, when true, substitutes the variable name for an unset
-	// required ${VAR} placeholder instead of erroring. Used by offline flows like
-	// skills-generate, where source env vars are needed only to satisfy config
-	// parsing/validation, never to connect. A non-empty placeholder is used (not
-	// "") so required string fields still pass validation. The served path leaves
-	// this false so missing config still fails fast.
+	// required ${VAR} placeholder instead of erroring. Used by flows that never
+	// connect during parsing — skills-generate, and serving with a deferred
+	// connect — where source env vars are needed only to satisfy config
+	// parsing/validation. A non-empty placeholder is used (not "") so required
+	// string fields still pass validation. Eager serving leaves this false so
+	// missing config still fails fast.
 	AllowMissingEnvVars bool
+
+	// MissingEnvVars names the variables that were substituted with a
+	// placeholder, so a caller can report them rather than let a placeholder
+	// stand in for real config silently.
+	MissingEnvVars []string
 }
 
 // parseEnv replaces environment variables ${ENV_NAME} with their values.
@@ -117,6 +123,9 @@ func (p *ConfigParser) parseEnv(input string) (string, error) {
 		} else {
 			if p.AllowMissingEnvVars {
 				p.EnvVars[variableName] = variableName
+				if !slices.Contains(p.MissingEnvVars, variableName) {
+					p.MissingEnvVars = append(p.MissingEnvVars, variableName)
+				}
 				output.WriteString(variableName)
 			} else if !seenMissing[variableName] {
 				seenMissing[variableName] = true
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -359,7 +359,7 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 			reloadedOpts.Configs = slices.Clone(opts.Configs)
 			reloadedOpts.Cfg.Version = versionString
 
-			if _, err := reloadedOpts.LoadConfig(ctx, &internal.ConfigParser{}); err != nil {
+			if _, err := reloadedOpts.LoadConfig(ctx, &internal.ConfigParser{AllowMissingEnvVars: opts.Cfg.DeferEnvVarParsing}); err != nil {
 				logger.WarnContext(ctx, fmt.Sprintf("Error reloading config: %s", err))
 				continue
 			}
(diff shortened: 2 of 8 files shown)


cmd/internal/config.go
```diff
@@ -51,12 +51,18 @@ type ConfigParser struct {
...
+	// connect during parsing — skills-generate, and serving with lazy source
+	// initialization — where source env vars are needed only to satisfy config
+	// parsing/validation. A non-empty placeholder is used (not "") so required
+	// string fields still pass validation. Eager serving leaves this false so
+	// missing config still fails fast.
 	AllowMissingEnvVars bool
+
+	// MissingEnvVars names the required variables that were substituted with a
+	// placeholder because AllowMissingEnvVars was set. Callers report these, so
+	// a placeholder standing in for real config is never silent.
```
helloeve: could you elaborate what do you mean by "Callers report these, so a placeholder standing in for real config is never silent"?

cmd/root.go
```diff
@@ -139,21 +139,36 @@ func handleDynamicReload(ctx context.Context, toolsFile internal.Config, s *serv
...
 		logger.WarnContext(ctx, errMsg.Error())
 		return err
 	}
 
+	// Install the reloaded source configs before swapping the primitives. The
+	// two updates take different locks, so a tool call can land between them;
+	// in this order the worst case is a connect whose result is discarded by
+	// the swap. Reversed, a caller would connect from the pre-reload config and
+	// cache it into the new store, pinning a stale source for the process
+	// lifetime.
```
helloeve: I am having a hard time understanding the doc string here -- what does it mean by "the two updates take different locks"?

Could you clarify what's the behavior we are trying to implement here? (when lazy source init is true, (1) what happens if an edit was made before the source was initialized; (2) what will happen if the source has already been materialized)

internal/group/group.go
```diff
@@ -54,7 +54,6 @@ type Group struct {
 // that the Group package needs.
 // This is implemented to prevent import cycles.
 type GroupManager interface {
-	GetSource(string) (sources.Source, bool)
```
helloeve, line 57: Why are we removing GetSource? Can we have a wrapper of Source as the follow?

```
type LazySource interface {
	Source
	Materialize(ctx context.Context) (Source, error)
}
```

The LazySource would be compatible with Source as it has that embedded, along with an additional Materialize function.

Sample implementation (generate by GenAI):
```
type lazySource struct {
	name        string
	raw         []byte
	materialize func(context.Context, []byte) (Source, error)

	mu       sync.Mutex
	resolved Source
}

func NewLazySource(name string, raw []byte, materialize func(context.Context, []byte) (Source, error)) LazySource {
	return &lazySource{name: name, raw: raw, materialize: materialize}
}

func (l *lazySource) SourceType() string { return "lazy" }

func (l *lazySource) ToConfig() SourceConfig { return nil }

func (l *lazySource) Materialize(ctx context.Context) (Source, error) {
	l.mu.Lock()
	defer l.mu.Unlock()
	if l.resolved != nil {
		return l.resolved, nil
	}
	s, err := l.materialize(ctx, l.raw)
	if err != nil {
		return nil, fmt.Errorf("unable to initialize source %q: %w", l.name, err)
	}
	l.resolved = s
	return s, nil
}

```

With such, for a lazy source, we can construct the materialize function in the following way:
```
materialize := func(mctx context.Context, rawBlock []byte) (sources.Source, error) {
	resolved, err := util.ResolveEnvVars(string(rawBlock))
	if err != nil {
		return nil, err
	}
	srcCfgs, _, _, _, _, _, err := UnmarshalPrimitiveConfig(mctx, []byte(resolved))
	if err != nil {
		return nil, fmt.Errorf("unable to decode source config: %w", err)
	}
	for _, sc := range srcCfgs {
		return sc.Initialize(mctx, tracer)
	}
	return nil, fmt.Errorf("no source found in config block")
}
```

internal/server/mcp/util/telemetry.go
helloeve, line 1: Why a separate telemetry.go over using https://github.com/googleapis/mcp-toolbox/blob/main/internal/telemetry/telemetry.go?

internal/group/group.go
helloeve: With this, it should also help reduce the "changes needed" (e.g., for different function signatures you don't need to introduce the addition source.Getter or Resolvers )

helloeve (changes requested):

internal/group/group.go
akangsha7: Good idea. moving the lazy state into the value GetSource returns makes sense It removes the sources.Getter and *SourceResolver, which should reduce most of these changes.

PR #3805, opened 2026-08-11
feat(groups): add ttlMs and cacheScope customization to config

AlexTalreja:
## Description

This PR allows users to configure time-to-live and cache scope parameters for groups through Toolbox's config file. Previously, these values were hardcoded to `300000ms` and `public`, which Toolbox will still default to if not directly set.


internal/group/group_test.go
```diff
@@ -113,6 +119,44 @@ func TestGroupConfig_Initialize(t *testing.T) {
...
+			},
+			wantTTLMs: intPtr(10000),
+		},
+		{
+			name: "empty ttlMs",
+			config: group.GroupConfig{
+				Name:      "g",
+			},
+			wantTTLMs: intPtr(300000),
+		},
```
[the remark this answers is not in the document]
AlexTalreja: Not sure about this, I assumed that we would want the test to fail if someone changed the Toolbox defaults from our previously agreed upon values (unless it was for a specific reason, and then they could update the tests too). Any thoughts?

cmd/internal/config_test.go
```diff
@@ -1032,6 +1032,26 @@ func TestParseConfigFailure(t *testing.T) {
...
+		{
+			description: "negative ttlMs in group config",
+			in: `
+            kind: group
+            name: my-group
+            description: "Some group"
+            ttlMs: -100
+			`,
+			wantError: "validation for 'TTLMs' failed on the 'gte' tag",
+		},
```
Yuan325: These should probably be covered in `group_test.go` since it's related to the groups unmarshaling~ Please refer to the other comment :)

internal/group/group_test.go
Yuan325, line 1: Can we add another `TestFailParseFromYaml()` similar to source and tools. 

https://github.com/googleapis/mcp-toolbox/blob/ac7708fb0793827f5ee11ba3837373c98ce4caee/internal/sources/alloydbadmin/alloydbadmin_test.go#L80

here we for cacheScope we should insert a value `cacheScope: foo`, and it should throw an error from the validator. Similar to TTLMs as well.

If you'd like we can also include the `TestParseFromYamlGroup()`

Yuan325 (approved):
Thank you!

internal/group/group_test.go
AlexTalreja: Added both the `TestFailParseFromYaml()` and `TestParseFromYamlGroup()`.

PR #3816, opened 2026-08-13
feat(core): dynamic tool annotations for multi-mode tools on read-only sources

anubhav756:
## Summary

Enables multi-mode tools (e.g., SQL execution tools) to dynamically advertise `readOnlyHint: true` and `destructiveHint: false` in MCP tool manifests when bound to a read-only data source, and centralizes tool suppression logic across the server.

## Context & Motivation

* Tools like `postgres-execute-sql` and `mysql-execute-sql` are multi-mode (capable of both reads and writes) and default to `readOnlyHint: false` / `destructiveHint: true`. When connected to a `readOnly: true` database source (where session-level locks prevent modifications), these tools should dynamically report `readOnlyHint: true` and `destructiveHint: false` to MCP clients without requiring separate read-only tool implementations.
* Refactored `ShouldSuppress` from a `BaseTool` method into a package-level function `tools.ShouldSuppress(ctx, t, src)` that operates on the `Tool` interface, allowing suppression to dynamically evaluate `t.GetAnnotations(src)` without requiring concrete tool method overrides.

## Changes

   * Updated `Tool.GetAnnotations(sources.Source) *ToolAnnotations` to make annotations source-aware (matching the design pattern of `GetParameters(sources.Source)` and `Manifest(sources.Source)`).
   * Updated `BaseTool.GetAnnotations(_ sources.Source)` to return static annotations by default.
   * Added `DynamicReadOnlyAnnotations(base *ToolAnnotations) *ToolAnnotations` in `internal/tools/tools.go` to safely copy base annotations and set `ReadOnlyHint: true` / `DestructiveHint: false` while preserving other custom hints (e.g. `idempotentHint`, `openWorldHint`).
   * Updated `postgres-execute-sql` and `mysql-execute-sql` to implement `GetAnnotations(src)` using `tools.DynamicReadOnlyAnnotations(t.BaseTool.GetAnnotations(src))` when `src.IsReadOnly()`.
   * Removed redundant `ShouldSuppress` method overrides from concrete SQL tool structs.
   * Updated all 5 MCP schema version generators (`v20241105`, `v20250326`, `v20250618`, `v20251125`, `v20260728`) in `internal/server/mcp/` to pass `src` into `tool.GetAnnotations(src)`.
   * Added table-driven tests for `tools.DynamicReadOnlyAnnotations` in `internal/tools/tools_test.go` verifying nil handling, default flipping, custom hint preservation, and pointer immutability.
   * Updated `internal/server/server_test.go` and `internal/tools/tools_test.go` to test `tools.ShouldSuppress` with both write and destructive tools using `tools.NewWriteAnnotations()` and `tools.NewDestructiveAnnotations()`.
   * Updated Looker unit test suites to pass `tool.GetAnnotations(nil)`.

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -127,7 +127,7 @@ type Tool interface {
 	GetSourceName() string
 	GetDescription() string
 	GetAuthRequired() []string
-	GetAnnotations() *ToolAnnotations
+	GetAnnotations(sources.Source) *ToolAnnotations
 	Invoke(context.Context, sources.Source, parameters.ParamValues, AccessToken) (any, util.ToolboxError)
 	EmbedParams(context.Context, parameters.ParamValues, PrimitiveManagerI) (parameters.ParamValues, error)
 	Manifest(sources.Source) (Manifest, error)
@@ -139,7 +139,6 @@ type Tool interface {
 	GetParameters(sources.Source) (parameters.Parameters, error)
 	GetScopesRequired() []string
 	ValidateSource(sources.Source) error
-	ShouldSuppress(context.Context, sources.Source) bool
 }
 
 // PrimitiveManagerI defines the minimal view of the primitives.PrimitiveManager
@@ -219,11 +218,11 @@ func NewBaseTool[T ToolMeta](cfg T, annotations *ToolAnnotations, metadata Manif
 	}
 }
 
-func (b BaseTool[T]) GetName() string                  { return b.Cfg.GetName() }
-func (b BaseTool[T]) GetDescription() string           { return b.Cfg.GetDescription() }
-func (b BaseTool[T]) GetAuthRequired() []string        { return b.Cfg.GetAuthRequired() }
-func (b BaseTool[T]) GetScopesRequired() []string      { return b.Cfg.GetScopesRequired() }
-func (b BaseTool[T]) GetAnnotations() *ToolAnnotations { return b.annotations }
+func (b BaseTool[T]) GetName() string                                  { return b.Cfg.GetName() }
+func (b BaseTool[T]) GetDescription() string                           { return b.Cfg.GetDescription() }
+func (b BaseTool[T]) GetAuthRequired() []string                        { return b.Cfg.GetAuthRequired() }
+func (b BaseTool[T]) GetScopesRequired() []string                      { return b.Cfg.GetScopesRequired() }
+func (b BaseTool[T]) GetAnnotations(_ sources.Source) *ToolAnnotations { return b.annotations }
 
 // Manifest returns the precomputed metadata. It and GetParameters stay trivial
 // and never call each other: embedded methods have no virtual dispatch, so a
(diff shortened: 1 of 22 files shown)


internal/tools/tools.go
```diff
@@ -103,6 +103,23 @@ func NewWriteAnnotations() *ToolAnnotations {
...
+	res := NewReadOnlyAnnotations()
+	if base != nil {
+		copied := *base
+		copied.ReadOnlyHint = res.ReadOnlyHint
+		copied.DestructiveHint = res.DestructiveHint
+		return &copied
+	}
+	return res
+}
+
```
Yuan325: let's have this directly within the mysqlexecutesql and postgresexecutesql for now since it seems like only 2 of them wil have this approach.

internal/tools/tools.go
```diff
@@ -265,16 +281,17 @@ func (b BaseTool[T]) EmbedParams(ctx context.Context, paramValues parameters.Par
 // By default, if the source is read-only and the tool's ReadOnlyHint is explicitly false,
 // the tool is suppressed. Unannotated tools (ReadOnlyHint == nil) or read-only tools
 // (ReadOnlyHint == true) are not suppressed.
-func (b BaseTool[T]) ShouldSuppress(ctx context.Context, src sources.Source) bool {
-	if src == nil || !src.IsReadOnly() {
+func ShouldSuppress(ctx context.Context, t Tool, src sources.Source) bool {
```
Yuan325, line 267: why do we remove this from BaseTool?

Yuan325 (changes requested):

internal/tools/tools.go
anubhav756: As mentioned [here](https://github.com/googleapis/mcp-toolbox/pull/3816/changes/BASE..2b6758afca5a38b99e9c27ad11c4dd396c02373c#diff-e607ec0aa5cc1bb54b418caa5ae0f5d10116ed19a4d016748dff76b68a11f8d4R246), (and honestly to my surprise) Go doesn't have virtual method dispatch. So when the `ShouldSuppress()` on `BaseTool`
 calls `GetAnnotations()`, it was only seeing BaseTool's static annotations and missed the concrete tool's override.

Multi-mode tools like `postgres-execute-sql` override `GetAnnotations(src)` (to dynamically report `readOnlyHint: true` when connected to a read-only source).

So I changed that to a standalone function `ShouldSuppress(ctx, t, src)` and it passes the `Tool` interface, which enables interface dispatch to whatever the concrete tool's dynamic implementation is.

internal/tools/tools.go
anubhav756: Yea I just prepared for making it reusable, so I thought let's make it reusable. And it's also one place to put testing. But yea, let's YAGNI it!

anubhav756: Done.

Yuan325 (approved):

shuzhou-gc (approved):

Issue #3639, opened 2026-07-16
Docs: complete field reference per tool kind (worked example incl. agentContextReference), parameter required/default semantics, and stale-host 404s


AndyFooBlah:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/mcp-toolbox/issues)

Related: #1189 (open — standardize the per-tool docs template, user-provided vs agent-provided inputs). This issue is complementary: #1189 is about template consistency; this one is about three concrete coverage gaps that cost us real debugging time. Happy to fold any of these into #1189 if maintainers prefer one tracking issue.

### What are you trying to do that currently feels hard or impossible?

Author a `tools.yaml` from the docs alone, without reading Toolbox source code or guessing.

Three specific gaps, each with the real cost we paid:

**1. Tool-kind pages don't enumerate all accepted fields.** For `cloud-gemini-data-analytics-query`, the `context.datasourceReferences.cloudSqlReference.agentContextReference.contextSetId` field (how you attach an authored context set — the primary accuracy lever for that tool) does not appear in the tool-kind reference. We eventually reconstructed the shape from search-result fragments quoting other pages. A complete worked YAML example per tool kind, showing every field including optional ones, would have made this a 30-second lookup.

**2. Parameter `required` / `default` semantics are underdocumented, and `default: null` is a silent trap.** Current behavior (verified against `internal/util/parameters/parameters.go`):

```go
func CheckParamRequired(required bool, defaultV any) bool {
    return required && defaultV == nil
}
```

- Omitting `required` means **required: true**.
- A non-nil `default` makes the parameter effectively optional even without `required: false`.
- **`default: null` parses to Go nil — indistinguishable from "no default at all" — so the parameter silently stays required.** An author writing `default: null` to mean "optional, defaults to SQL NULL" (a natural reading, and exactly what we did) gets runtime "parameter X is required" errors from agent calls that omit it. The working spelling is `required: false`.

Suggested docs addition: a short "optional parameters" section with the three-row truth table above and an explicit warning that `default: null` has no effect. (A config-time warning would be even better, but the current `Default any` field can't detect explicit-null vs absent without a presence flag — so docs are the practical fix, possibly plus a lint in the validation subcommand proposed in #3543/#3634.)

[...]

PR #3823, opened 2026-08-14
docs(configuration): document parameter required/default semantics and the default: null trap

AndyFooBlah:
## Description

Documents how `required` and `default` interact on tool parameters, and the `default: null` trap.

Parameters are required by default, and supplying a `default` silently makes one optional — the existing table implies both in single cells (`Default to true`, `If provided, required will be false`), but not the interaction, and not the failure mode. Adds an **Optional Parameters** section under *Specifying Parameters* with:

- the two ways to make a parameter optional (`default: <value>` vs `required: false`) and how they differ for an omitted argument;
- the full `required` × `default` matrix;
- a note that the effective requiredness is what reaches the agent in the tool manifest;
- a warning that **`default: null` does not make a parameter optional**.

On that last point: YAML parses `default: null` — and equally `default: ~` or a bare `default:` — to a null value, which `CheckParamRequired` cannot distinguish from the field being absent:

```go
func CheckParamRequired(required bool, defaultV any) bool {
	return required && defaultV == nil
}
```

So the parameter stays required, and an agent call that omits the argument fails at invocation time with `parameter "..." is required`. Writing `default: null` to mean "optional, no value when omitted" is a natural reading that produces a runtime error instead; `required: false` is the working spelling. The section also contrasts `default: ""`, which *is* a real default — only null is ignored.

Docs only; no behavior change.

## Verification

Every row of the matrix was exercised through the real YAML parse path (`yaml.UnmarshalContext` into `parameters.Parameters`, then `Manifest()` and `ParseParams` with the argument omitted) rather than read off the source, including all three null spellings:

| config | `manifest.Required` | omitted-argument call |
|---|---|---|
| `required` omitted, no `default` | `true` | `parameter "p" is required` |
| `required: true`, no `default` | `true` | `parameter "p" is required` |
| `required: false`, no `default` | `false` | passes `nil` |
| `required: true` + `default: fallback` | `false` | uses default |
| `required` omitted + `default: fallback` | `false` | uses default |
| `required` omitted + `default: ""` | `false` | uses `""` |
| `required: true` + `default: null` | `true` | `parameter "p" is required` |
| `required` omitted + `default: null` | `true` | `parameter "p" is required` |
| `required` omitted + `default: ~` | `true` | `parameter "p" is required` |
| `required` omitted + bare `default:` | `true` | `parameter "p" is required` |

The `{{< notice warning >}}` shortcode used here is already used elsewhere in the same page.

## PR Checklist

- [x] Code coverage does not decrease (if any source code was changed) — no source changes

## Issue Reference

Part of #3639 🦕 — this is gap (2) of the three in that issue, which @anubhav756 and @averikitsch both green-lit as a self-contained first step. Gap (1) is with @wangauone for `cloud-gemini-data-analytics-query`, and gap (3) was answered as not feasible, so I've left the issue open rather than auto-closing it.


averikitsch (approved):

docs/en/documentation/configuration/tools/_index.md
```diff
@@ -90,6 +90,76 @@ parameters:
...
+
+Providing a `default` also makes the parameter optional — it overrides
+`required: true` rather than conflicting with it. The full matrix:
+
+| `required` | `default`             | Effective behavior                                              |
+|:-----------|:----------------------|:----------------------------------------------------------------|
+| omitted    | omitted               | **Required.** Calls that omit the argument are rejected.         |
+| `true`     | omitted               | **Required.** Same as above.                                     |
+| `false`    | omitted               | Optional; omitted calls pass no value (`NULL` in SQL).           |
+| `true`     | a value               | **Optional** — the `default` wins over `required: true`.         |
```
averikitsch:
```suggestion
| `true`     | a value               | **Optional**;  the `default` wins over `required: true`.         |
```

PR #3837, opened 2026-08-17
feat: add Toolbox version check on startup

### LINKED ISSUE #2367: Notify user when new version is available
### Prerequisites

Developers are using older versions of toolbox; there is no logs/messages to update to a new version.

### Suggested Solution(s)

Add a lightweight check during startup to compare the current version against the latest release, logging a recommendation to update if a mismatch is found

### Alternatives Considered

_No response_

### Additional Details

It would also be beneficial to guide the users to download a new version.

AlexTalreja:
## Description

Adds a lightweight startup check that compares the running Toolbox version against the latest GitHub release, and logs a warning message if the running version is out of date. The check can be disabled with the `--disable-version-check` flag.

Even when the check is enabled, any network errors, rate limits, or timeouts (3s) will only log debug messages, and will never prevent Toolbox startup.

🛠️ Fixes https://github.com/googleapis/mcp-toolbox/issues/2367


cmd/internal/options.go
```diff
@@ -304,3 +314,57 @@ func (opts *ToolboxOptions) LoadConfig(ctx context.Context, parser *ConfigParser
...
+	var rel struct {
+		TagName string `json:"tag_name"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&rel); err != nil {
+		opts.Logger.DebugContext(ctx, fmt.Sprintf("Unable to query latest Toolbox version: %v (skipping version check)", err))
+		return
+	}
+	latest := rel.TagName
+	current := "v" + opts.VersionNum
```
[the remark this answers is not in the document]
AlexTalreja: `versionNum` is originally embedded from `version.txt` which is updated with a consistent format (without the 'v' prefix).

cmd/internal/options.go
```diff
@@ -304,3 +314,57 @@ func (opts *ToolboxOptions) LoadConfig(ctx context.Context, parser *ConfigParser
...
+
+	if semver.Compare(latest, current) > 0 {
+		opts.Logger.WarnContext(ctx, fmt.Sprintf("A newer version of MCP Toolbox is available: (%s -> %s). Download the latest version from https://github.com/googleapis/mcp-toolbox/releases", current, latest))
```
Yuan325: Should this be Info log instead? @averikitsch thoughts?

cmd/internal/options_test.go
```diff
@@ -148,3 +154,90 @@ func TestLoadConfig(t *testing.T) {
...
+		desc                string
+		currentVersion      string
+		remoteVersion       string
+		httpStatus          int
+		disableVersionCheck bool
+		wantWarnLog         string
+	}{
+		{
+			desc:           "outdated version logs warning",
+			currentVersion: "1.7.0",
```
Yuan325, line 169: Just to confirm, is this the value of our `opts.VersionNum`?

cmd/internal/options_test.go
```diff
@@ -148,3 +154,90 @@ func TestLoadConfig(t *testing.T) {
...
+			remoteVersion:  "v1.8.0",
```
Yuan325, line 170: And this is the value that will be returned? (with the `v`)?

cmd/internal/options_test.go
AlexTalreja: Yes, `opts.VersionNum` is taken from the `version.txt` file which contains the number without the leading "v".

cmd/internal/options_test.go
AlexTalreja: Yes, the remote version number is fetched from the GitHub API [endpoint](https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#get-the-latest-release) which returns the latest release with a leading "v". A leading "v" is also added to `opts.VersionNum` instead of removed from this value because Go's `semver` library defaults to requiring the leading "v" to perform comparisons.

cmd/internal/options.go
Yuan325: Update: let's use Info log for this

cmd/internal/options.go
```diff
@@ -304,3 +314,57 @@ func (opts *ToolboxOptions) LoadConfig(ctx context.Context, parser *ConfigParser
...
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&rel); err != nil {
+		opts.Logger.DebugContext(ctx, fmt.Sprintf("Unable to query latest Toolbox version: %v (skipping version check)", err))
+		return
+	}
+	latest := rel.TagName
+	current := "v" + opts.VersionNum
+
+	if semver.Compare(latest, current) > 0 {
```
averikitsch, line 367: If users are using NPX or UVX is this the right message?

cmd/internal/options.go
```diff
@@ -101,6 +109,8 @@ func (opts *ToolboxOptions) Setup(ctx context.Context) (context.Context, func(co
 	ctx = util.WithLogger(ctx, logger)
 	opts.Logger = logger
 
+	opts.checkVersion(ctx)
+
 	ctx = util.WithIgnoreUnknownTools(ctx, opts.Cfg.IgnoreUnknownTools)
 
 	logger.InfoContext(ctx, fmt.Sprintf("Starting MCP Toolbox for Databases version %s", opts.Cfg.Version))
```
averikitsch, line 114: We may want the check to be logged after this message

averikitsch (commented):
For `toolbox invoke ...` commands will this cause a 3s delay if timeout is used? We may want this to be async or turned off for toolbox invoke commands

cmd/internal/options.go
AlexTalreja: Good point, I can update the message to be package-manager agnostic instead of directly instructing them to download from GitHub. How does this sound? 

`A newer version of MCP Toolbox is available: (%s -> %s). Visit https://github.com/googleapis/mcp-toolbox/releases to view the latest release.`

cmd/internal/options.go
```diff
@@ -105,6 +113,8 @@ func (opts *ToolboxOptions) Setup(ctx context.Context) (context.Context, func(co
...
 
+	go opts.checkVersion(context.WithoutCancel(ctx))
```
Yuan325, line 116: Do we need `context.WithoutCancel`?

Yuan325 (approved):

cmd/internal/options.go
AlexTalreja: If we don't use `context.WithoutCancel`, then in cases where the parent context finishes before the GitHub API returns (like `toolbox invoke`) it may log the debug message `Unable to query latest Toolbox version` which may be misleading since technically the command executed as intended.

PR #3841, opened 2026-08-18
fix(server): make initialize handshake compliant with dual-era MCP spec

### LINKED ISSUE #3834: initialize with protocolVersion 2026-07-28 returns -32601 instead of negotiating (behaviour changed in 1.8.0)
## Summary

An `initialize` request carrying `"protocolVersion": "2026-07-28"` is answered with `-32601 invalid method initialize` on **1.8.0** and **1.9.0**. On **1.7.0** the same request negotiates normally and the server replies with `2025-11-25`.

The graceful downgrade is otherwise intact — `2026-07-28` is the only value that fails. Unknown versions still negotiate fine, which is what makes this look like a gap rather than a policy: the version is known well enough to be routed, and being known is exactly what removes it from the "unknown → downgrade" path.

## Reproduction

Server started from the published `linux/amd64` binary with a minimal BigQuery config, then:

```sh
curl -s -X POST http://localhost:8080/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"<VERSION>","capabilities":{},"clientInfo":{"name":"probe","version":"0"}}}'
```

| requested `protocolVersion` | 1.7.0 | 1.8.0 | 1.9.0 |
|---|---|---|---|
| `2026-07-28` | `result.protocolVersion: 2025-11-25` | `-32601 invalid method initialize` | `-32601 invalid method initialize` |
| `2025-11-25` | `2025-11-25` | — | `2025-11-25` |
| `2025-06-18` | `2025-06-18` | — | `2025-06-18` |
| `2999-01-01` | `2025-11-25` | — | `2025-11-25` |
| `banana` | `2025-11-25` | — | `2025-11-25` |

Same result with and without `--enable-draft-specs`, and with and without an `Mcp-Protocol-Version` header. Same result on `/mcp` and `/mcp/`.

## Likely cause

`ProcessMethod` in `internal/server/mcp/mcp.go` routes `mcputil.VERSION_20260728` to `v20260728.ProcessMethod`. `initialize` appears not to be handled on that path, so it falls through to the generic method-not-found rather than to the unsupported-version branch that `default:` reaches — which is where `NewUnsupportedProtocolVersionError` lives.

## Why the error code matters beyond the message

`toolbox-core`'s protocol fallback keys on `UNSUPPORTED_PROTOCOL_VERSION_ERROR_CODE` (`-32022`) or on the message containing `invalid protocol version` / `unsupported protocol version`. A `-32601` matches neither, so it is not recognisable as a negotiation failure by the client contract the server otherwise implements.

In practice clients do recover — we observe `Protocol fallback required. Switching from 2026-07-28 to 2025-11-25` in production logs against a 1.8.0 server, and `toolbox-core` 1.3.0 loads a full toolset against 1.9.0 without error. So this is not a broken-client report. It is that the rejection is not expressed in the vocabulary the negotiation contract defines, which makes the recovery incidental rather than guaranteed.

## Suggested resolution

Either implement `initialize` on the `v20260728` path, or have an unimplemented-but-routed version answer `-32022` with the supported list, so that the documented fallback contract applies. The second also covers any future version added to the routing table before its handlers land.

## Context

Found while bumping a downstream packaging repository from 1.8.0 to 1.9.0. Happy to test a patch against the same matrix.

anubhav756:
## Overview

Fixes an issue where sending an MCP `initialize` request carrying `"protocolVersion": "2026-07-28"` fails with JSON-RPC error `-32601 Method Not Found` (`"invalid method initialize"`) on `1.8.0` and `1.9.0`, rather than gracefully negotiating down to `2025-11-25` as it did in `1.7.0`.

We do this by making MCP Toolbox `initialize` handshake fully compliant with the official [MCP Dual-Era Server Specification](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#compatibility-matrix).

image

## Background

In `1.8.0` and `1.9.0`, `"2026-07-28"` was added to `SUPPORTED_PROTOCOL_VERSIONS` (https://github.com/googleapis/mcp-toolbox/pull/3699). When this was done, incoming `initialize` requests asking for `2026-07-28` were recognized as a supported version and routed directly to the 2026 handler ([here](https://github.com/googleapis/mcp-toolbox/blob/98c1213269522e4e50f879917bbc95634058da18/internal/server/mcp.go#L840-L845)). Because the 2026 handler is stateless and does not implement `initialize`, the request fell through to an unhandled method error (`-32601`).

> [!NOTE]
> Being included in the general supported versions list also bypassed the legacy downgrade path that unknown versions take.

## Specification

Under the official [MCP 2026-07-28 Spec](https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/2026-07-28/basic/versioning.mdx#L126-L184), we see the spec for the dual-server behaviour specified for this case:
image
image

## Changes

1. Partitioned protocol versions into explicit stateful versions (`2024-11-05` through `2025-11-25`) and stateless versions (`2026-07-28`), while keeping the composite list for server discovery.
2. Cleaned up unused legacy draft constants.

3. Updated the `initialize` message router to check only against stateful versions. If the version is not in the stateful list (such as `2026-07-28` or an unknown future version), it correctly downgrades to `2025-11-25`.

4. Added test coverage verifying that `2026-07-28`, unknown versions, malformed strings, and all legitimate legacy versions negotiate to their exact expected protocol versions under both standard and draft server modes.

Fixes https://github.com/googleapis/mcp-toolbox/issues/3834

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -838,10 +838,10 @@ func processMcpMessage(ctx context.Context, body []byte, s *Server, protocolVers
 
 		var version string
 		v := initReq.Params.ProtocolVersion
-		if slices.Contains(mcputil.GetSupportedVersions(s.enableDraftSpecs), v) {
+		if slices.Contains(mcputil.GetStatefulEraVersions(), v) {
 			version = v
 		} else {
-			version = mcputil.GetLatestSupportedVersion(s.enableDraftSpecs)
+			version = mcputil.GetLatestStatefulVersion()
 		}
 
 		result, err := mcp.ProcessMethod(ctx, version, baseMessage.Id, baseMessage.Method, group.Group{}, nil, body, nil)
--- a/internal/server/mcp/util/util.go
+++ b/internal/server/mcp/util/util.go
@@ -14,6 +14,8 @@
 
 package util
 
+import "slices"
+
 const (
 	VERSION_20241105 = "2024-11-05"
 	VERSION_20250326 = "2025-03-26"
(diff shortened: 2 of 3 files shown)


Yuan325 (changes requested):
This seems to say that if client is using the `initialize` method, we are responding with `2025-11-25` as if we don't support `2026-07-28`. I'm unsure if this is what dual-era means~

Let's wait for MCP conformance before implementing this. That way we are sure that the MCP is in the same page as our server's implementation.

PR #1975, opened 2025-11-18
feat(tools/bigquery): add per tool query label in BigQuery jobs

shobsi:
## Description

This changes adds job label in the BigQuery jobs submitted by the tools. The key of the label would be `genai-toolbox-tool` and the value be the tool kind (e.g. `bigquery-execute-sql`). This will help the customer identify jobs submitted via their MCP toolbox usage. Here is a quick demo: screen/BsWziBoetSexe6T.

### Tested locally

* Ran the toolbox server locally with UI
```shell
go run . --config internal/prebuiltconfigs/tools/bigquery.yaml --ui
```
* Executed a SQL via `execute_sql` tool in the toolbox UI http://127.0.0.1:5000/ui -- screen/7ZDt35Cq58mzhaq
* Verified that the job had `mcp-toolbox-tool: bigquery-execute-sql` label attached -- screen/3jdDFr22tbJKbS2

## PR Checklist

> Thank you for opening a Pull Request! Before submitting your PR, there are a
> few things you can do to make sure it goes smoothly:

- [x] Make sure you reviewed
  [CONTRIBUTING.md](https://github.com/googleapis/genai-toolbox/blob/main/CONTRIBUTING.md)
- [x] Make sure to open an issue as a
  [bug/issue](https://github.com/googleapis/genai-toolbox/issues/new/choose)
  before writing your code! That way we can discuss the change, evaluate
  designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)
- [ ] Make sure to add `!` if this involve a breaking change

🛠️ Fixes #1974

PR #3843, opened 2026-08-18
feat(source/bigquery): attach SQLCommenter attributes as BigQuery job labels

### LINKED ISSUE #3286: feat(tools/bigquery): Wire SQLCommenter for BigQuery source (mirror postgres/mysql pattern)
## What feature would you like to see?

Wire the SQLCommenter feature (added in v1.3.0 via #3064 closes #3061) into the **BigQuery** source so that `--sql-commenter` produces SQL comments on BigQuery queries — matching the pattern already implemented for postgres and mysql sources.

## Background

#3064 / v1.3.0 introduced the SQLCommenter infrastructure:

- `--sql-commenter` CLI flag (`cmd/internal/flags.go`)
- `internal/sources/sqlcommenter/sqlcommenter.go` with `AppendComment(ctx, statement, dbSystemName)`
- Context propagation via `util.WithSQLCommenterEnabled` and `extractMeta()` in `internal/server/mcp.go`

The feature was wired into:

- `internal/sources/postgres/postgres.go:106` → `statement = sqlcommenter.AppendComment(ctx, statement, SourceType)`
- `internal/sources/mysql/mysql.go:110` → same pattern

**BigQuery source was not updated**: `internal/sources/bigquery/bigquery.go RunSQL` calls `bqClient.Query(statement)` directly without invoking `AppendComment`. Verified in v1.3.0 release tag **and** in current `main` HEAD.

Issue #3061 was explicitly scoped to *\"cloudSQL databases\"*, so BigQuery was likely never in-scope for the original PR — this is a follow-up feature request to extend the SQLCommenter coverage to BigQuery, where the same observability problem #3061 originally described applies just as strongly.

## Current behavior (empirically validated)

We deployed Toolbox v1.3.0 to Cloud Run with `--sql-commenter --telemetry-gcp` args. After triggering a BigQuery query via the `execute_sql` tool:

- The query reached BQ correctly (177 chars, matching the SQL the MCP client sent — no commenter overhead appended).
- `INFORMATION_SCHEMA.JOBS_BY_PROJECT.query` shows the raw SQL only, with no SQLCommenter comments.
- The `--sql-commenter` flag is therefore a **no-op** for BigQuery in v1.3.0.

## Proposed fix

Mirror the postgres/mysql pattern in `internal/sources/bigquery/bigquery.go`:

1. Add import:

```go
\"github.com/googleapis/mcp-toolbox/internal/sources/sqlcommenter\"
```

2. In `RunSQL`, immediately before `bqClient.Query(statement)`:

```go
statement = sqlcommenter.AppendComment(ctx, statement, SourceType)
```

This is the same 2-line change applied for `postgres` and `mysql`. The `AppendComment` function is already context-aware (no-op if `SQLCommenterEnabledFromContext(ctx)` returns false), so the change is backwards-compatible by construction.

## Use case / motivation

For agent observability, SQLCommenter-emitted comments in BigQuery's `INFORMATION_SCHEMA.JOBS_BY_PROJECT.query` would enable:

- **Cost attribution** per agent tool invocation / trace ID (BQ pricing is per-query-byte; correlating `total_bytes_billed` to `tool.name` and `traceparent` is otherwise impossible without injecting comments client-side).
- **Audit trail** mapping each BQ query to the MCP client / agent context that triggered it.
- **Cross-system correlation** via W3C `traceparent` propagation — distributed tracing visible in Cloud Trace alongside the BQ jobs that resulted from each tool call.

Today, with the BQ source not wired, the observability chain breaks at the BQ tool boundary — the same problem #3061 originally described, applied to BigQuery.

## Reproduction

1. Deploy Toolbox v1.3.0 with `--sql-commenter --telemetry-gcp` Cloud Run args (or equivalent).
2. Configure a BigQuery source in the Toolbox config (`bigquery-tools.yaml`).
3. Execute any SQL query via the `execute_sql` tool from an MCP client.
4. Inspect `region-eu.INFORMATION_SCHEMA.JOBS_BY_PROJECT.query` (or the equivalent for your BQ region) for the executed job — observe no SQLCommenter comment block appended.
5. Compare to the same flag against a postgres or mysql source, where `AppendComment` is called and the comment block appears in the database query logs.

## Environment

- Toolbox version: v1.3.0 (verified against `main` HEAD as of 2026-05-25)
- BigQuery client: `cloud.google.com/go/bigquery` (as bundled in Toolbox v1.3.0)
- Deployment: Cloud Run with `--telemetry-gcp` and `--sql-commenter`, `cloudtrace.googleapis.com` API enabled

spark2ignite:
## Description

Enabling `--sql-commenter` currently has no effect on the BigQuery source:
the commenter is only wired into postgres/mysql-family sources, and BigQuery
would not retain SQL-text comments in job metadata anyway. As noted on the
issue, today exactly one attribute reaches the job — the automatic
`mcp-toolbox-tool` label added in v1.5.0 (#1975), which carries the tool
*type* (e.g. `bigquery-execute-sql`). The seven remaining SQLCommenter
attributes — per-user, per-agent, per-model attribution and the trace
linkage — never reach the job. This PR adds them. The SQLCommenter
`tool.name` attribute (the MCP tool *name*, e.g. `execute_sql`) is distinct
from the existing type label; both are kept, and explicit tool-supplied
labels always win on key collisions.

Following the direction discussed in the linked issue, this PR attaches the
SQLCommenter attributes as **native job labels** instead of prepending a
comment, so they surface first-class in `INFORMATION_SCHEMA.JOBS`, audit
logs, and billing exports — no query text parsing needed to recover them.

- New `sqlcommenter.Labels(...)` returns the exact attribute set
  `PrependComment` uses (`tool.name`, `client`, `client.model`,
  `client.user.id`, `client.agent.id`, `traceparent`, `server`,
  `db.system.name`), sanitized to BigQuery label constraints: lowercase
  `[a-z0-9_-]`, max 63 chars, keys start with a letter, dots mapped to
  underscores (`tool.name` → `tool_name`).
- Labels are merged in `Source.RunSQL`, so both `bigquery-execute-sql` and
  `bigquery-sql` are covered with no per-tool changes; ML/AI tools that don't
  flow through `RunSQL` are unaffected, matching the scope note on the issue.
  Explicit tool-supplied labels (e.g. `mcp-toolbox-tool`) win on collisions.
- Gated identically to other sources: global `--sql-commenter` flag plus an
  optional per-source `sqlCommenter` override in tools.yaml, default off.
- Docs: new "BigQuery: Job Labels" section in the SQL Commenter monitoring
  page (with an example `INFORMATION_SCHEMA.JOBS` attribution query) and a
  `sqlCommenter` row in the BigQuery source reference.

Tests: table-driven unit tests for key/value sanitization (uppercase,
illegal characters, truncation, leading non-letter keys), enable/override
precedence mirroring the existing `PrependComment` cases, label merge
precedence, and YAML config parsing. `go test -race ./cmd/... ./internal/...`
and `golangci-lint run` pass.

Also verified end-to-end against a live BigQuery project: with
`sqlCommenter: true`, a query issued through the MCP protocol (JSON-RPC
`tools/call` with `_meta["dev.mcp-toolbox/telemetry"]` set) produced a job
carrying all eight attributes as labels alongside the existing
`mcp-toolbox-tool`:

```json
"labels": {
  "client": "select-e2e-client_0_0_1",
  "client_agent_id": "agent-e2e-42",
  "client_model": "test-model-1",
  "client_user_id": "vadim-e2e",
  "db_system_name": "bigquery",
  "mcp-toolbox-tool": "bigquery-execute-sql",
  "server": "genai-toolbox_1_9_0_dev_darwin_arm64",
  "tool_name": "execute_sql",
  "traceparent": "00-f58710a2aea7bf757cf9292b026a2bce-de6a2f6214d3f085-01"
}
```

With the flag off, the job carries only `mcp-toolbox-tool` — byte-identical
to current behavior. The labels are queryable in
`INFORMATION_SCHEMA.JOBS_BY_PROJECT` (grouping bytes billed by `tool_name`
and `client` works as expected).

## PR Checklist

- [x] Make sure you reviewed
- [x] Ensure you have manually reviewed the entire diff before requesting a
  review

🛠️ Fixes #3286


internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,93 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+		}
+	}
+	return out
+}
+
+// sanitizeLabelValue sanitizes s for use as a BigQuery label value. Values
+// share the key character set but may be empty.
+func sanitizeLabelValue(s string) string {
+	return sanitizeLabelPart(s)
+}
```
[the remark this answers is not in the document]
spark2ignite: Verified against live BigQuery rather than docs alone: the leading-character requirement applies to label keys only: values beginning with _ or - are accepted and stored verbatim (bq query --label 'sanitize_check:_leading_underscore' 'SELECT 1' succeeds, and bq show -j returns the value unchanged; same for -leading-dash). 

Keys are already guarded: sanitizeLabelKey prefixes x when the first character isn't a lowercase letter. Added test cases asserting the verified behavior plus a comment on sanitizeLabelValue documenting the key/value asymmetry in 1db1e12.

internal/sources/sqlcommenter/sqlcommenter_test.go
```diff
@@ -243,3 +244,141 @@ func TestPrependComment_EmptyTelemetryAttributes(t *testing.T) {
...
+		in   string
+		want string
+	}{
+		{"plain", "search_user", "search_user"},
+		{"slash and dots replaced", "genai-toolbox/1.1.0", "genai-toolbox_1_1_0"},
+		{"uppercase lowered", "Test-Client", "test-client"},
+		{"spaces replaced", "a b c", "a_b_c"},
+		{"empty allowed", "", ""},
+		{"leading digit kept", "00-abc", "00-abc"},
+		{"truncated to 63", strings.Repeat("v", 100), strings.Repeat("v", 63)},
```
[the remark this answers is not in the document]
spark2ignite: Added in 1db1e12 - asserting the verified live-BigQuery behavior: leading _ and - in values are preserved (no prefixing; the leading-character rule applies to keys only), and a leading . sanitizes to _.

shobsi (approved):
Nit: SQL commenter will add ~5-8 labels to the job. Since BigQuery limits jobs to 64 labels, a user with ~55+ explicit labels might start seeing 400 Bad Request errors. Letting the BigQuery API naturally reject this is probably the right approach rather than silently dropping telemetry, but just wanted to call it out so we're aware of the edge case!

docs/en/documentation/monitoring/sql_commenter.md
```diff
@@ -57,6 +58,36 @@ block. For example:
...
+SELECT
+  (SELECT value FROM UNNEST(labels) WHERE key = 'tool_name') AS tool_name,
+  (SELECT value FROM UNNEST(labels) WHERE key = 'client') AS client,
+  SUM(total_bytes_billed) AS total_bytes_billed
+FROM `region-us`.INFORMATION_SCHEMA.JOBS
+WHERE EXISTS (SELECT 1 FROM UNNEST(labels) WHERE key = 'tool_name')
+GROUP BY tool_name, client
+ORDER BY total_bytes_billed DESC;
+```
+
```
Yuan325: Can we please move this to `docs/en/integrations/bigquery/source.md` after the `## Reference` header. Something like this: 

```
## Advanced Usage

### SQL Commenter: Job Labels

BigQuery does not retain ......
```

We'll need to update the links as well~

internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,95 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+	return out
+}
+
+// sanitizeLabelValue sanitizes s for use as a BigQuery label value. Values
+// share the key character set but may be empty, and unlike keys they have no
+// leading-character requirement: BigQuery accepts values beginning with an
+// underscore, dash, or digit.
+func sanitizeLabelValue(s string) string {
+	return sanitizeLabelPart(s)
+}
```
Yuan325: Seems like this function is just calling another function, please remove it if redundant.

internal/sources/bigquery/bigquery.go
```diff
@@ -642,6 +648,25 @@ func (s *Source) RunSQL(ctx context.Context, bqClient *bigqueryapi.Client, state
...
+	merged := make(map[string]string, len(explicit)+len(commenter))
+	for k, v := range commenter {
+		merged[k] = v
+	}
+	for k, v := range explicit {
+		merged[k] = v
+	}
+	return merged
+}
+
```
Yuan325: We should move this merging logic within `AppendLabels()` instead. To keep it consistent with the existing `PrependComment()` function.

internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,95 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+// The attribute set matches PrependComment exactly. Keys and values are
+// sanitized to satisfy BigQuery label constraints: only lowercase letters,
+// digits, underscores, and dashes, at most 63 characters each, and keys must
+// begin with a letter. Attribute names map punctuation to underscores, e.g.
+// tool.name becomes tool_name.
+//
+// sourceOverride behaves as in PrependComment: when non-nil it takes priority
+// over the global sql-commenter flag from the context. Returns nil when the
+// commenter is disabled or no attributes are available.
+func Labels(ctx context.Context, dbSystemName string, sourceOverride *bool) map[string]string {
```
Yuan325:
```suggestion
func Labels(ctx context.Context, labels []map[string]string, dbSystemName string, sourceOverride *bool) map[string]string {
```

internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,95 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+	if len(pairs) == 0 {
+		return nil
+	}
+
+	labels := make(map[string]string, len(pairs))
+	for k, v := range pairs {
+		key := sanitizeLabelKey(k)
+		if key == "" {
+			continue
+		}
```
Yuan325, line 171: Seems like we are deferring label's key and value validation to the API in our existing implementation. Are there any benefits of sanitizing it here?

If we are deferring it to the API, we can just append the labels and return the map back to the source.


Similarly, if we exceed the max label, will the BigQuery API return API Error? If yes, we should leave error handling to the API.

Yuan325 (dismissed):
Thank you, please let me know if you need more clarifications in any of the comments :)

docs/en/documentation/monitoring/sql_commenter.md
spark2ignite: Done in 2db7ac4 — moved to `docs/en/integrations/bigquery/source.md` under `## Advanced Usage` / `### SQL Commenter: Job Labels`, and updated the links (the supported-sources entry in `sql_commenter.md` now points there).

internal/sources/sqlcommenter/sqlcommenter.go
spark2ignite: Removed in 2db7ac4 — values now use `sanitizeLabelPart` directly (that is the complete rule for values, which may be empty and have no leading-character requirement); keys keep `sanitizeLabelKey` on top for the leading-letter guard.

internal/sources/bigquery/bigquery.go
spark2ignite: Done in 2db7ac4 — the merge now lives in `sqlcommenter.AppendLabels`, which takes the job's existing labels and returns the merged map (existing labels win on key collisions), mirroring `PrependComment`'s take-and-return shape. `mergeJobLabels` is gone from the BigQuery source and the call site is a single line.

internal/sources/sqlcommenter/sqlcommenter.go
spark2ignite: Adopted this shape in 2db7ac4, with one tweak: `labels` is a single `map[string]string` rather than `[]map[string]string`, since `RunSQL` carries exactly one label map per job. Happy to switch to a slice/variadic if you have a use case in mind for multiple maps.

internal/sources/sqlcommenter/sqlcommenter.go
spark2ignite: Character sanitization is load-bearing here, because unlike user-supplied labels these values are machine-derived and routinely invalid as-is: `client` is `name/version`, `server` is the Toolbox user agent (`genai-toolbox/1.9.0`), `client.model` is e.g. `gemini-2.5-flash`, and attribute keys like `tool.name` contain dots. Verified against live BigQuery: `bq query --label 'k:has/slash' ...` fails with `Label value "has/slash" has invalid characters.` — so without sanitization, enabling the flag would fail essentially every job and the feature would be unusable.

The max-labels limit is a different story and I agree it belongs to the API: verified that a job with 65 labels is rejected with `Too many labels on this resource. The maximum number of labels is 64.` — a clear error rather than silent dropping. 2db7ac4 leaves that to the API (no counting or truncation in Toolbox) and notes it in the docs.

docs/en/integrations/bigquery/source.md
```diff
@@ -157,3 +157,39 @@ useClientOAuth: true
...
+
+## Advanced Usage
+
+### SQL Commenter: Job Labels
+
+BigQuery does not retain SQL-text comments in its logs, so instead of
+prepending a comment, [SQL Commenter](../../documentation/monitoring/sql_commenter.md)
+attaches the same attributes as native
+[job labels](https://cloud.google.com/bigquery/docs/adding-labels#job-label)
+on every query job executed by the `bigquery-execute-sql` and `bigquery-sql`
```
anubhav756: nit: Should we also mention `bigquery-forecast`?

docs/en/integrations/bigquery/source.md
```diff
@@ -157,3 +157,39 @@ useClientOAuth: true
...
+tools. Labels appear in `INFORMATION_SCHEMA.JOBS`, the Jobs API, audit logs,
```
anubhav756: nit: Should we apply `AppendLabels` in `bigquery-analyze-contribution` as well (here or in a follow-up PR)?

internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,101 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+		return labels
+	}
+
+	pairs := collectAttributes(ctx, dbSystemName)
+	if len(pairs) == 0 {
+		return labels
+	}
+
+	merged := make(map[string]string, len(labels)+len(pairs))
+	for k, v := range pairs {
```
anubhav756: Assuming this iterates in randomized order, if any two attributes in `pairs` sanitize to the same label key (e.g. `foo.bar` and `foo_bar`), the one that survives in merged could be random across runs.

Just like we sorted keys in `PrependComment`, should we sort the keys of `pairs` before iterating, or sort after sanitizing to make label assignment deterministic?

internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,101 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+		switch {
+		case r >= 'a' && r <= 'z', r >= '0' && r <= '9', r == '_', r == '-':
+			b.WriteRune(r)
+		default:
+			b.WriteByte('_')
+		}
+	}
+	out := b.String()
+	if len(out) > maxLabelLength {
+		out = out[:maxLabelLength]
```
anubhav756, line 222: If two IDs share a long prefix and only differ at the end, they will collide and overwrite each other.

Should we either log a debug warning when a key or value is truncated, or document that keys and values exceeding 63 characters are silently truncated?

internal/sources/sqlcommenter/sqlcommenter.go
```diff
@@ -125,3 +125,101 @@ func collectAttributes(ctx context.Context, dbSystemName string) map[string]stri
...
+// begin with a lowercase letter, so a leading non-letter is prefixed with
+// "x". An empty input yields an empty result, which callers should treat
+// as an unusable key.
+func sanitizeLabelKey(s string) string {
+	out := sanitizeLabelPart(s)
+	if out == "" {
+		return ""
+	}
+	if out[0] < 'a' || out[0] > 'z' {
+		out = "x" + out
```
anubhav756, line 239: nit (rare case): If another key was already `x0key`, this can silently collide with `0key`. Does it make sense to make the collision handling same as above here as well (log warning or document)?

internal/sources/bigquery/bigquery_test.go
```diff
@@ -566,6 +586,10 @@ func TestNormalizeValue(t *testing.T) {
 	}
 }
 
+func boolPtr(b bool) *bool {
+	return &b
+}
```
anubhav756: Isn't this function already defined elsewhere? Can we make it DRY by moving to utils and reusing?

internal/sources/sqlcommenter/sqlcommenter_test.go
```diff
@@ -243,3 +244,172 @@ func TestPrependComment_EmptyTelemetryAttributes(t *testing.T) {
...
+func TestAppendLabels_EmptyContext(t *testing.T) {
+	ctx := sqlCommenterCtx()
+
+	// No attributes available at all — nil map, not an empty one
+	if labels := AppendLabels(ctx, nil, "", nil); labels != nil {
+		t.Errorf("expected nil labels for empty context, got: %v", labels)
+	}
+}
+
+func TestAppendLabels_FullAttributes(t *testing.T) {
```
anubhav756, line 304: nit: should we also test OpenTelemetry span `traceparent`, because it has a hyphenated structure (`00-<trace_id>-<span_id>-<flags>`)?

internal/sources/sqlcommenter/sqlcommenter_test.go
```diff
@@ -243,3 +244,172 @@ func TestPrependComment_EmptyTelemetryAttributes(t *testing.T) {
...
+	})
+	ctx = util.WithTelemetryAttributes(ctx, &util.TelemetryAttributes{
+		ClientName:    "toolbox-langchain-python",
+		ClientVersion: "v0.1.0",
+		ClientModel:   "gemini-2.5-flash",
+		ClientUserID:  "user-123",
+		ClientAgentID: "agent-456",
+	})
+
+	labels := AppendLabels(ctx, nil, "bigquery", nil)
```
anubhav756, line 318: Should we also test for empty value attribute?

internal/sources/sqlcommenter/sqlcommenter_test.go
```diff
@@ -243,3 +244,172 @@ func TestPrependComment_EmptyTelemetryAttributes(t *testing.T) {
...
+}
+
+// TestAppendLabels_MergePrecedence verifies that labels already present on
+// the job are returned unchanged when the commenter is disabled and always
+// win over commenter attributes on key collisions.
+func TestAppendLabels_MergePrecedence(t *testing.T) {
+	explicit := map[string]string{"mcp-toolbox-tool": "bigquery-execute-sql", "tool_name": "explicit-value"}
+
+	// Disabled: input map returned unchanged.
+	if got := AppendLabels(context.Background(), explicit, "bigquery", nil); !reflect.DeepEqual(got, explicit) {
```
anubhav756, line 399: Should we also add a test to exercise 63 character truncation?

anubhav756 (approved):

docs/en/integrations/bigquery/source.md
spark2ignite: Good catch — checked the tree and `bigquery-forecast` actually executes its forecast query through `source.RunSQL`, so it already receives the labels with no code change. Docs updated in 18b8861 to list all four SQL-executing tools.

docs/en/integrations/bigquery/source.md
spark2ignite: `bigquery-analyze-contribution` gets labels on its main path already — both `getInsightsSQL` and the insights query go through `source.RunSQL`. The one exception is its internal `CREATE TEMP MODEL` statement, which uses a direct `bqClient.Query` with only the `mcp-toolbox-tool` label; the tool would need access to the source's commenter setting to cover it, so I'd propose that as a small follow-up PR to keep this one scoped. Documented the exception in the source docs (18b8861).

internal/sources/sqlcommenter/sqlcommenter.go
spark2ignite: Done in 18b8861 — attributes are now applied in sorted name order (mirroring `PrependComment`), so any collision resolves deterministically. Worth noting the current fixed attribute set can't collide (all eight sanitized keys are distinct); the sort future-proofs it.

internal/sources/sqlcommenter/sqlcommenter.go
spark2ignite: Documented in 18b8861 — both on `AppendLabels` and in the source docs: values over 63 characters are silently truncated, so values sharing a long prefix can become indistinguishable, and sorted-order application makes the outcome deterministic. Kept it doc-only to match `PrependComment`'s no-logging style, but happy to add a debug log if you'd prefer.

internal/sources/sqlcommenter/sqlcommenter.go
spark2ignite: Documented in 18b8861 on `sanitizeLabelKey` — the `x` prefix can in principle collide with an existing `x`-prefixed key, and the sorted-order application above makes that deterministic too. Same doc-only treatment as the truncation case.

internal/sources/bigquery/bigquery_test.go
spark2ignite: It wasn't in a shared package (the other definitions are function-literal locals in a couple of test files), so moved it to `testutils.BoolPtr` in 18b8861 and reused it across this file.

internal/sources/sqlcommenter/sqlcommenter_test.go
spark2ignite: Added in 18b8861 — `TestAppendLabels_Traceparent` builds a real span context (fixed trace/span IDs, sampled flag) and asserts the label comes out as `00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01`, byte-identical: lowercase hex and hyphens are all label-legal.

internal/sources/sqlcommenter/sqlcommenter_test.go
spark2ignite: Added in 18b8861 — `TestAppendLabels_EmptyAttributeValues` asserts that empty telemetry attributes are omitted entirely (only `db_system_name` remains) rather than emitted as empty labels.

internal/sources/sqlcommenter/sqlcommenter_test.go
spark2ignite: Added in 18b8861 — `TestAppendLabels_ValueTruncation` exercises it end-to-end: a 100-character user ID lands as a 63-character label value.

PR #3851, opened 2026-08-20
feat(bigquery): add native read-only session locking

anubhav756:
## Overview
Add support for BigQuery in Toolbox's Read-Only architecture.

When BigQuery is configured with `writeMode: "blocked"`, the source is recognized as read-only (`IsReadOnly() == true`). `bigquery-execute-sql` remains available to AI agents for `SELECT` queries as a single polymorphic tool, dynamically updating its MCP annotations to `readOnlyHint: true` and `destructiveHint: false`. Any dedicated mutating tools without read-only annotations are automatically suppressed.

## Changes

   - Updated `BigQuerySource.IsReadOnly()` to return `true` when `WriteMode == "blocked"`.
   - Added unit tests covering all `writeMode` states (`blocked`, `allowed`, `protected`, empty default).
   - Implemented `ShouldSuppress(ctx, source) bool { return false }` so `execute_sql` remains exposed to agents for `SELECT` queries on read-only sources.
   - Implemented `GetAnnotations(src)` to dynamically return `readOnlyHint: true` and `destructiveHint: false` when connected to a `writeMode: "blocked"` source, preserving custom hints.
   - Added comprehensive unit tests (`TestGetAnnotations`, `TestShouldSuppress`).
   - Exposed `writeMode: ${BIGQUERY_WRITE_MODE:allowed}` on `bigquery-source` so prebuilt images can be launched in read-only mode via environment variables (`BIGQUERY_WRITE_MODE=blocked ./toolbox --prebuilt bigquery`).
   - Added `TestBigQuery_ReadOnlyVulnerabilityBlock` verifying that if a custom tool is falsely annotated with `readOnlyHint: true` on a `writeMode: "blocked"` source, BigQuery dry-run validation catches and rejects mutating SQL statements (like `CREATE TABLE`) before execution.
   - Updated the `writeMode` reference table to document coordination with MCP read-only annotations and tool suppression.


internal/prebuiltconfigs/tools/bigquery.yaml
```diff
@@ -17,6 +17,8 @@ name: bigquery-source
 type: bigquery
 project: ${BIGQUERY_PROJECT}
 location: ${BIGQUERY_LOCATION:}
+readOnly: ${BIGQUERY_READONLY:false}
```
shobsi: Consider `readOnly: ${BIGQUERY_READONLY:}` to let omitted environment variable result in a `nil` pointer, which safely bypasses the strict validation and let any `writeMode` dictate the behavior

shobsi (changes requested):

docs/en/integrations/bigquery/tools/bigquery-execute-sql.md
```diff
@@ -16,17 +16,19 @@ A `bigquery-execute-sql` tool executes a SQL statement against BigQuery.
...
-
-- **`allowed` (default):** All SQL statements are permitted.
-- **`blocked`:** Only `SELECT` statements are allowed. Any other type of
-  statement (e.g., `INSERT`, `UPDATE`, `CREATE`) will be rejected.
-- **`protected`:** This mode enables session-based execution. `SELECT`
-  statements can be used on all tables, while write operations are allowed only
-  for the session's temporary dataset (e.g., `CREATE TEMP TABLE ...`). This
-  prevents modifications to permanent datasets while allowing stateful,
-  multi-step operations within a secure session.
+The behavior of this tool is influenced by the `readOnly` and `writeMode`
```
shobsi: The below list made sense just for the `writeMode` config, but with the two configs, a table/matrix (like in the design doc) would read better

shobsi: The behavior of this tool is influenced by the combination of the `readOnly` and `writeMode` settings on its `bigquery` source:

| `readOnly` | `writeMode` | Tool Behavior | MCP Tool Annotations |
| :--- | :--- | :--- | :--- |
| `false` (default) | `allowed` (default) | All SQL statements are permitted. | (Default) |
| `true` | `blocked` (default if readOnly is true) | Only `SELECT` statements are allowed. Any other type of statement (e.g., `INSERT`, `UPDATE`, `CREATE`) will be rejected. | `readOnlyHint: true` |
| `true` | `protected` | Enables session-based execution. `SELECT` statements can be used on all tables, while write operations are allowed only for the session's temporary dataset (e.g., `CREATE TEMP TABLE ...`). | `readOnlyHint: true` |

*(Note: Conflicting configurations, such as `readOnly: true` combined with `writeMode: allowed`, are rejected during server startup).*

internal/sources/bigquery/bigquery.go
```diff
@@ -131,7 +132,24 @@ func (r Config) SourceConfigType() string {
...
+		} else {
+			r.WriteMode = WriteModeAllowed
+		}
+	}
+
+	if r.WriteMode != WriteModeAllowed && r.WriteMode != WriteModeBlocked && r.WriteMode != WriteModeProtected {
+		return nil, fmt.Errorf("invalid writeMode %q: must be one of %q, %q, or %q", r.WriteMode, WriteModeAllowed, WriteModeProtected, WriteModeBlocked)
+	}
+
+	if r.ReadOnly != nil {
```
shobsi, line 145: [nit] could this be simplified to:

```go
	if r.ReadOnly != nil {
		// A writeMode is considered a read-only mode if it is Blocked or Protected.
		isReadOnlyMode := (r.WriteMode == WriteModeBlocked || r.WriteMode == WriteModeProtected)
		
		// The declared readOnly boolean must match the writeMode's behavior.
		if *r.ReadOnly != isReadOnlyMode {
			return nil, fmt.Errorf("conflicting source configuration: readOnly is %v, but writeMode is %q", *r.ReadOnly, r.WriteMode)
		}
	}
```

for fewer branches, more readable?

internal/prebuiltconfigs/tools/bigquery.yaml
anubhav756: Thanks for catching that! Removed this default value :)

docs/en/integrations/bigquery/tools/bigquery-execute-sql.md
anubhav756: Thanks for the great suggestion! Updated

internal/sources/bigquery/bigquery.go
anubhav756: Absolutely! Thanks for the suggestion to simplify! Updated :)

shobsi (approved):

PR #3863, opened 2026-08-24
docs(skills): add reproduce-bug skill for mcp-toolbox

twishabansal:
## Description

Adds an Agent Skill for deciding whether a reported bug is real, and cross-links it from `triage-issues`.

Reading ~46 closed `type: bug` issues showed that about half of reports are not bugs in server code: they are misconfiguration, client-side problems, or something already fixed in a release the reporter has not upgraded to. So the skill checks the recurring false-alarm shapes first, then climbs a five-rung reproduction ladder only as far as needed, ending in one of seven verdicts. Most reports settle on the SQLite rung with no credentials, since config parsing, type validation, and parameter binding all sit upstream of the database connection.

Error text is traced with `grep` and `git log -S` rather than hardcoded, and pointers name symbols instead of line numbers, so the skill degrades less as wording and releases drift.


skills/maintainer/reproduce-bug/references/repro-env.md
```diff
@@ -0,0 +1,237 @@
...
+    kind: mysql               type: mysql
+```
+
+So when a report says `kind: mysql` under a `tools:` block, the fault is the
+*value* (`mysql` is a source type, the tool type is `mysql-sql`), not the field
+name. Do not tell a reporter to rename `kind` to `type` as the fix.
+
+Convert a nested config to flat with `go run . migrate --config tools.yaml`.
+
+## Zero-infrastructure repro: SQLite
```
averikitsch, line 74: Why just the SQLlite example?

twishabansal: SQLite is the only source that needs no credentials and setup. Therefore, it is easy to reproduce bugs using it.

Every other engine means a container, which is rung 3: covered by "What genuinely needs real infrastructure" at the end of the file.

averikitsch (approved):

PR #3873, opened 2026-08-26
docs: fix dead links to integration index pages

### LINKED ISSUE #3752: Dead links at `/integrations/*`

Ashfaqbs:
Every `/integrations/<name>/_index.md` page (47 of them) was missing the `type: docs` front matter key that its sibling pages already set explicitly: `source.md`, the tools pages, and the parent `/integrations/_index.md`.

Because of that, Hugo inferred content type `integrations` for these pages instead of `docs`, so they fell back to the Docsy theme's default section layout instead of this repo's own `layouts/docs/section.html`. That local override is what calls the `section-index` partial to auto-list a section's child pages (Source, Tools, Samples). Without it, visiting an integration's root — e.g. `/integrations/alloydb/` — rendered only the page chrome (title, nav, footer) with none of its own child links: a dead end, exactly as described in #3752.

3 integrations (`arcadedb`, `dataform`, `utility`) already had `type: docs` set and render correctly today — comparing those against the other 47 is what confirmed the root cause and the fix.

## Fix

Add `type: docs` to the front matter of the 47 affected `_index.md` files, matching the convention already used by their sibling pages. No template or layout changes — this makes the existing, already-imported Docsy `section-index` partial pick these pages up the same way it already does for the 3 that worked.

Fixes #3752

## Verification

Fetched the live site before/after reasoning about the fix:
- `https://mcp-toolbox.dev/integrations/alloydb/` (missing `type: docs`) — renders only page chrome, no links to Source/Tools.
- `https://mcp-toolbox.dev/integrations/arcadedb/` (already has `type: docs`) — renders a proper index with links to "ArcadeDB Source" and "Tools" (with individual tool pages listed).

This confirmed `type: docs` alone (no description or other params required — `utility` has none and still renders its Tools link) is the fix, so I applied it uniformly to all remaining integration index pages rather than writing per-page overview content or introducing a new redirect mechanism.

Scope: front-matter only, one line added per file, no other content changed.

### DIFF
--- a/docs/en/integrations/alloydb-admin/_index.md
+++ b/docs/en/integrations/alloydb-admin/_index.md
@@ -1,4 +1,5 @@
 ---
 title: "AlloyDB Admin"
+type: docs
 weight: 1
 ---
--- a/docs/en/integrations/alloydb/_index.md
+++ b/docs/en/integrations/alloydb/_index.md
@@ -1,4 +1,5 @@
 ---
 title: "AlloyDB PostgreSQL"
+type: docs
 weight: 1
 ---
--- a/docs/en/integrations/bigquery/_index.md
+++ b/docs/en/integrations/bigquery/_index.md
@@ -1,4 +1,5 @@
 ---
 title: "BigQuery"
+type: docs
 weight: 1
 ---
--- a/docs/en/integrations/bigtable/_index.md
+++ b/docs/en/integrations/bigtable/_index.md
@@ -1,4 +1,5 @@
 ---
 title: "Bigtable"
+type: docs
 weight: 1
 ---
--- a/docs/en/integrations/cassandra/_index.md
+++ b/docs/en/integrations/cassandra/_index.md
@@ -1,4 +1,5 @@
 ---
 title: "Cassandra"
+type: docs
 weight: 1
 ---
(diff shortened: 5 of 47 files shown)


Yuan325 (approved):
Thank you @Ashfaqbs ! :)

PR #3880, opened 2026-08-27
ci: add label-gated presubmit evals with changed-path targeting

omkargaikwad23:
## Description

Lets maintainers run the prebuilt config evals on a pull request, instead of
only on the schedule.

A maintainer applies the `evals: run` label and
`.github/workflows/evals_trigger.yml` comments `/gcbrun`, a comment being Cloud
Build's only pull request gate. The workflow posts the comment and nothing
else; it never checks out pull request code. The label is removed afterwards,
so a re-run needs a fresh label.

A `detect-changes` step then narrows the run to the databases the pull request
actually touched. Each step declares the paths it owns and skips green when
none of them changed; paths affecting every database live once in
`EVAL_COMMON_PATTERN`. On the scheduled run the diff is empty, so a fallback
path matches every step and the full matrix still runs.

`eval_group` becomes a substitution, so pull request results are tagged
separately and stay out of the scheduled accuracy trend.

Also bumps the `_CLAUDE_MODEL` substitution to `claude-opus-4-8`.

Documented for maintainers in `maintainer-playbook.md`, with the new per-step
`EVAL_CHANGED_PATTERN` in `DEVELOPER.md`.

CI configuration and docs only, no Toolbox source code is changed.

## PR Checklist
- [x] Make sure you reviewed
- [x] Ensure you have manually reviewed the entire diff before requesting a
  review

### DIFF
--- a/.github/workflows/evals_trigger.yml
+++ b/.github/workflows/evals_trigger.yml
@@ -1,0 +1,61 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Bridges the 'evals: run' label to Cloud Build, whose only pull request gate is
+# a /gcbrun comment. Posts the comment and nothing else; it never checks out PR
+# code, which is what makes pull_request_target safe here.
+
+name: evals trigger
+# zizmor flags pull_request_target, but it is safely gated by the 'evals: run' label manually applied by maintainers.
+on: # zizmor: ignore[dangerous-triggers]
+  pull_request_target:
+    types: [labeled]
+
+permissions: read-all
+
+jobs:
+  gcbrun:
+    if: "${{ github.event.label.name == 'evals: run' }}"
+    name: gcbrun
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: 'write'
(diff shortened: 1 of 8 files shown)


.github/workflows/evals_trigger.yml
```diff
@@ -0,0 +1,61 @@
...
+    permissions:
+      pull-requests: 'write'
+    steps:
+      - name: Comment /gcbrun
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            await github.rest.issues.createComment({
+              body: '/gcbrun',
```
prernakakkar-google, line 41: This trigger does not pass an _EVAL_GROUP override with /gcbrun. As a result, PR-triggered builds will use the default _EVAL_GROUP: "mcp-toolbox" from .ci/evals.cloudbuild.yaml, the same group as scheduled runs, so their results will still be included in the scheduled accuracy trend.

Please configure this invocation (or the corresponding Cloud Build trigger) to set a distinct presubmit group, e.g. mcp-toolbox-presubmit, and verify the supported /gcbrun syntax for supplying substitutions.

omkargaikwad23: _EVAL_GROUP is a substitution on the Cloud Build trigger, which overrides the default in substitutions:  /gcbrun is the whole comment syntax and can't carry substitutions.

.github/workflows/evals_trigger.yml
```diff
@@ -0,0 +1,61 @@
+# Copyright 2026 Google LLC
```
prernakakkar-google, line 1: labeled run can be run once and cannot be done more than once on PR.

prernakakkar-google (approved):

.github/workflows/evals_trigger.yml
omkargaikwad23: The workflow removes the label right after posting /gcbrun. Re-apply it to re-run, including after new commits. Same pattern as conformance.yml. You can also comment /gcbrun directly any number of times.

PR #3884, opened 2026-08-27
fix(prebuilt/cloud-sql-postgres): clarify execute_sql tool description

omkargaikwad23:
## Description

`execute_sql` in the Cloud SQL for PostgreSQL prebuilt config was described as
"Use this tool to execute a single SQL statement." That restates the tool name
and tells an agent nothing about the dialect, the return format, or when one of
the dedicated `list_*` tools would be the better choice — which matters because
`execute_sql` is easy to over-select.

The new description states the dialect and return format and points at the
schema-inspection tools first.

Scoped to this one config on purpose. `postgres.yaml` and `alloydb-postgres.yaml`
carry the same weak text and can follow.

## PR Checklist


internal/prebuiltconfigs/tools/cloud-sql-postgres.yaml
```diff
@@ -38,7 +38,7 @@ kind: tool
 name: execute_sql
 type: postgres-execute-sql
 source: cloudsql-pg-source
-description: Use this tool to execute a single SQL statement.
+description: Executes a single PostgreSQL statement and returns result rows as JSON. Prefer the dedicated schema tools (list_tables, list_schemas, list_views, list_indexes) for inspecting database objects; use this for queries they do not cover.
```
Yuan325:
```suggestion
description: Executes a single PostgreSQL statement and returns result rows as JSON. Use dedicated schema tools (list_tables, list_schemas, list_views, list_indexes) for inspecting database objects. Only use this tool for custom data queries not covered by those tools.
```

Yuan325 (approved):
Approve with a comment on the description :) Thank you!

PR #3901, opened 2026-08-28
test(postgres): stabilize list-active-queries test against shared instance concurrency

anubhav756:
## Overview
Fixes intermittent test failures in `list_active_queries` by making the test verify that our test query is present in the results, rather than assuming our test is the only thing running on the shared database.

## Background
* In #3872, we added native read-only support for AlloyDB.
* To test this, we upgraded our shared test database from PostgreSQL 15 to PostgreSQL 17, which supports read-only session locking.

## What Failure Happened After Upgrade?
* In PostgreSQL 17, internal background processes (like data sync and replication helpers) now stay permanently connected and show up as `"active"` in Postgres’s activity monitor (see [`pg_stat_activity`](https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW)).
* In PostgreSQL 15, these background helpers were not reported this way.
* Because our integration test originally expected the database to have zero running queries when idle, seeing these new Postgres 17 background helpers caused the test to fail.

## Previous Attempt (#3885)
* In #3885, we filtered out Postgres's internal helpers by only looking for regular user sessions (`backend_type = 'client backend'`).
* **Why it still flaked?** Other CI test runs also connect as regular user sessions (`client backend`). Because multiple CI jobs share the same test database, whenever two tests ran at the same time, the tool caught queries from both tests (2 queries instead of 1), causing the exact count check to flake.

## Solution
Instead of requiring the shared database to have only our query and nothing else:
* We check that our expected query is included in the returned list.
* This is also how other shared tests in this repo work.

## Links
* [PostgreSQL 17: `pg_stat_activity` Documentation](https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW)
* [PostgreSQL 17 Release Notes](https://www.postgresql.org/docs/17/release-17.html)

### DIFF
--- a/tests/tool.go
+++ b/tests/tool.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -2097,30 +2098,25 @@ func RunPostgresListActiveQueriesTest(t *testing.T, ctx context.Context, pool *p
 		Query            string `json:"query"`
 	}
 
+	testUUID := strings.ReplaceAll(uuid.New().String(), "-", "")[:8]
+	tc2AppName := fmt.Sprintf("test_sleep_tc2_%s", testUUID)
+	tc3AppName := fmt.Sprintf("test_sleep_tc3_%s", testUUID)
+
+	tc3Query := fmt.Sprintf("SELECT set_config('application_name', '%s', false), pg_sleep(10);", tc3AppName)
 	singleQueryWanted := queryListDetails{
-		ProcessId:        any(nil),
-		User:             "",
-		Datname:          "",
-		ApplicationName:  "",
-		ClientAddress:    "",
-		State:            "",
-		WaitEventType:    "",
-		WaitEvent:        "",
-		BackendStart:     any(nil),
-		TransactionStart: any(nil),
-		QueryStart:       any(nil),
-		QueryDuration:    any(nil),
-		Query:            "SELECT pg_sleep(10);",
+		ApplicationName: tc3AppName,
+		Query:           tc3Query,
 	}
 
 	invokeTcs := []struct {
 		name                string
 		toolName            string
 		args                map[string]any
+		appName             string
 		clientSleepSecs     int
 		waitSecsBeforeCheck int
 		wantStatusCode      int
-		want                any
+		want                []queryListDetails
 	}{
 		// exclude background monitoring apps such as "wal_uploader"
 		{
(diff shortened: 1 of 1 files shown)


tests/tool.go
```diff
@@ -2199,17 +2201,35 @@ func RunPostgresListActiveQueriesTest(t *testing.T, ctx context.Context, pool *p
...
 			if err := json.Unmarshal([]byte(resultString), &details); err != nil {
 				t.Fatalf("failed to unmarshal nested ObjectDetails string: %v", err)
 			}
-			got = details
 
-			if diff := cmp.Diff(tc.want, got, cmp.Comparer(func(a, b queryListDetails) bool {
-				return a.Query == b.Query
-			})); diff != "" {
-				t.Errorf("Unexpected result: got %#v, want: %#v", got, tc.want)
+			if tc.compareSubset {
```
Yuan325: With this implementation, it seems like test case 2 `tc.want` will not be verified. 
Also, it seems like the `tc.want` is always the same as we did not set any application name. (e.g. if we have 2 int tests running this at the same time, there will be 2 copies of the exact same `tc.want`).

Suggestion:
1. (preferred) in the waitGroup, lets do `SELECT pg_sleep_with_name(10, '<uuid>');` or `SELECT set_config('application_name', 'my_sleep_task', false), pg_sleep(10);`. With that, we can check for specific want response with the uuid to ensure that the pg_sleep that was just added is found within list_active_query.

2. Alternative if we really don't really care to look for the exact pg_sleep that was added within this test run, we can just check if `tc.want` is in `got`; without the `tc.compareSubset`.

Yuan325 (changes requested):

tests/tool.go
anubhav756: Interesting! I didn't know about the `set_config(application_name, ...` so thanks for the suggestion! This would definitely make it more robust. I've added the uuids and used tc.want for assertions.

PS: In addition to this, I also noticed that the test case 2 was running query sleep just for 1 second, which could be finished by the time the assertion query ran, so I also increased the timing of that sleep to 3 seconds as well to make this more secure.

tests/tool.go
```diff
@@ -2145,13 +2141,15 @@ func RunPostgresListActiveQueriesTest(t *testing.T, ctx context.Context, pool *p
...
 			args:                map[string]any{"min_duration": "1 seconds", "exclude_application_names": "wal_uploader"},
+			appName:             tc3AppName,
 			clientSleepSecs:     10,
 			waitSecsBeforeCheck: 5,
 			wantStatusCode:      http.StatusOK,
 			want:                []queryListDetails{singleQueryWanted},
 		},
 	}
 
+	parentT := t
```
Yuan325: Do we need to have a parentT?

Yuan325 (approved):

tests/tool.go
anubhav756: Fair enough! I originally had `parentT` because the `WaitGroup` was outside the loop, meaning subtests completed before background queries finished. I've updated it to make it faster as well, and also eliminated `parentT`.

PR #3905, opened 2026-08-28
feat(sources): add ConnectOnce, a helper for connecting on first use

akangsha7:
This PR adds `sources.ConnectOnce`, the helper behind `--lazy-source-init` (#3791). No source uses it yet and nothing changes behavior — the migrations follow in separate PRs, so the concurrency handling gets reviewed on its own rather than inside a 5,600 line diff.

A source that builds its handle on first use needs more than a `sync.Once`:
* Concurrent first calls coalesce through `singleflight`, so a burst of tool calls opens one pool rather than N.
* The connect runs under the context `Initialize` was called with, not the caller's. A request context carries that caller's auth claims and a user agent without `--user-agent-metadata`, and the handle every later caller reuses would otherwise inherit them. Only the span context crosses over, so the connect still appears in the trace of the request that paid for it.
* The attempt is bounded by `ConnectTimeout` while each caller returns on its own deadline. `WithMinConnectTimeout` raises that ceiling for a source whose own config allows a longer connect — Looker's default timeout is 600s, ten times the ceiling.
* Only successes are recorded, so a source that was briefly down starts working without a restart.

Storing the startup context on the struct is deliberate, and is what makes "nothing caller-specific ends up in the shared connect" a property of the helper rather than a rule each source has to remember.

Towards #3791 🦕


internal/sources/connect.go
```diff
@@ -0,0 +1,167 @@
...
+	userAgent, _ := util.UserAgentFromContext(ctx)
+	o := options{timeout: ConnectTimeout}
+	for _, opt := range opts {
+		opt(&o)
+	}
+	if o.timeout < ConnectTimeout {
+		o.timeout = ConnectTimeout
+	}
+	return &ConnectOnce[T]{name: name, sourceType: sourceType, tracer: tracer, userAgent: userAgent, timeout: o.timeout}
+}
```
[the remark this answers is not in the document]
akangsha7: Kept as a floor, but renamed to `WithMinConnectTimeout` in 5c73601 so the name says so.

The option exists because a source's own configured bound can be *longer* than the ceiling, not shorter: Looker's `timeout` defaults to 600s and bounds the login call the connect makes, so a 60s ceiling would silently cut the default config to a tenth of its documented value. Postgres' `connectTimeout` is the same shape. No source today configures a connect bound shorter than 60s, so honoring shorter values would change nothing except to make the ceiling itself defeatable — a source could set 1ms and pin the ceiling below what a cold cloud connector path needs.

Worth revisiting if a source ever wants fast-fail semantics, and at that point the right shape is probably a separate option rather than overloading this one.

internal/sources/connect_test.go
```diff
@@ -0,0 +1,318 @@
...
+	<-firstStarted
+
+	secondErr := make(chan error, 1)
+	go func() {
+		// Give the first caller time to win the singleflight.
+		time.Sleep(20 * time.Millisecond)
+		cancelFirst()
+		_, err := once.Do(context.Background(), c.connect)
+		secondErr <- err
+	}()
```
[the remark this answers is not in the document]
akangsha7: Fair — the sleep is a scheduling assumption rather than a synchronization point.

Leaving it for now because the assertion does not actually depend on the cancellation landing mid-connect: if the first caller cancels before winning the singleflight, the second caller simply starts the attempt itself and the test still asserts the right thing (the connect never sees a cancelled context, and the connection is cached). So a slow CI box makes the test less pointed, not flaky.

Happy to add an entered-connect channel to the mock if you would rather have it deterministic.

internal/sources/connect_test.go
```diff
@@ -0,0 +1,318 @@
...
+	if err := c.connectContextErr(); err != nil {
+		t.Fatalf("the connect saw a cancelled context: %s", err)
+	}
+	if _, ok := once.Get(); !ok {
+		t.Fatal("expected the shared attempt to finish and cache the connection")
+	}
+}
+
+func TestConnectOnceRespectsCallerDeadline(t *testing.T) {
+	c := &connector{delay: time.Minute}
```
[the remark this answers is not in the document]
akangsha7: Done in 5c73601 — the delay is now 2s, comfortably past the caller's 50ms deadline without leaving a minute-long goroutine parked for the rest of the package's tests.

internal/sources/connect.go
```diff
@@ -0,0 +1,158 @@
...
+		// source reads the same in a trace either way. It is not
+		// InitConnectionSpan, which sources emit around their own pool setup
+		// and which therefore nests underneath this one.
+		childCtx, span := c.tracer.Start(
+			connectCtx,
+			"toolbox/server/source/init",
+			trace.WithAttributes(attribute.String("source_type", c.sourceType)),
+			trace.WithAttributes(attribute.String("source_name", c.name)),
+		)
+		defer span.End()
```
Yuan325: This should be mirroring here instead: 
https://github.com/googleapis/mcp-toolbox/blob/bd444e92edaf838e229c4c25d2d19efd978e0eda/internal/sources/sources.go#L73-L80

This will remove from individual source's, example: 
https://github.com/googleapis/mcp-toolbox/blob/bd444e92edaf838e229c4c25d2d19efd978e0eda/internal/sources/alloydbpg/alloydb_pg.go#L207-L208

We'll still have the `toolbox/server/source/init` in server.go. Technically the source did initialize but not the db connection.

internal/sources/connect.go
Yuan325, line 1: Might be worth to add a `Close()` function. We can either add it here in ConnectOnce, or to reduce scope for now, we want to at least make sure that we add it within `connect func(context.Context) (T, error)`.

e.g.: 
```
func (s *Source) pool() (*pgxpool.Pool, error) {
    return s.conn.Do(....){
        ....
        if err := pool.Ping(ctx); err != nil {
            pool.Close() // here
            return nil, err
        }
        ....
    })
}
```

Yuan325 (changes requested):
Other than the comment LGTM

internal/sources/connect.go
akangsha7: Added, will be called from sources in next cahnge

internal/sources/connect.go
akangsha7: Done ConnectOnce calls sources.InitConnectionSpan now instead of starting its own toolbox/server/source/init.

Yuan325 (approved):

PR #3907, opened 2026-08-28
feat(sources): connect sources on first use and resolve env vars then

### LINKED ISSUE #2341: Skip connectivity checks on start up

akangsha7:
## Description

Toolbox connects to every source at startup and refuses to start if any connection fails, so a tool catalog cannot be inspected without every database being reachable, and a misconfigured source takes the server down instead of surfacing as an error an agent can read.

Sources now reach their connection through `sources.ConnectOnce` (#3905), which connects on first use, shares the attempt between concurrent callers, and bounds it independently of the request that triggered it. A failed attempt is not cached, so a source that becomes reachable later starts working without a restart.

Environment variables are resolved at that same point. Previously an unset `${VAR}` was replaced with the bare variable name, so a deferred source connected with a fabricated value and failed with whatever the driver made of it. The reference is now kept and resolved when the value is first needed, so an unset variable reports its own name. The undecoded config reaches the source through the context `Initialize` already runs under, and the registered factory is re-run with real values rather than fields being patched — preserving the defaults each factory applies and running every validator against what will actually be used.

Fields the tool catalog reads are excluded: `readOnly` and BigQuery's `writeMode` decide whether write-capable tools are listed at all, so deferring one would publish a catalog that contradicts the configuration. Those are reported together at startup, since each costs a restart to fix.

### Notable points

* `SourceConfig.Initialize` takes a `deferConnect` argument, wired to `--defer-source-connect`. It defaults to false, so by default every source still connects at startup and still reports the failure there.
* `--defer-env-var-parsing` requires `--defer-source-connect`; the combination is validated on both the root and `serve` commands.
* Config that needs no network is still validated at startup. The exception is value parsing that would choke on an unresolved reference — `baseUrl` and `timeout` (HTTP), `timeout` (Looker), `retryBaseDelay` (CockroachDB) moved into the connect path, keeping their error text.
* Each source holds a `ConnectOnce[T]` and reaches its handle through an accessor instead of an exported field, plus a `…Context(ctx)` accessor for callers needing a live handle. The existing typed accessor stays as the type discriminator tools assert on.
* Raise the connect ceiling wherever a source's own config permits a longer connect, so mysql, oceanbase, singlestore, mindsdb, trino and cockroachdb opt out the way postgres and looker already did.
* Fix the bigquery client creator capturing the connect's context, which is cancelled as soon as the connect returns.
* Stop `InitConnectionSpan` panicking on a nil tracer.

### Known limitations

Documented in `cli.md`. A few fields are read while the source is built rather than when it connects, so a deferred variable there has no effect and the startup value is used: `queryTimeout` (MySQL, MindsDB, OceanBase, SingleStore), the Couchbase cluster credentials, `allowedDatasets` (BigQuery), `allowedIpRanges`/`customBlockedIpRanges` (HTTP), and `protocol` (ClickHouse).

One cosmetic difference is left on purpose: mysql, oceanbase, singlestore, mindsdb and clickhouse lost the `unable to create pool: ` prefix on a malformed-config error, because the parse runs at startup. They still fail at startup with the same cause. Happy to restore the wrapping.

### Verification

Behaviour with neither flag set is byte-identical to `main` across nine configs (valid, env set, env missing, bad type, bad enum, missing required field, unknown key, bad source reference, and a value that itself contains `${...}`), covering startup output, `/api/toolset` and MCP `tools/list`. With both flags and no credentials at all, a Postgres and an HTTP source both initialise, `tools/list` returns the full catalog, and each invocation fails naming its own unset variables.

`go test -race -v ./cmd/... ./internal/...`

Carries #3791 (the flags); follows #3905 (the helper) and #3921 (release the handle on a failed connect).

## PR Checklist

## Issue Reference

Fixes #2341 🦕

### DIFF
--- a/cmd/internal/config.go
+++ b/cmd/internal/config.go
@@ -51,12 +51,20 @@ type ConfigParser struct {
 	requiredEnvVars []string
 
 	// AllowMissingEnvVars, when true, substitutes the variable name for an unset
-	// required ${VAR} placeholder instead of erroring. Used by offline flows like
-	// skills-generate, where source env vars are needed only to satisfy config
-	// parsing/validation, never to connect. A non-empty placeholder is used (not
-	// "") so required string fields still pass validation. The served path leaves
-	// this false so missing config still fails fast.
+	// required ${VAR} placeholder instead of erroring. Used by flows that never
+	// connect during parsing — skills-generate, and serving with a deferred
+	// connect — where source env vars are needed only to satisfy config
+	// parsing/validation. A non-empty placeholder is used (not "") so required
+	// string fields still pass validation. Eager serving leaves this false so
+	// missing config still fails fast.
 	AllowMissingEnvVars bool
+
+	// MissingEnvVars names the variables that were substituted with a
+	// placeholder. The substitution is final — nothing re-reads the environment
+	// afterwards, so any source depending on one of these fails when it
+	// connects. Callers report the list rather than let a placeholder stand in
+	// for real config silently.
+	MissingEnvVars []string
 }
 
 // parseEnv replaces environment variables ${ENV_NAME} with their values.
@@ -117,6 +125,9 @@ func (p *ConfigParser) parseEnv(input string) (string, error) {
 		} else {
 			if p.AllowMissingEnvVars {
 				p.EnvVars[variableName] = variableName
+				if !slices.Contains(p.MissingEnvVars, variableName) {
+					p.MissingEnvVars = append(p.MissingEnvVars, variableName)
+				}
 				output.WriteString(variableName)
 			} else if !seenMissing[variableName] {
 				seenMissing[variableName] = true
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -359,7 +359,7 @@ func watchChanges(ctx context.Context, watchDirs map[string]bool, watchedFiles m
 			reloadedOpts.Configs = slices.Clone(opts.Configs)
 			reloadedOpts.Cfg.Version = versionString
 
-			if _, err := reloadedOpts.LoadConfig(ctx, &internal.ConfigParser{}); err != nil {
+			if _, err := reloadedOpts.LoadConfig(ctx, &internal.ConfigParser{AllowMissingEnvVars: opts.Cfg.DeferEnvVarParsing}); err != nil {
 				logger.WarnContext(ctx, fmt.Sprintf("Error reloading config: %s", err))
 				continue
 			}
@@ -431,9 +431,25 @@ func run(cmd *cobra.Command, opts *internal.ToolboxOptions) error {
 		_ = shutdown(ctx)
 	}()
 
-	isCustomConfigured, err := opts.LoadConfig(ctx, &internal.ConfigParser{})
+	// A placeholder only makes sense for a source that has not connected yet.
+	// Substituting one and then connecting eagerly fails on the placeholder
+	// itself, so reject the combination rather than report that failure.
+	if opts.Cfg.DeferEnvVarParsing && !opts.Cfg.DeferSourceConnect {
+		errMsg := fmt.Errorf("--defer-env-var-parsing requires --defer-source-connect")
+		opts.Logger.ErrorContext(ctx, errMsg.Error())
+		return errMsg
+	}
+
+	parser := internal.ConfigParser{AllowMissingEnvVars: opts.Cfg.DeferEnvVarParsing}
+	isCustomConfigured, err := opts.LoadConfig(ctx, &parser)
 	if err != nil {
 		return err
+	}
+	if len(parser.MissingEnvVars) > 0 {
+		slices.Sort(parser.MissingEnvVars)
+		opts.Logger.WarnContext(ctx, fmt.Sprintf(
+			"Unset environment variables replaced with placeholders because --defer-env-var-parsing is set; any source using them will fail to connect: %s",
+			strings.Join(parser.MissingEnvVars, ", ")))
 	}
 
 	// Validate ToolboxUrl if MCP Auth is enabled
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -113,7 +113,7 @@ func InitializeConfigs(ctx context.Context, cfg ServerConfig) (
 				trace.WithAttributes(attribute.String("source_name", name)),
 			)
 			defer span.End()
-			s, err := sc.Initialize(childCtx, instrumentation.Tracer)
+			s, err := sc.Initialize(childCtx, instrumentation.Tracer, cfg.DeferSourceConnect)
 			if err != nil {
 				return nil, fmt.Errorf("unable to initialize source %q: %w", name, err)
 			}
@@ -129,6 +129,9 @@ func InitializeConfigs(ctx context.Context, cfg ServerConfig) (
 		sourceNames = append(sourceNames, name)
 	}
 	l.InfoContext(ctx, fmt.Sprintf("Initialized %d sources: %s", len(sourcesMap), strings.Join(sourceNames, ", ")))
+	if cfg.DeferSourceConnect {
+		l.InfoContext(ctx, "Source connections are deferred; each source connects on first use.")
+	}
 
 	// initialize and validate the auth services from configs
 	authServicesMap := make(map[string]auth.AuthService)
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -49,7 +49,7 @@ import (
 	"github.com/googleapis/mcp-toolbox/internal/server"
 	v20260728 "github.com/googleapis/mcp-toolbox/internal/server/mcp/v20260728"
 	"github.com/googleapis/mcp-toolbox/internal/sources"
-	"github.com/googleapis/mcp-toolbox/internal/sources/alloydbpg"
+	_ "github.com/googleapis/mcp-toolbox/internal/sources/alloydbpg"
 	_ "github.com/googleapis/mcp-toolbox/internal/sources/postgres"
 	_ "github.com/googleapis/mcp-toolbox/internal/sources/sqlite"
 	"github.com/googleapis/mcp-toolbox/internal/telemetry"
@@ -412,10 +412,10 @@ func TestUpdateServer(t *testing.T) {
 	}
 
 	newSources := map[string]sources.Source{
-		"example-source": &alloydbpg.Source{
-			Config: alloydbpg.Config{
-				Name: "example-alloydb-source",
-				Type: "alloydb-postgres",
+		"example-source": testutils.MockSource{
+			MockSourceConfig: testutils.MockSourceConfig{
+				Name: "example-source",
+				Type: "mock-source",
 			},
 		},
 	}
@@ -1497,7 +1497,7 @@ func TestInitializeConfigs(t *testing.T) {
 	ctx = util.WithInstrumentation(ctx, instrumentation)
 	t.Run("valid initialization", func(t *testing.T) {
 		sourceConfig1 := testutils.MockSourceConfig{Name: "my-source", Type: "mock-source"}
-		source1, _ := sourceConfig1.Initialize(ctx, nil)
+		source1, _ := sourceConfig1.Initialize(ctx, nil, false)
 		tools1 := testutils.NewMockTool("my-tool", "mock tool for offline config", "my-source", nil, false, false)
 		validCfg := server.ServerConfig{
 			Version: "0.0.0",
--- a/internal/sources/alloydbpg/alloydb_pg.go
+++ b/internal/sources/alloydbpg/alloydb_pg.go
@@ -68,26 +68,16 @@ func (r Config) SourceConfigType() string {
 	return SourceType
 }
 
-func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.Source, error) {
-	pool, err := initAlloyDBPgConnectionPool(ctx, tracer, r.Name, r.Project, r.Region, r.Cluster, r.Instance, r.IPType.String(), r.User, r.Password, r.Database, r.ReadOnly)
-	if err != nil {
-		return nil, fmt.Errorf("unable to create pool: %w", err)
-	}
-
-	err = pool.Ping(ctx)
-	if err != nil {
-		pool.Close()
-		if r.ReadOnly &&
-			strings.Contains(err.Error(), "unrecognized configuration parameter") &&
-			strings.Contains(err.Error(), "alloydb_session_read_only") {
-			return nil, fmt.Errorf("failed to initialize AlloyDB source in read-only mode: 'alloydb_session_read_only' is not supported on this instance version. See documentation for details: https://mcp-toolbox.dev/integrations/alloydb/source/#reference: %w", err)
-		}
-		return nil, fmt.Errorf("unable to connect successfully: %w", err)
-	}
-
+func (r Config) Initialize(ctx context.Context, tracer trace.Tracer, deferConnect bool) (sources.Source, error) {
 	s := &Source{
 		Config: r,
-		Pool:   pool,
+		conn:   sources.NewConnectOnce[*pgxpool.Pool](ctx, r.Name, SourceType, tracer),
+	}
+	if deferConnect {
+		return s, nil
+	}
+	if _, err := s.pool(ctx); err != nil {
+		return nil, err
 	}
 	return s, nil
 }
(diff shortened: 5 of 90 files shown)


internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -68,29 +68,47 @@ func (r Config) SourceConfigType() string {
...
-	s := &Source{
+func (r Config) newSource(ctx context.Context, tracer trace.Tracer) *Source {
+	return &Source{
 		Config: r,
-		Pool:   pool,
+		tracer: tracer,
+		conn:   sources.NewConnectOnce[*pgxpool.Pool](ctx, r.Name, SourceType, tracer),
 	}
-	return s, nil
 }
```
Yuan325, line 87: Why do we separate out newSource?

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -68,29 +68,47 @@ func (r Config) SourceConfigType() string {
...
 
 type Source struct {
 	Config
-	Pool *pgxpool.Pool
+	tracer trace.Tracer
+	conn   *sources.ConnectOnce[*pgxpool.Pool]
+}
+
+// pool returns the connection pool, creating it on first use.
+func (s *Source) pool(ctx context.Context) (*pgxpool.Pool, error) {
```
Yuan325, line 96: why do we need 3 pool related queries? `pool()`, `PostgresPool()`, and `PostgresPoolContext()`.

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -68,29 +68,47 @@ func (r Config) SourceConfigType() string {
...
+	return s.conn.Do(ctx, func(ctx context.Context) (*pgxpool.Pool, error) {
+		pool, err := initAlloyDBPgConnectionPool(ctx, s.tracer, s.Name, s.Project, s.Region, s.Cluster, s.Instance, s.IPType.String(), s.User, s.Password, s.Database, s.ReadOnly)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create pool: %w", err)
+		}
+
+		err = pool.Ping(ctx)
+		if err != nil {
```
Yuan325, line 104: might wanna add a pool.Close() here since we are implementing retries.

akangsha7:
Open design questions from the ongoing discussion, kept out of the description since they are not settled:

1. **Flag surface.** The states are strictly ordered — deferring env resolution while connecting eagerly is incoherent, since a connect needs resolved values — so this may want an enum such as `--init-strategy=eager|deferred-connect|deferred-all` rather than the current boolean.
2. **Where env resolution happens.** This PR resolves at parse time and substitutes a placeholder for unset variables. The alternative under discussion is deferring resolution into per-source accessors. Worth noting that all 39 prebuilt configs already parse with any combination of variables unset: every bare `${VAR}` lands in a `string` field, and every non-string field that is templated carries a default.
3. **What a manifest shows when a variable is unset.** Parse-time placeholders name the missing variable in the parameter description. Resolving in the accessor and swallowing the error omits the constraint instead. Both list the tool; they differ in what the agent is told.

internal/sources/alloydbpg/alloydb_pg.go
akangsha7: Done

internal/sources/alloydbpg/alloydb_pg.go
akangsha7: PostgresPool() is never called anywhere. It exists only because 33 compatibleSource interfaces in internal/tools still declare it.

internal/sources/alloydbpg/alloydb_pg.go
akangsha7: It's the build-without-connecting step, so Initialize reads the same in every source: construct, then either hand back the unconnected source or connect and hand it back.

cmd/internal/config.go
```diff
@@ -51,12 +51,18 @@ type ConfigParser struct {
...
+	// connect — where source env vars are needed only to satisfy config
+	// parsing/validation. A non-empty placeholder is used (not "") so required
+	// string fields still pass validation. Eager serving leaves this false so
+	// missing config still fails fast.
 	AllowMissingEnvVars bool
+
+	// MissingEnvVars names the variables that were substituted with a
+	// placeholder, so a caller can report them rather than let a placeholder
+	// stand in for real config silently.
+	MissingEnvVars []string
```
Yuan325, line 67: How will this var be used? How do these get resolved later?

cmd/root.go
```diff
@@ -431,10 +431,26 @@ func run(cmd *cobra.Command, opts *internal.ToolboxOptions) error {
...
+	parser := internal.ConfigParser{AllowMissingEnvVars: opts.Cfg.DeferEnvVarParsing}
+	isCustomConfigured, err := opts.LoadConfig(ctx, &parser)
 	if err != nil {
 		return err
 	}
+	if len(parser.MissingEnvVars) > 0 {
+		slices.Sort(parser.MissingEnvVars)
+		opts.Logger.WarnContext(ctx, fmt.Sprintf(
+			"Unset environment variables replaced with placeholders because --defer-env-var-parsing is set; any source using them will fail to connect: %s",
+			strings.Join(parser.MissingEnvVars, ", ")))
```
Yuan325, line 452: Will debug log be more suitable here? In user's POV, the env var resolution is just deferred instead of it actually being missing. assuming that they are aware that they should be setting those env var.

internal/server/server.go
```diff
@@ -128,7 +128,11 @@ func InitializeConfigs(ctx context.Context, cfg ServerConfig) (
 	for name := range sourcesMap {
 		sourceNames = append(sourceNames, name)
 	}
-	l.InfoContext(ctx, fmt.Sprintf("Initialized %d sources: %s", len(sourcesMap), strings.Join(sourceNames, ", ")))
+	if cfg.DeferSourceConnect {
+		l.InfoContext(ctx, fmt.Sprintf("Deferred initialization of %d sources, each connects on first use: %s", len(sourcesMap), strings.Join(sourceNames, ", ")))
+	} else {
+		l.InfoContext(ctx, fmt.Sprintf("Initialized %d sources: %s", len(sourcesMap), strings.Join(sourceNames, ", ")))
+	}
```
Yuan325, line 134: We might not need to update this. Technically the Source is initialized. Just that the db connection is deferred.

internal/server/server_test.go
```diff
@@ -432,7 +433,7 @@ func TestUpdateServer(t *testing.T) {
 	}
 
 	gotSource, _ := s.PrimitiveMgr.GetSource("example-source")
-	if diff := cmp.Diff(gotSource, newSources["example-source"]); diff != "" {
+	if diff := cmp.Diff(gotSource, newSources["example-source"], cmpopts.IgnoreUnexported(alloydbpg.Source{})); diff != "" {
```
Yuan325: Curious why do we need to update this? Tests should be using the mocks instead~

internal/sources/alloydbpg/alloydb_pg.go
Yuan325: Can we keep this inline within the `Initialize` function instead of a separate function? doesn't seem like this function is required elsewhere~

cmd/internal/config.go
akangsha7: It's write-once in parseEnv, read once in run() to emit the startup warning naming every substitution — that's its only consumer.
The placeholder is the variable's own name and it's substituted at parse time, permanently — nothing re-reads the environment afterward. So a source depending on one will fail when it connects, with whatever error the driver gives for a bogus credential. The flag name is a bit misleading that way; it's really "substitute instead of failing," not "resolve later." I've expanded the doc comment to say so explicitly.

internal/server/server.go
akangsha7: Makes sense reverting. I will add a separate line for when deferral is on, since this is the only place in the logs that recorded the mode is active and it's useful when diagnosing a first-call failure. Let me know if you'd rather drop that too.

internal/server/server_test.go
akangsha7: Good catch, fixing

internal/sources/alloydbpg/alloydb_pg.go
akangsha7: Done

cmd/root.go
akangsha7: The substitution is permanent rather than deferred — nothing re-reads the environment after parse, so any source using one of these placeholders will fail when it connects. This is the only place the operator learns which variables were replaced, and it's the first thing to check when a source fails on its first call. Happy to drop it to debug if you'd still prefer.

internal/sources/postgres/postgres.go
```diff
@@ -71,29 +71,49 @@ func (r Config) SourceConfigType() string {
...
+// pool returns the connection pool, creating it on first use.
+func (s *Source) pool(ctx context.Context) (*pgxpool.Pool, error) {
+	return s.conn.Do(ctx, func(ctx context.Context) (*pgxpool.Pool, error) {
+		pool, err := initPostgresConnectionPool(ctx, s.tracer, s.Name, s.Host, s.Port, s.User, s.Password, s.Database, s.QueryParams, s.QueryExecMode, s.ConnectTimeout)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create pool: %w", err)
+		}
+		if err := pool.Ping(ctx); err != nil {
+			return nil, fmt.Errorf("unable to connect successfully: %w", err)
+		}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/bigtable/bigtable.go
```diff
@@ -57,38 +57,64 @@ func (r Config) SourceConfigType() string {
...
+
+		instanceAdminClient, err := initBigtableInstanceAdminClient(ctx, s.tracer, s.Name, s.Project)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create instance admin client: %w", err)
+		}
+
+		adminClient, err := initBigtableAdminClient(ctx, s.tracer, s.Name, s.Project, s.Instance)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create admin client: %w", err)
+		}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/bigquery/bigquery.go
```diff
@@ -162,121 +176,143 @@ func (r Config) Initialize(ctx context.Context, tracer trace.Tracer) (sources.So
...
+		// credentials of its own, so it cannot check.
+		if cs.client != nil {
+			for fullID := range s.AllowedDatasets {
+				projectID, datasetID, _ := strings.Cut(fullID, ".")
+				if _, err := cs.client.DatasetInProject(projectID, datasetID).Metadata(ctx); err != nil {
 					if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == http.StatusNotFound {
 						return nil, fmt.Errorf("allowedDataset '%s' not found in project '%s'", datasetID, projectID)
 					}
 					return nil, fmt.Errorf("failed to verify allowedDataset '%s' in project '%s': %w", datasetID, projectID, err)
 				}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/alloydbpg/alloydb_pg.go
```diff
@@ -68,29 +68,47 @@ func (r Config) SourceConfigType() string {
...
+			return nil, fmt.Errorf("unable to connect successfully: %w", err)
+		}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/firestore/firestore.go
```diff
@@ -68,33 +68,57 @@ func (r Config) SourceConfigType() string {
...
+		client, err := initFirestoreConnection(ctx, s.tracer, s.Name, s.Project, s.Database)
+		if err != nil {
+			return nil, err
+		}
+
+		// Initialize Firebase Rules client
+		rulesClient, err := initFirebaseRulesConnection(ctx, s.Project)
+		if err != nil {
+			return nil, fmt.Errorf("failed to initialize Firebase Rules client: %w", err)
+		}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/mongodb/mongodb.go
```diff
@@ -58,30 +58,47 @@ func (r Config) SourceConfigType() string {
...
+	return s.conn.Do(ctx, func(ctx context.Context) (*mongo.Client, error) {
+		client, err := initMongoDBClient(ctx, s.tracer, s.Name, s.Uri)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create MongoDB client: %w", err)
+		}
+
+		// Verify the connection
+		if err := client.Ping(ctx, nil); err != nil {
+			return nil, fmt.Errorf("unable to connect successfully: %w", err)
+		}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/dataproc/dataproc.go
```diff
@@ -63,41 +63,63 @@ func (r Config) SourceConfigType() string {
...
+			return nil, fmt.Errorf("failed to create dataproc client: %w", err)
+		}
+		opsClient, err := longrunning.NewOperationsClient(ctx, option.WithEndpoint(endpoint), option.WithUserAgent(ua))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create longrunning client: %w", err)
+		}
+		jobClient, err := dataproc.NewJobControllerClient(ctx, option.WithEndpoint(endpoint), option.WithUserAgent(ua))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create dataproc job client: %w", err)
+		}
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/serverlessspark/serverlessspark.go
```diff
@@ -64,47 +64,73 @@ func (r Config) SourceConfigType() string {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create dataproc session template client: %w", err)
+		}
+		opsClient, err := longrunning.NewOperationsClient(ctx, option.WithEndpoint(endpoint), option.WithUserAgent(ua))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create longrunning client: %w", err)
+		}
+		sessionClient, err := dataproc.NewSessionControllerClient(ctx, option.WithEndpoint(endpoint), option.WithUserAgent(ua))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create dataproc session client: %w", err)
```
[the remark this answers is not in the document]
akangsha7: Done — the handle is released on every failure path now. This is load-bearing under lazy init since a failed connect isn't cached and gets retried per tool call.

internal/sources/connect.go
```diff
@@ -0,0 +1,145 @@
...
+	return c.value, c.ready
+}
+
+// Do returns the connection, making it on the first call. Concurrent callers
+// share one attempt, and a failed attempt is not remembered.
+func (c *ConnectOnce[T]) Do(ctx context.Context, connect func(context.Context) (T, error)) (T, error) {
+	var zero T
+	if value, ok := c.Get(); ok {
+		return value, nil
+	}
```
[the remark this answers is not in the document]
akangsha7: Closing this one. `Do` already selects on `ctx.Done()` at the bottom, so a cancelled caller returns without waiting. Adding the check above `c.Get()` would also make a cancelled caller fail even when the connection is already established and no work is needed, and it returns the bare context error instead of the `unable to connect to source %q` wrapping the other paths use. The attempt itself isn't wasted either — it's shared via singleflight, so other waiters get it.

cmd/internal/config.go
```diff
@@ -34,10 +34,15 @@ import (
...
 	"github.com/googleapis/mcp-toolbox/internal/server"
+	"github.com/googleapis/mcp-toolbox/internal/sources"
 	"github.com/googleapis/mcp-toolbox/internal/util"
 )
 
 type Config struct {
+	// SourceDocs holds each source's configuration as written, before decoding.
+	// A source that left an environment variable unresolved decodes without
+	// that field, so this is what it resolves against when it connects.
+	SourceDocs        map[string]map[string]any      `yaml:"-"`
```
Yuan325: why do we keep this? I thought we are storing it in the connectOnce? We probably wanna avoid having a `SourceDocs` and `Sources`. that will create 2 source of truth.

PR #3920, opened 2026-09-01
chore(deps): bump geminidataanalytics to v1.4.0 (drops Spanner region)

wangauone:
Stacked on #3919. Review that one first; this diff shows only the bump once #3919 merges.

**Breaking for Spanner users.** v1.4.0 removes `SpannerDatabaseReference.region`. `protojson` rejects unknown fields, so any `tools.yaml` with `spannerReference.databaseReference.region` will fail to load after upgrading. Worth a release note.

**New fields**, added to the reference tables #3919 introduced:

- `databaseTableReferences` (all kinds) — `tableId` plus optional `schema`; richer alternative to `tableIds`
- `priority`, `requestTag` (Spanner only)

Test fixtures drop the two `region` usages. `go build ./...` and the cloudgda unit tests pass.

Not marked `!` since a major bump seems disproportionate for a beta-API field removal, but happy to change that.


Yuan325:
Wondering if it's possible for us to manually remove the `region` field (during unmarshaling) so that it doesn't break existing users?

PR #3921, opened 2026-09-01
fix(sources): release the handle when a source fails to connect

akangsha7:
## Description

`Initialize` builds a pool or client, verifies it, and then returns the verification error **without releasing what it just built**. Nothing can reach that handle afterwards, so its sockets and file descriptors stay open until the process exits.

At startup that costs little, since a failed `Initialize` usually takes the process down with it. **Config reload is where it accumulates:** `handleDynamicReload` (`cmd/root.go:135`) re-runs `InitializeConfigs`, which rebuilds every source, and the reload is driven by a file watcher. A source that is misconfigured or briefly unreachable leaks one handle per reload rather than one per process.

Each source now releases whatever it had already built before it returns:

```go
err = pool.Ping(ctx)
if err != nil {
    pool.Close()            // <- added
    return nil, fmt.Errorf("unable to connect successfully: %w", err)
}
```

The call differs per source because the handles disagree on shape — `pgxpool.Pool.Close` takes no context and returns nothing, neo4j's driver takes one, mongo spells it `Disconnect`, and falkordb holds the closer a level down on `Conn`. The five sources that build several clients (`bigtable`, `dataproc`, `serverlessspark`, `firestore`, `bigquery`) release the ones that already succeeded.

`cloudsqlpg` and `dataplex` already did this and are unchanged.

26 files, +35/-0. Every hunk is a release call inserted before an existing error return; no control flow, signatures, or error messages change.

This also lands the release step that @Yuan325 asked for while reviewing #3905, and covers the resource-leak comments raised on #3907. It is independent of the lazy-source-init work — the leak exists on `main` today.

## PR Checklist

Verified locally:

```
go build ./...
go test -race ./cmd/... ./internal/...    # 405 packages, all pass
golangci-lint run ./internal/... ./cmd/...  # 0 issues
```


Yuan325 (approved):
Thank youuuu !! :)

PR #3933, opened 2026-09-03
fix(source/cockroachdb,source/redis): release the handle when a connect attempt fails

hktitof:
## Description

#3921 made every source release the handle it built when verification
fails, but two sources slip through that sweep bcs the handle is built
inside a helper instead of inline in Initialize

cockroachdb: the retry loop in initCockroachDBConnectionPoolWithRetry
creates a fresh pgxpool.Pool per attempt and drops the previous one
unclosed on ping failure, so one failed Initialize leaks up to
maxRetries+1 pools (default is 5 retries = 6 pools), each keeping its
health check goroutine alive. dynamic reload makes it accumulate, one
generation per reload attempt

redis: the cluster path returns on a failed ForEachShard ping and the
standalone path on a failed Ping without closing the client, and both
are built with MinIdleConns: 1, so the abandoned pool keeps dialing in
the background forever

now every failed attempt closes what it built before the next retry or
the error return, same shape as #3921, no control flow or error message
changes. regression tests cover both, verified they fail without the
fix (3+ leaked goroutines on cockroachdb retries, redis dialers that
never stop)

## PR Checklist

- [ ] Make sure you reviewed

Verified locally:

```
go build ./...
go test -race ./cmd/... ./internal/...    # 405 packages, all pass
golangci-lint run ./internal/... ./cmd/...  # 0 issues
```


Yuan325 (approved):
Hi @hktitof thank you for your contributions :)

PR #3959, opened 2026-09-04
ci: add BigQuery evaluation support with automated setup/teardown 

omkargaikwad23:
Adds evals for the `bigquery` prebuilt config.

The project has no stable BigQuery fixture, so `evals/setup/bigquery.sh` seeds a
per-build dataset (`toolbox_evals_${EVAL_RUN_ID}`) and `evals/teardown/bigquery.sh`
drops it. Teardown also has a sweep mode for datasets a killed build never
reached teardown for.

Four scenarios, covering 7 of the config's 9 tools:
* list the datasets, then describe the seeded one and its tables
  (`list_dataset_ids`, `get_dataset_info`, `list_table_ids`)
* inspect the `orders` schema and find the top customer by total amount
  (`get_table_info`, `execute_sql`)
* forecast the next 14 days of `daily_revenue` (`get_table_info`, `forecast`)
* find the segments behind a change in `signups`
  (`get_table_info`, `analyze_contribution`)

`search_catalog` and `ask_data_insights` are left out: the first goes through
Dataplex, which will not have indexed a dataset created minutes earlier, and the
second needs the Gemini Data Analytics API enabled on the build project.

The source runs with `readOnly: true`, so no scenario can write.


.ci/evals.cloudbuild.yaml
```diff
@@ -74,7 +74,7 @@ steps:
       - "TOOLBOX_PREBUILT=cloud-sql-postgres"
       - "EVAL_DATASET=evals/evalsets/cloud-sql-postgres.json"
       - "EVAL_ENV_PREFIX=CLOUD_SQL_POSTGRES"
-      - "EVAL_HARNESSES=claude_cli gemini_cli"
+      # - "EVAL_HARNESSES=claude_cli gemini_cli"
```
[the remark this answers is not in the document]
omkargaikwad23: We can ignore gemini_cli. Default claude_cli will be used & later will add agy cli & codex.

evals/evalsets/bigquery.json
```diff
@@ -0,0 +1,26 @@
...
+      "id": "bigquery-02",
+      "starting_prompt": "In the BigQuery dataset toolbox_evals_${EVAL_RUN_ID}, what columns does the orders table have, and which customer has the highest total order amount?",
+      "conversation_plan": "The user wants the schema first and then a figure derived from the data, so the assistant has to inspect the table before querying it. Accept the column list plus a named customer and end the conversation.",
+      "expected_trajectory": [
+        "toolbox__get_table_info",
+        "toolbox__execute_sql"
+      ],
+      "kind": "tools",
+      "max_turns": 1
+    }
```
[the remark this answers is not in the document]
omkargaikwad23: outdated

.ci/evals.cloudbuild.yaml
prernakakkar-google: nit: remove the comment

.ci/evals.cloudbuild.yaml
```diff
@@ -132,6 +132,27 @@ steps:
       - "CLOUD_SQL_POSTGRES_REGION=${_REGION}"
       - "CLOUD_SQL_POSTGRES_INSTANCE=${_CLOUD_SQL_POSTGRES_INSTANCE}"
 
+  # No secretEnv: the source authenticates through ADC. BIGQUERY_READONLY is not
```
prernakakkar-google: nit:remove

evals/evalsets/bigquery.json
```diff
@@ -0,0 +1,49 @@
+{
```
prernakakkar-google, line 1: Consider adding other complex CUJs as well.

evals/evalsets/bigquery.json
```diff
@@ -0,0 +1,49 @@
...
+  "scenarios": [
+    {
+      "id": "bigquery-data-01",
+      "starting_prompt": "I am getting oriented in this project's BigQuery data. What datasets are there, and for ${EVAL_BQ_DATASET} specifically, what location is it in and which tables does it hold?",
+      "conversation_plan": "The user wants a dataset listing, then the location and table list for ${EVAL_BQ_DATASET}. If the assistant covers only some of the three, ask for what is missing. Once all three are given, end the conversation.",
```
prernakakkar-google: Add success/fail criteria for all CUJ

evals/evalsets/bigquery.json
omkargaikwad23: Made the existing CUJs harder instead of adding new ones over the same tools.

evals/evalsets/bigquery.json
omkargaikwad23: Done

evals/evalsets/bigquery.json
prernakakkar-google: I think  we should add more CUJs of varying difficulty we can use this as example CUJs. Follow what is required in dataset quality

omkargaikwad23: Agreed, added three more CUJs:
- bigquery-recovery-01 — points at a table that doesn't exist, so the agent has to notice the failure and find the real one.
- bigquery-recovery-02 — asks for a DELETE against the read-only connection; doubles as a guardrail test that the agent reports the refusal instead of claiming success.
- bigquery-data-03 — asks for a month the history doesn't cover, so the query succeeds with zero rows and the agent has to report the gap rather than invent a number, then refine to the last covered month.

prernakakkar-google (approved):

PR #3966, opened 2026-09-06
feat(config): support Compose-style env-var interpolation modifiers

### LINKED ISSUE #3635: Env-var interpolation treats set-but-empty variables as valid values; support Compose-style ${VAR:-default} and ${VAR:?} semantics

GISWLH:
## Summary
- Add Compose/bash-style `${VAR:-default}` (use default when unset **or** empty) and `${VAR:?[message]}` (error when unset **or** empty).
- Keep legacy `${VAR}` and `${VAR:default}` semantics unchanged for back-compat (empty set values still interpolate).
- Expand table-driven `TestParseEnv` coverage across unset/empty/set for all forms; update configuration docs.

Fixes #3635.

## Test plan
- [x] `go test ./cmd/internal/ -run TestParseEnv -count=1`
- [x] `go test ./cmd/internal/ -count=1`

### DIFF
--- a/cmd/internal/config.go
+++ b/cmd/internal/config.go
@@ -59,18 +59,31 @@ type ConfigParser struct {
 	AllowMissingEnvVars bool
 }
 
-// parseEnv replaces environment variables ${ENV_NAME} with their values.
-// also support ${ENV_NAME:default_value}.
+// parseEnv expands ${ENV_NAME...} placeholders using Compose-style rules.
+//
+// Forms:
+//   ${VAR}            error if unset; empty set values are kept
+//   ${VAR:default}    default only when unset (empty set values are kept)
+//   ${VAR:-default}   default when unset or empty
+//   ${VAR:?[message]} error when unset or empty
 func (p *ConfigParser) parseEnv(input string) (string, error) {
-	re := regexp.MustCompile(`\$\{(\w+)(:([^}]*))?\}`)
+	re := regexp.MustCompile(`\$\{(\w+)(?:(:-|:\?|:)([^}]*))?\}`)
 
 	if p.EnvVars == nil {
 		p.EnvVars = make(map[string]string)
 	}
 
+	type missingEnvVar struct {
+		name             string
+		message          string
+		line             int
+		column           int
+		requiredNonEmpty bool
+	}
+
 	tokens := lexer.Tokenize(input)
 
-	var missing []string
+	var missing []missingEnvVar
 	seenMissing := make(map[string]bool)
 	matches := re.FindAllStringSubmatchIndex(input, -1)
 	var output strings.Builder
@@ -96,11 +109,14 @@ func (p *ConfigParser) parseEnv(input string) (string, error) {
 		output.WriteString(input[lastIndex:start])
 
 		variableName := input[match[2]:match[3]]
-		defaultValue := ""
-		defaultProvided := match[4] != -1 && match[5] != -1
-		if defaultProvided {
-			defaultValue = input[match[6]:match[7]]
+		operator := ""
+		operand := ""
+		if match[4] != -1 && match[5] != -1 {
+			operator = input[match[4]:match[5]]
+			operand = input[match[6]:match[7]]
 		}
+		defaultProvided := operator == ":" || operator == ":-"
+		requiredNonEmpty := operator == ":?"
 
 		if defaultProvided {
 			p.OptionalEnvVars = append(p.OptionalEnvVars, variableName)
@@ -108,20 +124,34 @@ func (p *ConfigParser) parseEnv(input string) (string, error) {
 			p.requiredEnvVars = append(p.requiredEnvVars, variableName)
 		}
 
-		if value, found := os.LookupEnv(variableName); found {
+		value, found := os.LookupEnv(variableName)
+		useDefault := defaultProvided && !found
+		if operator == ":-" && value == "" {
+			useDefault = true
+		}
+		missingRequired := requiredNonEmpty && (!found || value == "")
+
+		switch {
+		case found && !useDefault && !missingRequired:
 			p.EnvVars[variableName] = value
 			output.WriteString(value)
-		} else if defaultProvided {
-			p.EnvVars[variableName] = defaultValue
-			output.WriteString(defaultValue)
-		} else {
-			if p.AllowMissingEnvVars {
-				p.EnvVars[variableName] = variableName
-				output.WriteString(variableName)
-			} else if !seenMissing[variableName] {
+		case useDefault:
+			p.EnvVars[variableName] = operand
+			output.WriteString(operand)
+		case p.AllowMissingEnvVars:
+			p.EnvVars[variableName] = variableName
+			output.WriteString(variableName)
+		default:
+			if !seenMissing[variableName] {
 				seenMissing[variableName] = true
 				line, column := lineColumnAt(input, start)
-				missing = append(missing, fmt.Sprintf("%q (line %d, column %d)", variableName, line, column))
+				missing = append(missing, missingEnvVar{
+					name:             variableName,
+					message:          operand,
+					line:             line,
+					column:           column,
+					requiredNonEmpty: requiredNonEmpty,
+				})
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


cmd/internal/config.go
```diff
@@ -96,32 +109,49 @@ func (p *ConfigParser) parseEnv(input string) (string, error) {
...
-			} else if !seenMissing[variableName] {
+		case useDefault:
+			p.EnvVars[variableName] = operand
+			output.WriteString(operand)
+		case p.AllowMissingEnvVars:
+			p.EnvVars[variableName] = variableName
+			output.WriteString(variableName)
+		default:
+			if !seenMissing[variableName] {
 				seenMissing[variableName] = true
```
AlexTalreja, line 146: It seems like if a config uses a generic `${VAR_1}` first and later specifies a custom message for that same var like `${VAR_1:?custom message}`, the second occurrence is skipped and the custom error message is dropped. 

Is this intended behavior? We might want a later `:?` custom message to take precedence and upgrade the reported error.

AlexTalreja (changes requested):

AndyFooBlah:
@GISWLH thanks for implementing this — the shape matches what @he-yufeng proposed and what was discussed in #3635: `${VAR:-default}` and `${VAR:?message}` added, `${VAR}` and `${VAR:default}` left alone. That's the right split, and covering unset/empty/set across every form in the table-driven test is more thorough than the issue asked for.

One thing I'd flag for the description rather than the code: **three spellings change meaning**, and the summary says back-compat is preserved. It is for `${VAR}` and `${VAR:default}`, but the old regex `\$\{(\w+)(:([^}]*))?\}` treats everything after the first `:` as a literal default, so today these already parse — and produce values. I ran them against current `main` (v1.12.0, c97ca4d) rather than reasoning from the regex:

| config | behavior on main today | after this PR |
|---|---|---|
| `${VAR:-fallback}` (unset) | interpolates `-fallback` | default `fallback` |
| `${VAR:?}` (unset) | interpolates `?` | error |
| `${VAR:?must be set}` (unset) | interpolates `?must be set` | error |
| `${VAR:-}` (unset) | interpolates `-` | default `""` |

@he-yufeng called the `${VAR:?}` case out in [the issue](https://github.com/googleapis/mcp-toolbox/issues/3635#issuecomment-5004864015) as the one compat edge, and `:-` turns out to have the same property.

Realistically nobody writes `${VAR:-foo}` intending a literal `-foo`, so the blast radius is small — but it's a silent behavior change rather than a pure addition, and this repo marks those with `!` per the conventional-commits convention in CONTRIBUTING. I'd either add the `!` or, if the maintainers think that's too heavy for something this obscure, note it in the docs section and the PR body so it shows up in the release notes. Their call on which.

Separately, the `cla/google` check is red — first-time contributors need to sign the CLA before maintainers can look at it, so that's worth clearing regardless of the above.

Not a maintainer, just the person who filed the issue — take or leave any of this.

cmd/internal/config.go
GISWLH: Good catch — not intended. Fixed in 75c9139: when the same var is recorded as missing more than once, a later `${VAR:?[message]}` now upgrades the earlier entry (plain `${VAR}` or a weaker `:?` form), keeping the first occurrence’s line/column and preferring the later custom message. Table-driven coverage added for plain→`:?custom`, `:?)`→`:?custom`, later custom precedence, and no-downgrade when a plain form follows.

GISWLH:
Addressed @AlexTalreja’s review: later `${VAR:?custom}` now upgrades an earlier plain/weaker missing recording (75c9139).

Also took @AndyFooBlah’s note on the three spellings that change meaning — added a short back-compat note in the configuration docs (legacy `${VAR}` / `${VAR:default}` unchanged; `:-` / `:?` are the Compose-style behavior change). Happy to add `!` to the commit type if maintainers prefer that for the release notes.

PR #3970, opened 2026-09-07
feat(MCPApps): implement MCP Apps UI resource metadata schema

dishaprakash:
## Summary
This PR introduces support for MCP Ext-Apps by implementing the UI resource metadata schema. It allows configuration authors to expose  UI applications securely through the MCP server. 
By toggling `ui: true`, resource and template configs can now define Content Security Policies (CSP), requested browser permissions, and layout preferences.

**Reference Schema** - https://github.com/modelcontextprotocol/ext-apps/blob/main/src/spec.types.ts

## Key Changes
- **UI Resource Configuration & Schemas** (`internal/resources/ui.go`, `internal/resources/resources.go`):
  - Added `ui: bool` to `ConfigBase`, designating resources/templates as MCP Apps.
  - Implemented `CSPConfig` with validation for `connectDomains`, `resourceDomains`, `frameDomains`, and `baseUriDomains` (strictly enforcing valid absolute URIs with `http` or `https` schemes).
  - Implemented `PermissionsConfig` supporting `camera`, `microphone`, `geolocation`, and `clipboardWrite` device permission requests.
  - Added `domain` (view sandbox origin) and `prefersBorder` display boundary hint.
- **Strict MIME Type Validation & Canonicalization**:
  - Defined `const UIMimeType = "text/html;profile=mcp-app"`.
  - Automatically defaults `mimeType` to `UIMimeType` if omitted when `ui: true`.
  - Strictly validates that any explicitly configured `mimeType` conforms to `text/html;profile=mcp-app` using `mime.ParseMediaType` (permits formatting variations such as optional whitespace, while canonicalizing outgoing representation to `UIMimeType`).
  - Rejects non-conforming MIME types during configuration validation.
- **URI Scheme Enforcement**:
  - Automatically defaults URI to `ui://{name}` for UI resources if omitted.
  - Enforces `ui://` scheme for all UI resources and forbids non-UI resources from claiming the `ui://` scheme.
- **Resource Implementations**:
  - Implemented `GetResourceUIMetadata()` across `file` and `text` resource and resource template implementations.


internal/resources/ui.go
```diff
@@ -0,0 +1,143 @@
...
+	return meta
+}
+
+// ResourceUIMetadata represents the metadata for a UI Resource.
+type ResourceUIMetadata struct {
+	CSP           *CSPConfig     `yaml:"csp,omitempty" json:"csp,omitempty"`
+	Permissions   map[string]any `yaml:"permissions,omitempty" json:"permissions,omitempty"`
+	Domain        string         `yaml:"domain,omitempty" json:"domain,omitempty"`
+	PrefersBorder *bool          `yaml:"prefersBorder,omitempty" json:"prefersBorder,omitempty"`
+}
```
Yuan325, line 143: Might wanna move this to the `internal/server/mcp/<version>/types.go`? In cases if the next release have different type, we would want to version-control this value.

internal/resources/resources.go
```diff
@@ -89,10 +94,27 @@ func (c ConfigBase) GetDescription() string               { return c.Description
...
+		permMeta = c.Permissions.ToUIMetadata()
+	}
+	return ResourceUIMetadata{
+		CSP:           c.CSP,
+		Domain:        c.Domain,
+		PrefersBorder: c.PrefersBorder,
+		Permissions:   permMeta,
+	}
+}
+
```
Yuan325, line 116: If we move the `ResourceUIMetadata` to `internal/server/mcp/...`, we might also want to move this there.

Yuan325 (approved):
Please resolve the comment, otherwise LGTM :)

internal/resources/ui.go
dishaprakash: Makes sense. Following the same design as ResourceAnnotations, I will be adding the same struct in types.go as well to version-control this

internal/resources/resources.go
dishaprakash: Will be adding the tests in types.go in the RPC Endpoints PR

Yuan325: Seems like this is not moved~ We probably wanna remove this function from here overall~

dishaprakash: Ah I see we want to remove it completely. Will raise a cleanup PR for this after release as it is currently not a blocker

PR #3971, opened 2026-09-07
feat(MCPApps): add Tool UI metadata schema and internal validation

dishaprakash:
### Overview
This PR implements linking MCP Tools directly to UI resources. 
It adds the internal `ui` metadata schema, strong typing for visibility controls, and strict configuration validation to ensure tools and their bound resources are correctly scoped.

Following spec - https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

### Key Changes
- **Tool UI Configuration** (`internal/tools/tools.go`):
  - Added `ui` configuration block to `ToolConfig` with `resource` (name of referenced UI resource) and `visibility` (`model`, `app`, defaulting to `["model", "app"]`).
  - Added `GetToolUIMetadata() *ToolUIMetadata` getter to the `Tool` interface.
- **Global Scoping & Group Exclusion** (`internal/group/group.go`):
  - Enforced that UI resources cannot be placed in `groups[].resources` or `groups[].resourceTemplates`. UI resources are global by design and accessible across all groups.
  - Added boot-time validation rejecting group configurations containing UI resources:
    `UI resource "<name>" cannot be included in group "<group>": UI resources are globally accessible and cannot be scoped to groups`
- **Tool UI Reference Validation** (`internal/server/server.go`):
  - Validates during server startup that any `ui.resource` referenced by a tool exists in the registry and has `ui: true` enabled.
  - Rejects references to non-existent or standard non-UI resources:
    `resource "<name>" referenced by tool "<tool>" is not a UI resource (ui: true is required)`
  - Allows tools in any group to reference UI resources freely without group-scoping restrictions.


internal/group/group.go
```diff
@@ -103,6 +103,18 @@ func (gc GroupConfig) Initialize(toolsMap map[string]tools.Tool, promptsMap map[
 		templateNameSet[name] = struct{}{}
 	}
 
+	for _, toolName := range gc.ToolNames {
+		tool := toolsMap[toolName]
+		uiMeta := tool.GetToolUIMetadata()
+		if uiMeta != nil && uiMeta.Resource != "" {
+			_, hasRes := resourceNameSet[uiMeta.Resource]
+			_, hasTmpl := templateNameSet[uiMeta.Resource]
+			if !hasRes && !hasTmpl {
```
Yuan325: Should this be a Requirement? Or should we just return resources/read as failed when user query it?

internal/server/server.go
```diff
@@ -354,6 +364,21 @@ func InitializeOfflineConfigs(ctx context.Context, cfg ServerConfig) (
 	return toolsMap, groupsMap, nil
 }
 
+// validateToolUIResources validates that all UI resources referenced by tools exist in either resourcesMap or resourceTemplatesMap.
+func validateToolUIResources(toolsMap map[string]tools.Tool, resourcesMap map[string]resources.Resource, resourceTemplatesMap map[string]resources.ResourceTemplate) error {
```
Yuan325: mentioend as above, do we want to enforce this~?

internal/group/group.go
```diff
@@ -89,17 +89,25 @@ func (gc GroupConfig) Initialize(toolsMap map[string]tools.Tool, promptsMap map[
 
 	resourceNameSet := make(map[string]struct{}, len(gc.ResourceNames))
 	for _, name := range gc.ResourceNames {
-		if _, ok := resourcesMap[name]; !ok {
+		res, ok := resourcesMap[name]
+		if !ok {
 			return Group{}, fmt.Errorf("resource does not exist: %q", name)
 		}
+		if res.GetResourceUIMetadata() != nil {
```
Yuan325:
```suggestion
		if res.UI {
```

Should we check the `UI` field directly? Not sure if it's worth creating a new ResourceUIMetadata struct just for verification if its UI. 

Alternatively, we can also add a `func IsUI() bool` to the Resource.

internal/group/group.go
```diff
@@ -89,17 +89,25 @@ func (gc GroupConfig) Initialize(toolsMap map[string]tools.Tool, promptsMap map[
...
 	}
 
 	templateNameSet := make(map[string]struct{}, len(gc.ResourceTemplateNames))
 	for _, name := range gc.ResourceTemplateNames {
-		if _, ok := templatesMap[name]; !ok {
+		tmpl, ok := templatesMap[name]
+		if !ok {
 			return Group{}, fmt.Errorf("resource template does not exist: %q", name)
 		}
+		if tmpl.GetResourceUIMetadata() != nil {
```
Yuan325: same with this as well~

internal/server/server.go
```diff
@@ -354,6 +364,27 @@ func InitializeOfflineConfigs(ctx context.Context, cfg ServerConfig) (
...
+				return fmt.Errorf("resource %q referenced by tool %q is not a UI resource (ui: true is required)", uiMeta.Resource, toolName)
+			}
+			if hasTmpl && tmpl.GetResourceUIMetadata() == nil {
+				return fmt.Errorf("resource template %q referenced by tool %q is not a UI resource (ui: true is required)", uiMeta.Resource, toolName)
+			}
+		}
+	}
+	return nil
+}
+
```
Yuan325: We might not wanna include this since we are treating UI resources similar to auth service and embedding model right now. We do not check if those primitive exist during initialization~

internal/server/server.go
```diff
@@ -424,13 +455,17 @@ func initializeGroups(ctx context.Context, cfg ServerConfig, toolsMap map[string
 	slices.Sort(allPromptNames)
 
 	allResourceNames := make([]string, 0, len(resourcesMap))
-	for name := range resourcesMap {
-		allResourceNames = append(allResourceNames, name)
+	for name, res := range resourcesMap {
+		if res.GetResourceUIMetadata() == nil {
```
Yuan325: `res.UI` or `res.IsUI` will be better here~

internal/server/server.go
```diff
@@ -424,13 +455,17 @@ func initializeGroups(ctx context.Context, cfg ServerConfig, toolsMap map[string
...
+			allResourceNames = append(allResourceNames, name)
+		}
 	}
 	slices.Sort(allResourceNames)
 	allResourceTemplateNames := make([]string, 0, len(resourceTemplatesMap))
-	for name := range resourceTemplatesMap {
-		allResourceTemplateNames = append(allResourceTemplateNames, name)
+	for name, tmpl := range resourceTemplatesMap {
+		if tmpl.GetResourceUIMetadata() == nil {
```
Yuan325: same as above~

internal/server/server_test.go
Yuan325, line 1: Might needa update this according to the feedback above~

Yuan325 (changes requested):
Mostly LGTM :)

internal/group/group.go
dishaprakash: Changed to just return query as failed

internal/group/group.go
dishaprakash: Added IsUI()

internal/group/group.go
dishaprakash: Done

internal/server/server.go
dishaprakash: Removed check as per latest discussion

internal/server/server.go
dishaprakash: Done

internal/server/server.go
dishaprakash: Done

internal/server/server_test.go
dishaprakash: Done

Yuan325 (approved):

PR #3988, opened 2026-09-09
fix(source/oracle): enforce readOnly with an Oracle read-only transaction

### LINKED ISSUE #3987: Oracle `readOnly: true` does not prevent writes: UPDATE/DELETE run and commit
### Prerequisites

### Toolbox version

`toolbox version 1.10.0+dev.linux.amd64` 

### Environment

1. OS type and version: `Linux 6.8.0-106-generic #106~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC x86_64` (Ubuntu 22.04.5 LTS)
2. How are you running Toolbox: Compiled from source — `go build -o toolbox .`
3. Database: Oracle Database Free 23 (`gvenzl/oracle-free:slim`), PDB `FREEPDB1`

### Client

1. Client: plain HTTP against the MCP endpoint (`curl`), so this does not depend on any SDK
2. Version: n/a
3. Configuration used:

```yaml
sources:
  my-oracle:
    type: oracle
    connectionString: 127.0.0.1:1521/FREEPDB1
    user: toolbox
    password: ${PASSWORD}
tools:
  readonly-update:
    type: oracle-sql
    source: my-oracle
    readOnly: true
    description: A tool declared read-only that performs an UPDATE.
    statement: UPDATE repro SET "name" = 'HACKED' WHERE "id" = 1
```

### Expected Behavior

A tool configured with `readOnly: true` should not be able to modify data. A
write statement should be rejected, not executed.

### Current Behavior

`readOnly` only selects `QueryContext` over `ExecContext`. That is a choice of Go
function, not a constraint on the database: both send the same SQL over the same
connection, and Oracle executes whatever it is sent. Since both drivers run in
autocommit, the write is committed immediately.

The two tools return different response shapes, which confirms the field is
parsed and honored:

```
readOnly: false  ->  {"content":[{"text":"{\"rows_affected\":1,\"status\":\"success\"}"}]}
readOnly: true   ->  {"content":[]}
```

But both wrote. Read back from a separate sqlplus session after each call:

```
after control-update  (readOnly: false)  ->  WRITTEN_BY_RO_FALSE
after readonly-update (readOnly: true)   ->  WRITTEN_BY_RO_TRUE
```

So the `readOnly: true` call reports a successful, empty result while silently
committing the write — nothing in the response indicates anything was modified.

The write is durable, not pending on a pooled connection: after killing Toolbox
entirely and reading from a fresh session, the change is still there, along with
a row inserted the same way by a second `readOnly: true` tool:

```
	id name
	 1 WRITTEN_BY_RO_TRUE
	99 INSERTED_BY_RO_TRUE
```

`UPDATE` was verified on **both** drivers — `useOCI: true` (godror) and
`useOCI: false` (go-ora, the default). `INSERT` and `DELETE` behave the same
(verified with `useOCI: true`).

Relatedly, `SELECT ... FOR UPDATE` also runs through a `readOnly: true` tool, and
on go-ora (the default) it leaves the resulting transaction and row locks open on
the pooled connection — `v$transaction` stays at 1, still held 30+ seconds later,
and other sessions get `ORA-00054` until Toolbox is restarted. Same root cause.

### Steps to reproduce?

1. Start Oracle: `docker run -d --name oracle-free -e ORACLE_PASSWORD=<pw> -p 1521:1521 gvenzl/oracle-free:slim`
2. Seed a row:
   ```sql
   CREATE TABLE repro ("id" NUMBER PRIMARY KEY, "name" VARCHAR2(50));
   INSERT INTO repro ("id","name") VALUES (1,'ORIGINAL');
   COMMIT;
   ```
3. Save the config above as `tools.yaml`, then run `./toolbox --tools-file tools.yaml`.
4. Invoke the tool declared read-only:
   ```sh
   curl -s -X POST http://localhost:5000/mcp -H 'Content-Type: application/json' \
     -d '{"jsonrpc":"2.0","method":"tools/call","params":{"name":"readonly-update","arguments":{}},"id":1}'
   ```
   It returns `{"content":[]}` — a successful, empty result.
5. From a separate session, observe the committed write:
   ```sql
   SELECT "name" FROM repro WHERE "id" = 1;   -- WRITTEN_BY_RO_TRUE
   ```
   Stopping Toolbox first shows the change survives, so it is committed. Invoking
   `control-update` instead shows the flag is read: it returns `rows_affected`
   rather than `{"content":[]}`.

### Additional Details

The flag appears to have been introduced in #2323 to fix "incorrect array type"
errors when DML was forced through `QueryContext`, i.e. to choose the result
shape (rows vs. `rows_affected`). The name suggests a permission it never
enforced.

Closing this in Go by inspecting the statement text would not be reliable — any
client-side read-only guarantee based on pattern-matching SQL is a guess. Oracle
can enforce it directly with `SET TRANSACTION READ ONLY`, which rejects writes
and locking reads with `ORA-01456`.

A source-level option (the AlloyDB / Cloud SQL `readOnly` pattern, which pins a
session parameter at connect time) is not available here: Oracle has no
session-level read-only mode — `ALTER SESSION SET READ ONLY` fails with
`ORA-02248`. Transaction scope is the only mechanism Oracle offers.

I have a fix ready and will link a PR.

Luap2003:
## Description

`readOnly: true` on an Oracle tool was a naming convention rather than a
guarantee: it only chose `QueryContext` over `ExecContext`, which selects a Go
function rather than constraining the database. Verified against Oracle Free 23
on **both** drivers, a `readOnly: true` tool running
`UPDATE repro SET "name" = 'WRITTEN_BY_RO_TRUE' WHERE "id" = 1` reports success
with an empty result and the row is changed and committed — both drivers run in
autocommit, so there is no rollback that would undo it. The write survives
killing Toolbox, confirming it is committed rather than pending on a pooled
connection. `INSERT` and `DELETE` behave the same (verified with `useOCI: true`),
and `SELECT ... FOR UPDATE` returns rows.

On the pure-Go `go-ora` driver — the default, since `useOCI` defaults to `false`
— locking reads additionally leak: `SELECT ... FOR UPDATE` leaves an open
server-side transaction that `database/sql` knows nothing about, so the
connection returns to the pool with the locks still held. One invocation leaves
`v$transaction` at 1, still held 30s later, blocking a second session with
`ORA-00054`; a second invocation reuses the same connection, so the next caller
inherits it. With `useOCI: true` (godror) ODPI-C commits after the statement, so
the lock is released and only the missing enforcement applies.

This runs the statement inside an Oracle read-only transaction
(`SET TRANSACTION READ ONLY`) so the database itself rejects writes and locking
reads, and always ends the transaction before the connection goes back to the
pool. With the fix, the `UPDATE` above fails with `ORA-01456` on both drivers and
the row is unchanged.

### Why an explicit statement instead of `sql.TxOptions{ReadOnly: true}`

Neither supported driver implements that option usefully, verified against a
live Oracle Free 23 instance:

| | `BeginTx(TxOptions{ReadOnly:true})` | `BeginTx(nil)` + explicit `SET TRANSACTION READ ONLY` |
|---|---|---|
| godror | accepted, but an `UPDATE` still succeeds — **not enforced** | accepted, `UPDATE` → `ORA-01456` |
| go-ora | rejected: `readonly transaction is not supported` | accepted, `UPDATE` → `ORA-01456` |

godror issues its own `SET TRANSACTION` while the connection is still in
autocommit mode, committing away the transaction it just opened. Sending the
statement as the first statement inside the transaction works on both drivers
and raises no `ORA-01453`.

A source-level option (the AlloyDB/Cloud SQL `readOnly` pattern, which pins a
session parameter at connection time) is not available here: Oracle has no
session-level read-only mode — `ALTER SESSION SET READ ONLY` fails with
`ORA-02248`. Transaction scope is the only mechanism Oracle offers.

### Why `readOnly` stays a tri-state

`readOnly` is already a `*bool` (added in [#2323](https://github.com/googleapis/mcp-toolbox/pull/2323), commit `72146a4b`). Previously `nil` was
collapsed to `true` in the tool layer, so the source only ever saw two states.
It now reaches `RunSQL` as a third:

| value | behavior |
| --- | --- |
| `true` | read-only transaction; writes and locking reads rejected |
| `false` | run as DML, report `rows_affected` (unchanged) |
| `nil` | run as a query in a transaction that is committed at the end |

Neither two-state encoding works while `false` keeps its current meaning:

- folding `nil` into `true` breaks every configured tool that writes without
  setting `readOnly` — including this repo's own
  `insert-table-templateParams-tool`, whose `INSERT` would fail with
  `ORA-01456`;
- folding `nil` into `false` breaks every SELECT tool that does not set
  `readOnly`, since `ExecContext` returns no result set.

Spanner gets away with a plain `bool` because its `false` is a read-write
transaction that still returns rows. Oracle's `false` means "this is DML,
report affected rows". Collapsing this to a plain `bool` would mean redefining
`readOnly: false`, which is a breaking change and better handled separately —
happy to file a follow-up if maintainers prefer that direction.

### Known limits, and what to use instead

A read-only transaction is the strongest guarantee available *inside* a
connection, but it is not a security boundary, and the docs now say so plainly:

- **DDL is not blocked, and it is a bypass.** Oracle commits implicitly before
  and after DDL, which ends the read-only transaction, so `CREATE`, `TRUNCATE`
  and `ALTER` take effect under `readOnly: true`. Worse, a single statement that
  performs DDL before writing escapes entirely — verified: a PL/SQL block doing
  `EXECUTE IMMEDIATE '<DDL>'` followed by an `UPDATE` committed the update.
- **PL/SQL with `PRAGMA AUTONOMOUS_TRANSACTION`** likewise still writes
  (verified: a procedure called inside a read-only transaction inserted a row
  that survived the rollback).

So for tools that run untrusted or model-generated SQL — `oracle-execute-sql`
above all — the enforcement has to be in the database. `source.md` gains a
**Read-Only Access** section covering this: on Oracle Database 23ai and later,
`CREATE USER ... READ ONLY` makes every write from that user's sessions fail
with `ORA-28194`. Verified against the full battery, including the cases the
transaction cannot stop: DDL, `TRUNCATE`, the DDL-then-write bypass, autonomous
transactions, definer's-rights procedures, and DML over a database link. The one
gap documented is a procedure invoked *over* a database link, which runs on the
target database as the link's user. Pre-23ai guidance is least privilege.

One behavior change worth noting:

- **`ORA-01466`.** A read-only transaction takes a consistent snapshot, so
  selecting from a table whose DDL happened seconds earlier fails with
  `ORA-01466`. Dictionary views are unaffected, so the five prebuilt tools
  marked `readOnly: true` here are not exposed; a user-defined `readOnly: true`
  tool reading a just-created table would be.

### Follow-up not included here

Oracle's tools always default to `NewDestructiveAnnotations`, unlike
`spannersql`/`falkordb-execute-cypher`, which pick `NewReadOnlyAnnotations` when
`readOnly` is set. That predates this change and has no functional effect today
(the Oracle source's `IsReadOnly()` is hardcoded `false`, so the suppression
path in `tools.go` never fires), so it is left out to keep this PR focused.
Happy to send it separately.

### Breaking change?

Filed as `fix` rather than `fix!`. An explicit `readOnly: true` changes from
permissive to enforced, so a config that set `readOnly: true` on a writing
statement now fails — but the previous docs described that setting as
"**read-only** (SAFE mode)", so this makes the flag behave as documented.
Flagging it explicitly in case maintainers would rather label it breaking.

### Testing

- `go build ./...`, `go vet`, `gofmt -l`, `golangci-lint run` — all clean.
- `go test -race ./cmd/... ./internal/...` — 405 packages, all pass.
- `tests/oracle` against a live Oracle Free 23 instance, covering: `ORA-01456`
  for locking reads and for writes; no lock leak (a second session doing
  `SELECT ... FOR UPDATE NOWAIT` after each tool call, including after
  `oracle-execute-sql` takes locks on the permissive path); and no write
  regression — template-param `CREATE`/`INSERT` and execute-sql `CREATE`/`DROP`
  all still persist.
- Enforcement was additionally verified by hand on **both** drivers. Note the
  existing `tests/oracle` suite connects with `useOCI: true`, so the go-ora path
  is not covered by CI; with `useOCI: false`, `SET TRANSACTION READ ONLY` is
  accepted and `ORA-01456` still fires.
- The Oracle 23ai read-only user documented in `source.md` was verified on
  Oracle Database Free 23.26 with write privileges explicitly granted, so the
  `READ ONLY` attribute is isolated rather than confounded with privilege errors.
- The shared suite's Google-auth subtests (`RunToolInvokeTest`,
  `RunMCPToolCallMethod`, `RunExecuteSqlToolInvokeTest`) need gcloud/ADC and
  were not runnable locally; they are unmodified by this PR.

## PR Checklist

Fixes #3987


anubhav756:
Thanks for contributing the read-only feature for Oracle @Luap2003 . I will take a closer look and get back with any thoughts or questions! 🙂

anubhav756:
This is actually a known issue. Thanks for the PR, I will try and resolve this ASAP.

PR #3989, opened 2026-09-09
feat(resources): allow skill:// URIs on file resources

twishabansal:
## Description

A resource that belongs to an Agent Skill is addressed under `skill://` so resources have to accept that scheme alongside their own. Without this, a skill's files cannot be declared at all.

### DIFF
--- a/internal/resources/resources.go
+++ b/internal/resources/resources.go
@@ -36,6 +36,22 @@ const UIMimeType = "text/html;profile=mcp-app"
 // BaseDirKey is the context key for storing the base directory path during config parsing.
 const BaseDirKey contextKey = "baseDir"
 
+// SkillScheme is how a resource belonging to an Agent Skill is addressed,
+// as skill://<skill-name>/<path>, whichever resource type backs it.
+const SkillScheme = "skill"
+
+// ValidateScheme checks uri against the schemes a resource may be addressed by:
+// nativeScheme, which is the resource's own type (eg. file), or SkillScheme. The
+// returned error names the accepted set, so callers need only prefix it with the
+// resource they were validating.
+func ValidateScheme(uri, nativeScheme string) error {
+	parsed, err := url.Parse(uri)
+	if err != nil || (parsed.Scheme != nativeScheme && parsed.Scheme != SkillScheme) {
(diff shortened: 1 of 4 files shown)


internal/resources/resources.go
```diff
@@ -34,6 +34,23 @@ type contextKey string
...
+// after the resource's own type — or SkillScheme. Callers format their own
+// error so it can name the type; this only decides the rule, and it lives here
+// so every resource type answers the question the same way.
+func SchemeAllowed(uri, nativeScheme string) bool {
+	parsed, err := url.Parse(uri)
+	if err != nil {
+		return false
+	}
+	return parsed.Scheme == nativeScheme || parsed.Scheme == SkillScheme
+}
```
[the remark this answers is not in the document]
twishabansal: The scheme is already normalized before this comparison. `net/url.Parse` lowercases it internally ([url.go#L455](https://github.com/golang/go/blob/go1.26.7/src/net/url/url.go#L455)), and `ResourceConfigBase.Validate()` lowercases scheme and host again at `internal/resources/resources.go:193-194` before the type-specific check runs. The other operand is always the lowercase const `resourceType` (`internal/resources/file/file.go:38`).

Added test cases for validation.

internal/resources/resources.go
```diff
@@ -36,6 +36,22 @@ const UIMimeType = "text/html;profile=mcp-app"
...
 
+// SkillScheme is how a resource belonging to an Agent Skill is addressed,
+// as skill://<skill-name>/<path>, whichever resource type backs it.
+const SkillScheme = "skill"
+
+// ValidateScheme checks uri against the schemes a resource may be addressed by:
+// nativeScheme, which is the resource's own type (eg. file), or SkillScheme. The
+// returned error names the accepted set, so callers need only prefix it with the
+// resource they were validating.
+func ValidateScheme(uri, nativeScheme string) error {
```
dishaprakash, line 47: nit: Can we add unit test for this

dishaprakash (approved):

internal/resources/resources.go
twishabansal: Thanks for pointing out. Done!

anubhav756 (approved):

PR #4014, opened 2026-09-11
refactor(resources): key the runtime resource and template maps by URI

twishabansal:
## Description

The upcoming `directory` resource type enumerates one resource per file, and those derived resources have no operator-given name to key the map by. Rather than invent a key format for them, this keys the runtime resource and resource template maps by URI — every resource and template already has a unique, non-empty one. `ResourceConfigBase.Validate` fills in `<type>://<name>` when `uri` is omitted, `uriTemplate` is required, and config parsing already rejects duplicates across both.

### DIFF
--- a/internal/server/primitives/primitives.go
+++ b/internal/server/primitives/primitives.go
@@ -121,9 +121,7 @@ func (r *PrimitiveManager) GetPrompt(promptName string) (prompts.Prompt, bool) {
 }
 
 // GetResource returns a specific resource by name or by URI. Names are resolved
-// first: resource names are not validated, so a name is allowed to look like a
-// URI, and an operator's existing name must always win over a URI that collides
-// with it.
+// first.
 func (r *PrimitiveManager) GetResource(nameOrURI string) (resources.Resource, bool) {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
--- a/internal/server/primitives/primitives.go
+++ b/internal/server/primitives/primitives.go
@@ -43,6 +43,20 @@ type PrimitiveManager struct {
 	resources         map[string]resources.Resource
 	resourceTemplates map[string]resources.ResourceTemplate
 	groups            map[string]group.Group
+	// resourceNameIndex maps a resource's name to its URI, the key of resources.
+	resourceNameIndex map[string]string
+}
+
+// indexResourceNames builds the name to URI index for the resource map.
+func indexResourceNames(resourcesMap map[string]resources.Resource) map[string]string {
+	index := make(map[string]string, len(resourcesMap))
+	for uri, res := range resourcesMap {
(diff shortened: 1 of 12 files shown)


internal/server/primitives/primitives.go
```diff
@@ -43,6 +43,20 @@ type PrimitiveManager struct {
...
+func indexResourceNames(resourcesMap map[string]resources.Resource) map[string]string {
+	index := make(map[string]string, len(resourcesMap))
+	for uri, res := range resourcesMap {
+		if res == nil {
+			continue
+		}
+		index[res.GetName()] = uri
+	}
+	return index
 }
```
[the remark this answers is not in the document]
twishabansal: Empty names can't reach this — ConfigBase.Name has validate:"required", so name: "" is rejected at parse time.

internal/server/primitives/primitives.go
```diff
@@ -43,6 +43,20 @@ type PrimitiveManager struct {
...
+// indexResourceNames builds the name to URI index for the resource map.
+func indexResourceNames(resourcesMap map[string]resources.Resource) map[string]string {
+	index := make(map[string]string, len(resourcesMap))
+	for uri, res := range resourcesMap {
+		if res == nil {
+			continue
+		}
+		index[res.GetName()] = uri
+	}
+	return index
```
Yuan325, line 62: Instead, should we have the key as uri, and the value as the name?

That way we can use the map to find the name based on URI, and retrieve the resources via name.

internal/server/primitives/primitives.go
```diff
@@ -105,11 +120,16 @@ func (r *PrimitiveManager) GetPrompt(promptName string) (prompts.Prompt, bool) {
 	return prompt, ok
 }
 
-// GetResource returns a specific resource by name.
-func (r *PrimitiveManager) GetResource(name string) (resources.Resource, bool) {
+// GetResource returns a specific resource by name or by URI. Names are resolved
+// first.
+func (r *PrimitiveManager) GetResource(nameOrURI string) (resources.Resource, bool) {
```
Yuan325, line 142: Is there a place that we are calling `GetResource()` without knowing if its a name of URI?

I think it might be better to keep this function as is. And we can introduce a new function `GetResourceNameByURI()` or `GetResourceByURI()` that will check for the resource name via `uri`, then retrieve the resource with `GetResource()`.


With this being implemented, we might be able to also dedup and remove the `GetUIResourceFromURI()` function.

Yuan325 (changes requested):
Wondering if we can just build the map with using uri as key during initialization?

twishabansal:
For directory type resources, we will auto-generate names for the derived file resources. We can continue using names as the key.

PR #4017, opened 2026-09-11
feat(skills): validate skills at startup

twishabansal:
## Description

The server validates every skill at startup. Broken means: no SKILL.md, unparseable frontmatter, a frontmatter name that disagrees with the URI, or an unreadable file.

It also bounds what one skill loads. A skill caps at 16 MiB. A file resource accepts a maxSize up to 1 GiB.

Entries are discarded on purpose. skills/list and skills/get rebuild them later against current content, so an edit yields a new digest and the host recovers through the SEP's refresh path.


dishaprakash (approved):

internal/skills/discover.go
```diff
@@ -115,16 +115,29 @@ func buildEntry(ctx context.Context, root string, members []resources.Resource)
 
 	refs := make([]ResourceRef, 0, len(members))
 	var frontmatter map[string]any
+	// Manifest.Validate enforces the same limit, but only once every file is in
+	// memory. We sum as we read: this bounds what each skill loads.
+	var total int64
 	for _, res := range members {
+		// Subtraction, not addition: a huge hint would wrap the total negative.
+		if sz := res.GetSize(); sz != nil && *sz > MaxTotalSize-total {
```
anubhav756: optional: Do we also want to think about having `MaxTotalSize` and `MaxRefs` as config overridable? Note that the spec says: 

Image

anubhav756 (approved):

internal/skills/discover.go
twishabansal: The server says that for hosts, which I assume means client. For the server, it explicitly says should not. Do you think there's a misunderstanding here?

PR #1, opened 2024-07-12
chore(ci): set up ci

kurtisvg:
Sets up basic CI for the repo, including:
- CODEOWNERS
- header-checker-lint
- labels
- sync-repo-settings
- golangci-lint
- tests on pr/merge

PR #4019, opened 2026-09-11
fix(telemetry): circuit-break GCP trace export on persistent PermissionDenied

### LINKED ISSUE #3638: --telemetry-gcp: back off / disable trace export after persistent PermissionDenied instead of retrying every ~5s
### Prerequisites

Closest found: #2285 (closed — ADC env var required for GCP telemetry locally) and #2222 (closed — general telemetry enhancement). Neither covers exporter behavior under persistent permission failure.

Run Toolbox on Cloud Run with `--telemetry-gcp` under a least-privilege service account — and still have readable logs when one telemetry permission is missing.

Our service account had `roles/monitoring.metricWriter` (so metric export worked) but not the Cloud Trace equivalent. Result: the trace exporter retried forever, emitting this pair every ~5 seconds for the life of the container:

```
failed to export to Google Cloud Trace: rpc error: code = PermissionDenied
desc = Permission 'cloudtrace.traces.patch' denied on resource
'//logging.googleapis.com/projects/<project>' (or it may not exist).
```

The failure is *permanent* — IAM grants don't appear spontaneously — but the exporter treats it as transient. During a production debugging session (with `--log-level=DEBUG` enabled to diagnose an unrelated tool failure), this noise dominated the log stream and materially slowed the investigation. The permission error itself never needs more than one line to communicate.

### Suggested Solution(s)

1. **Circuit-break on persistent PermissionDenied:** after N consecutive `PermissionDenied` export failures (e.g. 3), log a single WARN naming the missing permission and the exact role to grant (`roles/cloudtrace.agent`), then disable that exporter for the process lifetime (or back off to something like once-per-hour re-probe). `PermissionDenied` is worth distinguishing from `Unavailable`/`DeadlineExceeded`, which genuinely are transient.
2. **Document the full IAM requirements of `--telemetry-gcp` in one place:** it needs *both* `roles/monitoring.metricWriter` (metrics) and `roles/cloudtrace.agent` (traces). The partial-grant state (metrics working, traces spamming) is exactly the one a least-privilege deployment lands in first.
3. Optional: split the flag (`--telemetry-gcp-metrics` / `--telemetry-gcp-traces`) or auto-disable the trace half when the role is absent at startup — a startup-time probe would surface the misconfiguration before any query traffic.

### Alternatives Considered

- **Grant the missing role** (what we did): correct end state, but the operator only learns the role name by reading the spam; nothing in the docs or startup output lists it.
- **Log-based exclusion filters:** hides the noise from humans while still paying to generate and ingest it (Cloud Logging bills the excluded volume's ingestion path in effort, and DEBUG sessions still drown).
- **Dropping `--telemetry-gcp`:** loses the metrics that were working fine.

### Additional Details

Observed with Toolbox on Cloud Run (v1.3.x, container built from the official binary). The metrics exporter behaved perfectly throughout — this is solely about the trace exporter's retry behavior under a permanent authorization failure, plus documentation of the roles pair.

Ashfaqbs:
**1. Description**

`--telemetry-gcp`'s trace exporter (wrapped via `tracesdk.WithBatcher`) treats every export failure as transient and just retries on the next batch interval (~5s by default). A service account missing `roles/cloudtrace.agent` produces a `PermissionDenied` that will never clear on its own — IAM grants don't appear spontaneously — so the exporter spams the same `failed to export to Google Cloud Trace: rpc error: code = PermissionDenied ...` line roughly every 5 seconds for the life of the process. This is exactly the state a least-privilege deployment (metrics working, traces missing the separate role) lands in first, and it drowned out an unrelated `--log-level=DEBUG` investigation for the reporter.

Wrap the GCP trace exporter in a small circuit breaker (`circuitBreakingTraceExporter` in `internal/telemetry/gcp_trace_exporter.go`): after 3 consecutive `PermissionDenied` failures it logs one `WARN` naming the required IAM role and disables itself for the rest of the process lifetime, silently dropping further export attempts instead of calling through to the real exporter. Any other error code (`Unavailable`, `DeadlineExceeded`, etc.) always passes through unchanged and never counts toward the threshold, since those genuinely are worth retrying. A success resets the counter, so a transient blip doesn't accumulate toward tripping the breaker.

The IAM-roles documentation half of #3638 (the pair of roles required, and a callout for this exact partial-grant scenario) was already covered by #3824 — this PR addresses only the remaining "stop the noise" half (issue's suggested fix #1).

**2. PR Checklist**
- [x] Appropriate docs were updated (if necessary) — N/A, the docs half is already merged in #3824
- [ ] Make sure to add `!` if this involves a breaking change — not applicable

**3. Issue Reference**

Fixes #3638 🦕

## Test plan

- `go test -race ./internal/telemetry/...` — new file `gcp_trace_exporter_test.go` covers: a success passing through and resetting the counter; a non-`PermissionDenied` error always passing through without ever tripping the breaker; the breaker tripping on exactly the 3rd consecutive `PermissionDenied` and returning `nil` from then on without calling the underlying exporter again; a success or a different error code resetting the consecutive-failure count so a later `PermissionDenied` streak starts over; `Shutdown` delegating to the wrapped exporter.
- `go build ./...` and `go test ./cmd/... ./internal/...` — full build and test suite pass, no regressions.
- `go vet ./internal/telemetry/...` — clean.

## Caveats

- The one-time warning is logged via the standard `log/slog` default logger rather than this project's own `internal/log.Logger`, since threading a logger into `telemetry.SetupOTel`/`newTracerProvider` would require an import of `internal/log` at that call depth and (via `internal/util`, which already imports `internal/telemetry`) risks an import cycle if wired through the existing context-based logger accessor. Given the very narrow scope (one WARN, once, ever, only on the `--telemetry-gcp` + missing-role path), I judged the plumbing not worth the added surface — happy to revisit if a maintainer prefers it routed through the app's structured logger.
- Chose "disable permanently after 3 consecutive failures" per the issue's own suggested fix #1, rather than the "once-per-hour re-probe" alternative it also mentions — simpler, and correct for the common case (a missing IAM grant doesn't get fixed without a restart anyway, since it's read at exporter construction time).

### DIFF
--- a/internal/telemetry/gcp_trace_exporter.go
+++ b/internal/telemetry/gcp_trace_exporter.go
@@ -1,0 +1,83 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package telemetry
+
+import (
+	"context"
+	"log/slog"
+	"sync/atomic"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// gcpTracePermissionDeniedThreshold is the number of consecutive
+// PermissionDenied export failures after which the exporter disables itself
+// for the rest of the process lifetime. A PermissionDenied on trace export
+// means the service account is missing an IAM grant, which does not appear
+// spontaneously, unlike Unavailable/DeadlineExceeded which are genuinely
+// worth retrying.
+const gcpTracePermissionDeniedThreshold = 3
+
+// gcpTracePermissionDeniedRole is named in the one-time warning so an
+// operator does not have to read the raw error to find it.
+const gcpTracePermissionDeniedRole = "roles/cloudtrace.agent"
+
+// circuitBreakingTraceExporter wraps a Cloud Trace span exporter and stops
+// calling it after gcpTracePermissionDeniedThreshold consecutive
+// PermissionDenied failures, logging a single WARN instead of the
+// underlying exporter's own per-export error log (the batch span processor
+// otherwise retries on every export interval, ~5s by default, for the life
+// of the process).
+type circuitBreakingTraceExporter struct {
+	sdktrace.SpanExporter
+	consecutivePermissionDenied atomic.Int32
+	disabled                    atomic.Bool
+}
+
+// newCircuitBreakingTraceExporter wraps exp with the PermissionDenied
+// circuit breaker described above.
+func newCircuitBreakingTraceExporter(exp sdktrace.SpanExporter) sdktrace.SpanExporter {
+	return &circuitBreakingTraceExporter{SpanExporter: exp}
+}
+
+// ExportSpans implements sdktrace.SpanExporter.
+func (e *circuitBreakingTraceExporter) ExportSpans(ctx context.Context, spans []sdktrace.ReadOnlySpan) error {
+	if e.disabled.Load() {
+		return nil
+	}
+	err := e.SpanExporter.ExportSpans(ctx, spans)
+	if err == nil {
+		e.consecutivePermissionDenied.Store(0)
+		return nil
+	}
+	if status.Code(err) != codes.PermissionDenied {
+		e.consecutivePermissionDenied.Store(0)
+		return err
+	}
+	if e.consecutivePermissionDenied.Add(1) < gcpTracePermissionDeniedThreshold {
+		return err
+	}
+	if e.disabled.CompareAndSwap(false, true) {
+		slog.WarnContext(ctx,
+			"disabling Google Cloud Trace export after repeated PermissionDenied failures; grant the missing role to the service account to re-enable",
+			"consecutiveFailures", gcpTracePermissionDeniedThreshold,
+			"requiredRole", gcpTracePermissionDeniedRole,
+			"error", err.Error(),
+		)
+	}
+	return nil
+}
--- a/internal/telemetry/gcp_trace_exporter_test.go
+++ b/internal/telemetry/gcp_trace_exporter_test.go
@@ -1,0 +1,166 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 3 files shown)


AndyFooBlah:
@Ashfaqbs thanks for picking this up, and sorry — there's context buried in #3638 that isn't visible from the issue body, and it may affect this PR's direction. Flagging it now rather than after a review round.

Your implementation is a faithful build of suggested fix #1 as I originally wrote it, and the details are right in ways the issue didn't spell out: distinguishing `PermissionDenied` from `Unavailable`/`DeadlineExceeded`, resetting the counter on success so a transient blip doesn't accumulate, and logging the role name once. No complaints about the code.

The problem is that I withdrew that proposal. In [this comment](https://github.com/googleapis/mcp-toolbox/issues/3638#issuecomment-5287246844) @averikitsch pushed back on the whole approach:

> Toolbox focuses on fast failing if sources are not configured correctly. If a developer wants to enable telemetry and the user permissions are not available should the server fail to start versus swallowing the error?

I agreed and dropped the circuit-breaker. A breaker is precisely the "swallow the error" behavior she was steering away from — it keeps a misconfigured server running quietly, where the project's preference is to fail loudly at startup. So the remaining work on that issue was scoped to a startup probe that exits non-zero on a permanent authorization failure, not a runtime breaker.

Two other things worth knowing:

1. **Fail-fast can't reuse this code path.** Exporter construction (`texporter.New`) succeeds without touching the API — the first `PermissionDenied` only appears when the batch processor exports, ~5s after the server is already serving. So it needs an explicit probe before serving begins, which is a different shape from a breaker wrapping the exporter.
2. **The issue was relabelled** by @Yuan325 on 08-25, dropping `type: feature request`/`type: question` for `type: process`, right after the docs half (#3824) merged. I asked [whether that means the behavior change is off the table entirely](https://github.com/googleapis/mcp-toolbox/issues/3638#issuecomment-5465362470) and haven't had an answer, so the status of *any* code change there is genuinely unclear right now.

To be clear, none of this is my call — I filed the issue, I don't speak for the maintainers, and it's entirely possible they'd rather have a breaker than a probe. I'd just rather you hear it from the thread than spend another round on it. If a maintainer confirms they want the breaker, disregard all of this; your PR is in better shape than my original sketch.

If it turns out they want the probe instead, I'm happy for you to take that too — I'd been holding off only because I couldn't get the unconditional-vs-escape-hatch question answered. Either way I'd rather it land than sit on my offer indefinitely.

Ashfaqbs:
Thanks for the thorough flag, @AndyFooBlah — appreciated, and no need to apologize. I read the issue body but missed the pivot buried in the comments, which is exactly the research-before-PR miss you're pointing at. That's on me, not a gap in how you wrote it up.

Given the actual state — @averikitsch's fail-fast preference, your withdrawal of the breaker, and the unanswered escape-hatch question after the `type: process` relabel — I don't think it's right for me to guess a direction and rebuild this PR a second time. So:

@averikitsch @Yuan325 — could one of you confirm which of these you'd actually want, now that it's in front of you as working code either way:
1. This PR as-is (the circuit-breaker), if the fail-fast preference from August has since changed, or
2. Closed in favor of a startup probe that exits non-zero on a permanent `PermissionDenied` (unconditional, or with an escape hatch — @AndyFooBlah's open question from #3638), or
3. Neither — #3824 already covers what this issue needed and the behavior half is closed as out of scope.

Happy to build whichever of 1/2 you'd rather have, or close this PR outright if it's 3. Just want a real answer this time before writing more code against it.

Issue #3706, opened 2026-07-24
Migrate all integration tests to `/mcp` endpoint


anubhav756:
We want to migrate integration tests in MCP Toolbox from the legacy `/api` endpoints to the `/mcp` endpoints.

# Background

Completing this migration is a prerequisite for removing the legacy `/api` code paths from the server source code (currently gated behind the `--enable-api` flag, [#2846](https://github.com/googleapis/mcp-toolbox/pull/2846)) and cleaning up the legacy test harness.

The migration of the integration test suite across 40+ database plugins has been challenging due to:

* Test fragmentation  
* Inner loop breakdown

These challenges heavily inform the current tiered strategy. Tier 1 aims to break free from this tech debt by creating clean, isolated tests, while Tier 2 accepts some complexity to ensure rapid coverage for non-core databases.

# Proposed Strategy

For this large migration, we propose assigning tiers to database sources based on their importance and the desired end-state quality.

## Tier 1: Native MCP optimized tests

Tests are migrated to native MCP integration tests using a thin test harness built solely for `/mcp` endpoints from the ground up for some of the P0 GCP DBs. It contains no traces of the `/api` endpoints or its legacy harness.

### Progress

* PRs merged:  
  * [\#2842](https://github.com/googleapis/mcp-toolbox/pull/2842) (http)  
  * [\#2843](https://github.com/googleapis/mcp-toolbox/pull/2843) (alloydb)  
  * [\#2921](https://github.com/googleapis/mcp-toolbox/pull/2921) (alloydbainl)  
  * [\#2828](https://github.com/googleapis/mcp-toolbox/pull/2828) (test harness)  
  * [\#2991](https://github.com/googleapis/mcp-toolbox/pull/2991) (refactor harness)  
* PRs in-flight:  
  * [\#2992](https://github.com/googleapis/mcp-toolbox/pull/2992) (alloydbpg)  
  * [\#2994](https://github.com/googleapis/mcp-toolbox/pull/2994) (cloudsqlpg)  
  * [\#2964](https://github.com/googleapis/mcp-toolbox/pull/2964) (alloydbomni)  
  * [\#3014](https://github.com/googleapis/mcp-toolbox/pull/3014) (achieve parity)

## Tier 2: Options-based dual-target tests

[...]

PR #4033, opened 2026-09-14
test(source/valkey): add MCP integration tests

smadikanti:
## Description

Added `valkey_mcp_test.go` with `TestValkeyMCPListTools` and `TestValkeyMCPCallTools`, following #2922. The API tests are still there. Both suites share the setup and call cases, and MCP uses `WithMCP()` plus the common manifest builder with Valkey's command-array schema.

Checked against local Valkey 8: API and MCP discovery passed, along with the public invocation cases on both endpoints. Command-package race tests and lint passed too.

Local integration runs used a temporary port overlay because port 5000 is occupied on my Mac. The public invocation checks also used a local-only overlay to bypass eager token setup and skip auth cases. Neither overlay is in the PR. Full integration still needs a Google audience and ADC, so I need CI for the authenticated cases.

## PR Checklist

- [ ] Ensure the tests and linter pass (local checks above pass; authenticated integration needs CI)
- [x] Code coverage does not decrease (existing cases retained; no production code changed)
- [x] Appropriate docs were updated (no docs changes needed)
- [x] Make sure to add `!` if this involves a breaking change (not applicable)

Related to #3706


AlexTalreja:
Hi @smadikanti, could you update this PR and other migrations to have the migrated `/mcp` tests in a new file similar to the convention in https://github.com/googleapis/mcp-toolbox/pull/2922? 

Additionally, here are some guidelines we've been given regarding the migration:
1. Every integration test should have this 2 test functions: `Test<>ListTools` and `Test<>CallTools`
2. Let's use the options method (ref: https://github.com/googleapis/mcp-toolbox/pull/2922) to refactor existing test functions to take both the /api and /mcp endpoint. With this, we won't have to worry if there are test cases where it was tested in the api but not the mcp endpoint. If needed, we can update the following: 
update `api string` or `url string` to `toolName string` so that it could be used for both api and mcp endpoint.
update `args string` to be `args map[string]any` so that it could be used for both api and mcp endpoint.
add `contentError bool` as needed if we want to check the `content[0].IsError` value.
3. utilize the `GetBaseMCPExpectedTools()`, `GetExecuteSQLMCPExpectedTools()`, and `GetTemplateParamMCPExpectedTools()` to retrieve common tools' `[]MCPToolManifest`. Every source should have the same items for these if they are tested fully. We can add source-specific ones as needed.

PR #4036, opened 2026-09-14
build: output OCI tarball to staging AR for MOSS compliance

camposperez-hub:
## Description

Updates `.ci/versioned.release.cloudbuild.yaml` to output the container image as an OCI tarball (`/workspace/image.tar`) instead of pushing directly, and uploads the tarball artifact to the staging Artifact Registry (`us-central1-docker.pkg.dev/database-toolbox/toolbox-containers-staging/toolbox`) for MOSS compliance.

## PR Checklist

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/mcp-toolbox/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

🛠️ Fixes #


.ci/versioned.release.cloudbuild.yaml
```diff
@@ -32,11 +32,7 @@ steps:
         --build-arg BUILD_TYPE=container.release \
         --build-arg COMMIT_SHA=$(git rev-parse --short HEAD) \
         $TAGS \
-        --push .
-
-      # Pull the image to ensure it's available for provenance generation.
-      docker pull ${_DOCKER_URI}:${VERSION}
-      docker tag ${_DOCKER_URI}:${VERSION} ${_DOCKER_URI}:latest
+        --output=type=oci,dest=/workspace/image.tar .
```
duwenxin99, line 36: Why do we need this

camposperez-hub: Ok, after investigating a bit more, since we changed docker buildx to output the image directly to a file (--output=type=oci,dest=/workspace/image.tar), it is no longer loaded into the local Docker daemon. Because of this, docker pull and docker tag will fail since the daemon can't find the image locally. 
Cloud Build's native artifacts: block now handles the upload directly from the tarball file instead

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -480,3 +473,8 @@ substitutions:
   _EXIT_GATE_PROJECT: "mcp-toolbox"
   _EXIT_GATE_MANIFEST_BUCKET: "oss-exit-gate-prod-projects-bucket"
   _PUSH_TO_EXIT_GATES: "false"
+
+artifacts:
+  oci:
+    - file: '/workspace/image.tar'
```
duwenxin99, line 480: how do we tag it with the version and `latest` tag?

camposperez-hub: The tags are embedded inside the OCI tarball during the docker buildx step (via the -t flags in $TAGS). 
When Cloud Build uploads the tarball using the artifacts block, Artifact Registry automatically extracts those embedded tags and applies both the version and latest tags to the image.

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -480,3 +474,8 @@ substitutions:
...
+      registry_path: '${_DOCKER_URI}'
```
duwenxin99: This should be `registryPath` right? See https://docs.cloud.google.com/build/docs/build-config-file-schema#oci Also seems like we need to use the absolute path.

duwenxin99 (changes requested):

.ci/versioned.release.cloudbuild.yaml
camposperez-hub: I will update the YAML to use registryPath (camelCase) to match the official Cloud Build schema. Also, the file path /workspace/image.tar is already an absolute path, so it should be good to go

duwenxin99:
@emagda97 Could you take a look at this PR and make sure your changes can be rebased on it?

duwenxin99 (approved):

duwenxin99 (commented):
Could you test the final script with a test trigger?

duwenxin99 (changes requested):
We should wait until the Exit Gate promotion config CL is merged before merging this PR.

PR #4044, opened 2026-09-15
feat(skills): publish SKILL.md under its frontmatter identity

twishabansal:
## Description

SEP-2640 fixes a SKILL.md as `text/markdown`, and takes its `name` and `description` from the frontmatter, not from the resource that backs it. Otherwise a client sees a resource called `SKILL.md`, which does not say which skill it belongs to.

Discovery already parses and validates the frontmatter, so this wraps the SKILL.md resource with the identity it declares. A wrapper keeps SKILL.md's shape out of `file`, `text`, and later `directory`, and applies on every protocol version.

The metadata is fixed until the next reload. Validation ties `frontmatter.name` to the final URI segment, so only `description` can drift.

Depends on #4014


twishabansal:
**TODO (rebase onto #4014):** collapse the apply loop in `InitializeConfigs`.

#4014 keys the runtime resource map by URI (`resourcesMap[r.GetURI()] = r`).
Once this rebases onto it, `key` and `r.GetURI()` are the same string, so the
loop translates a key into itself:

```go
for key, r := range resourcesMap {
    if doc, ok := docs[r.GetURI()]; ok {
        resourcesMap[key] = doc
    }
}

It reduces to maps.Copy(resourcesMap, docs).

Flagging it because
#4014 edits server.go:250 and this edits 260-274 — the rebase will likely
merge clean and leave the redundant loop in place.

File: internal/skills/metadata.go [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+		if !ok {
+			continue
+		}
+		name, nameOK := e.Frontmatter["name"].(string)
+		desc, descOK := e.Frontmatter["description"].(string)
+		if !nameOK || !descOK {
+			continue
+		}
+		docs[res.GetURI()] = skillDoc{Resource: res, name: name, description: desc}
+	}
```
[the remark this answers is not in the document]
twishabansal: Errored out!

dishaprakash (approved):

internal/skills/metadata.go
```diff
@@ -0,0 +1,70 @@
...
+// skillDoc presents a SKILL.md under the name and description its frontmatter
+// declares. Otherwise a client sees a file resource named "SKILL.md" which
+// does not identify the skill.
+type skillDoc struct {
+	resources.Resource
+	name        string
+	description string
+}
+
+func (s skillDoc) GetName() string        { return s.name }
```
anubhav756: Is this override really reaching the client? The `resources/list` seems to pass the config key, not `res.GetName()` (`GenerateListResourcesResult` in `manifests.go`). Other protocol versions also do the same.

I checked it with a real binary, using a `file` resource named `guide` at `skill://analytics-guide/SKILL.md` with `name: analytics-guide` in its frontmatter. `resources/list` returned:

`description` and `mimeType` get overridden correctly, but `name` doesn't.

The test in `server_test.go` also only calls `GetName()` on the in-memory resource, so it passes anyway. I think a test that goes through `GenerateListResourcesResult` would have caught this.

internal/skills/metadata_test.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	if len(docs) != 1 {
+		t.Fatalf("got %d doc resources, want 1: %v", len(docs), docs)
+	}
+
+	doc, ok := docs["guide"]
+	if !ok {
+		t.Fatalf("no replacement for %q", skillURI)
+	}
+	if got := doc.GetName(); got != "analytics-guide" {
```
anubhav756: This checks seems like in-memory wrapper, but not what a client receives. See my comment [above](https://github.com/googleapis/mcp-toolbox/pull/4044/changes#r4093261171).

anubhav756 (changes requested):

internal/skills/metadata.go
twishabansal: Thanks for checking this. This was done in a PR somewhere but might have gotten lost in the stack. But thank you for pointing this out. I have fixed this!

internal/skills/metadata_test.go
twishabansal: Added a test to server_test.go to ensure this is caught!

PR #4046, opened 2026-09-15
test(source/bigquery): add MCP integration tests

arijitbakshi15-stack:
## Description

Adds BigQuery integration tests for the `/mcp` endpoint in a new file, and refactors the BigQuery test helpers to run against both endpoints, following the convention in #2922 and the guidelines from the review. Part of #3706.

### New file

`tests/bigquery/bigquery_mcp_test.go`, all starting the server without `--enable-api`:

- `TestBigQueryMCPListTools` verifies the complete `tools/list` manifest for all 48 tools, built from `GetBaseMCPExpectedTools()`, `GetExecuteSQLMCPExpectedTools()` and `GetTemplateParamMCPExpectedTools()` plus the BigQuery-specific entries (data-type, client-auth, semantic search and the 27 prebuilt tools).
- `TestBigQueryMCPCallTool` runs the shared helpers with `WithMCP()`, `WithMCPTemplate()` and `WithMCPExec()`, and `RunMCPToolCallMethod`, which already ran over `/mcp`.
- `TestBigQueryMCPToolWithDatasetRestriction`, `TestBigQueryMCPWriteModeAllowed`, `TestBigQueryMCPWriteModeBlocked`, `TestBigQueryMCPWriteModeProtected` and `TestBigQueryMCPReadOnlyVulnerabilityBlock` cover the remaining scenarios over MCP, so no case is exercised on one endpoint only.

### Options method

Per the review, the BigQuery-specific helpers now take `...tests.ToolExecOption` and describe each case in an endpoint-neutral way (`toolName string`, `args map[string]any`), evaluated by one runner with a single `if config.IsMCP() { ... } else { ...legacy... }`. The legacy tests call them unchanged and keep running over `/api`.

Shared harness:

- `RunSearchCatalogToolTest` and `RunSemanticSearchToolInvokeTest` accept the same option. Defaults are unchanged, so Spanner, Postgres, Cloud SQL Postgres, ClickHouse, Neo4j, Elasticsearch and SingleStore are unaffected.
- `ToolExecConfig` gets an exported `IsMCP()` so source-specific helpers can select the endpoint the same way the shared helpers do.
- `RunToolInvokeTest`: over MCP, a client-OAuth tool called without a token, or with an invalid one, is rejected with a JSON-RPC error and HTTP 401 (`internal/server/mcp/v20251125/method.go`). `RunMCPToolCallMethod` already asserts this in BigQuery CI; the two cases previously expected HTTP 200. BigQuery is the only caller of `EnableClientAuthTest`.

### Validation

- The expected manifests come from real `tools/list` output. I started the Toolbox with this exact config, a placeholder `authorized_user` ADC file and a placeholder Gemini key, so no Google calls were made, and the strict check passed. A deliberately altered description failed.
- The error cases that never reach Google (missing parameter, `authRequired` without a token, client-OAuth without a token) pass through the new runner on **both** endpoints: once with `--enable-api` on `/api`, once over `/mcp`. Negative controls with wrong expected text failed on both paths.
- Case parity with `main` was checked helper by helper; the `/api` case tables are unchanged in content.
- `go build ./...`, `go vet ./tests/...`, `gofmt` and `golangci-lint run ./tests/bigquery/... ./tests/` are clean.

### Not yet verified

Every case that reaches BigQuery, Dataplex, Gemini or Google token verification needs real credentials and a project, which I don't have locally. Those need a maintainer CI run.

### Notes for reviewers

- #2561 changes `tests/bigquery/bigquery_integration_test.go` significantly and will likely conflict.
- On `main`, `RunToolInvokeWithTemplateParameters` sends only the select-fields case, because the others lack `enabled: true`. #4006 fixes that in the shared harness; the MCP expectations here are already in the right shape for it.
- The MCP scenario tests (restriction, write modes, read-only) duplicate their setup, as `cloud_sql_mysql_mcp_test.go` does. Happy to drop them from this PR if the extra CI time is a concern.

## PR Checklist

- [ ] Ensure the tests and linter pass (linter passes; integration tests need CI)

Part of #3706 🦕


AlexTalreja:
Hi @arijitbakshi15-stack, could you update the PR and other migrations to have the migrated `/mcp` tests in a new file similar to the convention in https://github.com/googleapis/mcp-toolbox/pull/2922? 

Additionally, here are some guidelines we've been given regarding the migration:
1. Every integration test should have this 2 test functions: `Test<>ListTools` and `Test<>CallTools`
2. Let's use the options method (ref: https://github.com/googleapis/mcp-toolbox/pull/2922) to refactor existing test functions to take both the /api and /mcp endpoint. With this, we won't have to worry if there are test cases where it was tested in the api but not the mcp endpoint. If needed, we can update the following: 
update `api string` or `url string` to `toolName string` so that it could be used for both api and mcp endpoint.
update `args string` to be `args map[string]any` so that it could be used for both api and mcp endpoint.
add `contentError bool` as needed if we want to check the `content[0].IsError` value.
3. utilize the `GetBaseMCPExpectedTools()`, `GetExecuteSQLMCPExpectedTools()`, and `GetTemplateParamMCPExpectedTools()` to retrieve common tools' `[]MCPToolManifest`. Every source should have the same items for these if they are tested fully. We can add source-specific ones as needed.

arijitbakshi15-stack:
@AlexTalreja I have made the changes

PR #4047, opened 2026-09-16
ci: parameterize exit gate configuration

duwenxin99:
This PR parameterizes the exit gate scripts to accept environment variables for bucket destinations and introduces a safety flag to control publishing. This allows us to safely test the release pipeline in a sandbox environment without accidentally triggering a production deployment.


.ci/versioned.release.cloudbuild.yaml
```diff
@@ -425,6 +425,9 @@ steps:
       - "publish-npm-to-ar"
     env:
       - "BUILD_ID=$BUILD_ID"
+      - "EXIT_GATE_PROJECT=$_EXIT_GATE_PROJECT"
```
wangauone, line 428: Both retry scripts (https://github.com/googleapis/mcp-toolbox/blob/main/.ci/pypi_retry.cloudbuild.yaml and https://github.com/googleapis/mcp-toolbox/blob/main/.ci/npm_retry.cloudbuild.yaml) will also use the shell. So let's add these variables there as well.

.ci/versioned.release.cloudbuild.yaml
```diff
@@ -471,3 +477,6 @@ substitutions:
   _ASSETS_BUCKET: toolbox-build-assets
   _DOCKER_URI: ${_AR_HOSTNAME}/${PROJECT_ID}/${_AR_REPO_NAME}/toolbox
   _PUSH_LATEST: "false" # Substituted in trigger
+  _EXIT_GATE_PROJECT: "mcp-toolbox"
+  _EXIT_GATE_MANIFEST_BUCKET: "oss-exit-gate-prod-projects-bucket"
+  _PUSH_TO_EXIT_GATES: "false"
```
wangauone, line 482: We are defaulting not publishing to exit gate here. Is this expected? I assume we would do PyPI and NPM publish during each versioned release?

.ci/trigger_exit_gate.sh
```diff
@@ -36,12 +36,18 @@ case "${REGISTRY}" in
...
-readonly EXIT_GATE_PROJECT="mcp-toolbox"
-readonly MANIFEST_BUCKET="oss-exit-gate-prod-projects-bucket"
+readonly EXIT_GATE_PROJECT="${EXIT_GATE_PROJECT:?EXIT_GATE_PROJECT must be set}"
+readonly EXIT_GATE_MANIFEST_BUCKET="${EXIT_GATE_MANIFEST_BUCKET:?EXIT_GATE_MANIFEST_BUCKET must be set}"
+
+PUSH_TO_EXIT_GATES="${PUSH_TO_EXIT_GATES:-false}"
+if [[ "${PUSH_TO_EXIT_GATES}" != "true" ]]; then
+  echo "Skipping exit gates for ${REGISTRY} (PUSH_TO_EXIT_GATES != true)"
+  exit 0
+fi
```
wangauone, line 42: nit: move this PUSH_TO_EXIT_GATES above the variable definitions? So  EXIT_GATE_PROJECT and EXIT_GATE_MANIFEST_BUCKET won't be required when PUSH_TO_EXIT_GATES is disabled.

wangauone (approved):

.ci/versioned.release.cloudbuild.yaml
duwenxin99: I substituted this variable as "true" in the official release trigger similar to what we are doing with "_PUSH_LATEST_". That will override this value.

.ci/versioned.release.cloudbuild.yaml
duwenxin99: Updated.

.ci/trigger_exit_gate.sh
duwenxin99: Done.

PR #4048, opened 2026-09-16
refactor(skills): derive skill membership through a registry

twishabansal:
## Description

Moves skill-membership grouping out of private helpers in `Discover` into a `Registry` type, so (`skills/list`, `skills/get`) and (`resources/directory/read`) share one implementation instead of three that drift.


dishaprakash (approved):

internal/skills/registry.go
```diff
@@ -0,0 +1,134 @@
...
+			continue
+		}
+		// Walk the URI's ancestors rather than every root, so the scan costs
+		// path depth instead of the number of skills.
+		for i := strings.LastIndex(uri, "/"); i > 0; i = strings.LastIndex(uri[:i], "/") {
+			// underSkill is the test Entry.Validate applies to every ref. A
+			// looser rule here admits a member the validation then rejects,
+			// which fails startup for the whole config.
+			root := uri[:i]
+			if segs, ok := rootSegs[root]; ok && underSkill(uri, resources.SkillScheme, segs) {
```
anubhav756: if a skill:// file doesn't match any `SKILL.md` (eg. because of a typo in the URI), it looks like it gets dropped without any error. Would it make sense to warn here, or maybe fail on startup? Or is there a case where this is expected that I'm not seeing?

anubhav756 (approved):

PR #4066, opened 2026-09-17
feat(tool/conversational-analytics-list-accessible-data-agents): return all data agents and support manual paging

Genesis929:
## Description

list_data_agents issued a single dataAgents:listAccessible request and returned the raw response, so projects with more agents than one page silently received a truncated list plus a nextPageToken the caller had no parameter to act on.

- Add optional page_size and page_token parameters; when either is set the tool returns that single page unchanged.
- When both are omitted, drain every page and return the merged list, bounded by 1000 agents / 100 pages / 5 minutes. A nextPageToken is returned only when a bound stopped the drain early.
- Preserve extra top-level fields (e.g. unreachable) by concatenating their JSON arrays across pages, and stop when the API repeats a page token.
- Propagate ctx into every request and surface a paging hint on drain timeout.
- Update the prebuilt config description and tool docs.

## PR Checklist

- [ ] Make sure you reviewed
- [ ] Ensure you have manually reviewed the entire diff before requesting a
  review


internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
```diff
@@ -147,38 +177,199 @@ func (t Tool) Invoke(ctx context.Context, s sources.Source, params parameters.Pa
...
+// far, concatenating JSON arrays so lists such as `unreachable` keep every page.
+func mergeExtras(extras, pageExtras map[string]json.RawMessage) {
+	for field, value := range pageExtras {
+		extras[field] = concatJSONArrays(extras[field], value)
+	}
+}
+
+// concatJSONArrays joins two JSON arrays, falling back to `next` when either
+// value is missing or is not an array.
+func concatJSONArrays(prev, next json.RawMessage) json.RawMessage {
```
shobsi, line 274: [Nit] You might want to flip the fallback to return `prev` if `next` fails to parse as an array, so that you don't lose the previously successfully aggregated fields (like the `unreachable` array). 

```go
func concatJSONArrays(prev, next json.RawMessage) json.RawMessage {
	if len(prev) == 0 {
		return next
	}
	var prevItems, nextItems []json.RawMessage
	if json.Unmarshal(prev, &prevItems) != nil {
		return next 
	}
	if len(next) == 0 || json.Unmarshal(next, &nextItems) != nil {
		return prev // Keep what we successfully parsed so far
	}
	merged, err := json.Marshal(append(prevItems, nextItems...))
	if err != nil {
		return prev
	}
	return merged
}
```

shobsi (approved):

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
Genesis929: Updated `concatJSONArrays` so an unmergeable `next` falls back to `prev` instead of discarding previously aggregated array entries.

Also added `prevItems == nil` / `nextItems == nil` checks alongside `json.Unmarshal` (and hoisted `len(next) == 0`), since Go's `encoding/json` unmarshals JSON `null` into a `nil` slice with `err == nil` — without the `nil` check, a `null` field on an earlier page would be treated as an array and block later pages from overwriting it. Added `TestConcatJSONArrays` covering these cases.

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
```diff
@@ -80,7 +91,20 @@ func (cfg Config) Initialize(context.Context) (tools.Tool, error) {
 		cfg.Location = "global"
 	}
 
-	params := parameters.Parameters{}
+	minPageSizeValue := minPageSize
```
Yuan325, line 94: Might wanna remove this~ Seems like it's not needed

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
```diff
@@ -34,6 +37,14 @@ import (
 
 const resourceType string = "conversational-analytics-list-accessible-data-agents"
 
+const (
+	minPageSize       = 1
+	maxAutoDataAgents = 1000
+	maxAutoPages      = 100
+)
+
+var maxAutoDuration = 5 * time.Minute
```
Yuan325: We can put this in the `const` above

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
```diff
@@ -147,38 +177,211 @@ func (t Tool) Invoke(ctx context.Context, s sources.Source, params parameters.Pa
...
+	if err != nil {
+		return nil, util.NewClientServerError("failed to create GDA client", http.StatusInternalServerError, err)
+	}
+	client.Timeout = 30 * time.Second
+
+	return listAccessibleDataAgents(ctx, client, util.GetGDAEndpoint(), source.GetProjectID(), t.Cfg.Location, pageSize, pageToken)
+}
+
+// parsePaginationParams extracts optional page_size and page_token parameters.
+func parsePaginationParams(params parameters.ParamValues) (*int, string, util.ToolboxError) {
```
Yuan325, line 190: Recommend moving this function inline within `Invoke()`

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
```diff
@@ -147,38 +177,211 @@ func (t Tool) Invoke(ctx context.Context, s sources.Source, params parameters.Pa
...
+		if !ok {
+			return nil, "", util.NewAgentError(fmt.Sprintf("error casting 'page_token' parameter: %v", pt), nil)
+		}
+		pageToken = s
+	}
+	return pageSize, pageToken, nil
+}
+
+// listAccessibleDataAgents returns a single page when pagination parameters are provided, or all pages by default.
+func listAccessibleDataAgents(ctx context.Context, client *http.Client, endpoint, projectID, location string, pageSize *int, pageToken string) (any, util.ToolboxError) {
```
Yuan325, line 217: Recommend moving this inline as well~

Yuan325 (approved):
Thank you!

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
Genesis929: Done — it was a `var` only so tests could shrink the budget, so I pass the budget into
`listAllAccessibleDataAgents` instead. Same 5 min in production, and the tests no longer
mutate package state.

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
Genesis929: I'd rather keep it out of `Invoke()` — as a free function it's unit-tested directly (3 table
tests); inlining would need a full source + token-source mock. Happy to move the pagination
helpers into a sibling `pagination.go` if it's about file organisation.

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
Genesis929: Same here, more so: this seam is what lets the 18 pagination tests drive the logic against an
`httptest` server. Inlining would cost that coverage and grow `Invoke()` to ~120 lines.

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
Genesis929: Unfortunately this one is load-bearing: `parameters.WithIntMinValue` takes a `*int`, and
`minPageSize` is an untyped constant, which isn't addressable in Go, so `&minPageSize`
won't compile:

    invalid operation: cannot take address of minPageSize (untyped int constant 1)

The local is just there to give the constant an address. The alternatives are a package-level
`var minPageSizeValue = minPageSize` (a mutable global for no real gain) or dropping the
`WithIntMinValue` bound entirely — I'd rather keep the bound, since it's what surfaces the
constraint in the tool manifest the model sees.

Happy to restructure if you had a specific shape in mind — or were you pointing at the fact
that `parsePaginationParams` also re-checks `size < minPageSize`? That second check is
deliberate: it's the layer the unit tests exercise directly, without going through the
parameter framework.

Yuan325: Got it, lets keep as is then~ :)

internal/tools/conversationalanalytics/conversationalanalyticslistaccessibledataagents/conversationalanalyticslistaccessibledataagents.go
Yuan325: Got it, let keep it for now then. We can move it to a different package when other tools need this.

